PR #57, opened 2021-02-07
Initialize Transact Warhead System

MortonPL:
### Transact Warhead

- It is now possible to increase, decrease or transfer specific "values" (referred to as Currency) between Technos.
  - To enable this feature, `Transact=true` must be set and at least one `X.Source` or `X.Target` tag needs to be non-zero.
  - Source is the Techno that fired the warhead, target is a Techno (or a group) that is affected by the warhead.
  - `X.Flat` values grant absolute Currency, while `X.Percent` values grant Currency calculated relatively to the unit's state.
    - In case of experience, value from `X.Percent` is calculated by multiplying it by current unit cost.
    - By default `X.Source.Percent` (percent of value that Source will receive) is calculated from Source unit cost, however it is possible to force calculation from Target cost with `CalcFromTarget` (Target and `CalcFromSource` behaves analogically).
    - When both `X.Flat` and `X.Percent` values are set, the one that yields more Currency is used.
  - If final Currency amount is positive or negative for both Source and Target, then all affected units gain or lose Currency.
  - When the final amount of Currency is negative only for either Source or Target, **Transfer** occurs.
    - In transfer, if the absolute value of Currency for Source and Target are different, then the smaller of them is used.
    - In transfer, if it is not possible to transfer the whole amount (i.e. a unit has not enough Currency), then only the possible amount is transferred.
  - Source values are ignored if the warhead is not created by a unit (i.e. via a super weapon, animation weapon or map action).
  - `Transact.SpreadAmongTargets` has effect on warheads with CellSpread. When set, instead of applying the full Currency amount to each Target separately, the value will be evenly distributed among the victims.
  - `Transact.Experience.IgnoreNotTrainable=true` will allow to give experience to units with `Trainable=false` set.
  - `Transact.RequiresValidTarget=false` will allow the source unit to lose/obtain Currency when the warhead doesn't detonate on a valid target (i.e. having 0% versus a unit). However, target remains unaffected and if Source was supposed to take part in a transfer, nothing will happen!
  - `Transact.RequiresAnyTarget=false` will allow the source unit to lose/obtain Currency when the warhead doesn't detonate on any target (i.e. force firing on the ground). However, if Source was supposed to take part in a transfer, nothing will happen! This tag takes priority over `Transact.RequiresValidTarget`.

```{warning}
It is impossible to revert veterancy-caused unit type conversion from Ares by demoting a unit!
```

Currently supported "values" are:
- Experience

In `rulesmd.ini`:
```ini
[SOMEWARHEAD]                                           ; Warhead
Transact=false                                          ; boolean
Transact.RequiresAnyTarget=true                         ; boolean
Transact.RequiresValidTarget=true                       ; boolean
Transact.SpreadAmongTargets=false                       ; boolean
Transact.Experience.IgnoreNotTrainable=false            ; boolean
Transact.Experience.Source.Flat=0                       ; integer
Transact.Experience.Source.Percent=0.0                  ; float, percents or absolute (-1.0 - 1.0)
Transact.Experience.Source.Percent.CalcFromTarget=false ; boolean
Transact.Experience.Target.Flat=0                       ; integer
Transact.Experience.Target.Percent=0.0                  ; float, percents or absolute (-1.0 - 1.0)
Transact.Experience.Target.Percent.CalcFromSource=false ; boolean
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -167,6 +167,7 @@ This page lists all the individual contributions to the project by their author.
   - Building `LimboDelivery` logic
   - Fix for `Image` in art rules
   - Power delta counter
+  - Transact Warhead
   - Help with docs
 - **ChrisLv_CN** (work relicensed under [following permission](https://github.com/Phobos-developers/Phobos/blob/develop/images/ChrisLv-relicense.png)):
    - General assistance
--- a/Phobos.vcxproj
+++ b/Phobos.vcxproj
@@ -84,6 +84,7 @@
     <ClCompile Include="src\Ext\TerrainType\Hooks.cpp" />
     <ClCompile Include="src\Ext\TerrainType\Hooks.Passable.cpp" />
     <ClCompile Include="src\Ext\Unit\Hooks.DisallowMoving.cpp" />
+    <ClCompile Include="src\Ext\WarheadType\Transact.cpp" />
     <ClCompile Include="src\Ext\WarheadType\Detonate.cpp" />
     <ClCompile Include="src\Ext\Techno\Hooks.Firing.cpp" />
     <ClCompile Include="src\Ext\Techno\Hooks.Shield.cpp" />
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ Credits
 - **Morton (MortonPL)** - developer
 - **Trsdy (chaserli)** - developer
 
+- **Morton (MortonPL)** - XDrawOffset, Shield passthrough & absorption, building LimboDelivery, fix for Image in art rules, power delta counter, transact warhead
 For all contributions see [full credits list](CREDITS.md).
 
 Thanks to everyone who uses Phobos, tests changes and reports bugs! You can show your appreciation and help project by displaying the logo (monochrome version can be found [here](https://github.com/Phobos-developers/Phobos/blob/develop/logo-mono.png)) in your client/launcher (make it open Phobos GitHub page for extra fanciness), linking to Phobos repository, contributing or donating to us via the links above.
--- a/README.md
+++ b/README.md
@@ -87,7 +87,6 @@ Credits
 - **Morton (MortonPL)** - developer
 - **Trsdy (chaserli)** - developer
 
-- **Morton (MortonPL)** - XDrawOffset, Shield passthrough & absorption, building LimboDelivery, fix for Image in art rules, power delta counter, transact warhead
 For all contributions see [full credits list](CREDITS.md).
 
 Thanks to everyone who uses Phobos, tests changes and reports bugs! You can show your appreciation and help project by displaying the logo (monochrome version can be found [here](https://github.com/Phobos-developers/Phobos/blob/develop/logo-mono.png)) in your client/launcher (make it open Phobos GitHub page for extra fanciness), linking to Phobos repository, contributing or donating to us via the links above.
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -919,6 +919,40 @@ In `rulesmd.ini`:
 BigGap=false   ; boolean
 ```
 
+### Transact Warhead
+
+- It is now possible to increase, decrease or transfer specific "values" between Technos.
+  - To enable this feature, `Transact=true` must be set and at least one source or target value needs to be non-zero.
+  - Source is the Techno that fired the warhead, Target is a Techno (or group) that is affected by the warhead.
+  - Flat values are amounts that are directly added/substracted, while percent values are calculated relatively to the unit's state.
+    - In case of experience, value from percent is calculated by multiplying it by current unit cost.
+    - By default Source.Percent (percent of value that Source will receive) is calculated from Source unit cost, however it is possible to force calculation from Target cost with `CalcFromTarget` (Target and `CalcFtomSOurce` behaves analogically).
+  - When both Flat and Percent values are set, the one that yields higher value is used.
+  - Transfer occurs only when either the calculated experience to Source or to Target is negative. If both values are positive or negative, then all units gain or lose experience, respectively.
+  - In transfer, if Source and Target values are different, then the smaller of them is used.
+  - In transfer, if it is not possible to transfer the whole amount (i.e. a unit has not enough experience), then only the maximum possible transfer amount is transferred.
+  - Source values are ignored if the warhead is not created by a unit (i.e. via a super weapon or map action). Target values are ignored if the warhead doesn't affect any units (fired at empty ground).
+    - In both cases, transfer is impossible.
+  - `Transact.SpreadAmongTargets` has effect on warheads with CellSpread. When set, instead of applying the full value to each Target separately, the value will be divided among the victims.
+
+It is impossible to revert veterancy-caused unit type conversion from Ares by demoting a unit!
+
+Currently supported "values" are:
+- Experience
+
+In `rulesmd.ini`:
+```ini
+[SOMEWARHEAD]                                           ; Warhead
+Transact=false                                          ; boolean
+Transact.SpreadAmongTargets=false                       ; boolean
+Transact.Experience.Source.Flat=0                       ; integer
+Transact.Experience.Source.Percent=0.0                  ; float, percents or absolute (-1.0 - 1.0)
+Transact.Experience.Source.Percent.CalcFromTarget=false ; boolean
+Transact.Experience.Target.Flat=0                       ; integer
+Transact.Experience.Target.Percent=0.0                  ; float, percents or absolute (-1.0 - 1.0)
+Transact.Experience.Target.Percent.CalcFromSource=false ; boolean
+```
+
 ### Trigger specific NotHuman infantry Death anim sequence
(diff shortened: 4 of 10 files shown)


Utilities/MortonUtils.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+#include <TechnoClass.h>
+
+class MortonUtils
```
Metadorius, line 4: Class would be an overkill. I introduced `GeneralUtilities.h/.cpp` file where the function could go until it's categorized.

src/Utilities/MortonUtils.cpp
```diff
@@ -0,0 +1,15 @@
+#include "MortonUtils.h"
```
Metadorius, line 1: Code duplication.

src/Ext/WarheadType/Hooks.cpp
```diff
@@ -20,6 +22,82 @@ void ReshroudMapForOpponents(HouseClass* pThisHouse) {
...
+			// ugly solution to 1/3 + 1/3 + 1/3 = 0.99 etc
+			vstruct->Add(toBeAdded + Math::sgn(toBeAdded) * 0.01);
+		}
+		return transffered;
+	}
+};
+*/
+
+void ModifyExperience(BulletClass* pThis, WarheadTypeExt::ExtData* pWHExt, bool isPercent) {
+	auto const pTarget = abstract_cast<TechnoClass*>(pThis->Target);
```
Metadorius, line 49: @Belonit told me `abstract_cast` is the reason why it wouldn't compile.

Metadorius (changes requested):
Looks good, left a bit of feedback in the code comments.

Otamaa:
Move "AddNoNegative" function to somewhere out "GeneralUtils" to "Hook.cpp" for example , this will fix the compiler problems

MortonPL:
Since the fork is automatically rechecked, I've got some things to explain here:

- cellspread support is here, but still requires additional testing to search and destroy possible bugs.
- that <abstract_cast>(TechnoClass*) is to stay - build errors are no more and normal cast causes fatals,
- as of now warhead doesn't work with GenericWarhead SW,
- in other cases everything's fine

SMxReaver:
This is a repost of the comments made in #phobos-project to keep feedback easily found:
- Experience has been working well so far, the only noticed issue I have currently is that the effect ignores Trainable=no and will promote even civilian structures.
- In addition, infantry firing inside an opentopped transport will promote as expected when Experience.FirerGetsExp=yes. When inside a garrisoned structure however, the building itself receives the experience and the firer receives none.

Metadorius:
For reference, the suggested tags for this are:
```INI
Transact.Source.Experience.Flat=0
Transact.Source.Experience.Percent=0
Transact.Source.Experience.Percent.CalcFromTarget=false

Transact.Target.Experience.Flat=0
Transact.Target.Experience.Percent=0
Transact.Target.Experience.Percent.CalcFromSource=false

Transact.SpreadAmongAffectedTargets=true
```
The way it works is that you calculate the availability of specified value changes for target and source in form of 0-1 coeffitient (or, say, percentage from 0% to 100%), and then take the lowest coeffitient and use it for transfer. If there's no transfer (only source or only targets are affected) then you just apply that one coeffitient.

`SpreadAmongAffectedTargets` is for switching the CellSpread mode: if it's true - then the transact values are spread amongst all targets affected by this warhead (so 20 exp. flat source/target transfer divided among 4 units would transfer as 5 to each, 20 total from source), if false - then full transfer would be applied to each affected target (same situation would result in 20 to each of 4 units, 80 from source).

This is to lay out the foundation for Transact warhead expansion.

MortonPL:
After a long-needed rework, the now Transact Warhead should be ready for preliminary testing.

ayylmaoRotE:
Tested 4 times all resulted in an IE.
7C0A74F6 - rookie source & rookie aircraft target
783C1AF1 - veteran source & rookie unit target
78811AF1 - veteran source & unit target
Uploaded dumps to discord.

Used default settings with Transact=true
Tried changing some of the exp tags but no luck, still get IEs.
All attempts were with rookie and veteran units alike.

ayylmaoRotE:
541F1BB2 on CellSpread

[ExperienceWH]
Verses=10%,10%,10%,10%,10%,10%,10%,10%,10%,10%,10%
CellSpread=5
Transact=true                                          ; boolean
Transact.SpreadAmongTargets=false                       ; boolean
Transact.Experience.Source.Flat=0                       ; integer
Transact.Experience.Source.Percent=0.0                  ; float, percents or absolute (-1.0 - 1.0)
Transact.Experience.Source.Percent.CalcFromTarget=false ; boolean
Transact.Experience.Target.Flat=0                       ; integer
Transact.Experience.Target.Percent=2.0                  ; float, percents or absolute (-1.0 - 1.0)
Transact.Experience.Target.Percent.CalcFromSource=false ; boolean

ayylmaoRotE:
Same IE happens when you smash too many EXPs into a single unit too quickly:
e.g. 
[PilotWH]
Verses=0%,0%,0%,10%,10%,10%,0%,0%,0%,0%,0%
Versus.t2mediump=0%
Transact=true                                          ; boolean
Transact.SpreadAmongTargets=false                       ; boolean
Transact.Experience.Source.Flat=0                       ; integer
Transact.Experience.Source.Percent=0.0                  ; float, percents or absolute (-1.0 - 1.0)
Transact.Experience.Source.Percent.CalcFromTarget=false ; boolean
Transact.Experience.Target.Flat=0                       ; integer
Transact.Experience.Target.Percent=10.0                  ; float, percents or absolute (-1.0 - 1.0)
Transact.Experience.Target.Percent.CalcFromSource=false ; boolean

even with Percent=2.0, though I can't reproduce every time, especially if you do it 1 by 1 initially. See vids below:

https://user-images.githubusercontent.com/69301373/158943007-f646fd80-938f-4911-905c-c63efd6d8d0d.mp4


https://user-images.githubusercontent.com/69301373/158943014-c6f61e9e-2bb0-4c05-8af3-04dfbd2996b5.mp4


https://user-images.githubusercontent.com/69301373/158943016-d46b0de5-cd2d-45d3-8a63-176e3891fa3c.mp4


https://user-images.githubusercontent.com/69301373/158943025-5303baba-46ce-4cdd-9868-1b0bdb5063af.mp4

:

ayylmaoRotE:
Some more findings (bugs):
1. The Transact.Experience aspect of the warhead will ignore verses/versus and whether the target is Trainable=yes or not 
e.g. if the WH has CellSpread=6 and Verses=0%,0%,0%,10%,10%,10%,0%,0%,0%,0%,0%, everything will be affected, infantry and buildings too.

> CellSpread=<1 or no CellSpread:
> - Flat: Works on source and target
> - Percent: Works on source and target
> 
> CellSpread=1>:
> - Flat: Works on source and target
> - Percent: Works on source and target
> - Only works on weapons, if triggered via SW (GenericWarhead) it will crash

3. If an infantry has IsDesolator=yes (AreaFire potentially affects this too), this breaks CellSpread on the Exp WH, e.g.:

> CellSpread=<1 or no CellSpread:
> - Flat: Does not work at all
> - Percent: Does not work at all
> 
> CellSpread=1>:
> - Flat: The correct value will be given to the source only, e.g. 500
> - Percent: The source will instantly gain the maximum experience possible regardless of the value used
> - CellSpread value is actually ignored, it won't affect any nearby targets

4. Trainable=no is ignored, e.g. MCVs, Engineers, buildings etc will also gain veterancy

mevitar:
> EDIT: some further tests, if the unit does not die mid EXP transfer, no crashes, but the Weapon I use kills the unit (not via Suicide but via Anim=), I think this is why the crash happens, maybe store the unit info like anim2unit directions is stored?

I can confirm the on-death crashes (although in my case i tried it with Death.NoAmmo=yes, so the units always die during the transfer, even though i used an Inviso=yes projectile).

> 4. Trainable=no is ignored, e.g. MCVs, Engineers, buildings etc will also gain veterancy

I would ask for it to be kept that way, because with it and Ares' conversion-on-promotion, modders can make units that don't gain experience, but still morph into a different one when affected by an SW or attacked by something. If someone wants exclude specific units from gaining experience, there's Versus.X for that.
But preferably, it might be best if whether Trainable=no units are affected was a separate tag.

mevitar:
> 3\. If an infantry has IsDesolator=yes (AreaFire potentially affects this too), this breaks CellSpread on the Exp WH, e.g.:

It's not AreaFire=yes, that one works fine. CellSpread also appears to works fine for me (tried with CellSpread=3).
Did you perhaps try it on a test infantry? I made that mistake and it didn't work at first because incorrect sequence forced it to use primary on deploy. Once i put the weapon on a proper YuriClone-like infantry, it worked fine (except for the fact that verses were still completely ignored and everything in the area got the experience...).

MortonPL:
Reopening after a long break. All reported crashes and issues (ignoring verses and Trainable) should be addressed, updated documentation is in PR description.

Metadorius (commented):
From a quick glance (can't do a full review at the time, sorry) I've found two things that bother me.

PR #63, opened 2021-02-20
TileSet 255 and above bridge repair fix

E1Elite:
To do: To include .cpp in vcxproj file. If needed .inj file entries for hooks.


src/Misc/tileset255_bridgerepairfix.cpp
```diff
@@ -0,0 +1,183 @@
...
+	bridges become non-repairable. The variable values of BridgeTopRight1 and BridgeBottomLeft2
+	that are parsed into, gets corrupted. To fix it, this patch allocates new addresses for those variables.
+*/
+
+#include <Helpers/Macro.h>
+
+int BridgeTopRight_1 = -1;
+int BridgeBottomLeft_2 = -1;
+
+DEFINE_HOOK(545BBA, tileset255_01, 5)
```
Metadorius, line 14: 1. My recommendation on how to write hook names is `[ClassName]_[MethodName]_[HookName]` or `[FunctionName]_[HookName]`.
2. Hook size is only used when you use `return 0;` so it's better to set size to `0` if you don't use it.

Metadorius (changes requested):
We don't use inj file in Phobos, so you only need to include it in vcxproj and then everyone can try the nightly DLL build. I've left a couple of comments in code.

E1Elite:
Updated, check if its fine now.

Metadorius:
Still not included in `.vcxproj`, can't push a fix to that because your PR doesn't allow edits by maintainers.

E1Elite:
Allow edits by maintainers is enabled from the beginning, I unchecked and checked the option again now. Latest clone or download doesn't compile straightaway for some reason, gives include errors. I compiled and tested with 2021-01-25 build, so I am avoiding change to .vcxprj file from latest commit.

Having .inj file has advantages for the modders, they can disable some hooks if they have trouble or don't want a specific change wothout going for a build from the source.

Metadorius:
Strange, I can't push to the branch anyways.

Latest `develop` compiles fine already. Perhaps you could rebase, start a new branch and reset `develop` to `upstream/develop` ? Something like `feature/bridge-repair-fix`.

Oh and also I highly suggest hopping in on Discord to stay in touch.

E1Elite:
May be problem could be me being on VS2019. Problem in compiling starts from 2021-02-12 commit onwards giving errors on selection.cpp and tacticalclass.h.

If you want to, go ahead and add the one line inclusion of the .cpp outside this PR.

Metadorius:
What's the problem? I can compile just fine, are you sure YRpp submodule has the correct branch checked out?

E1Elite:
Updated vcxprj. Problem was with YRpp branch, was using master. It would be better if YRpp URL in readme leads to correct branch.

E1Elite:
Updated with better fix.

PR #83, opened 2021-03-07
Add a config to control if the development commands should be enabled

secsome:
And some fixes on the templates

### DIFF
--- a/src/Commands/Commands.h
+++ b/src/Commands/Commands.h
@@ -3,6 +3,8 @@
 #include <YRPP.h>
 #include <CommandClass.h>
 
+#include "../Phobos.h"
+
 class PhobosCommandClass : public CommandClass
 {
 protected:
--- a/src/Commands/Commands.h
+++ b/src/Commands/Commands.h
@@ -5,7 +5,18 @@
 
 class PhobosCommandClass : public CommandClass
 {
-
+protected:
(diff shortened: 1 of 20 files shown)


secsome:
and RemoveMindControl warhead feature, I forgot to split it up again.

src/Commands/Commands.h
```diff
@@ -3,9 +3,22 @@
...
 
+#include "../Phobos.h"
+
 class PhobosCommandClass : public CommandClass
 {
-
+protected:
+	bool CheckDebugDeactivated() const {
+		if (!Phobos::Config::DevelopmentCommands) {
+			if (const wchar_t* text = StringTable::LoadString("TXT_COMMAND_DISABLED")) {
```
Metadorius, line 13: Not sure if there's a need to print anything at all in this case? Also wouldn't it return `MISSING:\`TXT_COMMAND_DISABLED\`` in case the CSF entry is missing? Could be better to use `LoadStringOrDefault`?

Metadorius (approved):
Looks good aside from one small question.

src/Commands/Commands.h
secsome: As Phobos requires Ares to run, we just leave this  call as Ares does.

PR #100, opened 2021-03-16
Feature/rad site

Otamaa:
Add HouseClass for RadSite
New tag introduce on weapon to set the owner to nullptr if user wanted to use old behaviour
[WEAPON]
RadNullOwner=false ;bool

:AffectsAllies=bool now can be use with RadWarhead
:Killing  techno using Radiation now counted as scored


Metadorius:
Squashed the commits, please force pull before doing any changes.
`AffectsAllies` doesn't work for me though.

PR #116, opened 2021-03-25
NoManualMove tag for TechnoTypes

### LINKED ISSUE #5: Permanently stationary units
Introduce INI key (name suggestion: `Stationary=true/false`) that allows any unit (or at least vehicles) to permanently exhibit behaviour identical / similar to what vehicles with `IsSimpleDeployer=true` when deployed (they don't accept movement commands, never move on their own accord but can fire weapons, acquire targets, deploy etc).

Primary use-case would likely be to combine with Ares' `Convert.Deploy` to simulate regular `IsSimpleDeployer` with much greater control over the deployed state's properties. May also be useful with certain dummy units, map objects etc.

Metadorius:
Allows to make TechnoTypes unable to be ordered with a movement command.

```ini
[SOMETECHO]      ; TechnoType
NoManualMove=no  ; boolean
```

Closes #5.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ Credits
 - **SMxReaver** - help with docs, extensive and thorough testing
 - **4SG** - help with docs
 - **wiktorderelf** - overhauled Unicode font
-- **Uranusian (Thrifinesma)** - Mind Control enhancement, custom warhead splash list, harvesters counter, promoted spawns, shields, death after dead fix, customizeable missing cameo, cameo sorting priority, placement mode responding of tab hotkeys fix, producing progress, custom ore gathering anim, overhauled Unicode font, help with docs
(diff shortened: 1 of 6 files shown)


Starkku:
On the above-linked nightly build, setting Speed=0 unit results in at least the following problems:

- Does not actually prevent issuing movement commands and the unit gets stuck like usual when you do.
- It displays 'no move' cursor all the time when the unit is selected, which also actually prevents manual firing etc. Deployed regular IsSimpleDeployer units actually display the default cursor unless targeting an unit and so on instead of this one.

Thrifinesma:
Allows to make TechnoTypes immobile.

[SOMETECHO]        ; TechnoType
NoManualMove=no  ; boolean

PR #119, opened 2021-03-28
Fix loader blowfish.dll

Belonit:
Fixed error `***FATAL*** String Manager failed to initilaized properly`, which occurred if `Blowfish.dll` could not be registered in the OS, for example, it happened when the player did not have administrator rights. With Phobos, if the game did not find a registered file in the system, it will no longer try to register this file, but will loading it bypassing registration. 

Screenshot of a bug in a vanilla game:
BlowfishError

### DIFF
--- a/Phobos.vcxproj
+++ b/Phobos.vcxproj
@@ -34,6 +34,7 @@
     <ClCompile Include="src\Ext\WeaponType\Body.cpp" />
     <ClCompile Include="src\Ext\WeaponType\Hooks.cpp" />
     <ClCompile Include="src\Ext\WeaponType\Hooks.DiskLaserRadius.cpp" />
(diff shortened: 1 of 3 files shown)


Belonit:
One way to test this:
1. Delete (or rename) all ``Blowfish.dll`` on your system until the vanilla game starts showing this error.
2. Disable administrator rights for your account.
3. Disable administrator rights for your account.
4. Create a new game folder with the correct ``Blowfish.dll``
5. Check if the error is still present. ---For an objective test, the error must be present and repeated more than once.
6. Add this version of Phobos to the folder, the game should start as usual, no problem.
7. If you delete Phobos, the game should still give an error.

Metadorius (approved):
Confirmed to be working by @FS-21.

PR #130, opened 2021-04-04
Specifies deploy facing for SimpleDeployers (ground units)

AutoGavy:
Also works with ares convertion. Units will turn to correct dir first instead of In-place deploy


Thrifinesma:
Ares already had one similar feature

PR #139, opened 2021-04-05
Fixed anim bug when deploying into structure

FS-21:
Also tested the undeploy anim and if the player explodes when the original owner of the MC unit died.

### DIFF
--- a/src/Ext/TechnoType/Hooks.VehicleToBuildingMCFix.cpp
+++ b/src/Ext/TechnoType/Hooks.VehicleToBuildingMCFix.cpp
@@ -5,26 +5,16 @@
 #include <InfantryClass.h>
 #include <BuildingClass.h>
 #include "../Phobos.h"
+#include "../../Misc/CaptureManager.h"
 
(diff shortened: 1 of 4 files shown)


src/Ext/TechnoType/Hooks.VehicleToBuildingMCFix.cpp
```diff
@@ -5,26 +5,16 @@
...
-				}
-			}
-			if (!FoundNode) { // say the link was broken beforehand, by inconsistently ordered code...
-				Manager->CaptureUnit(To);
-			}
+			CaptureManager::FreeUnit(Manager, From, true);
+			CaptureManager::CaptureUnit(Manager, To, true);
+
+			To->QueueMission(Mission::Construction, 0);
+			To->Mission_Construction();
```
Metadorius, line 17: Instead of unconditionally making techno construct perhaps it would be better to re-set the mission that was there before the unit was captured?

Metadorius (commented):
The code is good, but perhaps some change would be needed, refer to the code comment.

FS-21:
Code improvement, based on the feedback.

Issue #11, opened 2020-11-12
Customizable Chrono Logic


deathreaperz:
I hope these tags can be customized for TechnoTypes so that can give various Chrono ability.

ChronoDelay=
ChronoReinfDelay=
ChronoDistanceFactor=
ChronoTrigger=
ChronoMinimumDelay=
ChronoRangeMinimum=
WarpIn=
WarpOut=
WarpAway=

PR #142, opened 2021-04-05
Customizeable Teleport/Chrono Locomotor properties per attached Techno

Otamaa:
-Now fully customizeable per techno type direcly 
-Unfilled value default to Rules
-This properties only able to used by Techno that have Teleport Locomotor/Chrono attached

In `rulesmd.ini`:
```ini
[SOMETECHNO]			; TechnoType 
Locomotor={4A582747-9839-11d1-B709-00A024DDAFD1} ;required 
WarpOut					;Anim (played when Techno Warping out)			
WarpIn					;Anim (played when techno Warping in)
WarpAway				;Anim (played when Chronowarped by chronosphere)
ChronoTrigger			;boolean
ChronoDistanceFactor	        ;int
ChronoMinimumDelay		;int
ChronoRangeMinimum		;int
ChronoDelay			        ;int frame
```


deathreaperz:
test2

First, this is my default code from [General]
```
ChronoDelay=64
ChronoReinfDelay=180
ChronoDistanceFactor=48 ;0.56s
ChronoTrigger=yes
ChronoMinimumDelay=16 ;0.56s
ChronoRangeMinimum=0
```

Then this is my custom code for Ronco
```
[RONCO]
ChronoTrigger=yes
ChronoDistanceFactor=32
ChronoMinimumDelay=14
ChronoRangeMinimum=0
ChronoDelay=56
```
From what I read, Ronco's Chrono Teleportation delay should've been faster than CLeg's, but the result is not. Also, I haven't use WarpIn, WarpOut, and WarpAway but the default anim of that section kinda messed up because as you can see in the GIF, it uses WAKE2 Tiberian Sun leftover animation. Anyone else have tested this?

Otamaa:
> 
> 
> test2
> 
> First, this is my default code from [General]
> 
> ```
> ChronoDelay=64
> ChronoReinfDelay=180
> ChronoDistanceFactor=48 ;0.56s
> ChronoTrigger=yes
> ChronoMinimumDelay=16 ;0.56s
> ChronoRangeMinimum=0
> ```
> 
> Then this is my custom code for Ronco
> 
> ```
> [RONCO]
> ChronoTrigger=yes
> ChronoDistanceFactor=32
> ChronoMinimumDelay=14
> ChronoRangeMinimum=0
> ChronoDelay=56
> ```
> 
> From what I read, Ronco's Chrono Teleportation delay should've been faster than CLeg's, but the result is not. Also, I haven't use WarpIn, WarpOut, and WarpAway but the default anim of WarpIn kinda messed up because as you can see in the GIF, it uses WAKE2 Tiberian Sun leftover animation.
> Also, when I set Ronco's ChronoTrigger=no, all of Chrono units including CLeg will also have ChronoTrigger=no
> 
> Anyone else have tested this?

For anim it is actually intended behaviour , since WW coded everything to Warpout 
this patch is just separate them to 3 tags from [General] WarpIn WarpOut WarpAway like it should do 
but , well , if everyone doesnt like it i can restore it back 

and for delay , it is wonky indeed , for some reason it cant go faster , need to investigate that later 

and thanks for the feedback !

Metadorius:
Closes #11.

src/Ext/TechnoType/Hooks.Teleport.cpp
```diff
@@ -0,0 +1,99 @@
...
+	R->EDX<AnimTypeClass*>(pExt->WarpOut.Get(RulesClass::Instance->WarpOut));
+
+	return 0x71943F;
+}
+
+DEFINE_HOOK(719788, TeleportLocomotionClass_ILocomotion_Process_WarpInAnim, 6)
+{
+	GET_LOCO(ESI);
+
+	R->EDX<AnimTypeClass*>(pExt->WarpIn.Get(RulesClass::Instance->WarpOut));
```
chaserli, line 24: I think we should just use the default but unused `[General]WarpIn` for this one, instead of keep using `WarpOut`.

PR #151, opened 2021-04-08
Script 73 Wait until full ammo

FS-21:
Action 73,0 was created in mind for Aircrafts but could be used in other scenarios. It prevents a very rare bug that makes the AI aircraft going up & down in loop in the helipad.

Example of use:
[B0001066-G]
Name=Load and attack sructures in loop
0=73,0
1=0,2
2=6,1


FS-21:
All team members that are dead or limboed are ignored.

FS-21:
New changes.

PR #168, opened 2021-04-11
Modify Ammo and Gattling by Warheads

Coronia:
Allow to modify Techno's ammo and gattling with a warhead

`[WarheadType]`
`TargetGattlingStage (positive integer, default to 0)`
`TargetGattlingRateUp (integer, default to 0)`
`TargetReloadAmmo (integer, default to 0)`

Warhead with **TargetGattlingStage** can directly change the Techno's gattling stage to a specific value. The index is 1-based so 1 means the first stage, 2 means the second and so on. It'll also set the current GattlingValue of this Techno to Stage(X-2).

**TargetGattlingRateUp** will add a certain amount to the GattlingValue and change the weapon respectively. The value can be both positive and negative here. The value after calculation won't exceed the TechnoType's largest WeaponStage or below 0.

When fired by warheads with **TargetReloadAmmo** the Techno's ammo will be reloaded such amount. The value can also be negative to support a remove ammo attack.

Tested under MO 3.3.5 environment.


Metadorius:
Thanks for your contribution. We're going to implement a single system for modification/transfer of param values for the sake of consistency, reusability and ease of implementing support for new param value modifications, but this is gonna happen around 0.3 development, so this PR is gonna sit here for awhile.

mevitar:
What happened to this one? I wanted to test it but it says the build has expired.

Metadorius:
> What happened to this one? I wanted to test it but it says the build has expired.

GitHub automatically deletes old build artifacts unfortunately.

@Coronia could you rebase your PR onto newest develop?

mevitar:
I assumed this was committed to the devbuild so i tried testing it there, but i can't get any of this to work. Did i misunderstand something and the code for it actually isn't there?

Metadorius:
> I assumed this was committed to the devbuild so i tried testing it there, but i can't get any of this to work. Did i misunderstand something and the code for it actually isn't there?

It's actually not, we were planning to implement this differently but didn't get to it.

FS-21:
I cloned his repo & compiled it for testing:

TargetGattlingStage tag works as expected. Next time the weapon fires it will fire with the selected Gattling stage and you can't break it with negative values or higer values than the max stage the affected unit have.

TargetGattlingRateUp tag works as expected. The RateUp value gets affected and I did the 4 expected tests (negative value, positive value, an very very high value in positive and in negative for forcing what happens when is higher than the max unit RateUp values).

TargetReloadAmmo tag, like above two works as expected with the same tests for TargetGattlingRateUp.

Metadorius (changes requested):
Code needs to be refactored, docs, credits and changelog entries need to be added, otherwise good.

PR #174, opened 2021-04-13
Strafing aircraft weapon improvements.

Starkku:
- Add `Strafing.Shots` key to weapons that can be used to control number of shots fired during strafing run.
- Add `Strafing.SimulateBurst` key to weapons that can be used to simulate firing offset-altering effect of `Burst` on `Burst=1` strafing run weapons.
- Fix to take `Burst` into account for shots beyond the first one, this also applies to `Fighter=no` aircraft without strafing pattern weapon.


FS-21:
I've been testing this tag in a faulty aircraft that resembles the old TS Nod Banshee behaviour and it solves the 2 main problems it had without this tag  (removing all unnecessary strafing runs with no attacks & the possibility to attack Naval Yards without staying in a infinite loop of strafing runs without attacks).

Observations:
I tried "Strafing.SimulateBurst" with 2, 5 (default value) 6 and 10 and the max number of shots is 5. Probably is how it was designed so no objections here, simply works.
In each volley the ammo is decreased as it should be.
Fighter=yes seems ignored.

strafe lines improvement

Starkku:
> I tried "Strafing.SimulateBurst" with 2, 5 (default value) 6 and 10 and the max number of shots is 5. Probably is how it was designed so no objections here, simply works.

Currently it simply allows changing the length of the strafing run from between 1 and 5. Extending it to beyond 5 is not completely outside the realm of possibility but it comes with challenges of its own so probably best saved for future.

> Fighter=yes seems ignored.

Fighter=yes/no does not interact with strafing weapons (projectile ROT is 0/unset or 1) at all. What it does is determine whether or not the aircraft behaves like Harrier (Fighter=yes) or TS ORCA Fighter / Harpy (Fighter=no) when it doesn't have a strafing weapon.

Metadorius (approved):

Issue #165, opened 2021-04-10
Shields and Verses targeting


TAK02:
latest nightly build, I think. 
Description: I have a tank with an anti-inf and an anti-tank weapon (warheads with different verses), and a regular vehicle with infantry-grade shield. The tank refuses force-fire on the shielded unit, so it probably can't fire at all on such shields.

```
[ANTI_VEH_WH] ;for use against vehicles
Verses=0%,0%,0%,40%,40%,50%,25%,25%,25%,20%,0%

[ANTI_INF_WH] ;for use against infantry
Verses=20%,18%,14%,0%,0%,0%,0%,0%,0%,0%,40%

[SHIELDED_UNIT]
Shield.Armor=none
```

PR #185, opened 2021-04-20
Shield logic improvements

Thrifinesma:
Fix the known bug during shield implementation


TAK02:
If this was supposed to fix Shields and Verses targeting #165 it didn't.

TAK02:
Good news and bad news.
The good news is the proper weapon gets used to attack shields.
The bad news is the unit not switching back to main weapon. Or maybe I just set up my shield's regeneration improperly, I don't know.

PR #191, opened 2021-04-24
Dark theme for documentation

Metadorius:
Todo: add javascript switch to turn dark theme on and off.


Metadorius:
I think white elements here (current section, prev/next buttons) should be gray instead.

image

Metadorius:
I also think this part should have different color as it's the sidebar header. 
image

Belonit (approved):

PR #202, opened 2021-05-01
Critical damage chance; Projectile interception logic

TAK02:
Critical damage chance needs more info on what is, and Projectile interception logic doesn't work well non-AA weapons do the interception.


Metadorius:
Too storytelling. Implemented in 8702106.

PR #226, opened 2021-06-03
Spawned Negative Damage fix

### LINKED ISSUE #222: [Vanilla Bug] Fix spawned AircraftTypes's targeting (enable Repair Drones)
Description :
AircraftTypes spawned using the spawner logic will not stop attacking a target until the latter is destroyed, until the spawner is destroyed or until that AircraftType is destroyed. 

For instance, an AircraftType with a vehicle repair weapon will not stop healing a damaged unit, even once it's fully repaired. It will go back to its spawner, but will be spawned back to repair the vehicle forever. This means that it will not re-target a new damaged unit and will simply be stuck in a repairing loop.

To reproduce :
To reproduce the bug, simply give the [HORNET] a repair weapon (Damage=-1) and give the [CARRIER] the AttackFriendlies=yes tag.

Expected behavior :
The spawned AircraftType should be launched from its spawner unit, repair a nearby damaged vehicle, and once the latter is at full health, select a new damaged vehicle if it has any ammo left, ore go back to the spawner unit and scan for a new damaged unit.

Otamaa:
close  #222
Please test and confirm if the  fix working correcly for the issue mentioned!

### DIFF
--- a/Phobos.vcxproj
+++ b/Phobos.vcxproj
@@ -124,6 +124,7 @@
     <ClInclude Include="src\Utilities\Macro.h" />
     <ClInclude Include="src\Utilities\Parser.h" />
     <ClInclude Include="src\Utilities\Patch.h" />
+    <ClInclude Include="src\Utilities\PhobosTemplate.h" />
     <ClInclude Include="src\Utilities\Template.h" />
     <ClInclude Include="src\Utilities\TemplateDef.h" />
     <ClInclude Include="YRpp\GameTextManager.h" />
--- a/src/Ext/Techno/Body.cpp
+++ b/src/Ext/Techno/Body.cpp
@@ -161,6 +161,47 @@ bool TechnoExt::IsHarvesting(TechnoClass* pThis)
 	return false;
 }
 
+// [Vanilla Bug] Fix spawned AircraftTypes's targeting (enable Repair Drones) #222
+void TechnoExt::ApplyTargetHealthCheck(TechnoClass* pThis)
+{
+	if (const auto Targettech = abstract_cast<TechnoClass*>(pThis->Target))
+	{
+		if (auto manager = pThis->SpawnManager)
+		{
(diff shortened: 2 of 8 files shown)


Danielovich7:
Ok so I tested the build you sumitted, and made various tests (mostly adding/removing AttackFriendlies=, OpportunityFire= and CanPassiveAquire= on both the spawner and the spawned). So it works... kinda. Here's the sequence of events that I have :

1. Once the spawner get in GuardRange=6 of a vehicle, it will automatically spawn the first of its two Repair Drone, regardless of the vehicle's health pool. If no vehicles are within GuardRange=6, the drone lands back on the spawner and waits for a friendly vehicle to come close once more;
2.a If the vehicle is at full health, the spawner continously targets and launch its repair drone (but only the first one of two?), meaning that the drone will fly around "attacking" the vehicle, but will abandon mission instantly and try to land back on the spawner (it has a landing sound that indicate it tries to land on the spawner, so that's how I know);
2.b If the vehicle, or another nearby vehicle within GuardRange=6, is damaged, both repair drone will go into "attack" mission to repair it. It will only attack the required amount of time so that all damaged vehicle are at full health. Success!
3. Once all damaged vehicles are fully repaired, the second drone goes back to the spawner and lands, and the first one goes back into the attack mission loop described in point 2.a

I hope this helps! Thank for the time you're taking to fix this, very appreciated.

secsome:
I had refactored it to make the code cleaner, please test it when you are available @Danielovich7

Danielovich7:
After retesting, here's what happens : 

https://user-images.githubusercontent.com/79372870/120852345-12e13b80-c548-11eb-97de-681b69bbe859.mp4

secsome:
Currently, if a spawn with a too high speed, it will probably got stuck, such as the HORNET with Speed = 26.
I think we should make this PR in draft until all the hidden problems are all set.

PR #227, opened 2021-06-04
implement Power Plant Enhancer #131

secsome:
This is a prototype of #131 , please test it if anyone is available.
```ini
[SomeBuilding]
PowerPlantEnchancer.PowerPlants=<a list of buildingtypes>
PowerPlantEnchancer.Amount=integer
PowerPlantEnchancer.Factor=float
```

NOTICE, currently this feature add all amounts and then multiply the factors, both Amount and Factor need to be a positive number.

Example 1:
You have 1 A(PowerPlantEnchancer.Amount=10), 2 B(Power=100)
your total power = (100 + 10) * 2

Example 2:
You have 2 A(PowerPlantEnchancer.Amount=10), 2 B(Power=100)
your total power = (100 + 10 * 2) * 2

Example 3:
You have 1 A(PowerPlantEnchancer.Amount=10,PowerPlantEnchancer.Factor=.1), 2 B(Power=100)
your total power = (100 + 10) * 0.1 * 2

Example 4:
You have 3 A(PowerPlantEnchancer.Amount=10), 2 B(Power=100), 2 C(,PowerPlantEnchancer.Factor=.1)
your total power = (100 + 10 * 3) * 0.1 * 0.1 * 2

### DIFF
--- a/src/Ext/BuildingType/Hooks.cpp
+++ b/src/Ext/BuildingType/Hooks.cpp
@@ -1,0 +1,44 @@
+#include "Body.h"
+
+#include <BuildingClass.h>
+#include <HouseClass.h>
+
+int BuildingTypeExt::GetEnchancedPower(BuildingClass* pBuilding, HouseClass* pHouse)
+{
+	int nPower = pBuilding->GetPowerOutput();
+
+	for (const auto pBld : pHouse->Buildings)
+	{
+		const auto pSrcType = abstract_cast<BuildingTypeClass*>(pBuilding->GetType());
+		const auto pType = abstract_cast<BuildingTypeClass*>(pBld->GetType());
+		const auto pExt = BuildingTypeExt::ExtMap.Find(pType);
+
+		if (pExt->PowerPlantEnchancer_Buildings.Contains(pSrcType))
+			if (pExt->PowerPlantEnchancer_Amount > 0)
+				nPower += pExt->PowerPlantEnchancer_Amount;
+	}
+
+	for (const auto pBld : pHouse->Buildings)
+	{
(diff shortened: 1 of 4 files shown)


src/Ext/BuildingType/Hooks.cpp
```diff
@@ -0,0 +1,44 @@
...
+	for (const auto pBld : pHouse->Buildings)
+	{
+		const auto pSrcType = abstract_cast<BuildingTypeClass*>(pBuilding->GetType());
+		const auto pType = abstract_cast<BuildingTypeClass*>(pBld->GetType());
+		const auto pExt = BuildingTypeExt::ExtMap.Find(pType);
+
+		if (pExt->PowerPlantEnchancer_Buildings.Contains(pSrcType))
+			if (pExt->PowerPlantEnchancer_Factor > 0)
+				nPower = static_cast<int>(nPower * pExt->PowerPlantEnchancer_Factor);
+	}
```
Metadorius, line 30: You don't need to iterate twice on that, can do that in one loop. Also a better pattern in this case would be to use Nullables. Also multi-line braceless blocks are not allowed.

Metadorius (changes requested):
Functions as expected as confirmed by @Bu7loos. I suggest to apply factor before the flat bonus though, this would be more intuitive and allow for greater flexibility and interesting stuff like decreasing output of power plants with a factor and adding a flat bonus.

Also needs doc/readme entries and a rebase.

Metadorius (changes requested):
1. Apparently `.Amount` doesn't accept negative numbers?
2. It's "enhance", not "ehchance".

PR #233, opened 2021-06-10
Remove the limitation of Waypoint

secsome:
In vanilla YR, this number is limited to [0, 701], this PR will remove that limitation.

~~Who will use all 702 waypoints?~~

### DIFF
--- a/Phobos.vcxproj
+++ b/Phobos.vcxproj
@@ -29,6 +29,7 @@
     <ClCompile Include="src\Ext\RadSite\Body.cpp" />
     <ClCompile Include="src\Ext\RadSite\Hooks.cpp" />
     <ClCompile Include="src\Ext\Scenario\Body.cpp" />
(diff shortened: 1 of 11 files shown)


secsome:
For the map editor, see https://github.com/secsome/FA2sp/tree/develop

E1Elite:
It would be better to give the modders an option to use this feature or not. It could be with like UseUnlimitedWaypoints boolean flag in rules INI.

Metadorius:
> It would be better to give the modders an option to use this feature or not. It could be with like UseUnlimitedWaypoints boolean flag in rules INI.

Could you explain why would modders need that?

E1Elite:
Sample tested trigger actions, numbers - 7, 41, 42, 43, 48, 63, 80, 88, 99, 128.
TeamType field - Waypoint.
Script actions - most of those.
Event number 24 was tested for not crashing.
Player locations in multiplayer map.
HomeCell and AltHomeCell.

These were found to be working as expected.

PR #235, opened 2021-06-12
Possible fix buggy MoveToCell which uses 128 as leftovers

secsome:
This  possibly fix the Script Action 4 Move to cell from calculating the cell from dividing 128 to 1000.


E1Elite:
Fix works as expected.

Issue #112, opened 2021-03-23
[Vanilla Bug] Make FireOnce=yes with deploy weapons work on other TechnoTypes


Danielovich7:
Currently, the tag "FireOnce=" only seems to work on InfantryTypes. It would be helpful to have it work on other TechnoTypes. For example, this could make vehicle using a deployable weapon exit their attack mmission once the weapon is fired.

Thrifinesma:
It already works on all sorts of TechnoTypes

Danielovich7:
Then maybe there's an interaction with another tag that prevent it from working, because it does not work for me.

Falaka21:
It doesn't work for me,also.

Thrifinesma:
![FireOnce](https://user-images.githubusercontent.com/45818100/114292527-a7d43500-9ac1-11eb-854a-e622e0b6b8f4.gif)
I set `FireOnce=yes` to M60 105mm Vulcan2, all work

secsome:
Anyone can test it again?

PR #244, opened 2021-06-18
Vehicle DeployFire improvements / fixes

Starkku:
- Explicitly use weapon specified by `DeployFireWeapon` for firing deploy weapon.
- Respect `FireOnce` setting on the weapon.
- Respect stop commands issued during firing the deploy weapon.

If I am not mistaken this addresses #112.

### DIFF
--- a/src/Ext/TechnoType/Hooks.cpp
+++ b/src/Ext/TechnoType/Hooks.cpp
@@ -100,3 +100,25 @@ DEFINE_HOOK(6B7282, SpawnManagerClass_AI_PromoteSpawns, 5)
 
 	return 0;
 }
+
+DEFINE_HOOK(73DD12, UnitClass_Mission_Unload_DeployFire, 6)
+{
+	enum { ReturnFromFunction = 0x73DD3C };
+
+	GET(UnitClass*, pThis, ESI);
+
+	int weaponIndex = pThis->GetTechnoType()->DeployFireWeapon;
+
+	if (pThis->GetFireError(pThis->Target, weaponIndex, true) == FireError::OK)
+	{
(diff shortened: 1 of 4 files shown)


src/Ext/TechnoType/Hooks.cpp
```diff
@@ -99,4 +99,26 @@ DEFINE_HOOK(6B7282, SpawnManagerClass_AI_PromoteSpawns, 5)
 	}
 
 	return 0;
+}
+
+DEFINE_HOOK(73DD12, UnitClass_Mission_Unload_DeployFire, 6)
+{
+	enum { ReturnFromFunction = 0x73DD3C };
```
secsome, line 106: I don't think it's quite necessary do declare an enum as it is the only return addr you had used

src/Misc/Hooks.BugFixes.cpp
```diff
@@ -61,3 +64,16 @@ DEFINE_HOOK(737D57, UnitClass_ReceiveDamage_DyingFix, 7)
 
 	return 0;
 }
+
+DEFINE_HOOK(4C7518, EventClass_Execute_StopUnitDeployFire, 9)
+{
+	GET(TechnoClass*, pThis, ESI);
+
+	if (pThis->WhatAmI() == AbstractType::Unit && pThis->CurrentMission == Mission::Unload &&
+		pThis->GetTechnoType()->DeployFire && !static_cast<UnitClass*>(pThis)->Type->IsSimpleDeployer)
```
secsome, line 73: Try to use abstract_cast to make our code cleaner

Metadorius (approved):

PR #247, opened 2021-06-21
Allow to customize missing cameo

Thrifinesma:
This feature allows to custom missing cameo with any shp/pcx file. When using a PCX as XXICON, we don't need to change `cameo.pal` to extend colors for the cameo anymore. This feature is standalone, i.e. don't need Ares to work
```ini
[AudioVisual]
MissingCameo=XXICON.SHP   ; filename
```

### DIFF
--- a/src/Misc/Hooks.PCX.cpp
+++ b/src/Misc/Hooks.PCX.cpp
@@ -52,20 +52,21 @@ DEFINE_HOOK(6A99F3, StripClass_Draw_DrawMissing, 6)
 		_strlwr_s(pFilename);
 
 		if (!_stricmp(pCameoRef->Filename, "xxicon.shp")
-			&& _stricmp(pCameoRef->Filename, pFilename))
+			&& _stricmp(pCameoRef->Filename, pFilename)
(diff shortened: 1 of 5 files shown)


src/Misc/Hooks.PCX.cpp
```diff
@@ -37,3 +39,36 @@ DEFINE_HOOK(5535D0, PCX_LoadScreen, 6)
...
+
+		if (!_stricmp(pCameoRef->Filename, "xxicon.shp")
+			&& _stricmp(pCameoRef->Filename, pFilename)
+			&& strstr(pFilename, ".pcx"))
+		{
+			PCX::Instance->LoadFile(pFilename);
+			if (auto CameoPCX = PCX::Instance->GetSurface(pFilename))
+			{
+				GET(int, TLX, ESI);
+				GET(int, TLY, EBP);
```
Metadorius, line 62: Should rename `TLX` and `TLY` to something more sensible, and also make it camelCase.

Metadorius (changes requested):
Looks good aside from one minor issue.

PR #274, opened 2021-07-07
Burst logic improvements / additions

### LINKED ISSUE #46: Laser is mirrored relative to FireFLH
Muzzle flash is displayed correctly, but the laser is not (or rather, it is mirrored).
for example
`PrimaryFireFLH=170,100,120`

The weapon of the tank has a `Burst=2`
Muzzle flash is displayed in the required coordinate 170, 100, 120 and the laser is mirrored to 170, -100, 120

inage

### LINKED ISSUE #271: [New Feature] Separate burst delay for weapon type
### Description:
In the original game, we can use BurstdelayX on units to specify the delay between each burst, but it is not applicable to buildings. NPext has UseROFAsBurstDelays function to allow bursts to be fired using the weapon's ROF value for both unit and building. ARES does not appear to have such a function (thought can be simulated using ammo).  

### Possible Implementation:

Method 1: Burst delay on weapon
[Weapon]
Burst=...
Burst.Delay=Frame (should overwrite BurstdelayX)

Starkku:
### Firing offsets for specific Burst shots

- You can now specify separate firing offsets for each of the shots fired by weapon with `Burst` via using `(Elite)PrimaryFire|SecondaryFire|WeaponX|FLH.BurstN` keys, depending on which weapons your TechnoType makes use of. *N* in `BurstN` is zero-based burst shot index, and the values are parsed sequentially until no value for either regular or elite weapon is present, with elite weapon defaulting to regular weapon FLH if only it is missing. If no burst-index specific value is available, value from the base key (f.ex `PrimaryFireFLH`) is used.
- Burst-index specific firing offsets are absolute firing offsets and the lateral shifting based on burst index that occurs with the base firing offsets is not applied.

In `artmd.ini`:
```ini
[SOMETECHNO]                   ; TechnoType Image
PrimaryFireFLH.BurstN=         ; int - forward, lateral, height
ElitePrimaryFireFLH.BurstN=    ; int - forward, lateral, height
SecondaryFireFLH.BurstN=       ; int - forward, lateral, height
EliteSecondaryFireFLH.BurstN=  ; int - forward, lateral, height
WeaponXFLH.BurstN=             ; int - forward, lateral, height
EliteWeaponXFLH.BurstN=        ; int - forward, lateral, height
```

In addition effects like lasers are no longer drawn on wrong firing offsets with Burst.

### Burst.Delays

- Allows specifying weapon-specific burst shot delays. Takes precedence over the old `BurstDelayX` logic available on VehicleTypes, functions with Infantry & BuildingType weapons (AircraftTypes are not supported due to their weapon firing system being completely different) and allows every shot of `Burst` to have a separate delay instead of only first four shots.
- If no delay is defined for a shot, it falls back to last delay value defined (f.ex `Burst=3` and `BurstDelays=10` would use 10 as delay for all shots).
- Using -1 as delay reverts back to old logic (`BurstDelay0-3` for VehicleTypes if available or random value between 3-5 otherwise) for that shot.

In `rulesmd.ini`:
```ini
Burst.Delays=-1              ; int - burst delays (comma-separated) for shots in order from first to last.
```

Closes #271
Closes #46

### DIFF
--- a/src/Misc/Hooks.BugFixes.cpp
+++ b/src/Misc/Hooks.BugFixes.cpp
@@ -103,7 +103,7 @@ DEFINE_HOOK(0x702299, TechnoClass_ReceiveDamage_DebrisMaximumsFix, 0xA)
 				amountToSpawn = Math::min(amountToSpawn, totalSpawnAmount);
 				totalSpawnAmount -= amountToSpawn;
 
-				for ( ; amountToSpawn > 0; --amountToSpawn)
+				for (; amountToSpawn > 0; --amountToSpawn)
 				{
 					GameCreate<VoxelAnimClass>(pType->DebrisTypes.GetItem(currentIndex),
 						&cord, pThis->Owner);
@@ -162,6 +162,28 @@ DEFINE_HOOK(0x4FB2DE, HouseClass_PlaceObject_HotkeyFix, 0x6)
 	GET(TechnoClass*, pObject, ESI);
 
 	pObject->ClearSidebarTabObject();
+
+	return 0;
+}
+
+// Issue #46: Laser is mirrored relative to FireFLH
+// Author: Starkku
+DEFINE_HOOK(0x6FF2BE, TechnoClass_Fire_At_BurstOffsetFix_1, 0x6)
+{
+	GET(TechnoClass*, pThis, ESI);
+
+	--pThis->CurrentBurstIndex;
+
+	return 0x6FF2D1;
+}
+
+DEFINE_HOOK(0x6FF660, TechnoClass_Fire_At_BurstOffsetFix_2, 0x6)
+{
+	GET(TechnoClass*, pThis, ESI);
+	GET_BASE(int, weaponIndex, 0xC);
+
+	++pThis->CurrentBurstIndex;
+	pThis->CurrentBurstIndex %= pThis->GetWeapon(weaponIndex)->WeaponType->Burst;
 
 	return 0;
 }
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ Credits
 - **ChrisLv_CN** - interceptor logic, LaserTrails, laser fixes, general assistance (work relicensed under [following permission](images/ChrisLv-relicense.png))
 - **Xkein** - general assistance, YRpp edits
 - **thomassneddon** - general assistance
-- **Starkku** - Warhead shield penetration & breaking, strafing aircraft weapon customization, vehicle DeployFire fixes/improvements, stationary VehicleTypes
+- **Starkku** - Warhead shield penetration & breaking, strafing aircraft weapon customization, vehicle DeployFire fixes/improvements, stationary VehicleTypes, Burst logic improvements
 - **SukaHati (Erzoid)** - Minimum interceptor guard range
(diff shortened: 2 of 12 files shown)


src/Misc/Hooks.BugFixes.cpp
```diff
@@ -144,4 +144,24 @@ DEFINE_HOOK(73DD12, UnitClass_Mission_Unload_DeployFire, 6)
...
+
+	return 0x6FF2D1;
+}
+
+DEFINE_HOOK(6FF660, TechnoClass_Fire_At_BurstOffsetFix_2, 6)
+{
+	GET(TechnoClass*, pThis, ESI);
+	GET_BASE(int, weaponIndex, 0xC);
+
+    ++pThis->CurrentBurstIndex;
```
secsome, line 163: Check the align here.

Bu7loos:
image

Burst logic improvements / additions - Log 1:

Just found out that there is a problem with bursts logic attached to aircrafts most specifically with **burst delay**, the aircraft's weapon **does not respect it** and fire all 4 bursts together. Here is the code:

[Su80mmRockets]
Range=16
Burst=4
Burst.Delays=10 
Strafing.Shots=3
Strafing.SimulateBurst=false
TurboBoost=yes
FireOnce=no

Bu7loos:
Burst logic improvements / additions - Log 2:

Just found out that there is a problem with bursts logic when it comes the (EliteWeaponXFLH.BurstN) tag when the unit is in its non-Elite statues it responded nicely to normal custom burst fire port tags but when it gains elite statues it does not respond to it properly and it falls back the original lines that had been coded for normal non-Elite unit:

[LAV25] 
; Boris mode fire ports:
;Weapon26FLH=200,55,300    ; Boris Mode - Original

Weapon26FLH.Burst0=200,0,350 - no problems here - OK 
Weapon26FLH.Burst1=200,-55,350 - no problems here - OK 
Weapon26FLH.Burst2=200,55,350 - no problems here - OK 

EliteWeapon26FLH.Burst0=200,0,350 - problems here - reverted to non-Elite fire ports 
EliteWeapon26FLH.Burst1=200,-55,350 - problems here - reverted to non-Elite fire ports 
EliteWeapon26FLH.Burst2=200,55,350 - problems here - reverted to non-Elite fire ports 

EliteWeapon26FLH.Burst3=200,0,300 - problems here - reverted to non-Elite fire ports to which there was none
EliteWeapon26FLH.Burst4=200,-55,300 - problems here - reverted to non-Elite fire ports to which there was none
EliteWeapon26FLH.Burst5=200,55,300 - problems here - reverted to non-Elite fire ports to which there was none

Starkku:
Fixed issue with burst FLH's for EliteWeaponX not being parsed properly.

Metadorius (approved):
Add a what's new entry and we're good to go.

Starkku:
> Just found out that there is a problem with bursts logic attached to aircrafts most specifically with **burst delay**, the aircraft's weapon **does not respect it** and fire all 4 bursts together.

I missed this earlier on, just stating now for future reference that `Burst.Delays` does not (at least currently) function with AircraftType weapons (as their entire weapon firing code functions in a different way), a fact that is mentioned in documentation for this feature.

PR #291, opened 2021-07-12
Undeploy building into a unit plays EVA_NewRallyPointEstablished

### LINKED ISSUE #290: [Vanilla Bug] Fix Building-to-Vehicle building EVA Event
### Description:
When undeploying a building into a vehicle by giving it a move order (so for buildings that have UndeploysInto=), the player will get an EVA event saying "New rally point established", as if the he was setting a rally point from a factory building. This seems to be a vanilla bug.

### Conditions:
I think it requires to use the Vehicle-to-Building logic, using DeploysInto= and UndeploysInto=. Then, simply move the undeployable building with a move command.

### Steps To Reproduce:
1. As seen bellow, give a vehicle DeploysInto=THEBUILDING and give that building UndeploysInto=THEVEHICLE.
2. Move the deployed building so it undeploys. You will hear the EVA event.

### Expected Behaviour:
No EVA event should be played.

### Actual Behaviour:
EVA says "New rally point established".

### Additional Files / Links:

https://user-images.githubusercontent.com/79372870/125210635-136ba100-e26f-11eb-8db4-d14322a3a2c7.mp4

secsome:
Closes #290


Danielovich7:
I've tested this build and the bug appears to be fixed. Thanks!

PR #292, opened 2021-07-12
Feature/producing progress

Thrifinesma:
producing-progress-01

- You can now know your factories' status via sidebar!
- You need to draw your own assets (`tab0xpp.shp`, x is replaced by 0-3) and put them into `sidec0x.mix`.

In `uimd.ini`:
```ini
[Sidebar]
ProducingProgress.Show=no            ; boolean
```


src/Ext/Sidebar/Hooks.cpp
```diff
@@ -0,0 +1,74 @@
...
+				{
+					pFactory = pPlayer->GetPrimaryFactory(rtti, false, i == 1 ? BuildCat::Combat : BuildCat::DontCare);
+				}
+				else
+				{
+					pFactory = pPlayer->GetPrimaryFactory(AbstractType::UnitType, false, BuildCat::DontCare);
+					if (!pFactory || !pFactory->Object)
+						pFactory = pPlayer->GetPrimaryFactory(AbstractType::UnitType, true, BuildCat::DontCare);
+					if (!pFactory || !pFactory->Object)
+						pFactory = pPlayer->GetPrimaryFactory(AbstractType::AircraftType, false, BuildCat::DontCare);
```
Metadorius, line 57: This construct seems strange to me. Is there no way of checking which factory do you need to get without actually getting it?

Metadorius (commented):
Looks good aside from a few code style issues.
Also IIRC you were going to implement X/Y offsets?

src/Ext/Sidebar/Hooks.cpp
Thrifinesma: I think no

Thrifinesma: Each player has 6 factory entities (building, defense, infantry, vehicle, aircraft, naval), and the last three are sharing the fourth tab

PR #308, opened 2021-07-28
Re-enable obsolete [JumpjetControls]

Thrifinesma:
Re-enable obsolete [JumpjetControls], the keys in it will be as the default value of jumpjet units.
Moreover, added two tags for missing ones.

In `rulesmd.ini`: 
```ini
[JumpjetControls]
Crash=5.0       ; float
NoWabbles=no    ; bool
```

Note: `CruiseHeight` is for `JumpjetHeight`, `WobblesPerSecond` is for `JumpjetWobbles`, `WobbleDeviation` is for `JumpjetDeviation`, and `Acceleration` is for `JumpjetAccel`. All other corresponding keys just simply have no `Jumpjet` prefix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ Credits
 - **SMxReaver** - help with docs, extensive and thorough testing
 - **4SG** - help with docs
 - **wiktorderelf** - overhauled Unicode font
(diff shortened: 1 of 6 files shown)


Metadorius (changes requested):
Add docs and this should be merge-able.

Metadorius (approved):

PR #321, opened 2021-08-09
Unlimited Variables

secsome:
In vanilla YR, local variables are limited up to 100, now you can use up to INT_MAX waypoints.

### DIFF
--- a/src/Ext/Scenario/Body.h
+++ b/src/Ext/Scenario/Body.h
@@ -18,10 +18,12 @@ public:
 	public:
 		std::map<int, CellStruct> Waypoints;
 		std::map<int, Variable> LocalVariables;
+		std::map<int, Variable> GlobalVariables;
 
 		ExtData(ScenarioClass* OwnerObject) : Extension<ScenarioClass>(OwnerObject),
 			Waypoints {},
-			LocalVariables {}
+			LocalVariables {},
+			GlobalVariables {}
 		{
 
 		}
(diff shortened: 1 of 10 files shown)


Thrifinesma:
Unlimited Global Variables when :)

src/Ext/Scenario/Body.h
```diff
@@ -17,9 +17,13 @@ class ScenarioExt
...
 	public:
 		std::map<int, CellStruct> Waypoints;
+		std::map<int, Variable> LocalVariables;
+		std::map<int, Variable> GlobalVariables;
 
 		ExtData(ScenarioClass* OwnerObject) : Extension<ScenarioClass>(OwnerObject),
-			Waypoints {}
+			Waypoints {},
+			LocalVariables {},
+			GlobalVariables {}
```
Metadorius, line 26:
```suggestion
			Waypoints { },
			LocalVariables { },
			GlobalVariables { }
```

Metadorius (changes requested):
Overall the code is good but the code for globals and locals is pretty much duplicated. I suggest to remove duplication and merge if there are no any more issues. Also needs docs/readme entries.

SMxReaver:
I do not know the best way to test Save/Load since I am very unfamiliar with how locals function in-game, but I can confirm the game successfully fired Local Variable triggers that were indexed higher than 100. Attached is my test map, I use Local 103 to reveal the map and Local 90 to display "Boris" in the top left corner.
[variabletestmap.zip](https://github.com/Phobos-developers/Phobos/files/6995297/variabletestmap.zip)

Metadorius (approved):

PR #340, opened 2021-08-22
Ported `GiftBox` for Phobos

Otamaa:
Original Author : `ChrisLV-CN`
`https://github.com/ChrisLv-CN/PatcherExtension/blob/main/MyExtension/GiftBox.cs`

```This is not `TypeConversion` thing , there is no stat `Sync` from parent handled except `Destination` and `Focus` !```

No detail documentation yet  !
Tags: 
In `rulesmd.ini`:
```ini
[SOMETECHNO]
GiftBox.Types=;List TechnoType 
GiftBox.Nums=;List integer , Amount each Types will be created
GiftBox.Remove=;bool, Remove this Techno after creating Gifbox Types
GiftBox.Destroy=;bool, Destroy this Techno  after creating Gifbox Types
GiftBox.Delay=;integer , Giftbox Delay
GiftBox.RandomDelay=;integer (Min,Max) , Giftbox Random delay 
GiftBox.CellRandomRange=;integer 
GiftBox.EmptyCell=;bool
GiftBox.RandomType=;bool , Create random Types instead of all Types on the list 
```


RexDraco:
Some wishlists, not important:
Some type of auto scatter
a tag for auto selection of new unit(s) if giftbox was selected

When using "GiftBox.Destroy=" on infantry, they do not appear to actually be killed but rather just remove (Terrorists do not give off explosion). The differences between destroy and remove seems to be whether they follow the walking orders from giftbox or not. For units, it works as intended and in both cases they follow previous orders.

Infantry inside Bioreactors do not appear to keep track of delay.
Infantry cannot seem to convert to structures, did not know if this was by design or not, but they tend to disappear because of it

Infantry, when constructed, seem to walk into the cell of a converted infantry type without care. This is regardless if the infantry type is also converting to a new infantry or not. Having an GI walk to the rally point, convert to Brute, and then a GGI walk to same rally point and never convert will cause GGI to walk right on top of the Brute. Note I do not see a difference even if there's one Brute or ten, just that three GGI causes the cell to be full so he walks to the next.


When using this on SLAV types, they come out of the slave miner as expected and, even when converted to a new infantry, they travel to where the ore the SLAV was walking towards and then stand there. They continue to walk to destination regardless if removed or destroyed, in spite the common pattern from other scenarios with infantry where remove causes them to stop dead in their tracks. If delay is zero, the infantry simply exits the slave miner and then stops, not walking towards ore (probably because slaves do not have a destination set until finally exiting slave miner).

[SLAV]
...
Sight=5 ;!!
...
GiftBox.Types=SLAV
GiftBox.Nums=1
;GiftBox.Remove=yes
GiftBox.Destroy=yes;bool, Destroy this Techno  after creating Gifbox Types
GiftBox.Delay=50;integer , Giftbox Delay

Slaves spawn as usual from slave miner, then stop at Ore not mining due to convert to a non enslaved SLAV. Works as expected.
I do notice that, unlike tests with terrorists earlier, Slaves always go to last order given, even if it was their previous giftbox's order. Newly spawned slaves follows old orders.


Swapping to GiftBox.Remove:

GiftBox.Types=SLAV
GiftBox.Nums=1
GiftBox.Remove=yes
;GiftBox.Destroy=yes;bool, Destroy this Techno  after creating Gifbox Types
GiftBox.Delay=50;integer , Giftbox Delay

I notice no changes. Slaves follow last order given.


I notice that the SLAV has no sight in shroud however in both cases when being removed or "destroyed". Increasing the Delay to 500 fixes this.



Giftbox with jumpjets:

Insant transforms from units that concert into jumpjet work as expect, the rocketeer starts on the ground and flies of barracks
Transforms that have a delay seem to stop infantry from walking towards their location depending on if Destroy or Remove is used. Remove causes them to stop in their tracks.

Rocketeers that convert to GIs work as expected, the GI stands in the air until ordered to move.
Gis that convert to Rocketeers work as expected, the Rocketeer stands on the ground until ordered to move.
GIs deployed before convert does not negatively impact the jumpjets.


Giftbox with attackdogs:

When giving attack dogs a larger range and slower projectile speed, it can be observed that their delay isn't counted any longer. When given the attack dogs the ability to attack vehicles, their delay is also not counted. This applies to remove or destroy.



Giftbox With Units:

Work as expected, they continue to follow the orders given to the previous unit regardless if removed or destroyed.
Destroyed actually causes them to blow up with the explosion animation taking place.


Giftbox with Battlefortresses
They work as expected. However, infantry in original Battlefortress disappear and are lost with new battlefortress. 
-Infantry do not spawn even with the battlefortress being destroyed rather than removed, even with:
RookiePassengerChance= 100
Survivor.VeteranPassengerChance=100
Survivor.ElitePassengerChance=100

Infantry are unable to enter battlefortresses that are spawned this way, regardless if the previous one had infantry inside it or not, regardless if the previous unit was a battlefortress or not.


Giftbox with slave miner
When destroyed or removed, slaves go neutral, even if the SMIN turns into another SMIN



I will test Units more thoroughly next. This was primarily for infantry. Let me know if there's any questions or tests I should do, or if there's anything else I should have in mind.

src/New/Entity/GiftBoxClass.h
```diff
@@ -0,0 +1,146 @@
...
+	Valueable<bool> Remove;
+	Valueable<bool> Destroy;
+	Valueable<int> Delay;
+
+	Valueable<Point2D> DelayMinMax;
+	Valueable<int> RandomRange;
+	Valueable<bool> EmptyCell;
+	Valueable<bool> RandomType;
+
+	static const void LoadFromINI(GiftBoxData& nGiftboxData, INI_EX& parser, const char* pSection)
```
Metadorius, line 29: Could be made prettier as a class method.

src/New/Entity/GiftBoxClass.h
```diff
@@ -0,0 +1,146 @@
...
+	ValueableVector<TechnoTypeClass*> TechnoList;
+	ValueableVector<int> Count;
+	Valueable<bool> Remove;
+	Valueable<bool> Destroy;
+	Valueable<int> Delay;
+
+	Valueable<Point2D> DelayMinMax;
+	Valueable<int> RandomRange;
+	Valueable<bool> EmptyCell;
+	Valueable<bool> RandomType;
```
Metadorius, line 27: What if instead of RandomType we utilized the system me and Morton have invented for LimboDelivery? as in Types, Weights, Chances?

src/New/Entity/GiftBoxClass.h
```diff
@@ -0,0 +1,146 @@
...
+		Count(),
+		Remove(true),
+		Destroy(false),
+		Delay(0),
+		//min(X),max(Y)
+		DelayMinMax({ 0,0 }),
+		RandomRange(false),
+		EmptyCell(false),
+		RandomType(true)
+	{}
```
Metadorius, line 67:
```suggestion
	{ }
```

src/New/Entity/GiftBoxClass.h
```diff
@@ -0,0 +1,146 @@
...
+
+	explicit GiftBoxData(noinit_t) noexcept
+	{ }
+
```
Metadorius, line 71:
```suggestion
	explicit GiftBoxData(noinit_t) noexcept { }
```

src/New/Entity/GiftBoxClass.h
```diff
@@ -0,0 +1,146 @@
...
+};
+
+class GiftBoxClass
+{
+public:
+
+	GiftBoxClass() :
+		IsEnabled{ false },
+		Delay{ 0 }
+	{}
```
Metadorius, line 81:
```suggestion
	{ }
```

src/New/Entity/GiftBoxClass.h
```diff
@@ -0,0 +1,146 @@
...
+
+	GiftBoxClass(GiftBoxData& nGData) :
+		IsEnabled{ nGData },
+		Delay{ nGData.DelayMinMax.Get().Y == 0 ? abs(nGData.Delay.Get()) : abs(ScenarioClass::Instance->Random.RandomRanged(nGData.DelayMinMax.Get().X, nGData.DelayMinMax.Get().Y)) }
+	{}
```
Metadorius, line 86:
```suggestion
	{ }
```

src/New/Entity/GiftBoxClass.h
```diff
@@ -0,0 +1,146 @@
...
+public:
+
+	GiftBoxClass() :
+		IsEnabled{ false },
+		Delay{ 0 }
+	{}
+
+	GiftBoxClass(GiftBoxData& nGData) :
+		IsEnabled{ nGData },
+		Delay{ nGData.DelayMinMax.Get().Y == 0 ? abs(nGData.Delay.Get()) : abs(ScenarioClass::Instance->Random.RandomRanged(nGData.DelayMinMax.Get().X, nGData.DelayMinMax.Get().Y)) }
```
Metadorius, line 85: Split the line, too long.

Metadorius (changes requested):
I left some requests and suggestions, due to me doing this at 6AM feel free to argue against some of them. Overall the code style has greately improved since last PRs.

src/New/Entity/GiftBoxClass.cpp
```diff
@@ -0,0 +1,268 @@
...
+		}
+
+		return pTechno->Absorbed ||
+			pTechno->InOpenToppedTransport ||
+			pTechno->InLimbo ||
+			bIsGarrisoned ||
+			bIsOnWarfactory ||
+			pTechno->TemporalTargetingMe;
+
+	}
```
Metadorius, line 45:
```suggestion
	}
```

src/New/Entity/GiftBoxClass.cpp
```diff
@@ -0,0 +1,268 @@
...
+		{
+			for (CellSpreadEnumerator it((size_t)abs(nGiftBox.RandomRange.Get())); it; ++it)
+			{
+				auto const& offset = *it;
+				if (offset == CellStruct::Empty)
+					continue;
+
+				if (auto pCell = MapClass::Instance->TryGetCellAt(CellClass::Cell2Coord(nCellLoc->MapCoords + offset)))
+				{
+					if (nGiftBox.EmptyCell.Get() && (pCell->GetBuilding() || pCell->GetUnit(false) || pCell->GetInfantry(false)))
```
Metadorius, line 134:
```suggestion
					if (nGiftBox.EmptyCell.Get() && (pCell->GetBuilding()
							|| pCell->GetUnit(false) || pCell->GetInfantry(false)))
```

src/New/Entity/GiftBoxClass.cpp
```diff
@@ -0,0 +1,268 @@
...
+	{
+		auto pType = pTechno->GetTechnoType();
+		auto pTypeExt = TechnoTypeExt::ExtMap.Find(pType);
+		const auto newID = pTechno->get_ID();
+		int nDelay;
+
+		if (!IsTechnoChange && (strcmp(this->TechnoID, newID) != 0))
+		{
+			strcpy_s(this->TechnoID, newID);
+		}
```
Metadorius, line 182:
```suggestion
			strcpy_s(this->TechnoID, newID);
```

src/New/Entity/GiftBoxClass.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+			if (pTypeExt->GiftBoxData.Remove.Get())
+			{
+				// Limboing stuffs is not safe method depend on case
+				// maybe need to check if anything else need to be handle 
+				pTechno->Undiscover();
+				pTechno->Limbo();
+				pTechno->UnInit();
+
+			}
```
Metadorius, line 235:
```suggestion
				pTechno->UnInit();
			}
```

src/New/Entity/GiftBoxClass.cpp
```diff
@@ -0,0 +1,268 @@
...
+		}
+	}
+}
+
+DEFINE_HOOK(0x6F6CA0, TechnoClass_Put_Gbox, 0x7)
+{
+	GET(TechnoClass* const, pThis, ECX);
+	//	GET_STACK(CoordStruct*, pCoord, 0x4);
+	//	GET_STACK(int, Direction, 0x8);
```
Metadorius, line 259:
```suggestion
```

src/New/Entity/GiftBoxClass.h
```diff
@@ -0,0 +1,153 @@
...
+		Count.Read(parser, pSection, "GiftBox.Nums");
+		Remove.Read(parser, pSection, "GiftBox.Remove");
+		Destroy.Read(parser, pSection, "GiftBox.Destroy");
+		Delay.Read(parser, pSection, "GiftBox.Delay");
+		DelayMinMax.Read(parser, pSection, "GiftBox.RandomDelay");
+		RandomRange.Read(parser, pSection, "GiftBox.CellRandomRange");
+		EmptyCell.Read(parser, pSection, "GiftBox.EmptyCell");
+		RandomType.Read(parser, pSection, "GiftBox.RandomType");
+
+	}
```
Metadorius, line 43:
```suggestion
		RandomType.Read(parser, pSection, "GiftBox.RandomType");
	}
```

src/New/Entity/GiftBoxClass.h
```diff
@@ -0,0 +1,153 @@
...
+		return !TechnoList.empty() && !Count.empty();
+	}
+
+	GiftBoxData() noexcept :
+		TechnoList(),
+		Count(),
+		Remove(true),
+		Destroy(false),
+		Delay(0),
+		//min(X),max(Y)
```
Metadorius, line 56:
```suggestion
```

src/New/Entity/GiftBoxClass.h
```diff
@@ -0,0 +1,153 @@
...
+		DelayMinMax({ 0,0 }),
```
Metadorius, line 57:
```suggestion
		DelayMinMax({ 0, 0 }),

```

src/New/Entity/GiftBoxClass.h
```diff
@@ -0,0 +1,153 @@
...
+
+struct GiftBoxData
+{
+public:
+
+	ValueableVector<TechnoTypeClass*> TechnoList;
+	ValueableVector<int> Count;
+	Valueable<bool> Remove;
+	Valueable<bool> Destroy;
+	Valueable<int> Delay;
```
Metadorius, line 22: Why don't use a timer?

ChrisLv-CN:
`GiftBox` is a duplicate of Kratos, and I've received reports that Kratos' `GiftBox.RandomDelay` doesn't work, probably because of a conflict. Maybe need a new name.

PR #345, opened 2021-08-23
WIP Score counter, Score superweapon

MortonPL:
**Note: this is still quite WIP. Some extra things are unfinished, documentation incomplete and bugs may appear. PR ineligible for public testing.**

Implements a customizable score counter, SuperWeapon granting via score and possibly customizable text alignment for strings.

**Copy of documentation on SW granting:**

- It is now possible to award specific countries with single-use Super Weapons for reaching a defined milestone in game score.
   - `ScoreSuperWeaponN.Type` specifies the Super Weapon type that will be granted. This tag is mandatory.
   - `ScoreSuperWeaponN.Score` sets the amount of score players of this country have to earn to receive aforementioned Super Weapon. This tag defaults to 0, meaning that the Super Weapon will be instantly awarded.

In `rulesmd.ini`:
```ini
[SOMECOUNTRY]                                                  ; Country
ScoreSuperWeaponN.Type=SOMESWTYPE          ; SuperWeaponType
ScoreSuperWeaponN.Score=0                              ; integer
; where N = 0, 1, 2, ...
```

**Copy of documentation on score counter:**

- This opt-in counter displayes the in-game score (value of killed units minus value of dead units) for the player.
- The counter is displayed with the format of `(Points)Label`. The label is `★ U+2605` by default.
- You can adjust counter position by `Sidebar.ScoreCounter.Offset`, negative means left/up, positive means right/down.
- You can supply your own color through `Sidebar.ScoreCounter.Color`, in RGB values. The color is white (`255,255,255`) by default.
- You can specify text alignment with `Sidebar.ScoreCounter.Align`, accepted values are `Left`, `Right`, `Center`/`Centre`. Defaults to `Left`.
- You can also change the default position of the counter from the top of the sidebar to the command bar with `Sidebar.ScoreCounter.DrawOnCommandBar`. Defaults to `no`.

In `uimd.ini`:
```ini
[Sidebar]
ScoreCounter.Show=no						  ; boolean
ScoreCounter.Label=<none>					  ; CSF entry key
```

In `rulesmd.ini`:
```ini
[SOMESIDE]                                     ; Side
Sidebar.ScoreCounter.Offset=0,0                ; X,Y, pixels relative to default
Sidebar.ScoreCounter.Color=255,255,255		   ; R,G,B
Sidebar.ScoreCounter.Align=Left				   ; Left, Right, Center/Centre
Sidebar.ScoreCounter.DrawAtCommandBar=no		   ; boolean
```

***Note***: For the label to work, you need to download the improved font (v5 and higher) or draw your own icons.


ayylmaoRotE:
ScoreCounter.Label=<none> is borked, won't hide it and shows up as an "X"
Sidebar.ScoreCounter.Offset=0,0  would be good to get this somewhere outside of the top bar, it's very clogged up as is

ayylmaoRotE:
1. if a SW is granted by a building, then by points, if you remove the building the SW disappears and only becomes available via said building
2. they do not respect SW.NegBuildings. as in points take precedence over prerequisites
3. they also ignore SW.AuxBuildings and AuxBuilding

FS-21:
Suggestion:
What about new tags for using score points like consumable points each time is used a SW? By using 1 time a SW the needed points will get substracted to the total points count & the SW that doesn't have the needed consumable points should disappear from the sidebar until you reach again the needed points.

PR #352, opened 2021-08-29
Techno attachment logic

Metadorius:
- Technos now can be attached one to another in a tree like way. The attached units won't process any locomotion code and act like a part of a parent unit in a configurable.
  - Currently the attached techno may only be a vehicle.
  - When attached, the special `Attachment` (`{C5D54B98-8C98-4275-8CE4-EF75CB0CBE3E}`) locomotor is automatically casted on a unit. You may also specify it in the child unit types manually if the unit is not intended to move without a parent (f. ex. a turret).

In `rulesmd.ini`:
```ini
[AttachmentTypes]
0=MNT                                 ; (example)

[MNT]
RespawnAtCreation=true                ; boolean
RespawnDelay=-1                       ; integer, non-negative values enable the respawn timer
InheritOwner=true                     ; boolean, whether the child inherits owner of the parent while it's attached
InheritStateEffects=true              ; boolean (state effects = chaos, iron curtain etc.)
InheritCommands=true                  ; boolean
InheritCommands.StopCommand=true      ; boolean
InheritCommands.DeployCommand=true    ; boolean
LowSelectionPriority=true             ; boolean, whether the child is low priority while attached
YSortPosition=default                 ; Attachment YSort position enumeration - default|underparent|overparent
InheritDestruction=true               ; boolean
InheritHeightStatus=true              ; boolean, whether the layer and InAir/OnGround/IsSurfaced inherited from parent
OccupiesCell=true                     ; boolean
DestructionWeapon.Child=              ; WeaponType, detonated on child when parent is destroyed
DestructionWeapon.Parent=             ; WeaponType, detonated on parent when child is destroyed
ParentDestructionMission=             ; MissionType, queued to child when parent is destroyed
ParentDetachmentMission=              ; MissionType, queued to child when it's detached from parent

[SOMETECHNO]                          ; TechnoTypeClass
; used when this techno is attached
AttachmentTopLayerMinHeight=          ; integer
AttachmentUndergroundLayerMaxHeight=  ; integer
; used for attaching other technos
AttachmentX.Type=MNT                  ; AttachmentType (example)
AttachmentX.TechnoType=               ; TechnoType that can be attached, currently only units are supported
AttachmentX.FLH=0,0,0                 ; integer - Forward, Lateral, Height
AttachmentX.IsOnTurret=false          ; boolean
AttachmentX.RotationAdjust=0          ; rotation in DirType, from -255 to 255

[General]
AttachmentTopLayerMinHeight=500           ; integer,
AttachmentUndergroundLayerMaxHeight=-256  ; integer
```

- The semantic locomotor alias feature is also supported for Phobos locomotors.

| *Alias*  | *CLSID*                                  |
| -------: | :--------------------------------------: |
Attachment | `{C5D54B98-8C98-4275-8CE4-EF75CB0CBE3E}` |

TODO:
- [x] Handling different locomotors being attached
  - [x] TunnelLoco currently leaves the initial place as if it was left there, crashes the game on destruction
  - [x] Jumpjets stay invincible when attached
  - [x] Review the possibility of adding a custom attachment locomotor instead of a locomotor + skipping Loco process code and changing to it and back
  - [ ] *loosely related - turret is rendered wrong for TunnelLocoClass when going up*
- [x] Sync up the phasing in state for Teleport locomotor
- [x] IsOnTurret is broken visually
- [x] Shadows need to be fixed too
- [x] Latest build crashes on release/devbuild but not on debug when children is yeeted, most likely some piggybacking refcount screwery
- [x] Jumpjets crash improperly when they are crashable=yes and destroyed with children destruction
- [ ] Handle placing the detached children into the correct place (they are not centered at the cell or the correct position for inf atm)
- [ ] Add a flag to force same mission and target to attachments, shouldn't be mixed with commandability
- [ ] Make attached units untargetable and non-damageable using regular weapons via a tag
- [ ] Check AI shenanigans
- [x] Make attached units unselectable (transparent to selection) via a tag
- [x] Make attached units be transparent for pathfinding and building placement but still targetable and damageable
- [x] Handle movement command issued with command inheritance enabled (i.e. optionally reset targets like vanilla units do). possibly just reset the mission when movement command is given and recursed into children instead of doing nothing for that command
- [ ] The source of destruction on parent *unit* specifically appears to be null for cargo (and consequently the children), see 0x7384AB which later calls Kill_Cargo. Needs context and/or earlier Kill_Cargo/On_Destroyed call
- [x] Rotation adjustment tag
- [x] Non-omnifire units without turret rotate and snap back even though the code, test if still happens
- [x] Units now could clip into each other for some reason sometimes, need to review the passability hook again to see what changed
- [ ] When ordering a vehicle with attachments that offset the "center of mass" to move arrives at offset cell
- [ ] ConsideredAircraft still crashes when combined with AttachmentLocomotion, something related to AircraftTracker
- [x] Attachment loco does not provide sensors or sight currently
- [ ] Sort out speed=0 while being attached interaction (may just block speed=0 features if attached) and review similar hooks with Techno Attachment
- [ ] Investigate Per_Cell_Process for useful stuff for TA
- [ ] Check interaction with cloak inheritance and other state effect inheritance
- [x] Improve scatter when something tries to enter the cell on cell-occupying attached technos to make the parent scatter instead
- [ ] Weird bug when cell stays semi-occupied after the children gets destoyed with parent on that cell? might be related to RotE's husks
- [x] Implement handling of deploy
- [x] Implement handling of conversion
  - #1259 to be reviewed and worked upon
- [x] Implement building upgrades support
- [ ] Implement children appearing/disappearing on buildup end (or selldown start)
- [x] Fix bugged loading of unrelated units into transport when multiple units are on cell
- [ ] `AttachmentX.IsOnBarrel` (?)
- [x] Units still clip into each other sometimes (unconfirmed)
- [ ] Shadow scale on children of jumpjets is incorrect
- [x] Units with attachment at 0, 0, 0 can't exit War Factory once again for some reason despite the handling code existing and working previously

Known issues:
- The attachment jitter when they are offset from the 0,0,0 point. This is most likely a precision issue that is impossible to fix without a lot of code to draw the voxel relative to the main point and canvas enlargement. Prototype code exists but commented out and a bit outdated due to locomotors, it needs further work.
- Shadows currently don't behave like parts of the same unit (aka being on one single plane). If that's undesirable - disable children shadows and make shadows via main unit VXL (maybe shadow section hack with transparent voxels).
- Sight may be hacky as it abuses the in-air sight timer, might need testing, especially with height changes.


Metadorius:
Known issues:
- detached children are offset relative to cell center
- force detachment doesn't work / improper child locomotor behavior
- children turrets always face the same absolute direction

Metadorius:
Q: How do I make turrets that don't act like a separate unit? (can't be targeted, destroyed, don't have health)
A: Use `HealthBar.Hide`, `Selectable` and a custom armor type with some HP that is destroyed with `DestructionWeapon.Child`.

Metadorius:
TODO:
- sort out the damage/exp sync system (taking suggestions on this)
- child creation at certain health?
- proper state effects sync if it doesn't work
- sort out mission/order inheritance
- custom selection graphics?
etc.

Taking suggestions on logics, tags, hacks etc. for this.

Metadorius:
`<none>` on `*Weapon`s doesn't work.

Talafhah1:
chrono:
on parent: works but only children in the sw range actually teleport with the parent, rest get stuck in limbo; they die once the parent dies or is sent to a grinder
on child: doenst have any effect

ic:
on parent: works; affects all children too
on child: no effect

chaos: 
on parent: colours all children but only parents and children in range of the weap are affected, sometimes children who are really far away (3k leptons) are affected too even tho the weap only has a range and cell spread of 1
on children: only affects children in range of the weap, they dont get coloured tho, and they seem to target themselves

temporal weaps:
parent: targeting affects all children, stopping the attack makes them stuck in place, if u target them with a temporal weap then stop the go back to being attached like normally, if u kill the parent the game crashes
child: gets stuck in place, stopping the atk returns the child to the parent like youd expect

4SGX:
BuildingTypes have several bugs.

Attached BuildingTypes keep their anims when removed from the map if deployed and BuildingTypes attached to a movable techno will be placed at the position the parent was created until the parent techno deploys or is destroyed. Chronosphere-ing a unit with attached BuildingTypes will also rarely create impassible/unbuildable terrain when the parent deploys/is destroyed.
[Video](https://cdn.discordapp.com/attachments/96269045334032384/881612051171127316/Video.webm)

Having a child techno be important to the game ending (e.g. being the last building) the game will simply not end when destroyed.

4SGX:
BuildingTypes have several bugs.

Attached BuildingTypes keep their anims when removed from the map if deployed and BuildingTypes attached to a movable techno will be placed at the position the parent was created until the parent techno deploys or is destroyed. Chronosphere-ing a unit with attached BuildingTypes will also rarely create impassible/unbuildable terrain when the parent deploys/is destroyed.
[Video](https://cdn.discordapp.com/attachments/96269045334032384/881612051171127316/Video.webm)

Having a child techno be important to the game ending (e.g. being the last building) the game will simply not end when destroyed.

gangster-ppm:
Attach passanger to a projectiles:
+ Magnetron pulls its target, absorbs, then launches it away with inviso V3
+ RA3 Bullfrogs

4SGX:
Crash on having attached base destroyed and continuing-despite-short-game-on bug I mentioned earlier then creating a new attached building. Code, video, and snapshot in link below.
https://discord.com/channels/338055131666972683/773636942775582720/883466977761435668

FS-21:
Wish:
Can all units (master & attached) optionally share Ammo=?

Metadorius:
Added initial docs and stop command handling for all subtechnos.

Metadorius:
- Fixed rotation of attached technos 
- Fixed scatter crash
- Fixed Locomotor warheads screwing up attachments

a851903106:
I'm sorry to bother you.

I noticed that the technical attachment seems to crash and exit the game when the game reads the file, but it doesn't show any text prompts.

Metadorius:
Are you trying the save/load (that may be broken still), or just reading from the INI file? If the latter, then please show me how did you configure the attachment.

ststl-s:
Swizzle `TechnoClass*`, `TechnoTypeClass*` will be `nullptr`. And.. `AttachmentData*` is a pointer to `TechnoExt::ExtData`'s member, I'm sure there won't call `SwizzleManagerClass::Here_I_Am`, after swizzle it will be `nullptr` too.
As add `Debug::Log`, I find after loaded, it crashed in `AttachmentClass::AI`,  most members of `this` are nullptr.
I solve this problem like pr#658, load `ChildAttachments` and `ParentAttachment` after Ares, and add this hook.
```cpp
DEFINE_HOOK(0x6CF316, SwizzleManagerClass_Here_I_Am, 0x4)
{
	GET_STACK(void*, oldP, 0x8);
	GET_STACK(void*, newP, 0xC);
	PointerMapper::AddMapping(oldP, newP);
	return 0;
}
```
`PointerMapper` avoid ares clear mapping.

4SGX:
JumpjetTurn=no makes attached units move at random and attached units can create an invincibility cheat by being fired at by the unit they are attached to. Respectively: 
image image
Units attached to a Subterranean unit are still visible underground.
image

Metadorius:
TODO:
- Unbreak the drawing relative to parent's origin, currently a mess (check out earlier commits if you work on sth)
- Draw order specification relative to parent
- Handle movement command issued with command inheritance enabled (i.e. optionally reset targets like vanilla units do). possibly just reset the mission when movement command is given and recursed into children instead of doing nothing for that command
- If you have command inheritance, order to attack something (all children with inheritance will also get an order to attack), the waypoints will merge, and when you place a next waypoint to move - the child technos will try to execute that waypointed movement order constantly and ignore all subsequent waypoints. this needs to be handled somehow, either by skipping the movement if such occurs or by not merging the waypoints somehow
- Sync locomotor attacked status (not selectable, does not attack etc)
- Sync subterranean stuff (not being able to fire, not drawn, detach all etc)

NetsuNegi:
Right now, the child won't have any lag, great job!

Metadorius:
image

Metadorius:
TODO based on today's testing
- fix improper handling of TunnelLoco, it currently occupies tunnelling sequence start cell and crashes the game when the unit is destroyed
- physical position of a techno lags 1 frame behind, might be due to an order of draw/real stuff
- make attached units lose target on moving like turrets
- make AddSelectable work with real location as well, currently bugged

Metadorius:
selectables are fixed
new TODO: for some reason getting a crash when releasing a locomotor in Approach_Target (attack mission)

Metadorius:
More test are always welcome btw.

brsajo:
Pathfinding crash still happens by the way.

ststl-s:
image
This will be called when techno cloaked.

mevitar:
I don't know if it happens with other types of vehicles, but attaching a jumpjet to a ground vehicle and moving the parent makes the attachment invulnerable.

I wanted to check if it's possible to make a unit like the Colossus from SC2 (something that can be attacked by both anti-ground and anti-air) so i attached a jumpjet to a ground vehicle, then made ground anti-air attack the attachment. It all worked fine as long as the parent stayed immobile, but once i moved the parent outside of their range and they started to chase to be able to fire, their projectiles stopped dealing any damage to the attachment. At that point it didn't matter if the parent was immobile or moving, the attachment wasn't taking any damage from their attacks anymore.

Metadorius:
Oh that actually explains a bug Dmitry Volkov has shown me once and I couldn't reproduce. That means that I didn't handle jumpjet loco properly too (yet). Thanks for the report!

ststl-s:
In `Ext/Techno/Hoos.TechnoAttachment.cpp`
This pr and kratos conflict at `StopCommand_Context_Unset`  (crash when player send stop command)
```
DEFINE_HOOK(0x730F1C, StopCommand_Context_Unset, 0x5)
{
	TechnoAttachmentTemp::stopPressed = false;
	return 0;
}
``` 
Use this address, there will be no conflict.

Metadorius:
@ststl-s thank you, I will account for that when I work on it next time 👍

mevitar:
Did you change something relating to airstrikes or CanC4 code? Trying to airstrike a building crashes the game (to be precise, holding Ctrl and merely hovering the mouse button over my own building crashes the game).

The crash does not happen for buildings with CanC4=no or AllowAirstrike=no. Neither does it happen in the newest dev build.

Starkku:
> Did you change something relating to airstrikes or CanC4 code? Trying to airstrike a building crashes the game (to be precise, holding Ctrl and merely hovering the mouse button over my own building crashes the game).
> 
> The crash does not happen for buildings with CanC4=no or AllowAirstrike=no. Neither does it happen in the newest dev build.

Bug with AllowAirstrike hooks that has been fixed in develop branch now.

Metadorius:
ok yet another TODO with current things to finish
- [ ] Handling non-drive and non-ship locomotors being attached (is it even needed? except for TunnelLocomotor)
  - [ ] TunnelLoco currently leaves the initial place as if it was left there, crashes the game on destruction
  - [ ] Jumpjets stay invincible when attached
  - [ ] Review the possibility of adding a custom attachment locomotor instead of a locomotor + skipping Loco process code and changing to it and back
- [ ] Handle movement command issued with command inheritance enabled (i.e. optionally reset targets like vanilla units do). possibly just reset the mission when movement command is given and recursed into children instead of doing nothing for that command
- [ ] Add a flag to force same mission and target to attachments (piss easy, probably 6 lines in total), shouldn't be mixed with commandability
- [ ] Make attached units be transparent for pathfinding and building placement but still targetable and damageable

More distant TODO:
- [ ] Make attached units untargetable and non-damageable using regular weapons via a tag
- [ ] Make attached units unselectable (transparent to selection) via a tag

SCIPCION:
its working fine. 
1. Attached units doesnt share the same mission as the parent, this could be a problem especially with preplaced units.
2. There is way to make the attached unit as a "virtual unit"?, to make it work as a turret/multiturrets.
3. Attached to aircrafts is working fine, just the shadow dissapear over BibShapes

image

https://github.com/Phobos-developers/Phobos/assets/12478138/50501fb7-195e-4d8c-b025-05325d701075


https://github.com/Phobos-developers/Phobos/assets/12478138/8ae556fb-246b-4117-af2e-db713e50bd4f

Metadorius:
Ok so the new build now has a custom locomotor `{C5D54B98-8C98-4275-8CE4-EF75CB0CBE3E}` (or `Attachment` semantic loco name alias). The new behavior is as follows: if the attached unit has another locomotor - the attachment loco will be automatically casted onto it and automatically removed once the attachment is done.

Another TODO:
- [ ] Investigate why when destroying a jumpjet loco parent a random crash appears. Reproduction: make an attachment type, add it to a jumpjet unit, move the jumpjet at least once in the game, destroy the parent (currently InheritDestruction is temporarily always on).

Metadorius:
The outstanding locomotor issues appear to be fixed, I wasn't able to get the game to crash on basic tests. Pathfinding freezes also seem to be gone. I also updated the doc to reflect the current state of things and adjusted the available tags. Need more testing in different scenarios.

Metadorius:
The first post is updated with actual docs btw.

Starkku:
I get a crash when a parent unit or attachment is destroyed with following code (changing the owner or attached techno makes no difference.

```ini
[PARENT]
Attachment0.Type=ATTACHMENT
Attachment0.TechnoType=CHILD
Attachment0.FLH=150,125,230
Attachment0.IsOnTurret=true

[AttachmentTypes]
0=ATTACHMENT

[ATTACHMENT]
InheritOwner=true
InheritStateEffects=tru
InheritCommands=true
LowSelectionPriority=true
YSortPosition=overparent
InheritDestruction=true
```

Also what exactly is currently blocking targeting of attachments? They can still be damaged and selected (I am aware this is on the to-do-list) but I figure ideally all of this behaviour should be customizable. That said I can imagine there being some.. interesting cases with things like Temporal and Parasite against attachments.

Logs and dump for crash:
[logs.zip](https://github.com/Phobos-developers/Phobos/files/11859485/logs.zip)
https://mega.nz/file/puAFyKwJ#t7X8Lf-8HFKSqS41WVRN6ZQ6yEYiIraJlpN7L_pe3eQ

NetsuNegi:
> 

I got crash at last code when destroy parent unit too. I found game will not crash if don't use Attachment locomotor. Maybe something wrong with this loco?

NetsuNegi:
> I get a crash when a parent unit or attachment is destroyed with following code (changing the owner or attached techno makes no difference. If it changes things, I did only test this with preplaced objects so far.
> 
> ```ini
> [PARENT]
> Attachment0.Type=ATTACHMENT
> Attachment0.TechnoType=CHILD
> Attachment0.FLH=150,125,230
> Attachment0.IsOnTurret=true
> 
> [AttachmentTypes]
> 0=ATTACHMENT
> 
> [ATTACHMENT]
> InheritOwner=true
> InheritStateEffects=tru
> InheritCommands=true
> LowSelectionPriority=true
> YSortPosition=overparent
> InheritDestruction=true
> ```
> 
> Also what exactly is currently blocking targeting of attachments? They can still be damaged and selected (I am aware this is on the to-do-list) but I figure ideally all of this behaviour should be customizable. That said I can imagine there being some.. interesting cases with things like Temporal and Parasite against attachments.
> 
> Logs and dump for crash: [logs.zip](https://github.com/Phobos-developers/Phobos/files/11859485/logs.zip) https://mega.nz/file/puAFyKwJ#t7X8Lf-8HFKSqS41WVRN6ZQ6yEYiIraJlpN7L_pe3eQ

I got crash at last code when destroy parent unit too. I found game will not crash if don't use Attachment locomotor. Maybe something wrong with this loco?

mevitar:
I am also getting the crash-on-death, and i'm testing the newest build. Happens whether i kill the parent or the child.
If i set InheritDestruction=no, the crash is gone if the parent is killed, but still happens whenever the child is killed.

Metadorius:
For some reason latest build is crashing only in non-debug build, I'll have to debug that a bit later.

Metadorius:
Latest debug build for testing: https://cdn.discordapp.com/attachments/514505963412258831/1122876959345623061/Phobos.dll

Metadorius:
Once again, latest debug build: https://cdn.discordapp.com/attachments/773636942775582720/1122991166032465930/Phobos.dll

mevitar:
If the parent is destroyed, the child will still occupy the cell that the parent was on, and ordering units to attack that cell will make them attack the child. This was tested with Attachment0.FLH=0,0,512.

Otherwise, everything seems to be working fine. CellSpread now damages air addons when hitting the parent. I also noticed that the parent unit will even run away if its child attachment is attacked by an AircraftType, as if it was itself attacked by same aircraft.

Metadorius:
Thanks for the testing!

Hmm, well it's a bit weird in a sense because the game seems to have separate checks of whether the unit is in air or not, and the layer/airborne tracking checks. Currently there is absolutely no special handling for "in air" checks, only the layer stuff is handled on the locomotor level, so if the parent locomotor is considered airborne - so is the attachment loco, but apparently there is a need to handle the targeting checks separately. I will keep that in mind for later, for now I think a sufficient workaround would be to use TurretOffset where possible, want to fix the major issues first.

Metadorius:
- Fixed crash-on-destruction
- Fixed `Crashable=yes` jumpjets being stuck on death (probably potentially not only here)
- Fixed crashable units not disconnecting from the parent
- Fixed the missing source for crashable=yes handler function (bug?)
- Implemented `AttachmentX.RotationAdjust`

Metadorius:
@Starkku btw not sure if after your amendments in order to fix move command inheritance, but guard or area guard on jumpjets at least were making the units shoot in the ground instead.

brsajo:
> Pathfinding freeze still happens by the way. 

Haven't been able to reproduce this after using the latest nightly build.

One thing I'd like to see that might be out of the scope of this PR, is an option to hide the body of a unit so only the turret is visible for the purpose of things like turret attachments.
Right now I'm using an empty vxl file as the body for turrets but it seems like a round-about way of doing it if I just want to copy the turrets of existing units.

reedom114514:
We find two bugs here
1 If a passenger carrier (ex. battle fortress) is attached with unit who is not jumpjet, the infantries cannot go into it.


2 The unit A who has an attached unit C that deploys to convert itself to B, A would attached on B with C too.

Metadorius:
- Added `OccupiesCell` tag to control whether the childs make the terrain impassable.
- Fixed bib occupation bug when the childs of the built parent occupied the bib so WF couldn't kick out the unit.

reedom114514:
> We find two bugs here 1 If a passenger carrier (ex. battle fortress) is attached with unit who is not jumpjet, the infantries cannot go into it. 
> 
> 2 The unit A who has an attached unit C that deploys to convert itself to B, A would attached on B with C too. 

In this case I showed the battle fortress which cannot load the infs, now I find the reason: when the attached techno is in the same cell of the BT, infs cannot get into it

Metadorius:
- Added `InheritHeightStatus` and accompanying tags for top/underground min/max height (subject to change) for when the height status isn't inherited, along with custom code mostly replicating jumpjet behavior, also syncs OnBridge status when set to yes. @mevitar pls test
- Improved (un)limbo aircraft tracking

mevitar:
Good news: OccupiesCell works, and if i set it to "no" i can even deploy units with attachments into buildings (there is no memory leak here, right? i hope i'm not doing something that can cause problems).
Bad news: whenever the attachment disappears, the game crashes after around 30 seconds. And by "disappears" i don't mean just the unit->building deploy, i also mean destroying the parent or the attachment itself. Setting OccupiesCell=yes fixes this problem.
And in case you're wondering, if i set attachment FLH to like 256,0,0 then i can also deploy the unit into a building, and there is no crash as long as OccupiesCell is set to yes.
I can even move the parent unit in such way that attachment will overlap with a building, and then kill or deploy the parent (or the building), and there is still no crashes as long as OccupiesCell is set to yes.

In other news, with InheritHeightStatus=yes i am able to make units use anti-ground attacks against attachments for ground parents. Some weapons might miss if the height of the attachment is above the ground, though (for example: Arcing=yes projectiles).

Metadorius:
Interesting, wonder what causes the game to crash. Will investigate.

Think the weapons missing is kind of expected because they assume the attachment is on ground, not sure I can/should do much about it, something worth mentioning in docs though.

NetsuNegi:
DH}SFCU55C38K@M1`C4O05J
Maybe some error in AttachmentLoco, this crash happening almost every game. If I don't use AttachmentLoco, then it won't crash again.

Metadorius:
> DH}SFCU55C38K@M1`C4O05J Maybe some error in AttachmentLoco, this crash happening almost every game. If I don't use AttachmentLoco, then it won't crash again.

What's the specific conditions for the crash to happen?

NetsuNegi:
> > DH}SFCU55C38K@M1`C4O05J Maybe some error in AttachmentLoco, this crash happening almost every game. If I don't use AttachmentLoco, then it won't crash again.
> 
> What's the specific conditions for the crash to happen?

idk, I just choose observer mode, and game always crash about after 30 mins

Metadorius:
Added `RespawnTimer`, `TransparentToMouse` (thanks ChrisLv_CN), `RespawnAtCreation`.

mevitar:
[snapshot-20230719-233441.7z.001.zip](https://github.com/Phobos-developers/Phobos/files/12100443/snapshot-20230719-233441.7z.001.zip)
[snapshot-20230719-233441.7z.002.zip](https://github.com/Phobos-developers/Phobos/files/12100445/snapshot-20230719-233441.7z.002.zip)
[snapshot-20230719-233441.7z.003.zip](https://github.com/Phobos-developers/Phobos/files/12100446/snapshot-20230719-233441.7z.003.zip)
[snapshot-20230719-233441.7z.004.zip](https://github.com/Phobos-developers/Phobos/files/12100447/snapshot-20230719-233441.7z.004.zip)
[snapshot-20230719-233441.7z.005.zip](https://github.com/Phobos-developers/Phobos/files/12100450/snapshot-20230719-233441.7z.005.zip)
Crashdump. Remove .zip extension and then merge with 7zip.

NetsuNegi:
> [snapshot-20230719-233441.7z.001.zip](https://github.com/Phobos-developers/Phobos/files/12100443/snapshot-20230719-233441.7z.001.zip) [snapshot-20230719-233441.7z.002.zip](https://github.com/Phobos-developers/Phobos/files/12100445/snapshot-20230719-233441.7z.002.zip) [snapshot-20230719-233441.7z.003.zip](https://github.com/Phobos-developers/Phobos/files/12100446/snapshot-20230719-233441.7z.003.zip) [snapshot-20230719-233441.7z.004.zip](https://github.com/Phobos-developers/Phobos/files/12100447/snapshot-20230719-233441.7z.004.zip) [snapshot-20230719-233441.7z.005.zip](https://github.com/Phobos-developers/Phobos/files/12100450/snapshot-20230719-233441.7z.005.zip) Crashdump. Remove .zip extension and then merge with 7zip.

still crash ?

mevitar:
Yes, it crashed with the newest build.

mevitar:
Still crashes, same EIP.
However, the attachment had ConsideredAircraft=yes on it, so i commented it out. No longer crashes after that.

mevitar:
Does DestructionWeapon.Child= and DestructionWeapon.Parent= work properly for everyone? I used to be able to kill the child/parent with them, but i cannot do that anymore, nor do they do any damage to surrounding units.

Metadorius:
> Does DestructionWeapon.Child= and DestructionWeapon.Parent= work properly for everyone? I used to be able to kill the child/parent with them, but i cannot do that anymore, nor do they do any damage to surrounding units.

Haven't checked lately, but double check if the weapons have `AffectAllies=no` just to be sure.

mevitar:
Seems like DestructionWeapon.Child= and DestructionWeapon.Parent= are not exactly centered, as they can't damage the child/parent properly without CellSpread=. For comparison, Phobos' FeedbackWeapon= can damage the firer without any issues even without a CellSpread.

mevitar:
No, ignore the comment above. Of course it's not that. Weapons are perfectly aligned.

It's the default DamageSelf=no, because addon and parent are treated as the same unit, so they can't damage themselves with their DestructionWeapon. -_-

mevitar:
> Haven't checked lately, but double check if the weapons have `AffectAllies=no` just to be sure.
Not a single warhead used had AffectsAllies=no, and i tested it with multiple ones, including ones with high CellSpread. It's all because of DamageSelf=no.

src/New/Entity/AttachmentClass.cpp
```diff
@@ -0,0 +1,311 @@
...
+		}
+		else
+		{
+			Debug::Log("[" __FUNCTION__ "] Failed to create child %s of parent %s!\n",
+				pChildType->ID, this->Parent->GetTechnoType()->ID);
+		}
+	}
+}
+
+void AttachmentClass::AI()
```
Metadorius, line 75: Here I am not sure. Pretty sure part of this should be done on AttachmentLocomotion, but currently the logic processing order is built that way that this sync function runs after all instances of technos got their `TechnoClass::AI` called, so that it ensures the latest values are on parent. The `AttachmentLocomotion::Process` call is actually in `FootClass::AI`, so depending on the order of processing different technos there may be 1 frame lag if children are processed before parent.

I see 2 solutions to this:
1. Manually calling parent processing logic before sync and then patch the AI to not run twice on the same frame;
2. Partially sorting parent and children in techno array(s) so that parent is always processed first;
3. Maybe not using the AttachmentLoco process call for sync whatsoever?

Any more ideas on how that should be dealt with?

src/Ext/Techno/Hooks.TechnoAttachment.cpp
Metadorius, line 1: Elephant in the room: type conversion (on parent) handling and deploy transfer.

Possible ways to handle which all need to be implemented:
- no handling, keep as-is;
- convert attachment types and/or attached techno types recursively;
- limbo old attachments, unlimbo/create new attachments, aka cycle between different attachments;
- maybe something else.

Any proposed INI syntax for that which would be intuitive to use? I had something like this in mind:

```ini
[AttachmentTypes]
0=TURRET
1=CONTROLLABLETURRET

[TURRET]
InheritCommands=yes
Selectable=no
TransparentToSelection=yes

[CONTROLLABLETURRET]
InheritCommands=no
Selectable=yes
TransparentToSelection=no


[MOM]
Attachment0.FLH=128,0,0
Attachment0.Type=TURRET
Attachment0.TechnoType=MOMTURRET
Attachment0.ConversionMode=ConvertAttachmentType
Deploy.Convert=MOMDEPL

[MOMDEPL]
Attachment0.FLH=128,0,0
Attachment0.Type=CONTROLLABLETURRET
Attachment0.TechnoType=MOMTURRET
Attachment0.ConversionMode=LimboOldUnlimboNew  ; <-- coding error
Deploy.Convert=MOM

[MOMTURRET]
; ...
```

Please note the coding error (intended to be same between both technotypes), I am not sure how to design INI tags the way where such errors wouldn't be possible.

Metadorius (commented):
@Starkku @MortonPL @chaserli @Belonit @ZivDero specifically and whoever else, need your input.

docs/New-or-Enhanced-Logics.md
```diff
@@ -4,6 +4,57 @@ This page describes all the engine features that are either new and introduced b
...
+ParentDestructionMission=             ; MissionType, queued to child when parent is destroyed
+ParentDetachmentMission=              ; MissionType, queued to child when it's detached from parent
+
+[SOMETECHNO]                          ; TechnoTypeClass
+; used when this techno is attached
+AttachmentTopLayerMinHeight=          ; integer
+AttachmentUndergroundLayerMaxHeight=  ; integer
+; used for attaching other technos
+AttachmentX.Type=MNT                  ; AttachmentType (example)
+AttachmentX.TechnoType=               ; TechnoType that can be attached, currently only units are supported
```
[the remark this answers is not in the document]
Starkku: From the perspective of this semantic distinction that you've made where AttachmentType controls **how** the attachment and parent link together and any controls regarding **what** get attached are on the parent's type, `RespawnAtCreation` and `RespawnDelay` are technically type-agnostic and would still fall under controlling the link itself (in essence it is how it gets restored if severed). However it falls under things that people would likely want to adjust per parent type even in cases where a lot of the other properties are reused. `DestructionWeapon` etc. would likely fall under that same umbrella.

If you want to stick with the semantic distinction you've made, then they stay on the attachment type. If you are willing to relent and consider modder usability, you'll allow defining or overriding atleast the previously mentioned things on the parent's type.

src/New/Entity/AttachmentClass.cpp
Starkku: Either 1. or 3. Way too much room for advanced fuckery with 2.

If the entire idea behind `AttachmentLocomotionClass` is that it takes care of syncing every single movement (and only that, you're not gonna use it to sync other stuff from the parent hopefully) related property from the parent then yes, I'd probably go with the option 1. Otherwise 3. might be more straightforward.

src/Ext/Techno/Hooks.TechnoAttachment.cpp
Starkku: I could see the first three listed options (assuming the second one actually just means destroying any old attachments and replacing with the new ones if any) being actually used at the least, can't think of any others.

Just listing the conversion mode on the parent is probably good enough syntax.

Ain't got a clue what you mean by 'coding error' here though.

Metadorius: Well let's see what would happen. MOM deploys, the child is converted because the conversion mode is to convert children. MOMDEPL deploys again, but the child is hidden and the new one is created. This doesn't sound like a valid behavior to me in any case regardless of an application, because you will just hide the technos endlessly.

src/New/Entity/AttachmentClass.cpp
Metadorius: Not sure what you mean by >too much room for advanced fuckery?

src/Ext/Techno/Hooks.TechnoAttachment.cpp
Starkku: Not many ways to avoid that if the conversion mode is customizable like this. It is pretty much just a skill issue, you can do some pretty damn stupid things in this engine already if you don't know what you're doing. Game doesn't prevent infinite recursion in a lot of circumstances, for an instance. Only one I can think of is enforcing the first type's conversion mode for further conversions for that instance but idk how that would work once the instance is destroyed and/or recreated. I would probably personally just slap a warning in docs about problematic configurations and be done with it.

src/New/Entity/AttachmentClass.cpp
Starkku: Well I was thinking that shifting the order might throw game's assumptions off somewhere but now that I think about it that's probably not the case.

Not that it matters because that's not what you'd need to do anyway, AI calls are done in order the objects have been submitted to the logic layer e.g calls to `0x55BAA0`. Limboing deletes things from it, unlimboing adds them with few exceptions like objects that have unique ID of -2 (the movement ring animation from `MoveFlash` being a chief example) never being submitted and so on. Sorting that one is probably free game but maintaining the correct sort order might fuck with performance, idk.

RexDraco:
Using attached techno with an upgrade causes it to not appear and a "unit lost" message is played:
Nightly Build: 10981467c73405181fbdcd3f91c0951f4c015b25

When the below attachment type is placed directly upon the structure, it works perfectly as intended, but not when used on the upgrade. The repair claw weapon does work, it does repair units when they're nearby, when it is placed on the structure, and it is visible. When it is used on the upgrade instead and the upgrade is placed on the refinery, a message "Unit Lost" is played and no repair claw is present and no sign of an invisible attached techno is present, no repair weapon activates. 


`
[AttachmentTypes]
0=SMGUNP

[VehicleTypes]
...
WPAT04=REPAIRCLAWU

[BuildingTypes]
....
WPUP33=GAREFNREPAIRUP

[SMGUNP]
RespawnAtCreation=true                ; boolean
RespawnDelay=-1                       ; integer, non-negative values enable the respawn timer
InheritOwner=true                     ; boolean, whether the child inherits owner of the parent while it's attached
InheritStateEffects=true              ; boolean (state effects = chaos, iron curtain etc.)
InheritCommands=true                  ; boolean
LowSelectionPriority=true             ; boolean, whether the child is low priority while attached
YSortPosition=overparent;default                 ; Attachment YSort position enumeration - default|underparent|overparent
InheritDestruction=true               ; boolean
InheritHeightStatus=true              ; boolean, whether the layer and InAir/OnGround/IsSurfaced inherited from parent
OccupiesCell=false                     ; boolean
DestructionWeapon.Child=attachmentDelete

[GAREFN]
...
Upgrades=1

[GAREFNREPAIRUP]
UIName=NOSTR:Refinery Repair Upgrade
Name=Refinery Repair Upgrade
Image=INVISOX4x3
BuildCat=combat
Prerequisite=GAREFNPRE;,RADAR,FRENCHTHREEPR
Armor=wood
TechLevel=11
Sight=4
Owner=British,French,Germans,Americans,Alliance,Canada,Japan
RequiredHouses=French
AIBasePlanningSide=0
Cost=3;000
Points=10
Capturable=no
Crewed=no
Explosion=TWLT070,S_BANG48,S_BRNL58,S_CLSN58,S_TUMU60
DebrisAnims=DBRIS1LG,DBRIS1SM,DBRIS4LG,DBRIS4SM,DBRIS5LG,DBRIS5SM
MaxDebris=3
MinDebris=2
PowersUpBuilding=GAREFN
PowersUpToLevel=-1
ThreatPosed=0

Attachment0.Type=SMGUNP
Attachment0.FLH=0,0,15
Attachment0.IsOnTurret=no          ; boolean
Attachment0.RotationAdjust=0          ; rotation in DirType, from -255 to 255
Attachment0.TechnoType=REPAIRCLAWU`

Let me know if there's anything else I need to report.

src/New/Entity/AttachmentClass.cpp
Metadorius: @Starkku so what do you think about the idea of resubmission to the logic layer in the correct order in context of discussion then?

Metadorius:
- Implemented sight and sensors support. Sight may be hacky as it abuses the in-air sight timer, might need testing, especially with height changes.
- Improved scatter when something tries to enter the cell on cell-occupying attached technos to make the parent scatter instead.
- Found a weird bug when cell stays semi-occupied after the children gets destoyed with parent on that cell, although this might be related to RotE's husks and not this feature. Needs verification.

src/New/Entity/AttachmentClass.cpp
Metadorius: @Starkku bump

Starkku: Might work, not sure if it adds to performance overhead in any meaningful way but ultimately this is the kind of thing that'd you'd just have to try out.

Kn-geSponge:
The collisions are a little broken, they need to be fixed.
I’m posting a video to make it clearer.
-!!MR..NK

Uploading vid.mp4…

Kn-geSponge:
The collisions are a little broken, they need to be fixed.
I’m posting a video to make it clearer.
-!!MR..NK

https://github.com/Phobos-developers/Phobos/assets/86251338/4ace9714-c3f2-4544-bd2c-148ac28ff675

Kn-geSponge:
[compiled-dll-57594129a99096552a449eace1ffdf3527bd27ac.zip](https://github.com/Phobos-developers/Phobos/files/13808239/compiled-dll-57594129a99096552a449eace1ffdf3527bd27ac.zip)


> Nightly build for this pull request:
> 
>     * [compiled-dll-57594129a99096552a449eace1ffdf3527bd27ac.zip](https://nightly.link/Phobos-developers/Phobos/actions/artifacts/913163876.zip)
>       These artifacts will expire in 90 days and will not be available for download after that time.
> 
> 
> _This comment is automatic and is meant to allow guests to get latest nightly builds for this pull request without registering. It is updated on every successful build._

Metadorius:
Transports appear to be fixed. When the build finishes please test if anything that is linked with loading up passengers (initial passengers, reinforcements w/ paradrop etc).

mevitar:
> The collisions are a little broken, they need to be fixed. I’m posting a video to make it clearer.
> 

I just had an instance of this happening in the newest build, except for the "walker" locomotor. I'll see if i can trigger it again.

mevitar:
Yup, it definitely happens. Can't really get it to trigger when just moving units, but if they are moving and then they try to attack, then they might walk right on top of eachother trying to get in attack range.

mevitar:
> I just had an instance of this happening in the newest build, except for the "walker" locomotor. I'll see if i can trigger it again.

Good news, this might not be a Phobos or this build issue, as i managed to get it with just Ares alone. Bad news, that probably means that the walker locomotor is unsafe to use...

Metadorius:
- Infantry not being able to enter cells that contain multiple objects have been fixed. I am not sure if the implementation has side effects though; PLEASE test whether infantry can still enter transports, just vehicles (pilots/hijackers), buildings correctly like before *and* when there are multiple units on the cell (can be achieved via attachment).
- Fixed casemate units still being able to erratically turn when attached, credits to @ZivDero.
- Fixed residual occupations caused by an improperly ported Ares fix that I had to touch, ty @ZivDero for help.

SCIPCION:
first vehicle hasnt any TA.
Second one uses TA, cant leave the factory

https://github.com/user-attachments/assets/94b5e6e9-4959-4c48-93a6-37b646c6da8d

SCIPCION:
After months constantly using TA and making one of my factions totally dependent of TA, i need to ask for some requested features:

-If child techno die also parent will die, the child object as the main unit is usefull in many scenarios.
-Child and parent sharing Streng.
-TA doesnt follow parent IsTilted=no
-VirtualUnit=yes this is a kratos feature, making the child completely invisible or not interactive is not possible only with vanilla codes
-Child and parent needs to share missions, or at least sleep/stop missions

Metadorius:
@SCIPCION
@ayylmaoRotE figured out the TA unit getting stuck in factory bug, it now works fine

Metadorius:
> Still crashes, same EIP. However, the attachment had ConsideredAircraft=yes on it, so i commented it out. No longer crashes after that.

@mevitar could you repro that again on the latest build? @ayylmaoRotE could not repro

DeathFishAtEase:
Regarding `YSortPosition`: The current design does not seem to consider the occlusion relationships between multiple AttachmentUnits.

A user tested with the version in [issuecomment-3613072028](https://github.com/Phobos-developers/Phobos/pull/352#issuecomment-3613072028) (the immediately previous) and encountered a problem where when AttachmentUnit and ParentUnit overlap (`AttachmentX.FLH=0,0,0`), AttachmentUnit will be prioritized for target search while ignoring ParentUnit. This led to him setting an armor on the AttachmentUnit that cannot be attacked (`Versus=0%`) to use it as a virtual additional weapon, allowing the ParentUnit to swagger into the enemy base without any hindrance.

DeathFishAtEase:
> Unless the first post has been updated after this, it looks like it's a known issue.
> 
> > TODO:
> > Add a flag to force same mission and target to attachments, shouldn't be mixed with commandability

Well, I misunderstood it as something similar to `OpenTopped.AllowFiringIfDeactivated`.

momokama:
BUG:
when [parent] convert to other [techtypes],[DestructionWeapon.Child] will be triggered in an unexpected way.

src/Ext/Techno/Body.h
```diff
@@ -103,6 +105,7 @@ class TechnoExt
 
 		AttachmentClass* ParentAttachment;
 		ValueableVector<std::unique_ptr<AttachmentClass>> ChildAttachments;
+		std::map<int, ValueableVector<std::unique_ptr<AttachmentClass>>> DormantAttachments;
```
NetsuNegi, line 111: I think here can use `std::vector` instead of `ValueableVector`

Metadorius:
@momokama try if the latest build changes anything regarding your issue

DeathFishAtEase:
The root cause of the error is not yet clear. After merging the `develop` branch (the above build `compiled-dll-89c25d2c09bf8543ad013b6d9129ee0d99a22dc4.zip`), units with a larger `TurretOffset` will periodically twitch in a specific direction perpendicular to their orientation on the horizontal plane. This is not limited to Child Units, but it could not be reproduced on the latest Nightly (c0d59cd8857632350617d5bb629e066b0142a2bb) of the `develop` branch.
5a02ff

SethGekco:
**Bug: null dereference in `AttachmentClass::AI()` after `Unlimbo()`**

Repo: Phobos, PR #352 (`feature/techno-attachment`), file `src/New/Entity/AttachmentClass.cpp`.

**Crash**

`C0000005`, `READ at 0x00000081`. Offset `0x81` is `ObjectClass::InLimbo`. Ten identical dumps, same EIP.

**Cause**

`AI()` null-checks `this->Child`, then calls `this->Unlimbo()`, then dereferences `this->Child` again without re-checking:

```cpp
if (this->Child)
{
    if (this->Child->InLimbo && !this->Parent->InLimbo)
        this->Unlimbo();
    else if (!this->Child->InLimbo && this->Parent->InLimbo)
        this->Limbo();

    this->Child->SetLocation(this->GetChildLocation());   // <-- Child may be null here
```

`AttachmentClass::Unlimbo()` calls the engine's `TechnoClass::Unlimbo(coord, dir)`. A failed placement is destructive — the engine destroys the object rather than leaving it partly placed. Destruction runs the PR's own hook, which reaches `AttachmentClass::ChildDestroyed()`:

```cpp
void AttachmentClass::ChildDestroyed()
{
    if (this->Child)
    {
        ...
        this->Child = nullptr;   // clears the member re-entrantly, inside Unlimbo()
    }
}
```

So `this->Child` is nulled during the `Unlimbo()` call. The guard at the top of the block is stale on return, and the next dereference faults.

**Trigger conditions**

Anything that changes cell occupancy near the host so the child's placement fails — a nearby unit dying, a structure being sold, combat in the adjacent cells. The visible trigger is one causal step removed from the fault, so surface details differ between dumps while the EIP is identical.

**Fix**

Re-check the member after the calls that can destroy it:

```cpp
if (this->Child->InLimbo && !this->Parent->InLimbo)
    this->Unlimbo();
else if (!this->Child->InLimbo && this->Parent->InLimbo)
    this->Limbo();

if (!this->Child)
    return;

this->Child->SetLocation(this->GetChildLocation());
```

**Related, lower confidence**

`AttachmentClass::InvalidatePointer()` calls `AnnounceInvalidPointer(this->Parent, ptr)`, so `Parent` is nullable by design. `AI()` dereferences `Parent` roughly sixteen times with no null check (first at `this->Parent->InLimbo`). Not observed in any dump, but the same class of defect; an early `if (!this->Parent) return;` in `AI()` closes it.

**Verification status**

Confirmed in the PR's source as fetched from the live branch. Reproduced ten times in a standalone port of the PR, not on upstream Phobos itself.

PR #359, opened 2021-09-07
New Action 112 - Regroup around the Team Leader

FS-21:
Script action 74,n puts the TaskForce into Area Guard Mode for the given units of time around the Team Leader (this unit remains almost immobile until the action ends). 
The default guard area radius around the Leader is `[General] > CloseEnough` and the units will not leave that area. 
The countdown only works while all the units are inside the area.

Already tested inside PR 296 but I created it as a separate PR

### DIFF
--- a/Phobos.vcxproj
+++ b/Phobos.vcxproj
@@ -98,6 +98,7 @@
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="src\Commands\NextIdleHarvester.h" />
+    <ClInclude Include="src\Commands\NextTeam.h" />
     <ClInclude Include="src\Commands\ObjectInfo.h" />
     <ClInclude Include="src\Commands\Commands.h" />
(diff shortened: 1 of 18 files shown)


FS-21:
Now the countdown for jumping to the next action will start from the beginning to prevent infinite loops etc.

The only exception is argument 0 that remains without counter & launch the next action as soon as all the alive team members are gathered around the Leader.

Metadorius (changes requested):
Same deal as with other PRs, need to review the code style (newlines, braces) and add changelog entry.

FS-21:
I updated the code style with the previous feedback, if there is a mistake please notify me & I'll fix it.

FS-21 (commented):
I think I finished with the changes

FS-21 (commented):
done?

Metadorius (approved):

PR #296, opened 2021-07-17
New Script Actions

FS-21:
<Don't merge yet, testing is welcome>

** From Script Action 82 to 83 **
82,nn => Decrease Current Trigger Weight
83,nn => Increase Current Trigger Weight

Unlike script action 49,0 that only is evaluated at the end of the script execution these 2 updates the current weight of the trigger that belongs the Team in the moment that is executed the script action, so these 2 can be executed multiple times.
In both scripts are positive integers. if the nn value is 0 then AITriggerSuccessWeightDelta & AITriggerFailureWeightDelta are used for updating this value.

** From Script Action 74 to 81 **
[Script line]=aa,nn where "aa" can be one of these script actions:
=74,nn -> Threats that are far have more priority. Kill until no more targets.
=75,nn -> Threats that are close have more priority. Kill until no more targets.
=76,nn -> Farther targets from Team Leader have more priority. Kill until no more targets.
=77,nn -> Closer targets from Team Leader have more priority. Kill until no more targets.
=78,nn -> Threats that are far have more priority. 1 kill only (good for combos with action 49,0).
=79,nn -> Threats that are close have more priority. 1 kill only (good for combos with action 49,0).
=80,nn -> Farther targets from Team Leader have more priority. 1 kill only (good for combos with action 49,0).
=81,nn -> Closer targets from Team Leader have more priority. 1 kill only  (good for combos with action 49,0).

The argument "nn" is the type of filter for selecting an unit to attack:
· 1  // Any enemy living thing allowed
· 2  // Any enemy structure, but ignores deployed vehicles
· 3  // Any enemy harvesters
· 4  // Any enemy infantry
· 5  // Any kind of enemy vehicle (including Aircrafts & Jumpjets)
· 6  // Any enemy factories
· 7  // Any enemy defensive structures
· 8  // Any enemy non-building object like infantry, vehicles & Aircrafts & deployed units.
· 9  // Any enemy structures with Power >= 0
· 10 // Any enemy aircraft & flying units like jumpjets
· 11 // Harvesters, Refineries & Resource Purifiers
· 12 // Any enemy refineries
· 13 // Any enemy unit or structure with Mind Control weapon
· 14 // Any enemy structure with infantry inside garrisoned like bunkers
· 15 // Any enemy unit & structure in with Naval=yes
· 16 // Any enemy Stealth Generator (when uncloacked), Gap Generator, Radar Jammer & Inhibitor
· 17 // Any enemy ground vehicles (unlike argument "5" that included Air units)
[...]

PR #364, opened 2021-09-14
Created script action 111: Un-register Team success

FS-21:
Is just the opposite effect of 49,0


src/Ext/Script/Body.cpp
```diff
@@ -40,6 +40,10 @@ void ScriptExt::ProcessAction(TeamClass* pTeam)
 	case 73:
 		ScriptExt::WaitUntillFullAmmoAction(pTeam);
 		break;
+	case 74: // IMPORTANT: This action will be renamed from 74 to 111 or another number when PR 296 gets merged into Develop branch
+		// Un-register success for AITrigger weight adjustment (this is the opposite of 49,0)
+		ScriptExt::UnregisterGreatSuccess(pTeam);
+		break;
```
Metadorius, line 46: We shouldn't merge it like so, make it 111 or the number that you will implement without the need for other changes further.

Metadorius (changes requested):
Changing script action numbers further is a very bad idea, change it to a final or at least semi-final number (111 I guess?)

Metadorius:
Also no changelog mention.

src/Ext/Script/Body.cpp
FS-21: The problem is FA2 that script action numbers should be consecutive. The last one was 73 and the 74 and forward are the ones of PR #296, being the 111 the first free slot for any of the new script actions I'm creating.

FS-21: Ok, I leave the definitive number 111 and made the mention f the action.

PR #375, opened 2021-09-21
Fix the crash of TemporalTargetingMe related "stack dump starts with 0051BB7D"

secsome:


secsome:
Tested with a crashing save, and with this PR no longer crashes.

PR #380, opened 2021-09-25
Units retaining orders after changing ownership bugfix

### LINKED ISSUE #314: [Vanilla Bug] Units retaining orders after changing ownership.
### Description:
If you mind control or Hijack a unit, it sometimes retains the last order it had from the previous owner.

### Steps To Reproduce:
It is hard to reproduce it constantly but:
1. send a unit on an attack order (like a demotruck).
2. Mind control or hijack it.

### Expected Behaviour:
The unit should completely stop whatever it was told before changing ownership.

### Actual Behaviour:
The unit retains the order and can shoot at the new owner even if he owns it.

secsome:
Closes #314


Lylaktech:
I had a few cases with mind control where they still showed this bug happen.




interestingly in the 1st picture the unit was not showing an attack order, yet was still attacking the powerplant next to it.

Lylaktech:
Still came across issues with units retaining their old orders after mind control BUT it is MUCH less frequent and only seems to happen with infantry, didn't encounter a single tank doing it this time. 
Units are much more responsive now when they get initially mind controlled instead of moving a cell or 2 before responding to their new order.

Metadorius (approved):

PR #405, opened 2021-10-10
Mission actual timer is not accurate

secsome:
Fix the bug that WW take 1 second as 960 milliseconds in the singleplayer campaign actual time.


FS-21:
Ignorant question:
How should this PR be tested in-game if 1 second is 960 milliseconds or 1000?

PR #418, opened 2021-10-26
Misc. weapon-related additions / improvements

Starkku:
### Disabling fallback to (Elite)Secondary weapon

- It is now possible to disable the fallback to `(Elite)Secondary` weapon from `(Elite)Primary` weapon if it cannot fire at the chosen target by setting `NoSecondaryWeaponFallback` to true (defaults to false). This does not apply to special cases where `(Elite)Secondary` weapon is always chosen, including but not necessarily limited to the following:
  - `OpenTransportWeapon=1` on an unit firing from inside `OpenTopped=true` transport.
  - `NoAmmoWeapon=1` on an unit with  `Ammo` value higher than 0 and current ammo count lower or  equal to `NoAmmoAmount`.
  - Deployed `IsSimpleDeployer=true` units with`DeployFireWeapon=1` set or omitted.
  - `DrainWeapon=true` weapons against enemy `Drainable=yes` buildings.
  - Units with `IsLocomotor=true` set on `Warhead` of `(Elite)Primary` weapon against buildings.
  - Weapons with `ElectricAssault=true` set on `Warhead` against `Overpowerable=true` buildings belonging to owner or allies.
  - `Overpowerable=true` buildings that are currently overpowered.
  - Any system using `(Elite)WeaponX`, f.ex `Gunner=true` or `IsGattling=true` is also wholly exempt.
  
In `rulesmd.ini`:
```ini
[SOMETECHNO]                      ; TechnoType
NoSecondaryWeaponFallback=false   ; boolean
```

Primary use cases would likely be certain kinds of Overpowerable=true buildings, IsSimpleDeployer=true units or NoAmmoWeapon configurations.

### Weapon target type filter
(This is a companion to the already-existing target house filter)

- You can now specify which targets can fire at. This also affects weapon selection, other than certain special cases where the selection is fixed.

In `rulesmd.ini`:
```ini
CanTarget=all        ; list of Affected Target Enumeration (none|land|water|empty|infantry|units|buildings|all)
```

### AreaFire target customization

- You can now specify how AreaFire weapon picks its target. By default it targets the base cell the firer is currently on, but this can now be changed to fire on the firer itself or at a random cell within the radius of the weapon's `Range` by setting `AreaFire.Target` to `self` or `random` respectively.
- `AreaFire.Target=self` respects normal targeting rules (Warhead Verses etc.) against the firer itself.
- `AreaFire.Target=random` ignores cells that are ineligible or contain ineligible objects based on listed values in weapon's `CanTarget`.

In `rulesmd.ini`:
```ini
AreaFire.Target=base ; AreaFire Target Enumeration (base|self|random)
```

Primary use-case would be deploy weapons that should snap and center on the firer or pick a random cell in the vicinity, especially useful if combined with `CanTarget` for special requirements like empty cell or land / water distinction.


CommanderNour:
Every thing works exactly like it should except the Disabling fallback to (Elite)Secondary weapon
works on vehicles / infantry but couldn't get it to work on structures no matter how I try

CommanderNour:
everything is good to go

PR #427, opened 2021-11-08
Some simple addition to Terrain type 

Otamaa:
These Features doesnt work on tiberium tree , due to how game coded it , blame WWP !

`Rules.ini`
```
[SOMETREE]
DestroyAnim = ; AnimType
DestroySound= ; Sound
```

### DIFF
--- a/src/Ext/TerrainType/Body.h
+++ b/src/Ext/TerrainType/Body.h
@@ -4,6 +4,7 @@
 #include <Helpers/Macro.h>
 #include <Utilities/Container.h>
 #include <Utilities/TemplateDef.h>
+#include <Utilities/Macro.h>
 
 class TerrainTypeExt
 {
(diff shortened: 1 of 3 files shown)


src/Ext/TerrainType/Body.h
```diff
@@ -17,12 +18,16 @@ class TerrainTypeExt
 		Valueable<int> SpawnsTiberium_Range;
 		Valueable<Point2D> SpawnsTiberium_GrowthStage;
 		Valueable<Point2D> SpawnsTiberium_CellsPerAnim;
+		Valueable<AnimTypeClass*> DestroyAnim;
+		ValueableIdx<VocClass> DestroySound;
```
Starkku, line 22: Use `Nullable` / `NullableIdx` and check for `isset()` when needed instead of `Valueable` / `ValueableIdx` set to default to `nullptr` or some magic number.

Starkku (changes requested):
Feature appears to work as expected and code looks okay save for one very small thing that's not really even covered in style guide but would be more consistent with majority of other code.

Starkku (approved):
Made a minor style adjustment & rebased. Approved & merging.

PR #432, opened 2021-11-13
Add basic class so we can extend the Bullet Trajectories

secsome:


Falaka21:
Yes! After many years, we have been waiting for this. I tested this and the most I realized that, the projectiles hit all the solid technos throught their way to target.I mean, they think they hit a solid target by default and penetrate it while damaging it. I really like this.(Since I use Ares on my mods, even thought I didn't use any "SolidBuildings" feature by Ares, this happens...not only buildings,especially natural civilian vehicles ,and buildings.

secsome:
StraightTrajactory is ready for test now.
```ini
[SomeProjectile]
Trajactory.Straight=yes
```

secsome:
Now `StraightTrajectory` and `SampleTrajectory` can be tested.
```ini
[SOMEPROJECTILE]
Trajectory=Straight
```

```ini
[SOMEPROJECTILE]
Trajectory=Sample
Trajectory.Sample.ExtraHeight = 1145.14 ; double
```

Starkku:
`Trajectory.Straight` for most part works like expected, it seems to behave better in comparison to the previous `Gravity=0 & Gravity.HeightFix=true` setup as well, especially on anti-aircraft weapons.

However I am unable to discern any observable difference in projectile's travel speed regardless of what the weapon's `Speed` is set to.

Starkku:
Both `Straight` & `Bombard` trajectories seem to work as expected and did not notice any issues with Save / Load, or unusual interactions with Projectile `SubjectTo*` keys. Only issue on latest build seems to be`Trajectory.Speed` default value. Documentation says it defaults to `10.0`, code says it defaults to WeaponType `Speed` while the actual default seems to be 0 (WeaponType `Speed` isn't set yet or couldn't be acquired properly?).

src/Ext/Bullet/Trajectories/BombardTrajectory.h
```diff
@@ -0,0 +1,42 @@
...
+public:
+	BombardTrajectory() : PhobosTrajectory(TrajectoryFlag::Bombard)
+		, IsFalling { false }
+		, Height { 0.0 }
+	{}
+
+	BombardTrajectory(PhobosTrajectoryType* pType) : PhobosTrajectory(TrajectoryFlag::Bombard)
+		, IsFalling { false }
+		, Height { 0.0 }
+	{}
```
Metadorius, line 31: Can be merged into one ctor.

src/Ext/Bullet/Trajectories/PhobosTrajectory.h
```diff
@@ -0,0 +1,136 @@
...
+
+	static PhobosTrajectoryType* LoadFromStream(PhobosStreamReader& Stm);
+	static void WriteToStream(PhobosStreamWriter& Stm, PhobosTrajectoryType* pType);
+	static PhobosTrajectoryType* ProcessFromStream(PhobosStreamReader& Stm, PhobosTrajectoryType* pType);
+	static PhobosTrajectoryType* ProcessFromStream(PhobosStreamWriter& Stm, PhobosTrajectoryType* pType);
+
+	TrajectoryFlag Flag;
+};
+
+class PhobosTrajectory
```
Metadorius, line 37: The separation between Trajectory and TrajectoryType has no point considering you're not allowing the users to create custom types.

src/Ext/Bullet/Trajectories/PhobosTrajectory.h
```diff
@@ -0,0 +1,136 @@
...
+*          ##                                                                                   $#*                          
+*           ##*                                                                                *#*                           
+*            ^##$                                                                             ##^                            
+*              $##$                                                                         *##                              
+*                $##*^                                                                   ^###^                               
+*                  ^##  ^###**$$$$$$$$$$$   ^$$$$$$$$$$$$$$$$$$****$   *##############  ^##$                                 
+*                    #####$$*****#########^##*#########*#**###*****##$##$^$$$$^$$^^^^##*##                                   
+*                     ^$^                *##^                       ***               ^$                                     
+* 
+*/
```
Metadorius, line 136: xu

Metadorius (changes requested):
I am not at all convinced we should use this exact architecture. There is redundant separation of types and trajectories, for instance; there are also quite a lot of boilerplate save/load methods in the classes.

PR #438, opened 2021-11-16
Make WarpOut, WarpIn, and WarpAway Anim support facings

### LINKED ISSUE #437: [New Feature] Customized Chrono Locomotor Animation
### Description:

Make it so that animation entries for WarpOut, WarpIn, and WarpAway support 8 or 16 facings
### Possible Implementation:

```ini
[SomeSection]
ex: WarpOut=WarpOutN,WarpOutNE,WarpOutE,WarpOutSE,WarpOutS,WarpOutSW,WarpOutW,WarpOutNW
```

Otamaa:
Close #437 

Will pick this up automatically and use the largest power of 2 equal to or larger than 8 to select the best animation from for each firing direction.

Fewer than 8 defined animations will result in only the first animation being used.
Default to Rules->WarpOut Anim if empty.

```Rules.ini
[SOMETECHNO]
WarpIn=          ;Vector AnimType
WarpOut=       ;Vector AnimType
WarpAway=     ;Vector AnimType
```

### DIFF
--- a/src/Ext/TechnoType/Body.h
+++ b/src/Ext/TechnoType/Body.h
@@ -49,9 +49,9 @@ public:
 
 		Valueable<ShieldTypeClass*> ShieldType;
 
-		Nullable<AnimTypeClass*> WarpOut;
-		Nullable<AnimTypeClass*> WarpIn;
-		Nullable<AnimTypeClass*> WarpAway;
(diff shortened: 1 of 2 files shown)


snmiglight:
noticed that whenever a voxel runs into a slope, it snaps back as if it is on a flat surface when it stops.

PR #444, opened 2021-11-19
Animation & Jumpjet unit layer customization

Starkku:
- You can now customize whether or not animations attached to objects follow the object's layer or respect their own `Layer` setting. If this is unset, attached animations use `ground` layer.

In `artmd.ini`:
```ini
[SOMEANIM]                 ; AnimationType
Layer.UseObjectLayer=      ; boolean
```

Allows turning on or off jumpjet unit behaviour where they fluctuate between `air` and `top` layers based on whether or not their current altitude is equal / below or above `JumpjetHeight` or `[JumpjetControls] -> CruiseHeight` if former is not set on TechnoType. If disabled, airborne jumpjet units exists only in `air` layer. `JumpjetAllowLayerDeviation` defaults to value of `[JumpjetControls] -> AllowLayerDeviation` if not specified.

In `rulesmd.ini`:
```ini
[JumpjetControls]
AllowLayerDeviation=yes         ; boolean

JumpjetAllowLayerDeviation=yes  ; boolean
```

Together these can work as a sort of a fix for the issue where animations, especially attached ones like shields or AttachEffect don't show up on top of airborne jumpjet units. For some reason I was not quite able to decipher, things that normally fudge the rendering order or depth-related drawing stuff (namely `YSortAdjust` & `ZAdjust`) cease to matter when both objects are on `top` layer, at least in case of jumpjet units vs animations - hence the need to potentially constrain jumpjets to `air` layer. Disabling layer deviation may only be desirable on wobbling jumpjets, as that is where it comes into play most often. It also causes other visual changes, namely with how grouped up bunch of wobbling jumpjet units looks (normally which ones would get drawn in front of others would keep shifting, creating a sort of 'blinking' effect, but disabling the layer fluctuation will also stop this from happening which might not be desirable, but considering this doesn't happen with non-wobbling jumpjets it might not be a big deal).

### DIFF
--- a/src/Ext/AnimType/Body.cpp
+++ b/src/Ext/AnimType/Body.cpp
@@ -30,6 +30,7 @@ void AnimTypeExt::ExtData::LoadFromINIFile(CCINIClass* pINI)
 	this->CreateUnit_RandomFacing.Read(exINI, pID, "CreateUnit.RandomFacing");
 	this->XDrawOffset.Read(exINI, pID, "XDrawOffset");
 	this->HideIfNoOre_Threshold.Read(exINI, pID, "HideIfNoOre.Threshold");
+	this->Layer_UseObjectLayer .Read(exINI, pID, "Layer.UseObjectLayer");
 }
 
 const void AnimTypeExt::ProcessDestroyAnims(UnitClass* pThis, TechnoClass* pKiller)
@@ -104,6 +105,7 @@ void AnimTypeExt::ExtData::Serialize(T& Stm)
 		.Process(this->CreateUnit_RandomFacing)
 		.Process(this->XDrawOffset)
 		.Process(this->HideIfNoOre_Threshold)
+		.Process(this->Layer_UseObjectLayer)
 		;
 }
 
--- a/src/Ext/AnimType/Body.cpp
+++ b/src/Ext/AnimType/Body.cpp
@@ -30,7 +30,7 @@ void AnimTypeExt::ExtData::LoadFromINIFile(CCINIClass* pINI)
 	this->CreateUnit_RandomFacing.Read(exINI, pID, "CreateUnit.RandomFacing");
 	this->XDrawOffset.Read(exINI, pID, "XDrawOffset");
 	this->HideIfNoOre_Threshold.Read(exINI, pID, "HideIfNoOre.Threshold");
-	this->Layer_UseObjectLayer .Read(exINI, pID, "Layer.UseObjectLayer");
+	this->Layer_UseObjectLayer.Read(exINI, pID, "Layer.UseObjectLayer");
 }
 
 const void AnimTypeExt::ProcessDestroyAnims(UnitClass* pThis, TechnoClass* pKiller)
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ Credits
 - **ChrisLv_CN** - interceptor logic, LaserTrails, laser fixes, general assistance (work relicensed under [following permission](images/ChrisLv-relicense.png))
 - **Xkein** - general assistance, YRpp edits
 - **thomassneddon** - general assistance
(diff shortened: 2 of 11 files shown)


mevitar:
AllowLayerDeviation=no or JumpjetAllowLayerDeviation=no causes jumpjets to be untargetable while on the ground. Weapons that can attack both air and ground will be able to target them, but will do damage instead.

mevitar:
As for updating mods and unmodded YR to adjust for this build, i noticed that the default layering for DrainAnimationType= changed now, as it will be drawn over the Floating Disk (or similar unit) instead of under. It needs Layer.UseObjectLayer= to be manually set to "no" to work as it used to.
Something that might need to be added in the documentation.

Didn't notice other issues so far.

Starkku:
Fixed jumpjets using wrong layer when on ground problem and changed the behaviour of `Layer.UseObjectLayer`. If it is not set then attached animations use `ground` layer like before. Setting it to `true` or `false` is what makes it follow the object's layer or respect the animation `Layer` setting, respectively so there are no backwards compatibility issues whatsoever now.

YRIDZE (changes requested):
Code looks good.  Please, fix just one issue.

mevitar:
Didn't find any issues anymore.

src/Ext/AnimType/Body.cpp
```diff
@@ -30,6 +30,7 @@ void AnimTypeExt::ExtData::LoadFromINIFile(CCINIClass* pINI)
 	this->CreateUnit_RandomFacing.Read(exINI, pID, "CreateUnit.RandomFacing");
 	this->XDrawOffset.Read(exINI, pID, "XDrawOffset");
 	this->HideIfNoOre_Threshold.Read(exINI, pID, "HideIfNoOre.Threshold");
+	this->Layer_UseObjectLayer .Read(exINI, pID, "Layer.UseObjectLayer");
```
Metadorius, line 33:
```suggestion
	this->Layer_UseObjectLayer.Read(exINI, pID, "Layer.UseObjectLayer");
```

Metadorius (approved):

mevitar:
I can't find AllowLayerDeviation and JumpjetAllowLayerDeviation in the newest doc. Was this feature entirely removed, or is it undocumented?

Starkku:
> I can't find AllowLayerDeviation and JumpjetAllowLayerDeviation in the newest doc. Was this feature entirely removed, or is it undocumented?

Was removed in favour of fixing the layering properly by default.

mevitar:
> If the old approach was more desirable in certain cases, must've flown over my head.

I was trying to do it the old way and i jumpjets were covering the AE animation when they were in the top position of the wobble, so i was wondering whether something broke or did i use it incorrectly.

But no worries, since i can use YSortAdjust now i will do it with Layer.UseObjectLayer=yes on the animation (it seems to work fine so far, so that solves my problem).

mevitar:
Ugh, and right after i wrote that it works fine, i get the jumpjets clipping over the anim again. Worst part is, it doesn't always happen. Sometimes it always stays on top, and sometimes it vanishes when the aircraft is in top position. Same thing happens with YSortAdjust=20 and default value.
I even tried attaching the anim while the unit is on the ground, and then the anim went under the unit once it lifted up, and stayed like that forever.

I think it might have something to do with the position that the unit is in at the moment the animation is attached to it - if it's on top, then the animation will stay always on top, but if it's at the bottom, then it will vanish if the unit cycles to the top. I guess Layer.UseObjectLayer=yes might not be updating the layer properly.

Of course if i set Layer.UseObjectLayer=no and Layer=top then the animation always stays on top of the unit, but i wanted to avoid that because the animation can be attached to ground units.

Starkku:
>  I guess Layer.UseObjectLayer=yes might not be updating the layer properly.

Correct. It doesn't work correctly in all cases, bit of a design fault. Updating it it do that might be bit of a task but I'll take a look.

mevitar:
Checked the newest devbuild, and this time I couldn't find any issues with Layer.UseObjectLayer=yes on jumpjets.
(i hope i won't have to come here again in 10 minutes to say i was wrong)

PR #446, opened 2021-11-20
IsSimpleDeployer improvements

Starkku:
- `DeployDir` can be used to set the facing the vehicle needs to turn towards before deploying if it has `DeployingAnim` set. This is works the same as Ares flag of same name other than allowing use of negative numbers to disable the direction-specific deploy and that it only applies to units on ground. If not set, it defaults to `[General] -> DeployDir`.
- In addition there are some new options for `DeployingAnim`:
  - `DeployingAnim.KeepUnitVisible` determines if the unit is hidden while the animation is playing.
  - `DeployingAnim.ReverseForUndeploy` controls whether or not the animation is played in reverse for undeploying.
  - `DeployingAnim.UseUnitDrawer` controls whether or not the animation is displayed in the unit's palette and team colours.

In `rulesmd.ini`:
```ini
[SOMEVEHICLE]                          ; VehicleType
DeployDir=                             ; integer, facing or a negative number to disable direction-specific deploy
DeployingAnim.KeepUnitVisible=false    ; boolean
DeployingAnim.ReverseForUndeploy=true  ; boolean
DeployingAnim.UseUnitDrawer=true       ; boolean
```

In addition, `IsSimpleDeployer` units now only play `DeploySound` and `UndeploySound` once, when done with (un)deploying instead of repeating it over duration of turning and/or `DeployingAnim`.

### DIFF
--- a/src/Ext/TechnoType/Hooks.cpp
+++ b/src/Ext/TechnoType/Hooks.cpp
@@ -215,24 +215,49 @@ DEFINE_HOOK(0x73CF46, UnitClass_Draw_It_KeepUnitVisible, 0x6)
 	return 0;
 }
 
+DEFINE_HOOK(0x739B32, UnitClass_Deploy_Turning, 0x6)
+{
+	enum { StartAnim = 0x739B9E, Return = 0x739CBE };
+
+	GET(UnitClass*, pThis, ESI);
+
+	if (auto const pExt = TechnoExt::ExtMap.Find(pThis))
+	{
+		if (pExt->Deploying_TargetFacing.isset())
+		{
+			if (pThis->PrimaryFacing.current() != pExt->Deploying_TargetFacing.Get())
+			{
+				return Return;
+			}
+			else
+			{
+				pExt->Deploying_TargetFacing.Reset();
+				return StartAnim;
+			}
+		}
+	}
+
+	return 0;
+}
+
 // Ares hooks in at 739B8A, this goes before it and skips & basically reimplements Ares code with some changes.
 DEFINE_HOOK(0x739B7C, UnitClass_Deploy_DeployDir, 0x6)
 {
-	enum { Skip = 0x739C70, Continue = 0x739B9E };
+	enum { Skip = 0x739C70, Continue = 0x739B9E, Return = 0x739CBE };
 
 	GET(UnitClass*, pThis, ESI);
 
 	if (!pThis->InAir)
 	{
-		if (pThis->Type->DeployingAnim)
+		if (pThis->Type->DeployingAnim && !pThis->Deploying)
 		{
-			int deployDir = TechnoTypeExt::ExtMap.Find(pThis->GetTechnoType())->DeployDir.Get(RulesClass::Instance->DeployDir);
+			int deployDir = TechnoTypeExt::ExtMap.Find(pThis->GetTechnoType())->DeployDir.Get(RulesClass::Instance->DeployDir >> 5);
 
 			if (deployDir >= 0)
 			{
-				auto newFacing = DirStruct(deployDir << 13);
+				auto targetFacing = DirStruct(deployDir << 13);
 
-				if (pThis->PrimaryFacing.current() != newFacing)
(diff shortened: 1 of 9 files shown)


src/Ext/TechnoType/Hooks.cpp
```diff
@@ -180,4 +181,101 @@ DEFINE_HOOK(0x6F421C, TechnoClass_DefaultDisguise, 0x6) // TechnoClass_DefaultDi
...
+		if (auto const pAnim = GameCreate<AnimClass>(pThis->Type->DeployingAnim, pThis->Location, 0, 1, 0x600, 0, !isDeploying ? pExt->DeployingAnim_ReverseForUndeploy : false))
+		{
+			pThis->DeployAnim = pAnim;
+			pAnim->SetOwnerObject(pThis);
+
+			if (pExt->DeployingAnim_UseUnitDrawer)
+				return isDeploying ? DeployUseUnitDrawer : UndeployUseUnitDrawer;
+		}
+		else
+			pThis->DeployAnim = nullptr;
```
YRIDZE, line 261:
```suggestion
		else
		{
			pThis->DeployAnim = nullptr;
		}
```

YRIDZE (changes requested):
Code looks good. Just please correct a couple of points.

mevitar:
DeployingAnim.X= additions and sound changes work fine, didn't notice any issues there.

mevitar:
As for DeployDir, i'll be a downer here, but did it really have to be changed?
AFAIK it already worked _any_  VehicleType with a DeployingAnim= before build, including type conversion ones (i know, i use it extensively). Furthermore, it used to default to the [General]DeployDir= setting, but now it doesn't read that one and everything faces 0 direction unless i manually change it for each unit. So this change feels more like a downgrade.

Starkku:
> As for DeployDir, i'll be a downer here, but did it really have to be changed? AFAIK it already worked _any_ VehicleType with a DeployingAnim= before this build, including type conversion ones (i know, i use it extensively). Furthermore, it used to default to the [General]DeployDir= setting, but now it doesn't read that one and everything faces 0 direction unless i manually change it for each unit. So this change feels more like a downgrade.

It worked but it was also *forced* on any unit with DeployingAnim, allowing no direction-agnostic deploy combined with animation. With regular animation settings this isn't particularly useful but with the additions here that's another story. Not defaulting to `[General] -> DeployDir` is a bug, it should default to it and it not being mentioned in documentation is an oversight.

mevitar:
Fair enough. I thought it was intentional since the doc says that DeployDir is set to 0 (which is an actual ingame direction) by default.

Starkku:
Implemented code style suggestions and following tweaks:

- Units now automatically start deploy process/DeployingAnim once they are done turning to the correct direction.
- Make TechnoType->DeployDir default to General->DeployDir properly.

Both are also now mentioned in documentation.

mevitar:
I think that making deploy process start immediately causes a rare bug that i'm not able to reproduce: the deploying vehicle vanishes and gets stuck in undeployed form, unable to move, and without an image (it was a unit-unit conversion).

On the left there is the unit that lost its image and is stuck (you can faintly see the AE animation on it still active, though). On the right there is a bunch of that unit type but they deployed succesfully. No entries from this build were used for this unit, everything is the same as i used before i did the tests.
nanidafuq

Also noticed that the same unit, if initially facing a direction different than the deploying one (specified as being ingame south here), it will vanish for like 1 frame when the deploy animation didn't start yet, making it look as if it blinked for a moment. This doesn't happen if it already faces the correct direction, nor did it happen before (since they had to turn to be facing the correct direction first).

This obviously doesn't happen with DeployingAnim.KeepUnitVisible=yes, but as the said unit is supposed to completely change its form, i can't keep the unit visible or the transformation won't look the way it should.

Metadorius:
I assume this needs more work? I am marking it as draft then, if I mistakenly assumed it or you fix it - un-draft it then.

Starkku:
Reverted the automatic deploy after turning to required facing, it might be doable without issues but it'd require more thought and might be best saved for later.

PR #450, opened 2021-11-24
Simple EBolt trail

secsome:


Metadorius (changes requested):
I propose to not choose between the laser trail and ebolt trail but just add `IsLaser` and `IsEBolt`, and then just draw the corresponding trails. Maybe someone wants to combine those.

ayylmaoRotE:
Works, as far as the single Ebolt function goes, other settings seem to be ignored though

PR #461, opened 2021-12-04
Weapon targeting filter improvements

Starkku:
- `CanTarget` now explicitly requires `all` or `empty` to be listed to be able to fire at cells containing no TechnoTypes.
  - `AreaFire.Target=base` & `AreaFire.Target=self` (when firing the weapon, not for targeting or weapons selection) weapons are exempt from this.
- `AreaFire.Target=random` now respects `CanTargetHouses`.
- Fixed an issue where `AreaFire.Target=self` would override the firing techno's current target with itself.

Uses cases include changing force-fire weapon from Primary to Secondary (which can even be done conditionally based on whether cell is land or water) and allowing `AreaFire.Target=random` weapons to home into (specific) objects (owned by specific houses) instead of always allowing empty cells to be targeted.

### DIFF
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -420,6 +420,7 @@ Rad.NoOwner=no  ; boolean
 ### Weapon targeting filter
 
 - You can now specify which targets or houses a weapon can fire at. This also affects weapon selection, other than certain special cases where the selection is fixed.
+  - Note that `CanTarget` explicitly requires either `all` or `empty` to be listed for the weapon to be able to fire at cells containing no TechnoTypes.
 
 In `rulesmd.ini`:
 ```ini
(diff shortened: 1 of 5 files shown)


mevitar:
While not mentioned as an intended use-case, I wanted to see if it's possible to prevent weapons from force-firing at the ground so cloaked units cannot be killed without having a cloak detector. This is what i found:
_CanTarget=infantry,units,buildings_ forbids the player from force-firing at empty cells, but still allows to force-fire at cells that cloaked objects are on.
_CanTarget=empty_ only allows to force-fire on empty cells, but cells with cloaked objects on them aren't forbid at all (as long as the cell under them is targeted).

This means that it's impossible to do that. At best, the weapon will simply not fire when there is nothing on that cell, and fire when there is (so it's actually even easier to kill a cloaked object).

Not sure if this is something that needs fixing or not, though.

Starkku:
> While not mentioned as an intended use-case, I wanted to see if it's possible to prevent weapons from force-firing at the ground so cloaked units cannot be killed without having a cloak detector. This is what i found: _CanTarget=infantry,units,buildings_ forbids the player from force-firing at empty cells, but still allows to force-fire at cells that cloaked objects are on.
_CanTarget=empty_ only allows to force-fire on empty cells, but cells with cloaked objects on them aren't forbid at all (as long as the cell under them is targeted).
> 
> This means that it's impossible to do that. At best, the weapon will simply not fire when there is nothing on that cell, and fire when there is (so it's actually even easier to kill a cloaked object).
> 
> Not sure if this is something that needs fixing or not, though.

The problem is that these restrictions are applied on per target basis. If you force-fire on a cell (even if it has objects on it), the target is the cell and only the cell is checked. It would be possible to make it always check both cell and any & all objects on it (although it would likely need an exception for AreaFire deploy weapons to not break them as for those the target when considering if the weapon is allowed to fire is always the firer's cell, and the actual target is determined later) but this would still result in what you described as best-case scenario where the weapon can either be fired at empty cells & not at objects or vice-versa. It would not be outside the realm of possibility to consider cloaked objects separately but there is no real way to discern between cloaked object that is not currently revealed and one that is, only whether or not the object is cloaked or not. All in all, not entirely sure what could or should be changed here either.

mevitar:
> All in all, not entirely sure what could or should be changed here either.

I guess this can be marked as ready, then (unless someone finds a different issue - i haven't).

Metadorius (approved):

PR #465, opened 2021-12-12
Occupied Buildinng MuzzleFlashX Bug Fix

### LINKED ISSUE #439: [Vanilla Bug] MuzzleFlashX Bug on Occupied Buildings
### Description:

In art.ini, you can specify muzzle flash offsets from MuzzleFlash00 to MuzzleFlash09 for buildings with CanBeOccupied=yes and CanOccupyFire=yes. But for MuzzleFlash10 and onwards, the animations will be drawn on the center of the building, regardless of FLH offsets entered.

### Steps To Reproduce:

Just put an FLH value on MuzzleFlash10= or any higher number. Muzzle anims will show on the center of the building instead of the FLH offsets, as shown on MO's Concrete Fortress.

### Expected Behaviour:

Animation on MuzzleFlash10 and beyond should be drawn on the FLH offset entered.
### Actual Behaviour:

Muzzle flashes for eleventh or more occupants are drawn on the center. (Bug doesn't seem to affect the offsets for MuzzleFlash09 and below though.)

Otamaa:
close #439


FS-21:
I added increased the passengers from 10 to 12 in rulesmd and in artmd I added MuzzleFlash10 and MuzzleFlash11 (with extreme values just for noticing if new annims appears) and tried before and after using this PR binary. in my test looks that this work as planned.

; artmd.ini
[CARUS01]
Normalized=yes
Remapable=no
Foundation=4x4
Height=8
NewTheater=yes
DemandLoad=false
MuzzleFlash0=40,42 ; these are pixel offsets for muzzle flashes from center of base
MuzzleFlash1=-40,64
MuzzleFlash2=-2,-11
MuzzleFlash3=-31,0
MuzzleFlash4=30,64
MuzzleFlash5=-33,35
MuzzleFlash6=-72,46
MuzzleFlash7=0,-51
MuzzleFlash8=-5,79
MuzzleFlash9=-58,47
MuzzleFlash10=65,235 <----- new
MuzzleFlash11=-65,235 <----- new
CanBeHidden=False
OccupyHeight=5
DamageFireOffset0=-15,38
DamageFireOffset1=46,49
AddOccupy1=-4,-4
RemoveOccupy1=-2,1
RemoveOccupy2=-3,0
RemoveOccupy3=-3,-1
RemoveOccupy4=2,-1
RemoveOccupy5=1,-2
RemoveOccupy6=0,-3

imagen

CommanderNour:
eiffle1

src/Ext/BuildingType/Hooks.cpp
```diff
@@ -0,0 +1,61 @@
+#include "Body.h"
+
+#include <BuildingClass.h>
+
+DEFINE_HOOK(0x460285, BuildingTypeClass_LoadFromINI_Muzzle, 0x6)
+{
+	GET(BuildingTypeClass*, pThis, EBP);
+
+	R->Stack(STACK_OFFS(0x368, 0x358), 0);
+	R->EDX(0);
```
Metadorius, line 10: Would be good to have an explanation of what happens here.

src/Ext/BuildingType/Hooks.cpp
```diff
@@ -0,0 +1,61 @@
...
+	GET(BuildingTypeClass*, pThis, EBP);
+
+	R->Stack(STACK_OFFS(0x368, 0x358), 0);
+	R->EDX(0);
+
+	auto nCount = pThis->MaxNumberOccupants;
+
+	//manipulate this so it can disable itself use phobos one instead
+	nCount = Math::clamp(nCount, 0, 11);
+	return !nCount || nCount == 11 ? 0x460388 : 0x460299;
```
Metadorius, line 16: Use anonymous enums for user-readable addresses.

src/Ext/BuildingType/Hooks.cpp
Metadorius: Or even better, maybe you can make references to those locations with human-readable names and assign values to them?

src/Ext/BuildingType/Body.h
```diff
@@ -22,12 +22,14 @@ class BuildingTypeExt
 		Nullable<int> PowerPlantEnhancer_Amount;
 		Nullable<float> PowerPlantEnhancer_Factor;
 
+		DynamicVectorClass<Point2D> PhobosMuzzleFlash;
```
Metadorius, line 25: How about just `MuzzleFlashes` instead?

src/Utilities/TemplateDef.h
```diff
@@ -50,6 +50,34 @@
 #include <VoxClass.h>
 #include <ArmorType.h>
 
+namespace Templated
```
Metadorius, line 53: I am not entirely sure this function needs to be here, with that name and that we need it and can't use something that already exists? @secsome
@Thrifinesma

Metadorius (changes requested):
Some minor issues that need to be addressed.

src/Utilities/TemplateDef.h
```diff
@@ -50,6 +50,34 @@
...
+
+		for (int i = nVec.Count; i < nBaseSize; ++i)
+			nVec[i] = nDefaultVal;
+
+		nVec.Count = nBaseSize;
+
+		for (int i = 0; i < nBaseSize; ++i)
+		{
+			Valueable<Ta> nBuffer;
+			std::string nFlag(pBaseFlag + std::to_string(i));
```
Metadorius, line 69: There were implementations of this stuff without usage of std::string, look up LaserTrails for example.

src/Utilities/TemplateDef.h
Otamaa: hmm , i can use one from laser trail , but writing them on `LoadFromINI` like that is big chunk of code , maybe better template it for further usage and look cleaner

Metadorius (approved):

PR #476, opened 2021-12-18
Refactor ExtendedToolTip

secsome:
Just reimplement it and prepare for implement more modern tooltip


Metadorius:
Did anyone except secsome test if tooltips work the same way?

Metadorius (approved):

Metadorius:
This is what happens when secsome "tests" his own branches

Metadorius (changes requested):
Fix the bug and we can merge this.

PR #486, opened 2021-12-27
Shield / misc. fixes & improvements

### LINKED ISSUE #462: [Phobos Bug] Cloak + Shield = object does not reveal itself if damaged
### Description:
RA2/YR usually reveals cloaked objects that receive damage, even if they don't get hit directly. All that has to happen is that the cell they are on is hit and the weapon deals damage, then they decloak for a brief moment and cloak back. This, however, does not happen for objects with shields.

### Expected Behaviour:
Since that's how it worked forever, one could expect that a shielded object will reveal itself each time it is being hit, just like unshielded objects do.

### Actual Behaviour:
Cloaked objects without a shield (or those that are damaged with a weapon that bypasses their shield) decloak for a brief moment each time a cell they are on is hit and they receive damage.
Cloaked objects with a shield do not decloak, even if they are damaged through force-firing the ground underneat them. As long as the shield is active, they will stay cloaked.

Starkku:
- Fixed a number of issues with shields interacting with TechnoType changes (Undeploys/DeploysInto, Ares type conversion). If new TechnoType has `ShieldType` set, that shield will always be used, otherwise new `ShieldType` setting `AllowTransfer` determines if the (old) shield carries over or not. If `AllowTransfer` is not set, it defaults to `true` if shield was created through Warhead `Shield.AttachTypes`, otherwise `false`.
- Cloaked objects with shields are now uncloaked if the shield gets damaged by a warhead.
- Introduced a new Warhead setting `DecloakDamagedTargets`, which controls whether or not the warhead will decloak targets if it damages them (this applies to both shielded and unshielded targets. Defaults to `true`.
- Negative shield self-healing will now work as expected, deducting hitpoints and breaking the shield once there are no hitpoints left.

Closes #462

### DIFF
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -37,7 +37,8 @@ HitAnim=                             ; animation
 BreakWeapon=                         ; WeaponType
 AbsorbPercent=1.0                    ; double, percents
 PassPercent=0.0                      ; double, percents
-                                     
+AllowTransfer=                       ; boolean
+
 [SOMETECHNO]                         ; TechnoType
 ShieldType=SOMESHIELDTYPE            ; ShieldType; none by default
                                      
@@ -85,6 +86,7 @@ Shield.InheritStateOnReplace=false   ; boolean
 - `BreakWeapon`, if set, will be fired at the TechnoType once the shield breaks.
 - `AbsorbPercent` controls the percentage of damage that will be absorbed by the shield. Defaults to 1.0, meaning full damage absorption.
 - `PassPercent` controls the percentage of damage that will *not* be absorbed by the shield, and will be dealt to the unit directly even if the shield is active. Defaults to 0.0 - no penetration.
+- `AllowTransfer` controls whether or not the shield can be transferred if the TechnoType changes (such as `(Un)DeploysInto` or Ares type conversion). If not set, defaults to true if shield was attached via `Shield.AttachTypes`, otherwise false.
 - A TechnoType with a shield will show its shield Strength. An empty shield strength bar will be left after destroyed if it is respawnable.
   - Buildings now use the 5th frame of `pips.shp` to display the shield strength while other units uses the 16th frame by default.
   - `Pips.Shield` can be used to specify which pip frame should be used as shield strength. If only 1 digit set, then it will always display it, or if 3 digits set, it will respect `ConditionYellow` and `ConditionRed`. `Pips.Shield.Building` is used for BuildingTypes.
--- a/src/New/Entity/ShieldClass.cpp
+++ b/src/New/Entity/ShieldClass.cpp
@@ -17,13 +17,14 @@ ShieldClass::ShieldClass() : Techno { nullptr }
 	, Timers { }
 { }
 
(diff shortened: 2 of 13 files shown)


docs/New-or-Enhanced-Logics.md
```diff
@@ -84,7 +85,8 @@ Shield.InheritStateOnReplace=false   ; boolean
 - `HitAnim`, if set, will be played when the shield is attacked, similar to `WeaponNullifyAnim` for Iron Curtain.
 - `BreakWeapon`, if set, will be fired at the TechnoType once the shield breaks.
 - `AbsorbPercent` controls the percentage of damage that will be absorbed by the shield. Defaults to 1.0, meaning full damage absorption.
-- `PassPercent` controls the percentage of damage that will *not* be absorbed by the shield, and will be dealt to the unit directly even if the shield is active. Defaults to 0.0 - no penetration.
+- `PassPercent` controls the percentage of damage that will *not* be absorbed by the shield, and will be dealt to the unit directly even if the shield is active. Defaults to 0.0 - no penetration0.
```
Metadorius, line 88: Added typo.

src/New/Entity/ShieldClass.cpp
```diff
@@ -13,19 +13,20 @@
...
+ShieldClass::ShieldClass(TechnoClass* pTechno, bool isAttached) : Techno { pTechno }
+, IdleAnim { nullptr }
+, Timers { }
+, Cloak { false }
+, Online { true }
+, Temporal { false }
+, Available { true }
+, Attached { isAttached }
+, SelfHealing_Rate_Warhead { -1 }
+, Respawn_Rate_Warhead { -1 }
```
Metadorius, line 29: Tabs removed?

src/New/Entity/ShieldClass.cpp
```diff
@@ -365,33 +373,49 @@ void ShieldClass::TemporalCheck()
...
-	{ // Update this shield for the new owner
-		const auto pFrom_TechnoType = TechnoTypeClass::Find(this->TechnoID);
-		const auto pFromType = TechnoTypeExt::ExtMap.Find(pFrom_TechnoType)->ShieldType;
+	// Update shield type.
+	if (!allowTransfer && !pTechnoTypeExt->ShieldType->Strength)
+	{
+		this->KillAnim();
+		pTechnoExt->CurrentShieldType = ShieldTypeClass::FindOrAllocate(NONE_STR);
+		pTechnoExt->Shield = nullptr;
+		return true;
```
Metadorius, line 394:
```suggestion
		this->KillAnim();
		pTechnoExt->CurrentShieldType = ShieldTypeClass::FindOrAllocate(NONE_STR);
		pTechnoExt->Shield = nullptr;

		return true;
```

src/New/Entity/ShieldClass.cpp
Starkku: Auto-formatting the document or selection seems to do this with the current settings used.

mevitar:
The mentioned issues appear to be fixed for techno-owned shields, including type conversions that switch shields. New tags also appear to be working as specified.

PR #487, opened 2021-12-28
Shield / new features about shield stole

Loritas:
works when the ShieldDamage > 0

```ini
[SOMEWARHEAD]
Shield.Assimilate=false ; boolean, whether this warhead will recover its owner's shield by damaging another shield
Shield.Assimilate.Multiplier=1.0 ; double, recovered multiplier by assimilation, values from [0.0,10.0]
Shield.Steal=false ; boolean, whether this warhead will steal target's shield to its owner, when its owner doesn't have any shield
Shield.Steal.Multiplier=0.0 ; double, how many hp percents are the shield after stolen, values from [0.0,1.0]

[SOMESHIELDTYPE]
CanBeAssimilated=true ; boolean, whether this shield can be assimilated
CanBeStolen=true ; boolean, whether this shield can be stolen
```


Metadorius (changes requested):
The PR needs a few improvements. 
1. The assimilation and stealing logic aren't explained, and deducing them from code is a bit hard.
2. The doc entries, changelog and credits entries are missing.
3. Not sure how does stealing interact with already existing shields?
4. Some minor issues on the code style side.

Phobos.vcxproj
```diff
@@ -189,8 +189,8 @@
     <VCProjectVersion>15.0</VCProjectVersion>
     <ProjectGuid>{3FAF7126-F38C-4D1E-9973-C21A37870F60}</ProjectGuid>
     <RootNamespace>Phobos</RootNamespace>
-    <WindowsTargetPlatformVersion>7.0</WindowsTargetPlatformVersion>
     <ProjectName>Phobos</ProjectName>
+    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
```
Starkku, line 193: Is there a reason behind this change?

docs/New-or-Enhanced-Logics.md
```diff
@@ -113,6 +113,8 @@ BreakWeapon=                         ; WeaponType
 AbsorbPercent=1.0                    ; double, percents
 PassPercent=0.0                      ; double, percents
 AllowTransfer=                       ; boolean
+CanBeStolen=true   					 ; boolean
+CanBeStolenType=true  				 ; boolean
```
Starkku, line 117: Indentation of the commented bits should be fixed to line up (two spaces from the end of longest line to ;, and rest lining up with that).

docs/New-or-Enhanced-Logics.md
```diff
@@ -745,6 +757,7 @@ OpenTopped.WarpDistance=8          ; integer
   - `Overpowerable=true` buildings that are currently overpowered.
   - Any system using `(Elite)WeaponX`, f.ex `Gunner=true` or `IsGattling=true` is also wholly exempt.
   
+
```
Starkku, line 760: Introduction of unnecessary empty lines.

docs/New-or-Enhanced-Logics.md
```diff
@@ -1112,6 +1125,7 @@ RemoveMindControl=no                 ; boolean
     - `Crit.AnimOnAffectedTargets`, if set, makes the animation(s) from `Crit.AnimList` play on each affected target *in addition* to animation from Warhead's `AnimList` playing as normal instead of replacing `AnimList` animation.
   - `ImmuneToCrit` can be set on TechnoTypes to make them immune to critical hits.
   
+
```
Starkku, line 1128: Same as previous comment.

docs/New-or-Enhanced-Logics.md
```diff
@@ -140,6 +142,10 @@ Shield.ReplaceOnly=false             ; boolean
 Shield.ReplaceNonRespawning=false    ; boolean
 Shield.MinimumReplaceDelay=0         ; integer, game frames
 Shield.InheritStateOnReplace=false   ; boolean
+Shield.Steal=false  				 ; boolean
+Shield.Assimilate.Rate=1.0 			 ; double, percents
+Shield.StealTargetType=false 		 ; boolean
+Shield.StealTargetType.InitShieldHealthRate=0.0 ; double, percents
```
Starkku, line 148: Same issue as in previous comment. Also `Shield.StealTargetType.InitShieldHealthRate` is obnoxiously long key name, maybe atleast drop the second `Shield` from it to make it slightly less long if nothing else.

Starkku (changes requested):
There are some remaining, mostly smaller issues that need to be addressed (including the change Kerbiter has requested earlier still).

Phobos.vcxproj
Loritas: Just because vs changed it when I have a problem about compiling project (actually it doesn't change anything)

SMxReaver:
Shield stealing to restore your own health as well as stealing another object's entire ShieldType both seem to be working correctly. I was able to use:
- non-respawning zero initial health shields were respawned and filled up by shield.steal
- shields can be successfully stolen based on their type, and this works between technotypes (infantry can steal tank shields, buildings can steal aircraft shields, etc.)
- shields not specified to be stolen cannot be stolen
- shields not 

I wonder if the naming for turning on warhead-based stealing could be a bit better. This is just personal opinion, but perhaps Shield.Steal= could perhaps be Shield.Assimilate= to match second tag (Shield.Assimilate.Rate). Or Shield.Assimilate.Rate renamed to be Shield.Steal.Rate=?

PR #493, opened 2022-01-07
MakeInfantry logic on BombClass resulting on `Neutral` side infantry BugFix

### LINKED ISSUE #10: Enhance MakeInfantryOwner Logic
Allow `MakeInfantryOwner=` to designate infantry owning house invoked by animations apart from `InfDeathAnim`, `DeathAnims`, `InfantryExplode`, `FlamingInfantry`, `InfantryElectrocuted`,` InfantryHeadPop`, `InfantryNuked`, `InfantryVirus`, `InfantryMutate`, `InfantryBrute` and map triggers.

Expected use-case would be applying MakeInfantry logic on ivan deathbombs which current environment would always gain the resulted infantry to the Neutral side instead.

Otamaa:
close #10

- Fix the issue when `BombClass` giving nullptr `Techno` owner resulting on Some Warhead affect no applied properly . 
- Added Ownership for `BombClass` explosion anim to fix problem with `CreateUnit` and `MakeInf` .


Otamaa:
> Could you explain what did you do to fix it? Looks unconventional and bit hard to read.

1. Due to Ares hooking Above when `SelectCombatAnim`, that long code which replace `call` is one i can think of , to add ownership to the animation created by bomb explosion .
2. Second hook is fix when `BombClass` lose its techno ownership , thus we cant get any ownership with `Techno->GetOwningHouse()` .

deathreaperz:
wip-bio-infestation-2

Maybe a late review, I can confirm that "Added Ownership for BombClass explosion anim to fix problem with CreateUnit and MakeInf" works.

Starkku:
Tested, appears to work. Also fixes another related issue where the vanilla code would attempt to create `AnimClass` even if `AnimTypeClass` is `nullptr`.

Took a liberty of fixing issue preventing the code from building as well as couple of other things and pushing it.

Metadorius (approved):

PR #507, opened 2022-01-15
Replace DeployDir customization with DeployingAnim.AllowAnyDirection

Starkku:
This is to remove parsing conflicts with Ares, due to Ares treating any other facing than 0-8 as invalid when parsing INI content and thus generating warnings such as:

`[Developer warning]Failed to parse INI file content: [ATCRT]DeployDir=-1 (Expected a facing between 0 and 8)`

Besides requiring INI changes, no actual functionality should have been lost with this change. This will still only apply to ground units due to the Ares hook for jumpjet unit DeployDir being at an address that can't be hooked around.

Translated documentation will need to be adjusted to match the changes here.

### DIFF
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -96,18 +96,17 @@ Grinding.Weapon=           ; WeaponType
 
 ## Vehicles
 
-### Deploy direction for IsSimpleDeployer vehicles & deploy animation customization
+### IsSimpleDeployer vehicle deploy animation / direction customization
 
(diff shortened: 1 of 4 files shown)


mevitar:
I think it's not necessarily only ground units that this works for, but specifically anything that doesn't have to land to deploy.

The reason that distinction is important here is because hover units (technically ground units) with DeployToLand=yes still turn to one direction before deploying, even if they have DeployingAnim.AllowAnyDirection=yes. However, with DeployToLand=no, the same unit deploys properly without turning. So this tag works fine for hover->hover unit conversions (hover->landed require DeployToLand=yes to not levitate in the air after deployment, so it doesn't work for them).

Metadorius (approved):

PR #511, opened 2022-01-19
Critical damage chance additions

### LINKED ISSUE #505: Custom warhead for Crit.Damage=
This allows players to customize damage percent of Critical damage and add any effects.

[Warhead]
Crit.Warhead=; (warhead)

Starkku:
- `Crit.ApplyChancePerTarget` can be set to make the critical hit chance be determined individually for each target instead of only once when the Warhead is detonated.
- `Crit.Warhead`, if set, will be detonated at every affected target using `Crit.ExtraDamage`. Otherwise only damage will be dealt using current Warhead. **NOTE: Avoid using same Warhead and looping Warhead chains, they lead to freezes & crashes.**
- `Crit.AffectBelowPercent` can be used to set maximum percentage of hitpoints targets (if applicable) can have left to be affected.
- `Crit.AnimList.PickRandom` determines if a random animation from `Crit.AnimList` is picked or not. Defaults to `AnimList.PickRandom`.
- `Crit.AnimOnAffectedTargets`, if set, draws animation on every affected target (randomized separately every time if `Crit.AnimList.PickRandom` is in effect) instead of once at Warhead's impact location if critical damage chance roll was successful. This is in addition to Warhead's `AnimList`, contrary to the behaviour if this isn't set.

Additionally fixes a 'bug' where `Crit.AnimList` would be used even if no critical hits were dealt due to `ImmuneToCrit` or `Crit.Affects` configuration.

Closes #505


Metadorius:
@zenoniations requesting your testing (rhyme unintended).

Starkku:
Warhead is used purely for directly applying the `Crit.Damage` to the affected units.. It is not actually detonated, so any effects like animations or AE are not applied from it.

Metadorius:
I suggest to split that into another pull request, I believe it would be easy to make this into a full warhead via nested weapons (sorta like DeathWeapon works but with proper firer, target and explosion location).

Starkku:
Added `Crit.ApplyChancePerTarget`, made `Crit.Warhead` fully detonate the warhead if set (avoid using same Warhead and looping Warhead chains) and expanded documentation.

CommanderNour:
Works :thumbsup:

Metadorius (approved):

PR #516, opened 2022-01-29
Update Laser Trails after Convert.Deploy transformation

FS-21:
Convert.Deploy is an Ares tag so is necessary an extra check for loading the right Laser Trails when the unit is transformed.


Metadorius (approved):

FS-21:
I tested it with;

Unit "A" without LaserTrails.
Unit "B" with 2 LaserTrails.
Checked that each time I deploy the unit from A->B and B->A the LaserTrails get updated (each time the unit becomes "A" it loose the LaserTrails it had as "B").


Unit "A" with red LaserTrails.
Unit "B" with 2 while LaserTrails.
Checked that after the Unit transformations it gets the right LaserTrails.

PR #519, opened 2022-01-30
Lasertrails fixes when units enter/leave transports

FS-21:
- Fix: Units garrisoned inside Opentopped units generated LaserTrails.
- Fix: A LaserTrailer line is generated between the original location the units had when entered into the transport and the location when they left the transport.


FS-21:
Re-made the hooks code due to the feedback.

PR #524, opened 2022-02-07
Customization of the storage resource

FS-21:
[General] > Storage.TiberiumIndex=-1
This new tag set which type of Tiberium (from [Tiberiums] list) is used for storage, making it more easy to respect/calculate money limitations.

It needs Ares Storage logic activation for working.


SMxReaver:
oretest

Tested this feature out, pretty neat. Essentially it converts all other resource types into the specified resource by the tag. In this test, you see a Chrono Miner with Gems deposit the resource as stored ore into the refinery, mine ore offscreen, return to deposit and it is stored ore as well. This persists through save games, the feature works on multiple refineries and if one is destroyed the others remain working as intended.

src/Ext/Building/Body.cpp
```diff
@@ -3,6 +3,33 @@
...
+
+	if (amount > 0.0)
+	{
+		auto pBuildingType = pThis->Type;
+
+		if (pBuildingType)
+		{
+			auto const pExt = BuildingTypeExt::ExtMap.Find(pBuildingType);
+
+			if (pExt)
```
Thrifinesma, line 20:
```suggestion
		if (auto pBuildingType = pThis->Type)
		{
			if (auto const pExt = BuildingTypeExt::ExtMap.Find(pBuildingType))
```

Thrifinesma (changes requested):

FS-21:
Applied the last requested change from the review.

Metadorius (approved):

PR #529, opened 2022-02-18
Minor Save fix

secsome:
This PR has been already tested by hundreds of MO336 players which is known as `Extra Save Fix` in it, it provides fixes like:
Map lighting get corrupted after load the save
BuildOfAlly is reset after load the save

It also include a new trigger action called AdjustLighting which is more detailed. Which means you can choose to only update CellDrawers, Objects, and CustomPalettes. That has been tested already too, by many mappers like Sandman, Esther and so on.


secsome:
This dll conflicts with the savefix dll in mo336, do not test them together.

src/Ext/TAction/Hooks.cpp
```diff
@@ -70,49 +72,15 @@ DEFINE_HOOK(0x6E427D, TActionClass_CreateBuildingAt, 0x9)
...
-
-			pRadSite->LightSource = GameCreate<LightSourceClass>(coord,
-				visibility, intensity, color);
-
-			pRadSite->LightSource->Activate();
-		}
-	}
+	ScenarioExt::Global()->CurrentTint_Tiles =
+	ScenarioExt::Global()->CurrentTint_Schemes =
+	ScenarioExt::Global()->CurrentTint_Hashes = ScenarioClass::Instance->NormalLighting.Tint;
```
Metadorius, line 83: Not sure what I am even looking at.

Metadorius (changes requested):
Lacks docs, changelog, credits.

src/Ext/TAction/Hooks.cpp
chaserli, line 83: May I ask what's the status of this PR currently? Is it equivalent to savefixmo.dll in mo336?

secsome: > May I ask what's the status of this PR currently? Is it equivalent to savefixmo.dll in mo336?

Mo336's fix is a bugged fix version, this is the complete version of it

Metadorius:
@secsome add the damn docs and let's merge before 0.3 comes out ffs
@chaserli may help with testing for example?

chaserli:
> @secsome add the damn docs and let's merge before 0.3 comes out ffs @chaserli may help with testing for example?
After the branch gets updated. I don't quite understand the code here.

chaserli:
Which part is responsible for BuildOffAlly S/L fix? I don't quite get it

secsome:
> Which part is responsible for BuildOffAlly S/L fix? I don't quite get it Is it the serialization of `SessionClass::Instance->Config`?

yes, `.Process(SessionClass::Instance->Config)` does the fix, as YR didn't do  serialization on `SessionClass`, I decided to simply serialize it in `ScenarioClass`

chaserli:
AED040XDG} 3DCLKP(@UGM6
I didn't S/L, just after time-freeze it becomes like this

src/Ext/TAction/Body.cpp
```diff
@@ -261,6 +264,48 @@ bool TActionExt::BinaryOperation(TActionClass* pThis, HouseClass* pHouse, Object
...
+	if (pThis->Param4 != -1)
+		ScenarioClass::Instance->NormalLighting.Tint.Green = pThis->Param4;
+	if (pThis->Param5 != -1)
+		ScenarioClass::Instance->NormalLighting.Tint.Blue = pThis->Param5;
+
+	const int r = ScenarioClass::Instance->NormalLighting.Tint.Red * 10;
+	const int g = ScenarioClass::Instance->NormalLighting.Tint.Green * 10;
+	const int b = ScenarioClass::Instance->NormalLighting.Tint.Blue * 10;
+
+	if (pThis->Value & 0b001) // Update Tiles
```
chaserli, line 280: Can you add the docs related to this TAction? and how to give `TActionClass::Value` in a map editor?

src/Ext/Scenario/Body.cpp
```diff
@@ -57,6 +61,53 @@ void ScenarioExt::ExtData::ReadVariables(bool bIsGlobal, CCINIClass* pINI)
...
+			GameDelete(pBld->LightSource);
+			if (pBld->Type->LightIntensity)
+			{
+				TintStruct color { pBld->Type->LightRedTint, pBld->Type->LightGreenTint, pBld->Type->LightBlueTint };
+
+				pBld->LightSource = GameCreate<LightSourceClass>(pBld->GetCoords(),
+					pBld->Type->LightVisibility, pBld->Type->LightIntensity, color);
+
+				if (wasActivated)
+					pBld->LightSource->Activate();
```
chaserli, line 81: I get the whole picture of what you have done here, but I can't find the possible cause of some light posts failing to update light sources as I showed in the conversation before.

chaserli (commented):
I updated your branch with upstream a few days ago and refactored a few stuff for my better comprehension, can you check if that still conforms to what you would have meant?

secsome:
screenshot
I cannot reproduce the issue you mentioned here.

This feature should have a tag to switch it on, I will add a tag later.

src/Ext/Scenario/Body.cpp
secsome: I cannot reproduce it even I've did more than 10 tries. I'm not sure if it has some problem here.

src/Ext/TAction/Body.cpp
secsome: Well, in map editor we just write a number, 1 stands for `Tiles`, 2 stands for `Units & Buildings`, 3 stands for `Tiles &  Units and Buildings`, ... and so on. The document should be implemented later.

Metadorius:
Err why exactly is the switch needed?

chaserli:
> Err why exactly is the switch needed?

I think he may refer to the issue I reported above, although I think it's actually not quite needed too

secsome:
There are some existing maps set the lighting without the fix, if you don't give an option  to disable the lighting fix, the existing maps will looks bad. So this is just a forward compatibility.

Metadorius:
so what's the actual state😵 
image

src/Ext/Rules/Body.cpp
```diff
@@ -102,6 +102,8 @@ void RulesExt::ExtData::LoadBeforeTypeData(RulesClass* pThis, CCINIClass* pINI)
 
 	this->IronCurtain_KeptOnDeploy.Read(exINI, "CombatDamage", "IronCurtain.KeptOnDeploy");
 
+	this->AdjustLightingFix.Read(exINI, "Map", "AdjustLightingFix");
```
chaserli, line 105: I prefer `[Basic]` or even `[Lighting]` for the sake of clarity

secsome:
What needs to be tested:
1. Reveal All Command
2. `BuildOfAlly`
3. TAction 507 `AdjustLighting`
4. `[Basic]>AdjustLightingFix`

docs/Fixed-or-Improved-Logics.md
```diff
@@ -4,8 +4,9 @@ This page describes all ingame logics that are fixed or improved in Phobos witho
 
 ## Bugfixes and miscellaneous
 
-- Fixed the bug when reading a map which puts `Preview(Pack)` after `Map` lead to the game fail to draw the preview
-- Fixed the bug when retinting map lighting with a map action corrupted light sources.
+- Fixed the bug when reading a map which puts `Preview(Pack)` after `Map` lead to the game fail to draw the preview.
+- Fixed the bug when retinting map lighting with a map action corrupted light sources. You need to manually enable it by setting `[Basic]>AdjustLightingFix` to **true**.
+- Fixed the bug that `GameModeOptions` are not being saved. For example, `BuildOffAlly` will be corrupted when loading the game.
```
chaserli, line 9:
```suggestion
- Fixed the bug that `GameModeOptions` are not being saved. For example, `BuildOffAlly` will no longer be corrupted when loading the game.
```
Is that the correct logic?

src/Commands/RevealAll.h
```diff
@@ -0,0 +1,48 @@
+#pragma once
```
chaserli, line 1: You broke the convention set by yourself 0efa15c23f91eb51abc658585a325dc0f6c25e14

chaserli (changes requested):
It seems we won't be able to make it for v0.3, but at least we can add that BuildOffAlly part into v0.3.

src/Ext/Scenario/Body.cpp
```diff
@@ -81,10 +132,15 @@ void ScenarioExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
...
-
+	// Initialize
+	DefaultAmbientOriginal = ScenarioClass::Instance->AmbientOriginal;
+	DefaultAmbientCurrent = ScenarioClass::Instance->AmbientCurrent;
+	DefaultAmbientTarget = ScenarioClass::Instance->AmbientTarget;
+	DefaultNormalLighting = ScenarioClass::Instance->NormalLighting;
 
+	CurrentTint_Tiles = ScenarioClass::Instance->NormalLighting.Tint;
 
+	AdjustLightingFix = pINI->ReadBool("Basic", "AdjustLightingFix", AdjustLightingFix);
```
chaserli, line 143: Are you sure that your parsing is successful in this `ScenarioClass_LoadFromINI` hook? `pINI` seems to be released somehow. Should we move it a bit earlier? such as
```cpp
DEFINE_HOOK(0x68AD2F, ScenarioClass_LoadFromINI, 0x5)
{
	GET(ScenarioClass*, pItem, ESI);
	GET(CCINIClass*, pINI, EDI);
	ScenarioExt::LoadFromINIFile(pItem, pINI);
	return 0;
}
```

PR #530, opened 2022-02-18
Refactor the FogOfWar

### LINKED ISSUE #28: [Major] Fix vanilla YR Fog of War bugs & issues
Fog of War exists in code YR, but was disabled due to not working correctly. CNCNet Client spawner allows to force enable FoW, however this causes several issues:

- Units are still visible under the FoW,
- Structures under the FoW (mostly high civilian buildings) render incorrectly, usually with transparent triangles, This also happens when the top of a building is visible, but the foundation is not,
- Structures under the FoW cannot be properly targeted (units issue move command instead of attack/garrison/repair), unless the cursor is on the top edge of the top foundation cell.
- Jumpjet units and aircraft provide vision only when moving or landed (possibly shroud uncovering check is not tied to ground cells, but FoW refresh is?),
- Jumpjet and aircraft have a global vision range specified in rules->[General]->AircraftFogReveal, could be deglobalised to work with unit sight,
- High fog refresh rate (set in rules->[AudioVisual]->FogRate) causes flickering and effectively makes air units provide no vision,
- On the other hand, low fog refresh rate causes units to leave a decaying trail of vision behind them (this happens with any unit, but for slow units disappears very quickly, might be additionally linked to movement-into-two-cells issue?),
- Temporary drawing glitches, similar to ones that rarely happen with uncovered shroud in vanilla YR (black outline of cell remains until we scroll away and back, or black gradient around "old" vision that quickly disappears)
- Possibly more yet unknown bugs caused by Ares' new features

secsome:
resolves #28

### DIFF
--- a/Phobos.vcxproj
+++ b/Phobos.vcxproj
@@ -21,6 +21,7 @@
     <ClCompile Include="src\Commands\Commands.cpp" />
     <ClCompile Include="src\Commands\Dummy.h" />
     <ClCompile Include="src\Ext\AnimType\Hooks.cpp" />
+    <ClCompile Include="src\Ext\Cell\Body.cpp" />
     <ClCompile Include="src\Ext\Scenario\Hooks.Variables.cpp" />
     <ClCompile Include="src\Ext\Sidebar\Body.cpp" />
     <ClCompile Include="src\Ext\Sidebar\Hooks.cpp" />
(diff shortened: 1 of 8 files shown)


CatTanker:
Looking Good!

My first comment is ive noticed quite a few issues when drawing overlay (especially bridges) from underneath Shroud/Fog
I also saw some weird ore/gem flickering, but it wasnt as repeatable so i have no evidence
image


Animations such as smoke appear underneath the fog


If you look closely into the shroud, you can see the terrain behind it to an extent
However, in my image the heights of the terrain are 4-8, and are very dim
image

In other maps which have lower ground on 0, you can barely see anything under the shroud
image

Not sure if this corresponds with map lighting, but either way it is fairly inconsistent

ayylmaoRotE:
Here are my findings:
1. SpySat=yes does not work and fog remains intact.
2. Engineers can be ordered to enter structures (tech structures) through fog, but when they get to the structure they just stand still.
3. Units attacking my building were concealed by fog, after they destroyed the building they also got concealed by shroud, guessing this is to do with RevealOnFire on weapons - since it defaults to yes:
image
4. I am able to see laserlinetrails through fog (projectiles and units):
image
5. Superweapons with RevealToAll show up but are not visible, if that makes sense:
image
6. Got a Exception code: C0000005 at 2D8854E0, uploaded file to #phobos-chat
7. I think a SW.FireIntoFog tag needs to be implemented as this is such a prominent feature, SW.FireIntoShroud is completely ignored through fog, and if you have a Generals style mod where there is no shroud, only fog, this tag becomes useless

ayylmaoRotE:
Played online with Cat today, no desyncs or FoW related crashes. Same issues as above still apply.

Another thing we noticed is that you can place buildings in fog provided enough adjacency, should probably respect fog as it does shroud.

CatTanker:
To add, performance wise it was fine
On an average/standard connection/tunnel, we achieved around 43fps on average, aka no noticeable performance drops in comparison to normal gameplay.

PR #532, opened 2022-02-18
Units retaining orders after changing ownership bugfix

### LINKED ISSUE #314: [Vanilla Bug] Units retaining orders after changing ownership.

secsome:
Resolves #314

### DIFF
--- a/src/Misc/CaptureManager.cpp
+++ b/src/Misc/CaptureManager.cpp
@@ -124,6 +124,9 @@ bool CaptureManager::CaptureUnit(CaptureManagerClass* pManager, TechnoClass* pTa
 					}
 				}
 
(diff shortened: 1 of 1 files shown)


Lylaktech:
Problem persists:





I also saw infantry being ejected from Drakuv Prison Vehicle showing the same bug.

PR #536, opened 2022-02-20
Allow centering attached anims on object center

Starkku:
### Attached animation position customization

- You can now customize whether or not animations attached to objects are centered at the object's actual center rather than the bottom of their top-leftmost cell (cell #0).

In `artmd.ini`:
```ini
[SOMEANIM]                       ; AnimationType
UseCenterCoordsIfAttached=false  ; boolean
```

The effect is most noticeable on buildings larger than 1x1, but even on 1x1 buildings and units it will shift the animation up half a cell (15 pixels or so).
animcenter


CommanderNour:
Works as intended :thumbsup:

PR #537, opened 2022-02-21
New tag for overriding an underwater attack case

FS-21:
Default YR behavior with `NavalTargeting=1` is always attack underwater (and uncloaked underwater submarines) with secondary weapon and anything else with primary weapon.
This overrides a part of the vanilla YR logic for allowing naval units to use a different weapon if the underwater unit is uncloaked.
Useful if you want a unit with 1 weapon for underwater objects and another only for surface objects.

[SOMETECHNO]
ForceWeapon.Naval.Decloacked=-1   ; 0 for primary weapon, 1 for secondary weapon


FS-21:
Visual example
underwater-new-attack-tag

Metadorius (approved):

PR #538, opened 2022-02-26
Shrapnel enhancement

secsome:
> Allow the behavior to trigger regardless of whether it hits a unit

```ini
[SOMEPROJECTILE]
Shrapnel.AffectsGround=no ; boolean
Shrapnel.AffectsBuildings=no ; boolean
```


SMxReaver:
In my tests, I disabled Shrapnel effects from buildings and disabled Shrapnel effects from the ground. I tested the inverse as well, and even through a Save/Load and being inside Bunkers. 
- The only effect that is not changed by this feature are Shrapnel effects on Vehicles/Infantry, as the effect happens regardless of AffectsGround=yes/no. (I imagine this is out of scope for this feature)

test1
In this test, I used:
[SOMEPROJECTILE]
Shrapnel.AffectsGround=yes
Shrapnel.AffectsBuildings=no

PR #539, opened 2022-02-27
Shared Ammo from OpenTopped transports to passengers

FS-21:
OpenTopped Transports with this Ammo.Shared tag refill ammo to passengers that have the same Ammo.Shared tag.

A transport & passengers with the same Ammo.Shared.Group = <integer> value is used to filter passengers that can be refilled, ignoring other passengers with different groups values.

Ammo.Shared.Group=-1 in transports means that any fillable passenger is supported by the transport.

Transports must have Ammo.


FS-21:
I already made all kind of tests for confirming that the generated ammo in the OpenTopped transport can be transfered to compatible passengers.

Metadorius (approved):

PR #540, opened 2022-03-01
New case 36 for some new Attack/Move actions -

FS-21:
Returns true if the analyzed object is a building that isn't a defense.

We have case 2 (any building) and case 7 (Only defenses) but not "Buildings that aren't defenses".

Moreover, added missing scripts documentation.


Falaka21:
This is more likely “ an aggressive stance “ like in RA3 or GenZH guard area command ? 🤔

FS-21:
> This is more likely “ an aggressive stance “ like in RA3 or GenZH guard area command ? 🤔

Sorry, you made a mistake.

New actions of the type attack/move have 35 possible cases (the original RA2 action 0,x had only 9). This is simply a new case number 36 for evaluating only non-neutral structures that doesn't have "IsBaseDefense=yes".

For information about the original RA2 action 0,x check it at Modenc:
https://modenc.renegadeprojects.com/ScriptTypes/ScriptActions

For more Phobos information:
[Read the documentation](https://phobos--540.org.readthedocs.build/en/540/New-or-Enhanced-Logics.html#generic-target-type-attack-action)

Also, I tested the feature and AI (in the log file) always picks structures that aren't defenses, if anyone want to check I appreciate it :-)

Metadorius (approved):

SMxReaver:
This appears to be working as intended, I both notice the AI change from the original script action into a script action targeting only non-defensive buildings. Log file confirms this behavior, I think this one is also good to go!

Metadorius (approved):

PR #543, opened 2022-03-04
Fixed the bug when reading a map which puts `Preview(Pack)` after `Map` lead to the game fail to draw the preview

secsome:
In vanilla YR map, If you put the `[Preview]` or `[PreviewPack]` after `[Map]` section, the game won't read the preview.
This Pull Request fixed that stupid Westwood genius `optimization` and make it work as a real INI file.


Starkku:
Code's fine and it works so I'll merge. I still think that having the order FA2sp saves the map sections in be compatible with all versions of the game would also be a good idea but eh.

PR #545, opened 2022-03-10
Critical hits vs. empty cells change

Starkku:
This allows critical hits to occur when weapon is fired at empty cell if `Crit.Warhead` is set and `Crit.Affects` contains either `all` or `empty`.


CommanderNour:
Works as intended

Starkku:
Closing as critical hits occuring on empty cells has been deemed out of scope by consensus. Will look into implementing target-agnostic extra warhead detonation with chance/weight controls as a substitute instead.

PR #548, opened 2022-03-10
Building Placement Preview 

Otamaa:
- No documentation included atm 
```ini
;RA2MD
[Phobos]
ShowBuildingPlacementPreview =  ;bool , globally setting

[AUDIOVISUAL]
BuildingPlacementGrid.TranslucentLevel= ;int , default to 0 , {0=0% , 1=25%  , 2=50% , 3=75%}
BuildingPlacementPreview.DefaultTranslucentLevel = ;int , default to 3 , {0=0% , 1=25%  , 2=50% , 3=75%}

[BUILDINGTYPE]
PlacementPreview.Show=  ;bool , default to [Phobos]->ShowBuildingPlacementPreview
PlacementPreview.Shape=  ;SHPFile  (default to BuildupImage / Building->Image)
PlacementPreview.ShapeFrame = ;int (default to BuildupImage last frame - 1 / Building->Image frame 0)
PlacementPreview.Offset=  ;int int int , default to {0 , -15 , 1}
PlacementPreview.Remap=   ; bool , default to true 
PlacementPreview.Palette=  ; .PAL file , default to nullptr , this option not used  if remap
PlacementPreview.TranslucentLevel = ;int , default to `AUDIOVISUAL->BuildingPlacementPreview.DefaultTranslucentLevel`
```
image


CommanderNour:
Every tag works except 
**PlacementPreview.Remap=false**
SCRN 20220311-202529-00441
for some reason it shows wrong palette if used
**PlacementPreview.Shape=NAREFN**
SCRN 20220311-202857-00562
it should show me a soviet refinery instead of the allied one right ?!
I tried it on different buildings but it didn't work as intended
also is it possible to show the building BibShape with the building itself

Otamaa:
> Every tag works except **PlacementPreview.Remap=false** SCRN 20220311-202529-00441 for some reason it shows wrong palette if used **PlacementPreview.Shape=NAREFN** SCRN 20220311-202857-00562 it should show me a soviet refinery instead of the allied one right ?! I tried it on different buildings but it didn't work as intended also is it possible to show the building BibShape with the building itself

it is better to use `PlacementPreview.Shape=` option 
including Bib , it mean i need to handle separate drawing function and calculation for that , i dont want to do that 
too much code for a simple feature !

src/Ext/BuildingType/Hooks.cpp
```diff
@@ -59,4 +64,108 @@ DEFINE_HOOK(0x458623, BuildingClass_KillOccupiers_Replace_MuzzleFix, 0x7)
 	}
 
 	return 0;
-}
\ No newline at end of file
+}
+
+DEFINE_HOOK(0x6D528A, TacticalClass_DrawPlacement_PlacementPreview, 0x6)
+{
+	ObjectClass* pObject = Make_Global<ObjectClass*>(0x88098C);
```
secsome, line 71: `DisplayClass::Instance->CurrentBuilding`

src/Ext/BuildingType/Hooks.cpp
```diff
@@ -59,4 +64,108 @@ DEFINE_HOOK(0x458623, BuildingClass_KillOccupiers_Replace_MuzzleFix, 0x7)
...
+
+				if (!pCell)
+					return 0x0;
+
+				auto const nHeight = pCell->GetFloorHeight({ 0,0 });
+				Point2D nPoint{ 0,0 };
+				TacticalClass::Instance->CoordsToClient(CellClass::Cell2Coord(pCell->MapCoords, nHeight),&nPoint);
+
+				auto nFlag = BlitterFlags::Centered | BlitterFlags::Nonzero | BlitterFlags::MultiPass;
+				auto const nFlag_ = Math::clamp(pTypeExt->PlacementPreview_Transculency.Get(), 0, 3);
```
secsome, line 99: I noticed that this code block is duplicated, maybe wrap it as a function would be a wise decision.

src/Ext/BuildingType/Hooks.cpp
```diff
@@ -59,4 +64,108 @@ DEFINE_HOOK(0x458623, BuildingClass_KillOccupiers_Replace_MuzzleFix, 0x7)
...
+//Make Building placement Grid tranparent
+static void __fastcall CellClass_Draw_It_Shape(Surface* Surface, ConvertClass* Palette, SHPStruct* SHP, int FrameIndex,
+	const Point2D* const Position, const RectangleStruct* const Bounds, BlitterFlags Flags,
+	int Remap,
+	int ZAdjust, // + 1 = sqrt(3.0) pixels away from screen
+	ZGradient ZGradientDescIndex,
+	int Brightness, // 0~2000. Final color = saturate(OriginalColor * Brightness / 1000.0f)
+	int TintColor, SHPStruct* ZShape, int ZShapeFrame, int XOffset, int YOffset)
+{
+	auto const nFlag_ = Math::clamp(RulesExt::Global()->PlacementGrid_Transculency.Get(), 0, 3);
```
secsome, line 150: What I've mentioned above

secsome (changes requested):

CommanderNour:
The remap works
SCRN 20220312-055359-00662
For the preview shape image I really don't know 
here I used **PlacementPreview.Shape=NARADR**
SCRN 20220312-061050-00391
didn't work as shown in the image above
and here I used **PlacementPreview.Shape=EAREFN**
SCRN 20220312-061035-00680
works but displays a wrong theatre image it should pick the  second filename character A for snow ,D for desert and G for everything etc....
right ?!
I think you should wait for someone else to test this tag so you can make sure it works fine
also there is a bug with the allied barracks a 3x2 building
SCRN 20220312-055511-00582
if I went to look elsewhere in the map and get back to it the bug disappears as shown with one of the barracks
also why it shows in the Construction yard it doesn't related to the barrack

CommanderNour:
Works as intended

src/Ext/BuildingType/Hooks.cpp
```diff
@@ -59,4 +65,74 @@ DEFINE_HOOK(0x458623, BuildingClass_KillOccupiers_Replace_MuzzleFix, 0x7)
...
+				TacticalClass::Instance->CoordsToClient(CellClass::Cell2Coord(pCell->MapCoords, nHeight),&nPoint);
+
+				auto const nFlag = BlitterFlags::Centered | BlitterFlags::Nonzero | BlitterFlags::MultiPass | EnumFunctions::GetTranslucentLevel(pTypeExt->PlacementPreview_TranslucentLevel.Get());
+				auto const nREct = DSurface::Temp()->GetRect();
+				auto const nScheme = ColorScheme::Array()->GetItem(HouseClass::Player->ColorSchemeIndex);
+				auto const nPalette = pTypeExt->PlacementPreview_Remap.Get() ? nScheme->LightConvert : pTypeExt->PlacementPreview_Palette.GetOrDefaultConvert(FileSystem::UNITx_PAL());
+				Point2D nDefault{ 0 , -15 };
+				nPoint += pTypeExt->PlacementPreview_Offset.Get(nDefault);
+
+				DSurface::Temp()->DrawSHP(
```
Starkku, line 104: I don't think there's any need to lay out the params on their own lines here. This type of approach is useful in locations that are modified often, but here it at the best provides tiny bit of extra clarity at expense of taking a lot of vertical space.

Starkku (changes requested):
Code for most part looks okay, only some small things to point out.

As for the actual feature, I think using the `BuildUp` shape's last regular frame as the default if `PlacementPreview.Shape` is undefined, instead of simply the building's basic shape is the most sensible default option from modder's standpoint, as it would require the least amount of additional code to allow the feature to be used.

src/Ext/BuildingType/Hooks.cpp
```diff
@@ -59,4 +65,74 @@ DEFINE_HOOK(0x458623, BuildingClass_KillOccupiers_Replace_MuzzleFix, 0x7)
...
+					nPalette,
+					pImage,
+					nFrame,
+					&nPoint,
+					&nREct,
+					nFlag,
+					0,
+					0, ZGradient::Ground ,
+					 1000, 0, nullptr, 0, 0, 0
```
Starkku, line 113: I don't think there's any need to lay out the params on their own lines here. This type of approach is useful in locations that are modified often, but here it at the best provides tiny bit of extra clarity at expense of taking a lot of vertical space.

Starkku, line 113: I don't think there's any need to lay out the params on their own lines here. This type of approach is useful in locations that are modified often, but here it at the best provides tiny bit of extra clarity at expense of taking a lot of vertical space.

Otamaa:
> Code for most part looks okay, only some small things to point out.
> 
> As for the actual feature, I think using the `BuildUp` shape's last regular frame as the default if `PlacementPreview.Shape` is undefined, instead of simply the building's basic shape is the most sensible default option from modder's standpoint, as it would require the least amount of additional code to allow the feature to be used nicely with existing assets.

hmm , yeah 
i just dont wat to bother to figuering out which frames i need to use from build up 
easier to test with building image

Metadorius:
I believe trsdy on discord has already made a change to accomodate that.
```cpp
DEFINE_HOOK(0x6D528A, TacticalClass_DrawPlacement_PlacementPreview, 0x6)
///......
auto const pImage = pType->LoadBuildup(); //pTypeExt->PlacementPreview_Shape.Get(pType->GetImage());
if (!pImage)
    return 0x0;
auto const nFrame = (int)((pImage->Frames)/2-1); //Math::clamp(pTypeExt->PlacementPreview_ShapeFrame.Get(), 0, (int)pImage->Frames);
```

Otamaa:
> I believe trsdy on discord has already made a change to accomodate that.
> 
> ```c++
> DEFINE_HOOK(0x6D528A, TacticalClass_DrawPlacement_PlacementPreview, 0x6)
> ///......
> auto const pImage = pType->LoadBuildup(); //pTypeExt->PlacementPreview_Shape.Get(pType->GetImage());
> if (!pImage)
>     return 0x0;
> auto const nFrame = (int)((pImage->Frames)/2-1); //Math::clamp(pTypeExt->PlacementPreview_ShapeFrame.Get(), 0, (int)pImage->Frames);
> ```

yeah , done 
you can check latest code i push

Metadorius:
Another improvement from trsdy, support for building upgrades:
```cpp
const bool isUpgrade = pType->PowersUpBuilding[0] || (!pTypeExt->PowersUp_Buildings.empty());
auto const pImage = isUpgrade ? pType->GetImage() : pType->LoadBuildup();
if (!pImage) return 0x0;
const int nFrame = isUpgrade ? 0 : (int)((pImage->Frames)/2-1);
```

Metadorius:
Another thing that should be in - it should be disable-able via a flag in RA2MD, because this is a UI feature that not all players may necessarily like. See how I made it for smart selection filtering.

Otamaa:
> Another improvement from trsdy, support for building upgrades:
> 
> ```c++
> const bool isUpgrade = pType->PowersUpBuilding[0] || (!pTypeExt->PowersUp_Buildings.empty());
> auto const pImage = isUpgrade ? pType->GetImage() : pType->LoadBuildup();
> if (!pImage) return 0x0;
> const int nFrame = isUpgrade ? 0 : (int)((pImage->Frames)/2-1);
> ```

i mean , what is the point now 
my first implementation is open enough you can change the shape image manually if the thing not supported and or user dont like it 
Now you want add another check for upgrades ? 
yeah , single boolean wont be too much but , why ,..

Metadorius:
Because things working automagically where possible is a good thing for the end user of Phobos. This was already implemented by someone so why not add it if it's good enough? A few lines of code like this saves a lot of people's time, because with this they can just add a toggle to show the preview and be done with it instead of having to look up how to fix the wrong image.

This is not meant as a criticism in any way.

Otamaa:
> Another thing that should be in - it should be disable-able via a flag in RA2MD, because this is a UI feature that not all players may necessarily like. See how I made it for smart selection filtering.

new catagory ? or should i put it in inside `Sidebar` ?

Metadorius:
I am pretty sure Sidebar is a category in UIMD.ini, not in RA2MD.ini which is a user settings file.

chaserli:
This is Trsdy from discord. I noticed that you've changed the type of `BuildingTypeExt::ExtData::PlacementPreview_ShapeFrame ` from `Valuable<int>` to `Nullable<int>`. As far as I've tested, it always get 0 when `this->PlacementPreview_ShapeFrame.Read(exINI, pSection, "PlacementPreview.ShapeFrame")` is executed

chaserli:
> > Another improvement from trsdy, support for building upgrades:
> > ```c++
> > const bool isUpgrade = pType->PowersUpBuilding[0] || (!pTypeExt->PowersUp_Buildings.empty());
> > auto const pImage = isUpgrade ? pType->GetImage() : pType->LoadBuildup();
> > if (!pImage) return 0x0;
> > const int nFrame = isUpgrade ? 0 : (int)((pImage->Frames)/2-1);
> > ```
> 
> i mean , what is the point now my first implementation is open enough you can change the shape image manually if the thing not supported and or user dont like it Now you want add another check for upgrades ? yeah , single boolean wont be too much but , why ,..

The reason why I suggest such change is that I want to use the last frame of buildup or the first frame of powersup-workinganim to be the preview frame, and this seems pretty natural to a modder's logic. If I want to do so with your previous versions, I'll have to know the frame number of every buildup anim a prior. 

Maybe the case when buildup anim is absent should also be taken into consideration in this case.

Starkku (approved):
Rebased and fixed some issues like custom paletted buildings being displayed in unit*.pal with the remap option on, and some code style tweaks. Added documentation by @SMxReaver from his PR #591.

On the upgrade support, they work fine by default if you omit `Buildup` which usually is omitted for upgrades anyway.

The only real remaining issue there could be to fix is remappable previews tinting under superweapon or map trigger tint lighting

image

However I believe that fixing this and maintaining the ability to keep the image remapped to house's color scheme would be somewhat complex and require more research so it could be left as it is for now.

Not seeing any real issues left to fix so approving for merge.

src/Ext/BuildingType/Hooks.cpp
```diff
@@ -59,4 +66,79 @@ DEFINE_HOOK(0x458623, BuildingClass_KillOccupiers_Replace_MuzzleFix, 0x7)
...
+				SHPStruct* Selected = nullptr;
+				bool bBuildupExist = false;
+
+				if (pType->LoadBuildup())
+				{
+					bBuildupExist = true;
+					Selected = pType->LoadBuildup();
+				}
+				else
+				{ Selected = pType->GetImage(); }
```
Metadorius, line 94:
```suggestion
				{
					Selected = pType->GetImage();
				}

```

src/Ext/BuildingType/Hooks.cpp
```diff
@@ -59,4 +66,79 @@ DEFINE_HOOK(0x458623, BuildingClass_KillOccupiers_Replace_MuzzleFix, 0x7)
...
 
 	return 0;
-}
\ No newline at end of file
+}
+
+static SHPStruct* GetBuildingShape(BuildingTypeClass* pThis)
+{
+
+}
```
Metadorius, line 74:
```suggestion
```

PR #549, opened 2022-03-10
Mastermind Overload characteristic deharcoded 

Otamaa:
```ini
[MASTERMIND]
Overload.Count=               ; list of integer, default to [CombatDamage]->OverloadCount
Overload.Damage=              ; list of integer, default to [CombatDamage]->OverloadDamage
Overload.Frames=              ; list of integer, default to [CombatDamage]->OverloadFrames
Overload.DeathSound=          ; Sound entry, default to [AudioVisual]->MasterMindOverloadDeathSound
Overload.ParticleSys=         ; ParticleSystemType, default to [CombatDamage]->DefaultSparkSystem
Overload.ParticleSysCount=5   ; integer
MindControlLink.VisibleToHouse=all ; Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
```


Otamaa:
> **Overload.Count=5,7,9,11** **Overload.Damage=0,25,50,1000** **Overload.Frames=30,60,60,600** **Overload.DeathSound=NukeSiren** **Overload.ParticleSys=SmallGreySSys** **MindControll.DrawLink=false**
> 
> If my codes here are correct the only tags that worked for me is **Overload.Count** & **MindControll.DrawLink** the reset didn't work but they don't fall back to the old numbers meaning this SCRN 20220311-200637-00632 The Master Mind is at full HP

it seems the parser is broken , they not parsing the values , should be fixed now !

CommanderNour:
the **Overload.DeathSound** should play when the Master Mind dies right ? it plays when it reach any of the Overload Counts

beside that every thing works as intended

Otamaa:
> the **Overload.DeathSound** should play when the Master Mind dies right ? it plays when it reach any of the Overload Counts
>

It is not apparently ,but i why WW name this like that , omg  :kekw:

SMxReaver:
Using YR 1.001, I've added the following tags to the Lasher Tank for this test. I've found the following based on the newest available Nightly.

Overload.Count does not appear to be working correctly. When set to 'Overload.Count=5,7,11,55' , the game will start inflicting damage on the 8th captured object. In Vanilla YR, an 'OverloadCount=5,7,11,55' would start inflicting damage on the 6th captured object. 

Overload.Damage seems to be working correctly even with the issues described in Overload.Count. _Observed behavior: if you set the first value to above 0 (Overload.Damage=1,50,100,500 for example), the unit will immediately take damage upon construction without any captured victims._

Overload.Frames seems to be working correctly even with the issues described in Overload.Count.

Overload.DeathSound= is also working. Is it possible that a tag rename to 'Overload.VoiceFeedback' to indicate this is played when the unit is taking damage? The current game uses VoiceFeedback for units that hit yellow health and I think it may indicate it's functionality clearer.

Overload.ParticleSys and Overload.ParticleSysCount both seem to be working. Overload.ParticleSys= confirmed to work on Spark, Smoke and does not work with Gas, Fire, did not test Railgun, but I have my doubts.

MindControll.DrawLink= is also working. It has a typo in the tag name, maybe it is supposed to be MindControl.DrawLink= without the second L in MindControl?

src/Ext/CaptureManager/Hooks.cpp
```diff
@@ -49,3 +52,117 @@ DEFINE_HOOK(0x519F71, InfantryClass_UpdatePosition_BeforeBuildingChangeHouse, 0x
...
+			if (OverloadCount.empty())
+				return;
+
+			int nCurIdx = 0;
+			int const nNodeCount = pThis->ControlNodes.Count;
+
+			for (int i = 0; i < (int)(OverloadCount.size()); ++i)
+			{
+				if(nNodeCount > OverloadCount[i])
+					nCurIdx = i + 1; //select the index !
```
CrimRecya, line 113: This is different from the original logic, as it is possible to obtain an index value equal to size when the maximum value is exceeded, and may out of bounds.

src/Ext/CaptureManager/Hooks.cpp
```diff
@@ -49,3 +52,117 @@ DEFINE_HOOK(0x519F71, InfantryClass_UpdatePosition_BeforeBuildingChangeHouse, 0x
...
+			}
+
+			// prevent nCurIdx selecting out of bound index !
+			constexpr auto FixIdx = [](const Iterator<int>& iter, int nInput) {
+				return iter.empty() ? 0 : iter[nInput > (int)iter.size() ? (int)iter.size() : nInput];
```
CrimRecya, line 118: [size()] should be out of bounds.

src/Ext/CaptureManager/Hooks.cpp
```diff
@@ -49,3 +52,117 @@ DEFINE_HOOK(0x519F71, InfantryClass_UpdatePosition_BeforeBuildingChangeHouse, 0x
...
+						auto const nRandomY = ScenarioClass::Instance->Random.RandomRanged(-200, 200);
+						auto const nRamdomX = ScenarioClass::Instance->Random.RandomRanged(-200, 200);
+						CoordStruct nParticleCoord{ pOwner->Location.X + nRamdomX, nRandomY + pOwner->Location.Y, pOwner->Location.Z + 100 };
+						GameCreate<ParticleSystemClass>(pParticle, nParticleCoord, nullptr, nullptr, CoordStruct::Empty, nullptr);
+					}
+				}
+
+				if (nCurIdx > 0 && pOwner->IsAlive && pOwner->Health > 0 && !pOwner->InLimbo)
+				{
+					double const nBase = (nCurIdx != 1) ? 0.015 : 0.029999999;
```
CrimRecya, line 155: I think it's 0.03, just a small thing.

src/Ext/CaptureManager/Hooks.cpp
```diff
@@ -49,3 +52,117 @@ DEFINE_HOOK(0x519F71, InfantryClass_UpdatePosition_BeforeBuildingChangeHouse, 0x
...
+static void __stdcall DrawALinkTo(CoordStruct nFrom, CoordStruct nTo, ColorStruct color)
+{
+	JMP_STD(0x704E40);
+}
+
+DEFINE_HOOK(0x4721E6, CaptureManagerClass_DrawLinkToVictim, 0x6)
+{
+	GET(CaptureManagerClass*, pThis, EDI);
+	GET(TechnoClass*, pVictim, ECX);
+	GET_STACK(int, nNodeCount, STACK_OFFSET(0x30, 0x1C));
```
CrimRecya, line 65: GET_STACK(int, nNodeCount, STACK_OFFSET(0x30, -0x1C));

CrimRecya (changes requested):

Coronia:
The crash is fixed, and the logic works well in testing

CrimRecya (approved):

Coronia:
Profiler showed that the draw link hook cost a lot when the game REALLY had a lot of mind control links, but since it's taking over the vanilla link drawing process it should be acceptable, and numerous links is a laggy thing even in vanilla anyway. It won't contribute to lag visibly when there're not that many links, which is the case for most of the games

Though, it would be appreciated if it can be further optimized in the future

PR #553, opened 2022-03-16
Team Leader search tweaks

FS-21:
For some new Attack/Move script actions now the Team Leader search will be done if this is the first time or the Leader is missing.

Currently that Leader search routine was executed in each game frame in each AI Team.


FS-21:
This is only a code tweak, not a feature. So the only test could be that this doesn't do weird things or early crashes.

Metadorius (approved):

PR #563, opened 2022-03-19
Display "Banners" on screen via trigger actions

MortonPL:
TODO Description


Metadorius:
IMO would need to be able to do percent from battlefield view/pixel offset from different corners of viewport.

PR #569, opened 2022-04-05
Allow vehicles to use custom palettes

Starkku:
This fix should allow VehicleTypes, both SHP (including turreted) and voxels to fully respect `Palette` setting in `artmd.ini` and use custom palettes for drawing. For the record, against common knowledge, AircraftTypes already respected this even on voxel art.

Do note that voxel art is still bound to one global lookup table / VPL and that any restrictions that affect other custom-paletted objects regarding things like parachutes, disguise logic, map lighting etc. are likely to apply to vehicles as well.


AlliedG:
Tested Palette=TEST with testtem.pal on Heartland map and worked as described for vehicles https://imgur.com/a/AFSVYpK (top row of tanks with new palette) 

Also can confirm AircraftTypes Palette= also worked

Metadorius (approved):

PR #572, opened 2022-04-06
Single-color lasers (that respect thickness) & laser glow improvements

### LINKED ISSUE #410: [Phobos Bug] Prism Forwarding Laser draw low quality laser
Prism Forwarding Laser draw low quality laser

Conditions: Hardcode & INI Code Changement

### Phobos version
Build 24 and higher

### Expected Behaviour:
- The Laser from the Prism Forwarding (both Default and Prism Forwarding by Ares). When a unit fire a laser (mostly Building, Prism Tower in this case) supposed to respect the laserthickness tag as well as the quality of the laser itself.

### Actual Behaviour:
- The Laser come from forwarded Prism draw a low quality laser and does not respect the laserthickness tag (both the supporting building and supported building). Can be observe in the detailed picture from the attachment.
- If no value is set on LaserThickness tag, it still draw a poor quality laser

### Additional Files / Links:
- no crashes or lagging causes by the issues just the quality of the laser.
laser3

laser4

Starkku:
- You can now set laser to draw using only `LaserInnerColor` by setting `IsSingleColor`, in same manner as `IsHouseColor` lasers do using player's team color. These lasers respect laser thickness.

In `rulesmd.ini`:
```ini
IsSingleColor=false  ; boolean
```

Additionally, fixed single-color laser (IsHouseColor, IsSingleColor, LaserTrails) laser glow falloff to match the vanilla appearance.
laserfalloffn
*In order from left to right: original game behaviour, the old Phobos behaviour, new Phobos behaviour.*

**NOTE:** This change will affect appearance of existing laser trails, modders may need to review their settings / values.

Closes #410

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -136,6 +136,7 @@ This page lists all the individual contributions to the project by their author.
   - Weapon owner detachment
   - Feedback weapon
   - TerrainType & ore minimap color customization
+  - Laser fixes & improvements
 - **Morton (MortonPL)**:
   - `XDrawOffset`
   - Shield passthrough & absorption
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -351,6 +351,16 @@ In `rulesmd.ini`:
 DetachedFromOwner=false  ; boolean
 ```
 
+### Single-color lasers
+
+- You can now set laser to draw using only `LaserInnerColor` by setting `IsSingleColor`, in same manner as `IsHouseColor` lasers do using player's team color. These lasers respect laser thickness.
+
+In `rulesmd.ini`:
+```ini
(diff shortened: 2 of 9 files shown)


Metadorius:
I think this one will break any such laser from using `LaserOuterColor` and `Spread` because housecolor lasers are different and don't use those.

Starkku:
> I think this one will break any such laser from using `LaserOuterColor` and `Spread` because housecolor lasers are different and don't use those.

Yeah that is probably the case. Still this only applies if `LaserThickness` is set.

Metadorius:
@skyhigh3110 can you test the pull request?

sKyHigh3110:
> @skyhigh3110 can you test the pull request?

i will be back to the pull request when finish testing tmr tks!

PR #588, opened 2022-04-10
Shield animation & pip customizations

Starkku:
Added damage-state specific shield idle animations.

In `rulesmd.ini`: (Only new keys shown)
```ini
[SOMESHIELDTYPE]                   ; ShieldType name
IdleAnim.ConditionYellow=          ; animation
IdleAnim.ConditionRed=             ; animation
IdleAnimDamaged=                   ; animation
IdleAnimDamaged.ConditionYellow=   ; animation
IdleAnimDamaged.ConditionRed=      ; animation
```
An excerpt from adjusted documentation for shield idle animations:
- `IdleAnim`, if set, will be played while the shield is intact. This animation is automatically set to loop indefinitely.
  - `IdleAnim.ConditionYellow` and `IdleAnim.ConditionRed` can be used to set different animations for when shield health is at or below the percentage defined in `[AudioVisual]`->`ConditionYellow`/`ConditionRed`, respectively. If not set, `IdleAnim` is used.
  - `IdleAnimDamaged`, `IdleAnimDamaged.ConditionYellow` and `IdleAnimDamaged.ConditionRed` are used in an identical manner, but only when health of the object the shield is attached to is at or below `[AudioVisual]`->`ConditionYellow`. If none are set, then regular `IdleAnim` or variants are used.

There are also additional customizations for shield pips:

In `rulesmd.ini`: (Only new keys shown)
```ini
[AudioVisual]
Pips.Shield.Background=PIPBRD.SHP  ; filename - including the .shp/.pcx extension
Pips.Shield.Building.Empty=0       ; int, frame of pips.shp (zero-based) for empty building pip

[SOMESHIELDTYPE]                   ; ShieldType name
Pips=                              ; int, frames of pips.shp (zero-based) for Green, Yellow, Red
Pips.Building=                     ; int, frames of pips.shp (zero-based) for Green, Yellow, Red
Pips.Background=                   ; filename - including the .shp/.pcx extension
Pips.Building.Empty=               ; int, frame of pips.shp (zero-based) for empty building pip
```
An excerpt from adjusted documentation for shield pips:
- A TechnoType with a shield will show its shield Strength. An empty shield strength bar will be left after destroyed if it is respawnable. Several customizations are available for the shield strength pips.
  - By default, buildings use the 6th frame of `pips.shp` to display the shield strength while others use the 17th frame.
  - `Pips.Shield` can be used to specify which pip frame should be used as shield strength. If only 1 digit is set, then it will always display that frame, or if 3 digits are set, it will use those if shield's current strength is at or below `ConditionYellow` and `ConditionRed`, respectively. `Pips.Shield.Building` is used for BuildingTypes.
  - `Pips.Shield.Background` can be used to set the background or 'frame' for non-building pips, which defaults to `pipbrd.shp`. 4th frame is used to display an infantry's shield strength and the 3th frame for other units, or 2nd and 1st respectively if not enough frames are available.
  - `Pips.Shield.Building.Empty` can be used to set the frame of `pips.shp` displayed for empty building strength pips, defaults to 1st frame of `pips.shp`.
  - The above customizations are also available on per ShieldType basis, e.g `[ShieldType]`->`Pips` instead of `[AudioVisual]`->`Pips.Shield` and so on. ShieldType settings take precedence over the global ones, but will fall back to them if not set.
  - `BracketDelta` can be used as additional vertical offset (negative shifts it up) for shield strength bar. Much like `PixelSelectionBracketDelta`, it is not applied on buildings.

Additionally fixed an issue that caused crashes with shield idle animations under certain circumstances. This should hopefully fix the crashes some people were having with shield animations and cloaked units.

### DIFF
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -165,8 +165,9 @@ Shield.InheritStateOnReplace=false   ; boolean
   - If you want shield recovers/respawns 1 HP per time, currently you need to set tag value to any number between 1 and 2, like `1.1`.
 - `SelfHealing.Rate` and `Respawn.Rate` respect the following settings: 0.0 instantly recovers the shield, other values determine the frequency of shield recovers/respawns in ingame minutes.
 - `IdleAnim`, if set, will be played while the shield is intact. This animation is automatically set to loop indefinitely.
-  - `IdleAnim.ConditionYellow` and `IdleAnim.ConditionRed` can be used to set different animations for when shield health is at or below the percentage defined in `[AudioVisual]`->`ConditionYellow`/`ConditionRed`, respectively. If not set, `IdleAnim` is used.
-  - `IdleAnimDamaged`, `IdleAnimDamaged.ConditionYellow` and `IdleAnimDamaged.ConditionRed` are used in an identical manner, but only when health of the object the shield is attached to is at or below `[AudioVisual]`->`ConditionYellow`. If none are set, then regular `IdleAnim` or variants are used.
+  - `IdleAnim.ConditionYellow` and `IdleAnim.ConditionRed` can be used to set different animations for when shield health is at or below the percentage defined in `[AudioVisual]`->`ConditionYellow`/`ConditionRed`, respectively. If `IdleAnim.ConditionRed` is not set it falls back to `IdleAnim.ConditionYellow`, which in turn falls back to `IdleAnim`.
+  - `IdleAnimDamaged`, `IdleAnimDamaged.ConditionYellow` and `IdleAnimDamaged.ConditionRed` are used in an identical manner, but only when health of the object the shield is attached to is at or below `[AudioVisual]`->`ConditionYellow`. Follows similar fallback sequence to regular `IdleAnim` variants and if none are set, falls back to the regular `IdleAnim` or variants thereof.
+  - `Bouncer=true` and `IsMeteor=true` animations can exhibit irregular behaviour when used as `IdleAnim` and should be avoided.
 - `IdleAnim.OfflineAction` indicates what happens to the animation when the shield is in a low power state.
 - `IdleAnim.TemporalAction` indicates what happens to the animation when the shield is attacked by temporal weapons.
 - `BreakAnim`, if set, will be played when the shield has been broken.
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -105,6 +105,11 @@ Respawn=0.0                          ; double, percents or absolute
 Respawn.Rate=0.0                     ; double, ingame minutes
 BracketDelta=0                       ; integer - pixels
 IdleAnim=                            ; animation
+IdleAnim.ConditionYellow=            ; animation
+IdleAnim.ConditionRed=               ; animation
+IdleAnimDamaged=                     ; animation
+IdleAnimDamaged.ConditionYellow=     ; animation
+IdleAnimDamaged.ConditionRed=        ; animation
 IdleAnim.OfflineAction=Hides         ; AttachedAnimFlag (None, Hides, Temporal, Paused or PausedTemporal)
 IdleAnim.TemporalAction=Hides        ; AttachedAnimFlag (None, Hides, Temporal, Paused or PausedTemporal)
 BreakAnim=                           ; animation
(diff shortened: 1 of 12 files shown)


ayylmaoRotE:
[SHIELD]
Pips=n is bugged, documentation states that all condition stages should respect a single entry, it does not. It starts off as the default value in AudioVisual then at 50% it returns the 6th pip regardless of value entered
Pips=n,n,n works as described
BracketDelta= works as described

if n exceeds the number of frames available it returns nothing, which is good i suppose if you simply want to hide the shield

Starkku:
Fixes the issue with `Pips`. Both `[AudioVisual]`->`Pips_Shield(_Building)` and `[ShieldType]`->`Pips(_Building)` will now fall back to the first defined value if second or third one is not defined. If first value of  `[ShieldType]`->`Pips(_Building)` is not defined it falls back to `[AudioVisual]`->`Pips_Shield(_Building)` as a whole.

mevitar:
Are IdleAnimDamaged.X= animations supposed to always take priority over IdleAnim.X= ones? The latter will never be used if the unit's health is below green, even when shields are fully restored.

In fact, they won't switch once both health and shields are back to full either, and shields will be stuck at the IdleAnimDamaged= animation. The only way to get IdleAnim= back again is to let the shield regenerate while the unit's regular health is green. Anything but that and it won't activate.

mevitar:
```
Strength=250
AbsorbOverDamage=no
SelfHealing.Rate=.0165
SelfHealing=5
Respawn.Rate=.033
Respawn=10
IdleAnim=WARPOUT
IdleAnim.ConditionYellow=NUKEBALL
IdleAnim.ConditionRed=PDFXCLD
IdleAnim.OfflineAction=None
IdleAnim.TemporalAction=None
IdleAnimDamaged=WARPIN
IdleAnimDamaged.ConditionYellow=NUKEANIM
IdleAnimDamaged.ConditionRed=PDFXLOC
```

shieldbug
Still happens in the newest build.

Starkku:
Okay that GIF clarified it to me, there was a missing check so it should work now that I amended it.

mevitar:
I guess a gif is worth a thousand words after all. :D
Woks fine in the newest build.

docs/New-or-Enhanced-Logics.md
```diff
@@ -154,7 +165,8 @@ Shield.InheritStateOnReplace=false   ; boolean
   - If you want shield recovers/respawns 1 HP per time, currently you need to set tag value to any number between 1 and 2, like `1.1`.
 - `SelfHealing.Rate` and `Respawn.Rate` respect the following settings: 0.0 instantly recovers the shield, other values determine the frequency of shield recovers/respawns in ingame minutes.
 - `IdleAnim`, if set, will be played while the shield is intact. This animation is automatically set to loop indefinitely.
-  - `Bouncer=yes` animations are not supported at the moment.
```
Metadorius, line 157: I guess that needs to be mentioned in the changelog.

docs/New-or-Enhanced-Logics.md
```diff
@@ -154,7 +165,8 @@ Shield.InheritStateOnReplace=false   ; boolean
...
+  - `IdleAnim.ConditionYellow` and `IdleAnim.ConditionRed` can be used to set different animations for when shield health is at or below the percentage defined in `[AudioVisual]`->`ConditionYellow`/`ConditionRed`, respectively. If not set, `IdleAnim` is used.
+  - `IdleAnimDamaged`, `IdleAnimDamaged.ConditionYellow` and `IdleAnimDamaged.ConditionRed` are used in an identical manner, but only when health of the object the shield is attached to is at or below `[AudioVisual]`->`ConditionYellow`. If none are set, then regular `IdleAnim` or variants are used.
```
Metadorius, line 169: Maybe if `ConditionRed` ones are not set but `ConditionYellow` are then it would be better to fall back to `ConditionYellow`, not the regular one?
image

Metadorius: Alternative idea, not sure which would be better to use this though because the first one feels more natural: 
image

src/New/Entity/ShieldClass.cpp
```diff
@@ -13,20 +13,20 @@
...
-	, Respawn_Rate_Warhead { -1 }
+, IdleAnim { nullptr }
+, Timers { }
+, Cloak { false }
+, Online { true }
+, Temporal { false }
+, Available { true }
+, Attached { isAttached }
+, SelfHealing_Rate_Warhead { -1 }
+, Respawn_Rate_Warhead { -1 }
```
Metadorius, line 29: Not sure of a reason for this change?

Metadorius (changes requested):
Good to go after addressing the comments.

src/New/Entity/ShieldClass.cpp
Starkku: Visual studio auto-formatting based on the current .editorconfig settings.

Metadorius: > Visual studio auto-formatting based on the current .editorconfig settings.
hm, not sure if that can be configured to behave other way, either way that is undesirable.

docs/New-or-Enhanced-Logics.md
Starkku: The first diagram looks like a sensible approach, I can probably make it follow that.

docs/New-or-Enhanced-Logics.md
Starkku: It doesn't actually work per se, it just doesn't instantly crash the game. That and the fact that `IsMeteor=true` animations *do* crash the game makes me think either the parsing check should be brought back (it is just going to be more complicated due to the multiple animations parsed) or there needs to be a warning in the docs.

Starkku:
Addressed the issues.

src/Utilities/TemplateDef.h
```diff
@@ -852,15 +995,58 @@ void __declspec(noinline) ValueableIdxVector<Lookuper>::Read(INI_EX& parser, con
...
+{
+
+	// read the common flag, with the trailing dot being stripped
+	char flagName[0x40];
+	auto const pSingleFormat = pSingleFlag ? pSingleFlag : pBaseFlag;
+	auto res = _snprintf_s(flagName, _TRUNCATE, pSingleFormat, "");
+	if (res > 0 && flagName[res - 1] == '.')
+	{
+		flagName[res - 1] = '\0';
+	}
```
Metadorius, line 1027:
```suggestion
	// read the common flag, with the trailing dot being stripped
	char flagName[0x40];
	auto const pSingleFormat = pSingleFlag ? pSingleFlag : pBaseFlag;
	auto res = _snprintf_s(flagName, _TRUNCATE, pSingleFormat, "");

	if (res > 0 && flagName[res - 1] == '.')
		flagName[res - 1] = '\0';
```

Metadorius (approved):

PR #592, opened 2022-04-13
Fire superweapon using warhead

chaserli:
### Fire superweapons by warheads

- Superweapons can be launched when a standard warhead is detonated. Animation warheads and weapons from _Ares_ are **not** supported in this case. Nevertheless, the warhead from an animation which is used using `AttachEffect.Animation` is supported.

    - `SpawnSuperWeapons` specifies the superweapons to launch when the warhead is detonated.
    - `SpawnSuperWeapons.RealLaunch` controls whether the owner who fired the warhead must own all listed superweapons and sufficient fund to support `Money.Amout`. Otherwise they will be launched out of nowhere.
    - In either case, the superweapon is launched without any constraint. For example, `SW.Designators` and `SW.Inhibitors` are ignored. The user should not abuse such behavior.
   
In `rulesmd.ini`

```ini
[Warhead]                           
SpawnSuperWeapons=                                                        ; list of superweapons
SpawnSuperWeapons.RealLaunch= no                                ; bool
```

I've also tried animation superweapons, but the ownership issue cannot be resolved for certain cases.


ststl-s:
AbstractType TechnoClass::Target
if (pThis->Target->WhatAmI() == AbstractType::Cell)
		abstract_cast<CellClass*>(pThis->Target);
	else
		abstract_cast<ObjectClass*>(pThis->Target);

ststl-s:
AbstractClass* TechnoClass::Target;
if (pThis->Target->WhatAmI() == AbstractType::Cell)
abstract_cast<CellClass*>(pThis->Target);
else
abstract_cast<ObjectClass*>(pThis->Target);

secsome:
It's better to use Anim or WHs to implement the SW firing, instead of using Technos.

chaserli:
> It's better to use Anim or WHs to implement the SW firing, instead of using Technos.

I'm looking into anims right now. I'll refactor this when I have a clue

src/Ext/WarheadType/Body.h
```diff
@@ -74,6 +75,8 @@ class WarheadTypeExt
 		Valueable<int> Shield_MinimumReplaceDelay;
 		ValueableVector<ShieldTypeClass*> Shield_AffectTypes;
 
+		NullableVector<SuperWeaponTypeClass*> SpawnSuperWeapons;
```
Thrifinesma, line 78: I guess this is a wrong usage for NullableVector, I'd suggest to change it to ValueableVector

src/Ext/WarheadType/Detonate.cpp
```diff
@@ -36,6 +38,27 @@ void WarheadTypeExt::ExtData::Detonate(TechnoClass* pOwner, HouseClass* pHouse,
 
 		if (this->TransactMoney)
 			pHouse->TransactMoney(this->TransactMoney);
+
+		if (this->SpawnSuperWeapons.HasValue()) {
+			for (const auto pSWType : this->SpawnSuperWeapons.GetElements()) {
```
Thrifinesma, line 43: After changing it to ValueableVector, here only needs
```suggestion
		for (const auto pSWType : this->SpawnSuperWeapons)
		{
```

src/Ext/WarheadType/Detonate.cpp
```diff
@@ -36,6 +38,27 @@ void WarheadTypeExt::ExtData::Detonate(TechnoClass* pOwner, HouseClass* pHouse,
...
+				SuperClass* pSuper = nullptr;
+				bool canLaunch = false;
+				if (this->SpawnSuperWeapons_RealLaunch.Get()) {
+					pSuper = pHouse->Supers.GetItem(SuperWeaponTypeClass::Array->FindItemIndex(pSWType));
+					if (pSuper && pSuper->IsCharged) {
+						canLaunch = true;
+					}
+				}
+				else if (pSuper = GameCreate<SuperClass>(pSWType, pHouse)) {
```
Thrifinesma, line 52: Quite courageous method. Game will accumulate more and more `SuperClass` while running this feature. Change it to some other implementation please

Thrifinesma (changes requested):

src/Ext/WarheadType/Detonate.cpp
chaserli: Deleted the `GameCreate` part and made the code more concise. However, if the owner already has the superweapon in charge, this will enforce its reset as well, but this would not be a problem for modders.

src/Ext/WarheadType/Detonate.cpp
```diff
@@ -36,6 +38,22 @@ void WarheadTypeExt::ExtData::Detonate(TechnoClass* pOwner, HouseClass* pHouse,
 
 		if (this->TransactMoney)
 			pHouse->TransactMoney(this->TransactMoney);
+
+
```
Starkku, line 42: Remove one of the two empty lines.

Starkku (changes requested):
Lacks documentation, and there's one very small style error but besides that looks fine.

docs/New-or-Enhanced-Logics.md
```diff
@@ -1202,6 +1202,24 @@ In `rulesmd.ini`:
...
 ```
 
+### Launch superweapons on impact
+
+
+- Superweapons can be launched when a standard warhead is detonated. This means that animation warheads and weapons from _Ares_ are **not** supported. Nevertheless, the warhead from an animation which is used using `AttachEffect.Animation` still works.
+
+    - `SpawnSuperWeapons` specifies the superweapons to launch when the warhead is detonated. 
+    - `SpawnSuperWeapons.RealLaunch` controls whether the owner who fired the warhead must own all listed superweapons and sufficient fund to support `Money.Amout`. Otherwise they will be launched out of nowhere.
+    - In either case, the superweapon is launched without any constraint. For example, `SW.Designators` and `SW.Inhibitors` are ignored. The user should not abuse such behavior.
```
Starkku, line 1212: - Remove the empty line between the first level list item and the first child item.
- Fix the list indentation. The second level list items should be prefixed with only two whitespaces, add more two for every sublevel after that.

docs/New-or-Enhanced-Logics.md
```diff
@@ -1202,6 +1202,24 @@ In `rulesmd.ini`:
...
+   
+In `rulesmd.ini`:
+
+```ini
+[Warhead]                           
+SpawnSuperWeapons=                        ; list of superweapons
+SpawnSuperWeapons.RealLaunch= no          ; bool
```
Starkku, line 1219: - Remove the space between `SpawnSuperWeapons.RealLaunch=` and `no`.
- Fix indentation of the comments/variable descriptions. Two whitespaces between the end of longest line in section and ; character, and rest to line up with that.

Starkku (changes requested):
Couple of more style issues with the documentation.

Also missing contribution mention from `CREDITS.MD` and mention of the feature in `docs\Whats-New.md` still.

chaserli:
I need testing and feedback on this feature from the community first.

DmitryVolkov666:
So. Everything works great. Found a problem with a weather storm. He simply, when launching the next weapon, disappeared at the launch of the previous weapon and moved to the next blow. 

There was also a problem with the chronosphere for conventional weapons. see this too. I think it will amuse you
She simply created a starting point and that's it. and you could teleport something with the cursor, but in fact, nothing happened.
Yes, it is understand that not a single modder will use it because of the "complexity" of the chronosphere.
It would be nice to ban the teleportation superweapon. And in the future add separate "keys" to control the chronosphere

for example through one weapon
1 hit - generates the starting point of teleportation
2 - teleports already from 1 point to the one hit by the projectile 2 times. then the counter resets to zero and everything starts all over again.

but 2 weapons (as for me) is better. like how...

Primary = starting point
PipScale=Ammo
Ammo=1
PipWrap=1
EmptyReload=6000
reload=6000
NoAmmoWeapon=1
ReloadAmount=1
Secondary = endpoint (no ammo used)

CONDITIONALLY.

The tank shoot underneath him - creating a starting point for teleporting the object/objects through the primary weapon

Secondary = shooting already in the area you need. and teleport the object/objects to the endpoint

Yet again. This is only a wish regarding the chronosphere for the future. Since I understand that she will have her own problems (do not forget to watch this)

But in general, everything works fine if you do not abuse any complex superweapons. e.g. airstrikes, paradrop, genetic mutator, nuclear strikes, iron curtains, force shields work more excellently.

and most importantly - the superweapon respects the owner of the object.

moreover, if a super weapon is paid, it takes money

if the weapon can de-energize the base (for example, a force shield), then it works as expected. 
phew. looks like that's all

MortonPL:
"She simply created a starting point and that's it. and you could teleport something with the cursor, but in fact, nothing happened."

That happens because Chronosphere SW is actually two Super Weapons, one for source and one for destination: http://ares-developers.github.io/Ares-docs/new/superweapons/types/chronosphere.html

DmitryVolkov666:
> "She simply created a starting point and that's it. and you could teleport something with the cursor, but in fact, nothing happened."
> 
> That happens because Chronosphere SW is actually two Super Weapons, one for source and one for destination: http://ares-developers.github.io/Ares-docs/new/superweapons/types/chronosphere.html


I know. just described the problem

chaserli:
Close due to un-updated nightly build

PR #594, opened 2022-04-13
Implement passable & buildable-upon TerrainTypes

Starkku:
- TerrainTypes can now be made passable or buildable upon by setting `IsPassable` or `CanBeBuiltOn`, respectively.
  - Movement cursor is displayed on `IsPassable` TerrainTypes unless force-firing.
  - `CanBeBuiltOn=true` terrain objects are removed when building is placed on them.

In `rulesmd.ini`:
```ini
[SOMETERRAINTYPE]   ; TerrainType
IsPassable=false    ; boolean
CanBeBuiltOn=false  ; boolean
```

### DIFF
--- a/src/Ext/TerrainType/Hooks.Passable.cpp
+++ b/src/Ext/TerrainType/Hooks.Passable.cpp
@@ -1,0 +1,201 @@
+#include "Body.h"
+
+#include <OverlayClass.h>
+#include <TerrainClass.h>
+
+#include <Utilities/GeneralUtils.h>
+
+// Passable TerrainTypes Hook #1 - Do not set occupy bits.
+DEFINE_HOOK(0x71C110, TerrainClass_SetOccupyBit_PassableTerrain, 0x6)
+{
+	enum { Skip = 0x71C1A0 };
+
+	GET(TerrainClass*, pThis, ECX);
+
+	if (auto const pTypeExt = TerrainTypeExt::ExtMap.Find(pThis->Type))
+	{
+		if (pTypeExt->IsPassable)
+			return Skip;
+	}
+
+	return 0;
+}
+
+// Passable TerrainTypes Hook #2 - Do not display attack cursor unless force-firing.
+DEFINE_HOOK(0x7002E9, TechnoClass_WhatAction_PassableTerrain, 0x5)
+{
+	enum { Skip = 0x70020E };
+
+	GET(ObjectClass*, pTarget, EDI);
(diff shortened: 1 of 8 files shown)


Metadorius:
Do those two features really depend on each other? Maybe it would be a nice gimmick for someone to have an impassable terrain object that could be built on?

Starkku:
Fixed an issue that caused vehicles to not consider `IsPassable` TerrainTypes passable.

Also fixed Foundation=0x0 crashing game if used on TerrainTypes, so it is available as an alternative if there is a need for 'ghost' terrain object that cannot be interacted at all but is still visible.

Starkku:
Fixed a faulty hook that generated crashes.

Starkku:
Separated the buildable-upon setting from the passable one and made buildable-upon terrain objects deleted on building placement.

DmitryVolkov666:
everything seems to work without problems

it would be nice to add such a thing so that you can enumerate what can go through the "tree". and what is not.
that is the list
units, infantry, buildings.

and try to fix the parameter that allows you to crush trees (well, more precisely, it should have worked but does not work)
https://modenc.renegadeprojects.com/Crushable

Starkku:
> everything seems to work without problems
> 
> it would be nice to add such a thing so that you can enumerate what can go through the "tree". and what is not. that is the list units, infantry, buildings.
> 
> and try to fix the parameter that allows you to crush trees (well, more precisely, it should have worked but does not work) https://modenc.renegadeprojects.com/Crushable

Conditional passability would be rather difficult because of how scattered the passability checks are. Not necessarily impossible but not something I am interested in looking into. For what it is worth it is already possible to make trees passable for infantry only by setting `Temperate/SnowOccupationBits` to 0.

`Crushable` is something would have to look at separately. I cannot even make trees etc. crushable at all on my end so it makes testing bit difficult.

src/Ext/TerrainType/Hooks.Passable.cpp
```diff
@@ -0,0 +1,201 @@
...
+	GET(AbstractClass*, pTarget, ESI);
+
+	if (auto const pTerrain = abstract_cast<TerrainClass*>(pTarget))
+	{
+		auto const pTypeExt = TerrainTypeExt::ExtMap.Find(pTerrain->Type);
+
+		if (pTypeExt->IsPassable)
+		{
+			R->EBP(0);
+			return 0x73FD37;
```
Metadorius, line 84:
```suggestion
			return Return;
```

Metadorius (approved):

PR #620, opened 2022-05-03
Random Product, Action 9931, modify Action 505,506

ststl-s:
Random product for infantry and vehicle
Action 9931 message for specified house
Action 505,506 use SuperWeaponType


Starkku:
Unrelated features should be split into separate PR's.

PR #622, opened 2022-05-03
Misc. additions / improvements

### LINKED ISSUE #574: [New Feature]Debris Play ExpireAnim On Water
### Description:
Debris always plays splash animation when he touches the water
Although we can use BounceAnim= to play the animation, BounceAnim will have a visible offset in the direction of debris flight
So I think it's necessary to allow ExpireAnim played on the water

### Possible Implementation:

artmd.ini
1.
[Debris]
Debris.ExplodeOnWater=(boolean)
2.
[Debris]
SplashAnim=(animation)
(def=Warhead > SplashList)
Phobos has now allowed Warhead to custom the SplashList, so we can use that as the default value of Debris's SplashAnim=
3.
Based on 2 ,Without adding any tags, always read the SplashList= data of Warhead and select splash animation according to Damage/35
4.
Fix BounceAnim so that it has no offset, or make this offset can custom

### Additional Files / Links:
BounceAnim

Starkku:
### Spawner spawn delay customization

- `Spawner.DelayFrames` can be used to set the minimum number of game frames in between each spawn ejecting from the spawner. By default this is 9 frames for missiles and 20 for everything else.

In `rulesmd.ini`:
```ini
[SOMETECHNO]          ; TechnoType
Spawner.DelayFrames=  ; integer, game frames
```

### Cluster scatter distance customization

- `ClusterScatter.Min` and `ClusterScatter.Max` can be used to set minimum and maximum distance, respectively, in cells from the original detonation coordinate any additional detonations if `Cluster` is set to value higher than 1 can appear at.

In `rulesmd.ini`:
```ini
[SOMEPROJECTILE]         ; Projectile
ClusterScatter.Min=1.0  ; float, distance in cells
ClusterScatter.Max=2.0  ; float, distance in cells
```

### FlakScatter distance customization

- By default `FlakScatter=true` makes `Inviso=true` projectiles scatter by random distance (in cells) from 0 to `[CombatDamage]` -> `BallisticScatter`. This distance range can now be customized by setting `BallisticScatter.Min` & `BallisticScatter.Max` on the projectile. If not set, the default values are used.

In `rulesmd.ini`:
```ini
[SOMEPROJECTILE]      ; Projectile
BallisticScatter.Min= ; float, distance in cells
BallisticScatter.Max= ; float, distance in cells
```

### Customizable debris & meteor impact and warhead detonation behaviour

- `ExplodeOnWater` can be set to true to make the animation explode on impact with water. `ExpireAnim` will be played and `Warhead` is detonated or used to deal damage / generate light flash.
- `Warhead.Detonate`, if set to true, makes the `Warhead` fully detonate instead of simply being used to deal damage and generate light flash if it has `Bright=true`.
- `SplashAnims` contains list of animations used if `ExplodeOnWater` is not set and the animation impacts with water. Defaults to `[CombatDamage]` -> `SplashList` if `IsMeteor` is set, otherwise to a single animation specified in `[General]` -> `Wake`.
  - If `SplashAnims.PickRandom` is set to true, picks a random animation from `SplashAnims` to use on each impact with water. Otherwise last listed animation from `SplashAnims` is used.

In `artmd.ini`:
```ini
[SOMEANIM]                    ; AnimationType
ExplodeOnWater=false          ; boolean
Warhead.Detonate=false        ; boolean
SplashAnims=                  ; list of animations
SplashAnims.PickRandom=false  ; boolean
```

Also available for VoxelAnims, behaviour and INI keys are identical other than the need to list them in the `rulesmd.ini` entry rather than `artmd.ini`.

### Custom debris animations and additional debris spawn settings

- You can now use `DebrisAnims` to specify a list of debris animations to spawn instead of animations from `[General]` -> `MetallicDebris` when Warhead with `MaxDebris` > 0 and no `DebrisTypes` (VoxelAnims) listed is detonated.
- `Debris.Conventional`, if set to true, makes `DebrisTypes` or `DebrisAnims` only spawn if Warhead is fired on non-water cell.

In `rulesmd.ini`:
```ini
[SOMEWARHEAD]              ; WarheadType
DebrisAnims=               ; List of animations
Debris.Conventional=false  ; boolean
```

--------------

Some technical notes regarding these:
- Spawner (*SpawnManagerClass*) logic is by default tied to 10 frame timer. Nothing related to it gets processed in faster intervals than that. To allow faster spawn delay than this, the timer is set to `Spawner.DelayFrames` instead of 10 when the spawner is spawning aircraft. This shouldn't cause any unusual behaviour but I figured I'd mention it just in case.
- `BallisticScatter.Min` / `BallisticScatter.Max` use same key names as Ares does, but none of the other logic besides the `FlakScatter=true` case (which Ares does not implement) is handled on Phobos' side.
- The debris impact behaviour change also includes a minor change that makes the `Warhead.Detonate` use `Bright` setting from `Warhead` to determine if it generates lighting flash. This change also affects `Crit.Warhead`. Previously it would never generate flash.
- Warhead DebrisAnims/Conventional fix also makes MaxDebris on warhead work as expected, previously it would use MaxDebris-1 to calculate the debris count, meaning that configurations like MinDebris=0 & MaxDebris=1 would never spawn any debris.

--------------

Closes #574


AlliedG:
**Spawner spawn delay customization**
---

Tested spawning 10 hornets with Spawner.DelayFrames=0 and all of them were deployed at same time rather than individually taking off so working as expected.

**Cluster scatter distance customization**
---

Cluster.Scatter.Min and Cluster.Scatter.Max worked as expected

Setting values to 0 overrides the Cluster.Scatter behavior as expected

**FlakScatter distance customization**
---

BallisticScatter.Min= & BallisticScatter.Max= worked as expected

Setting values to 0 overrides the BallisticScatter behavior as expected

**Customizable debris & meteor impact and warhead detonation behaviour**
---

ExplodeOnWater= and Warhead.Detonate= worked as expected with the latter working with Ares tag Ripple.Radius and normal Particle=

SplashAnims= and SplashAnims.PickRandom= worked as expected when ExplodeOnWater=yes was omitted

**Custom debris animations and additional debris spawn settings**
---

DebrisAnims= and Debris.Conventional=false worked as expected. This also worked with debris animation warheads to spawn debris upon impact.

src/Ext/Bullet/Hooks.cpp
```diff
@@ -222,4 +224,104 @@ DEFINE_HOOK(0x469A75, BulletClass_Logics_DamageHouse, 0x7)
...
+
+DEFINE_HOOK(0x469D1A, BulletClass_Logics_Debris_Checks, 0x6)
+{
+	enum { SkipGameCode = 0x469EBA, SetDebrisCount=0x469D36 };
+
+	GET(BulletClass*, pThis, ESI);
+
+	auto pWHExt = WarheadTypeExt::ExtMap.Find(pThis->WH);
+
+	bool isLand = pThis->GetCell()->LandType != LandType::Water;
```
FS-21, line 290:
```suggestion
	auto pWHExt = WarheadTypeExt::ExtMap.Find(pThis->WH);
	bool isLand = pThis->GetCell()->LandType != LandType::Water;
```

src/Ext/Techno/Hooks.cpp
```diff
@@ -478,6 +479,39 @@ DEFINE_HOOK(0x6FD446, TechnoClass_FireLaser_IsSingleColor, 0x7)
...
+DEFINE_HOOK(0x6B7265, SpawnManagerClass_AI_UpdateTimer, 0x6)
+{
+	GET(SpawnManagerClass* const, pThis, ESI);
+
+	if (pThis->Owner && pThis->Status == SpawnManagerStatus::Launching && pThis->CountDockedSpawns() != 0)
+	{
+		if (auto const pTypeExt = TechnoTypeExt::ExtMap.Find(pThis->Owner->GetTechnoType()))
+		{
+			if (pTypeExt->Spawner_DelayFrames.isset()) { }
+				R->EAX(std::min(pTypeExt->Spawner_DelayFrames.Get(), 10));
```
FS-21, line 491: Looks a mistake with the brackets "{ }".

FS-21 (dismissed):

mevitar:
Tested everything except DebrisAnims, it worked fine.
However, i noticed that if ExplodeOnWater and Warhead.Detonate are both set to true, then both ExpireAnim and warhead's AnimList appear on hit. Is that intended?

Starkku:
> Tested everything except DebrisAnims, it worked fine. However, i noticed that if ExplodeOnWater and Warhead.Detonate are both set to true, then both ExpireAnim and warhead's AnimList appear on hit. Is that intended?

`Warhead.Detonate` will properly detonate the warhead if the debris explodes instead of simply dealing damage, this includes displaying `AnimList/SplashList`. `ExpireAnim` will still play regardless of this so yes, it is intended in a manner of speaking.

Starkku:
Fixed remaining issues and renamed `Cluster.Scatter.X` to `ClusterScatter.X`.

src/Ext/AnimType/Hooks.cpp
```diff
@@ -68,4 +72,79 @@ DEFINE_HOOK(0x424C49, AnimClass_AttachTo_BuildingCoords, 0x5)
 	}
 
 	return 0;
+}
+
+DEFINE_HOOK(0x423CC1, AnimClass_AI_HasExtras_Expired, 0x6)
```
Otamaa, line 77: overriden instruction at `0x423CC3` is requiredd for `IsTiberium` !

src/Ext/VoxelAnim/Hooks.cpp
```diff
@@ -26,4 +31,80 @@ DEFINE_HOOK(0x74A70E, VoxelAnimClass_AI_Additional, 0xC)
 	}
 
 	return 0;
+}
+
+DEFINE_HOOK(0x74A021, VoxelAnimClass_AI_Expired, 0x6)
```
Otamaa, line 36: Overriden instruction at `0x74A023` is required for `IsTiberium`

Starkku:
Fixed the hooks.

PR #625, opened 2022-05-05
Toggle for automatic passenger owner change to match transport

Starkku:
### Automatic passenger owner change to match transport owner

- Transports with `Passengers.SyncOwner` set to true will have the owner of their passengers changed to match the transport if transport's owner changes.
  - On `OpenTopped=true` transports this will also disable checks that prevent target acquisition by passengers when the transport is temporarily mind controlled.
  - `Passengers.SyncOwner.RevertOnExit`, if set to true, changes the passengers' owner back to whatever it was originally when they entered the transport when they are ejected.

In `rulesmd.ini`:
```ini
Passengers.SyncOwner=false               ; boolean
Passengers.SyncOwner.RevertOnExit=false  ; boolean
```

### DIFF
--- a/src/Ext/Techno/Body.h
+++ b/src/Ext/Techno/Body.h
@@ -31,6 +31,9 @@ public:
 		int Death_Countdown;
 		AnimTypeClass* MindControlRingAnimType;
 		int DamageNumberOffset;
+
+		// Used for Passengers.SyncOwner.RevertOnExit instead of TechnoClass::InitialOwner / OriginallyOwnedByHouse,
+		// as neither is guaranteed to point to the house the TechnoClass had prior to entering transport and cannot be safely overridden.
 		HouseClass* OriginalPassengerOwner;
 
 		ExtData(TechnoClass* OwnerObject) : Extension<TechnoClass>(OwnerObject)
--- a/src/Ext/Techno/Body.h
+++ b/src/Ext/Techno/Body.h
@@ -31,6 +31,7 @@ public:
 		int Death_Countdown;
 		AnimTypeClass* MindControlRingAnimType;
 		int DamageNumberOffset;
+		HouseClass* OriginalPassengerOwner;
 
 		ExtData(TechnoClass* OwnerObject) : Extension<TechnoClass>(OwnerObject)
 			, InterceptedBullet { nullptr }
@@ -45,6 +46,7 @@ public:
 			, Death_Countdown(-1)
 			, MindControlRingAnimType { nullptr }
 			, DamageNumberOffset { INT32_MIN }
+			, OriginalPassengerOwner {}
 		{ }
 
 		virtual ~ExtData() = default;
--- a/src/Ext/TechnoType/Body.h
+++ b/src/Ext/TechnoType/Body.h
@@ -198,7 +198,7 @@ public:
 			, Ammo_Shared_Group { -1 }
 			, SelfHealGainType()
 			, Passengers_SyncOwner { false }
-			, Passengers_SyncOwner_RevertOnExit { false }
+			, Passengers_SyncOwner_RevertOnExit { true }
 		{ }
 
 		virtual ~ExtData() = default;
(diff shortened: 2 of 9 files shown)


Metadorius:
Tested by Dmitry Volkov, everything seems to be working as intended.

src/Ext/Techno/Hooks.cpp
```diff
@@ -460,3 +460,57 @@ DEFINE_HOOK(0x6FD446, TechnoClass_FireLaser_IsSingleColor, 0x7)
 
 	return 0;
 }
+
+DEFINE_HOOK_AGAIN(0x6F7EC2, TechnoClass_ThreatEvals_OpenToppedOwner, 0x6)
+DEFINE_HOOK(0x6F8FD7, TechnoClass_ThreatEvals_OpenToppedOwner, 0x5)
+{
+	enum { SkipCheckOne = 0x6F8FDC, SkipCheckTwo=0x6F7EDA };
```
Metadorius, line 467:
```suggestion
	enum { SkipCheckOne = 0x6F8FDC, SkipCheckTwo = 0x6F7EDA };
```

Metadorius (commented):
Looks good except one minor nitpick. Also perhaps it would be good to also have a possibility to revert the owner change on passenger ejection?

Starkku:
- Changed key name from `Passengers.ChangeOwnerWithTransport` to `Passengers.SyncOwner` to make it less long.
- Introduced `Passengers.SyncOwner.RevertOnExit` which, if set to true alongside `Passengers.SyncOwner` will revert the passenger's owner to whatever it was when it entered the transport (InitialPayload should work with this) when it is ejected, be it manually or on destruction.
- Covered all threat scan etc. cases regarding OpenTopped and being mind controlled, instead of just two.
- Separated the passenger/transport hooks to a new file from general TechnoClass hooks.

src/Ext/Techno/Body.h
```diff
@@ -31,6 +31,7 @@ class TechnoExt
 		int Death_Countdown;
 		AnimTypeClass* MindControlRingAnimType;
 		int DamageNumberOffset;
+		HouseClass* OriginalPassengerOwner;
```
Metadorius, line 34: Is this one really needed? I recall there's already such a field in the original class.

src/Ext/TechnoType/Body.h
```diff
@@ -195,6 +197,8 @@ class TechnoTypeExt
 			, Ammo_Shared { false }
 			, Ammo_Shared_Group { -1 }
 			, SelfHealGainType()
+			, Passengers_SyncOwner { false }
+			, Passengers_SyncOwner_RevertOnExit { false }
```
Metadorius, line 201: Think `RevertOnExit` should default to true?

src/Ext/Techno/Body.h
Starkku: There's two similar `HouseClass` pointers in `TechnoClass`:
- `InitialOwner`, which is only set in constructor (so it points to the house that created the object. It is mostly unused too, only seemingly used in one check for buildings on owner change regarding upgrades.
- `OriginallyOwnedByHouse` which is the house the object had prior to being mind-controlled.

What I could have done is use `InitialOwner` for this. However I could not just use the value as-is because as stated before, it stores pointer to the house that created the object, which is not necessarily same as the house it has when it enters a transport. Given that it is mostly unused in vanilla game I could just override it instead but I decided to go the safe route because I have no guarantee that, say Ares, doesn't use it for something else.

Metadorius: May be good to leave a brief comment with such explanation on the code so people know not to "refactor" it in future.

src/Ext/TechnoType/Body.h
Starkku: Probably would make more sense yes, in hindsight.

Starkku:
Fixed the issues, merging.

PR #627, opened 2022-05-05
Mind Control Threshold

FS-21:
Now MC weapons can replace the Mind Control stuff by alternative warheads if the enemy health percentage is over the MindControl.threshold value.

More information in the documentation.


Metadorius:
Can't such functionality be implemented using `Crit.AffectsBelowPercent`?

Starkku:
> Can't such functionality be implemented using `Crit.AffectsBelowPercent` already without any modifications to the code?

MindControl is special logic that doesn't quite work through the critical hit system. The crit warhead wouldn't count as mind control warhead and the unit wouldn't have the CaptureManagerClass set up or the MC pips if it is its only MC weapon either.

Also there's no need to reinvent the wheel with the odd percentage parsing and health calculations. INI parser happily parses percentages to `double` and `ObjectClass::GetHealthPercentage()` exists.

FS-21:
> Also there's no need to reinvent the wheel with the odd percentage parsing and health calculations. INI parser happily parses percentages to `double` and `ObjectClass::GetHealthPercentage()` exists.

The only reason to reinvent the wheel was that in the first implementations that method returned unrealistic huge numbers so I made calculations by hand as you can see.
After reading this message and with the current checks the object returns valid values between 0.0 to 1.00000000... so I'll do the suggested changes.

> Use `MapClass::SelectDamageAnimation()` to pick the animation to display from the new warhead instead of reimplementing all the checks yourself. Current implementation doesn't even consider `Conventional` and excludes stuff like `SplashList`.

Phobos develop branch doesn't use this method so the only 2 valid examples are located in Ares sources... I'll try later (because ask for values I never used like land Type etc) but the current animation code is C&P from Crit system so if is incomplete here the same there.

> Having the threshold without having to set the Damage and Warhead could also be desirable.

I don't understand this, can you explain it, please?

FS-21:
> Perhaps having an option to invert threshold would be good? I already see two usecases:
> - weaken enemy until mind controllable;
> - kill enemy if he's too weak to be controlled.

Yes, the 2 user cases you mentioned are covered with the current tags.

Invertig the threshold means:
- Mind control units until the unit HP % is lower than the threshold, then the unit will ve hurt to death.

FS-21:
Applied feedback with those 2 internal changes.

MaxLenarc:
Max Lenarc here, just reporting my attempt to test the Mind Control Threshold from a while ago since I have not posted here in Github. I doubt it will be useful though. Ultimately failed to get the feature to work, not sure why though. I have the latest nightly build with the code given to me on Mod Haven. I think it may be a case of the code being put in the wrong spot in the INI?
INI looks like this.
"[Controller]
Verses=100%,100%,100%,100%,100%,100%,0%,0%,0%,0%,100%
MindControl=yes
AnimList=YURICNTL
AffectsAllies=no

Versus.mcvheavy=0%
Versus.mcvheavy.Retaliate=no
Versus.mcvheavy.ForceFire=no
Versus.mcvheavy.PassiveAcquire=no

Versus.t2mediump=0%
Versus.t2mediump.Retaliate=no
Versus.t2mediump.ForceFire=no
Versus.t2mediump.PassiveAcquire=no

Versus.t2heavyp=0%
Versus.t2heavyp.Retaliate=no
Versus.t2heavyp.ForceFire=no
Versus.t2heavyp.PassiveAcquire=no

Versus.t3heavyp=0%
Versus.t3heavyp.Retaliate=no
Versus.t3heavyp.ForceFire=no
Versus.t3heavyp.PassiveAcquire=no

Versus.mindarm=0%
Versus.mindarm.Retaliate=no
Versus.mindarm.ForceFire=no
Versus.mindarm.PassiveAcquire=no
Versus.sbanearmor=0%

Transact=true  
DamageAirThreshold=-1                                        ; boolean
Transact.SpreadAmongTargets=false                       ; boolean
Transact.Experience.Source.Flat=0                       ; integer
Transact.Experience.Source.Percent=0.40                  ; float, percents or absolute (-1.0 - 1.0)
Transact.Experience.Source.Percent.CalcFromTarget=true ; boolean
Transact.Experience.Target.Flat=0                       ; integer
Transact.Experience.Target.Percent=0.0                  ; float, percents or absolute (-1.0 - 1.0)
Transact.Experience.Target.Percent.CalcFromSource=false ; boolean

MindControl.Threshold= 25%  ; positive percentage. Represents a percentage from 0% to 100%
MindControl.AlternativeDamage= 50    ; integer
MindControl.AlternativeWarhead= PsychoFire; warhead
MindControl.CanKill= false           ; boolean
MindControl.Threshold.Inverse= false ; boolean

[PsychoFire]
Wall=no
Wood=no
Sparky=yes
Verses=100%,100%,100%,100%,100%,100%,50%,50%,50%,100%,100%
Conventional=yes
InfDeath=10
AnimList=UCINIT
ProneDamage=75%"
Both of these are under the Warheads list. Apologies in advance if this ends of being a waste of time, also for the wall of text.

brsajo:
> Ultimately failed to get the feature to work, not sure why though. I have the latest nightly build with the code given to me on Mod Haven. I think it may be a case of the code being put in the wrong spot in the INI?

There are some errors in the documentation. It says "AlternativeDamage/Warhead" but the actual tags are "AlternateDamage/Warhead".
It also says "the health’s victim" when it should say "the victim's health" or to be consistent with the ini tags, "the victim's strength".

FS-21:
> > Ultimately failed to get the feature to work, not sure why though. I have the latest nightly build with the code given to me on Mod Haven. I think it may be a case of the code being put in the wrong spot in the INI?
> 
> There are some errors in the documentation. It says "AlternativeDamage/Warhead" but the actual tags are "AlternateDamage/Warhead". It also says "the health’s victim" when it should say "the victim's health" or to be consistent with the ini tags, "the victim's strength".

I updated the tag names in the docs with your feedback.

FS-21:
> Max Lenarc here, just reporting my attempt to test the Mind Control Threshold from a while ago since I have not posted here in Github. I doubt it will be useful though. Ultimately failed to get the feature to work, not sure why though. I have the latest nightly build with the code given to me on Mod Haven. I think it may be a case of the code being put in the wrong spot in the INI? INI looks like this. "[Controller] Verses=100%,100%,100%,100%,100%,100%,0%,0%,0%,0%,100% MindControl=yes AnimList=YURICNTL AffectsAllies=no
> 
> Versus.mcvheavy=0% Versus.mcvheavy.Retaliate=no Versus.mcvheavy.ForceFire=no Versus.mcvheavy.PassiveAcquire=no
> 
> Versus.t2mediump=0% Versus.t2mediump.Retaliate=no Versus.t2mediump.ForceFire=no Versus.t2mediump.PassiveAcquire=no
> 
> Versus.t2heavyp=0% Versus.t2heavyp.Retaliate=no Versus.t2heavyp.ForceFire=no Versus.t2heavyp.PassiveAcquire=no
> 
> Versus.t3heavyp=0% Versus.t3heavyp.Retaliate=no Versus.t3heavyp.ForceFire=no Versus.t3heavyp.PassiveAcquire=no
> 
> Versus.mindarm=0% Versus.mindarm.Retaliate=no Versus.mindarm.ForceFire=no Versus.mindarm.PassiveAcquire=no Versus.sbanearmor=0%
> 
> Transact=true DamageAirThreshold=-1 ; boolean Transact.SpreadAmongTargets=false ; boolean Transact.Experience.Source.Flat=0 ; integer Transact.Experience.Source.Percent=0.40 ; float, percents or absolute (-1.0 - 1.0) Transact.Experience.Source.Percent.CalcFromTarget=true ; boolean Transact.Experience.Target.Flat=0 ; integer Transact.Experience.Target.Percent=0.0 ; float, percents or absolute (-1.0 - 1.0) Transact.Experience.Target.Percent.CalcFromSource=false ; boolean
> 
> MindControl.Threshold= 25% ; positive percentage. Represents a percentage from 0% to 100% MindControl.AlternativeDamage= 50 ; integer MindControl.AlternativeWarhead= PsychoFire; warhead MindControl.CanKill= false ; boolean MindControl.Threshold.Inverse= false ; boolean
> 
> [PsychoFire] Wall=no Wood=no Sparky=yes Verses=100%,100%,100%,100%,100%,100%,50%,50%,50%,100%,100% Conventional=yes InfDeath=10 AnimList=UCINIT ProneDamage=75%" Both of these are under the Warheads list. Apologies in advance if this ends of being a waste of time, also for the wall of text. Tested this on Rise of the East.

The problem was the tag names, I updated the docs.

MaxLenarc:
So that was the problem. Thanks for the info!

MaxLenarc:
Got to testing it again. Everyone works as intended. Admittedly I misunderstood what MindControl.Threshold.Inverse did at first; I thought it killed Mind Controlled units under the Threshold, not kill unit enemy units under the Threshold. I think it is good to go. One of the coolest logics I've seen here!

src/Ext/Bullet/Hooks.cpp
```diff
@@ -137,6 +138,124 @@ DEFINE_HOOK(0x4692BD, BulletClass_Logics_ApplyMindControl, 0x6)
...
+					&& pWarheadExt->MindControl_AlternateWarhead.isset())
+				{
+					int damage = pWarheadExt->MindControl_AlternateDamage;
+					WarheadTypeClass* pAltWarhead = pWarheadExt->MindControl_AlternateWarhead;
+					int realDamage = MapClass::GetTotalDamage(damage, pAltWarhead, pTargetType->Armor, 0);
+
+					if (!pWarheadExt->MindControl_CanKill && pTarget->Health <= realDamage && realDamage > 1)
+						pTarget->Health = realDamage;
+
+					return 0x469343;
```
Metadorius, line 181: I am not sure why you don't explode it.

src/Ext/Bullet/Hooks.cpp
```diff
@@ -137,6 +138,124 @@ DEFINE_HOOK(0x4692BD, BulletClass_Logics_ApplyMindControl, 0x6)
...
+		return 0;
+
+	auto pBulletWH = pBullet->WH;
+	auto pTarget = generic_cast<TechnoClass*>(pBullet->Target);
+
+	if (pTarget 
+		&& pBullet->Owner 
+		&& pBulletWH 
+		&& pBulletWH->MindControl)
+	{
```
Metadorius, line 205: Lots of code duplicated, think of how to avoid it please.

Metadorius (changes requested):

src/Ext/Bullet/Hooks.cpp
FS-21: Becasue I don't want to mess with the code between, these 2 hooks are 2 different (similar but not equal) cases

FS-21:
Finally I remade it, simplified it and in the way fixing the only nasty bug it had :-D

Starkku (changes requested):
The entire implementation could be generalized and massively simplified. There is nothing about this feature that would inherently require tying it to `MindControl`. In addition the current implementation reimplements a lot of things that it really does not need to.

In fact, because we now have `ExtraWarheads` this feature can be simplified down *a lot*. You can generalize the threshold system to not be tied to `MindControl`, remove all of the unnecessary (nullptr) checks that just bloat and slow down the code - only thing you realistically need to check is that the bullet's target is instance of `ObjectClass` or derived classes, e.g it has health. There's no need for damage, there's no need to Warheads or kill switches. The only thing needed is skipping Warhead detonation if the threshold check passees. Because we have `ExtraWarheads`, you can combine two or more Warheads to create health-based conditional Warhead detonation this way without requiring any other bespoke code.

Metadorius (changes requested):
I think the way this feature is designed is too specific. IMO the best would be reworking it into generic threshold logic.

After some brainstorming with @Starkku we came to conclusion that this whole logic can be made much simpler and much more usable. Per idea you would only implement two tags:

```ini
[SOMEWARHEAD]
AffectsBelowPercent=100%  ; means HP of the target
AffectsAbovePercent=0%    ; means HP of the target
```
(both conditions must be fulfilled for the warhead to activate so it works like a range, for example you could have a range between 20% and 80% percents. The code would be just  a two-line check at the very beginning of `BulletClass::Logics` that would bail from func if the conditions pass)

Why? Because we should be doing features that mesh with other features like bricks in the wall, and not overly specific and hardly reusable ones. This way every small feature is simple, and together the features allow for great flexibility.

As such, if you combine it with `Crit` or `ExtraWarhead` logics you can do an arbitrary amount of warheads activating at arbitrary HP ranges. `CanKill` can be done with relative damage logic, or if needed it could be implemented as a separate feature. 

As for the mind control itself it only works if `CaptureManagerClass` is inited for the firer, which currently only happens if one of the weapons is MC one, so to fix that you could iterate all the weapons with their normal, extra and crit (and other possible) warheads that the unit has on init (how the game does it, won't work in edge cases like interceptor replcaing WH with a MC WH), or check for `CaptureManagerClass` when the MC occurs, and if not inited and the firer is there - init it. Or both, which would probably be the best.

FS-21:
Due to the well explained feedback is much better if I close this PR and I rewrite the feature from 0 following the suggestions.

Metadorius:
Well you can simply reuse this PR and just re-code, this shouldn't take much time or changes.

FS-21:
Because I need it for releasing my next mod release I'll update with latest develop commit and then mark it as "draft" because I need to rewrite all the code.

Metadorius:
I'll close this PR for the time being, let's move onto the #1398.

PR #628, opened 2022-05-07
Idle CMIN counter bug fix

### LINKED ISSUE #603: [Phobos Bug] HarvesterCounter believes that the harvester in Chrono ready state is not working
### Description:
If a CMIN on its way back is walking because GAREFN is occupied, the HarvesterCounter believes that the CMIN is not working

### Phobos Build:
[Build 27](https://github.com/Phobos-developers/Phobos/releases/tag/build-27)

### Steps To Reproduce:
Make the number of CMIN more than XAREFN

### Expected Behaviour:
CMIN is still on its way back to the mine and has not changed to a non mission state, which should not be considered as not working. And the resulting high-frequency flashing of the HarvesterCounter is very annoying and meaningless (because the harvester is actually still working and does not need players to give it new instructions)

### Actual Behaviour:
The CMIN in this state is regarded as not working, and make the HarvesterCounter flashes at a high frequency

### Additional Files / Links:
HarvesterCounter_CMIN_Bug

chaserli:
Attempt to fix #603 

Count non-selected moving harvester in guard mission as harvesting state

### DIFF
--- a/src/Ext/Techno/Body.cpp
+++ b/src/Ext/Techno/Body.cpp
@@ -184,10 +184,16 @@ bool TechnoExt::IsHarvesting(TechnoClass* pThis)
 		return true;
 
 	auto mission = pThis->GetCurrentMission();
-	if ((mission == Mission::Harvest || mission == Mission::Unload || mission == Mission::Enter)
-		&& TechnoExt::HasAvailableDock(pThis))
+	if (TechnoExt::HasAvailableDock(pThis))
 	{
-		return true;
+		if (mission == Mission::Harvest || mission == Mission::Unload || mission == Mission::Enter)
+			return true;
+		else if (pThis->WhatAmI() == AbstractType::Unit && mission == Mission::Guard
+			&& !pThis->IsSelected)
+		{
+			if (auto pFoot = abstract_cast<UnitClass*>(pThis))
+				return pFoot->Locomotor->Is_Really_Moving_Now();
+		}
 	}
 
 	return false;


DeathFishAtEase:
> Attempt to fix #603

I've tested it and now it won't flash most of the time. Even the CMIN that has actually been unable to work due to the bug of the RepairBay is considered to be working when it in Enter state :P
why

But if you simply regard Enter as working, they will also be regarded as working when they enter the transport ship, RepairBay and Grinder xD

chaserli:
> > Attempt to fix #603
> 
> I've tested it and now it won't flash most of the time. Even the CMIN that has actually been unable to work due to the bug of the RepairBay is considered to be working when it in Enter state :P
> 
> But if you simply regard Enter as working, they will also be regarded as working when they enter the transport ship, RepairBay and Grinder xD

What you listed are existing issues in current and previous releases. I just excluded the major case which makes that harvester counter keeps blinking. Maybe the one who implemented this has a better criterion @Thrifinesma @Metadorius

src/Ext/Techno/Body.cpp
```diff
@@ -184,10 +184,16 @@ bool TechnoExt::IsHarvesting(TechnoClass* pThis)
...
 	{
-		return true;
+		if (mission == Mission::Harvest || mission == Mission::Unload || mission == Mission::Enter)
+			return true;
+		else if (pThis->WhatAmI() == AbstractType::Unit && mission == Mission::Guard
+			&& !pThis->IsSelected)
+		{
+			if (auto pFoot = abstract_cast<UnitClass*>(pThis))
+				return pFoot->Locomotor->Is_Really_Moving_Now();
+		}
```
FS-21, line 196: If you use if-else you should either have all of the code blocks braced or braceless to keep things consistent.

FS-21 (changes requested):
Needs a small change regarding "{" "}"

chaserli:
If you're cleaning up PRs, this one is a minor code adjustment, and was tested by @DeathFishAtEase as stated above.
It's just about redefining idle harvester, so the consensus is needed from @Metadorius @Thrifinesma.

PR #630, opened 2022-05-08
[4 features]Unit self-destruction logic fully extended+ Slaves' house decision when owner killed+ Building Sold EVA&Sound dehardcode

chaserli:
A total refactor of the existing unit self-destruction logic.
4 new features standalone complex:

### Kill Object Automatically

- Objects can be destroyed automatically if *any* of these conditions is met:
  - `OnAmmoDepletion`: The object will die if the remaining ammo reaches 0.
  - `AfterDelay`: The object will die if the countdown (in frames) reaches 0.

- The auto-death behavior can be chosen from the following:
  - `kill`: The object will be destroyed normally. Please notice that if the unit carries passengers, they will not be released. This might change in the future.
  - `vanish`: The object will be directly removed from the game peacefully instead of actually getting killed.
  - `sell`: If the object is a **building** with buildup, it will be sold instead of killed.

If this option is not set, the self-destruction logic will not be enabled.

In `rulesmd.ini`:
```ini
[SOMETECHNO]                  ; TechnoType
AutoDeath.Behavior=           ; enumeration (kill | vanish | sell), default not set

AutoDeath.OnAmmoDepletion=no  ; boolean
AutoDeath.AfterDelay=0        ; positive integer
```

### Customize building specific `SellSound`

- When a building is sold, the sell sound is played to the owner. This global sound has been deglobalized. Default to `[AudioVisual]->SellSound`.

In `rulesmd.ini`:
```ini
[SOMETECHNO]  ; TechnoType
SellSound=    ; sound entry
```

### Customize building specific `EVA_StructureSold`

- `EVA_StructureSold` can be customized or even disabled for specific building types.

In `rulesmd.ini`:
```ini
[SOMEBUILDING]              ; BuildingType
EVA.Sold=EVA_StructureSold  ; EVA entry
```

### Slaves' house decision customization when owner is killed

- You can now decide the slaves' house when the corresponding slave miner is killed using `Slaved.OwnerWhenMasterKilled`:
  - `suicide`: Kill each slave if the slave miner is killed.
  - `master`: Free the slaves but keep the house of the slave unchanged.
  - `neutral`: The slaves belong to civilian house.
  - `killer`: Free the slaves and give them to the house of the slave miner's killer. (vanilla behavior)

In `rulesmd.ini`
```ini
[SOMEINFANTRY]                     ; Slave type
Slaved=yes
Slaved.OwnerWhenMasterKilled=killer  ; enumeration (suicide | master | killer | neutral), default to killer

### CODE REVIEW

**Comment by IgortheBylarge** [comment]
Tested all new parameters: `Death.NoAmmo`, `Death.Countdown`, `Death.WithSlaveOwner`and `Death.Peaceful`. All of them are working stable, without any issues.

File: src/Ext/TechnoType/Body.cpp [hunk 1]
```diff
@@ -101,6 +101,7 @@ void TechnoTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 	this->Death_NoAmmo.Read(exINI, pSection, "Death.NoAmmo");
 	this->Death_Countdown.Read(exINI, pSection, "Death.Countdown");
 	this->Death_Peaceful.Read(exINI, pSection, "Death.Peaceful");
+	this->Death_WithMaster.Read(exINI, pSection, "Death.WithSlaveOwner");
```
FS-21, line 104: Class fields that can be set via INI tags should be named exactly like ini tags with dots replaced with underscores.

FS-21 (changes requested):
Small changes needed.

chaserli:
@Metadorius Please mark this thread as Needs Testing plz
@IgortheBylarge Are you available to test it again?

src/Ext/TechnoType/Body.h
```diff
@@ -52,9 +52,12 @@ class TechnoTypeExt
 		NullableIdx<VocClass> PassengerDeletion_ReportSound;
 		Valueable<bool> PassengerDeletion_Rate_SizeMultiply;
 		Nullable<AnimTypeClass*> PassengerDeletion_Anim;
+
 		Valueable<bool> Death_NoAmmo;
 		Valueable<int> Death_Countdown;
-		Valueable<bool> Death_Peaceful;
+		Valueable<HowToSuicide> Death;
```
Metadorius, line 58: 1. Could be Nullable instead of having to add a new enum value but not sure if that would be useful much, up to you.
2. `Death` is not a very descriptive name. I propose to name all the death tags with `AutoDeath` prefix and give the following names (or something like this):
```
AutoDeath.Behavior   <- this one
AutoDeath.OnAmmoDepletion
AutoDeath.AfterDelay
```

src/Ext/Techno/Body.cpp
```diff
@@ -481,58 +481,85 @@ bool TechnoExt::CanFireNoAmmoWeapon(TechnoClass* pThis, int weaponIndex)
...
+					}
+					else
+					{
+						// Countdown ended. Kill the unit
+						pData->Death_Countdown = -1;
+						TechnoExt::KillSelf(pThis, howToDie);
+
+						return;
+					}
 				}
```
Metadorius, line 554: Pretty sure this could use a timer, see shield code if you're upt to improving this piece of code.

src/Ext/TechnoType/Body.h
```diff
@@ -52,9 +52,12 @@ class TechnoTypeExt
...
+		Valueable<SlavesGiveTo> Slaved_OwnerWhenMasterDead;
+		ValueableIdx<VocClass> SellSound;
```
Metadorius, line 60: If it's a building-exclusive thing why is it in `TechnoType`? BTW could also be `NullableIdx`.

src/Utilities/Enum.h
```diff
@@ -149,6 +149,21 @@ enum class AreaFireTarget
 	Random = 2
 };
 
+enum class SlavesGiveTo
+{
+	Killer = 0, // default
+	Master = 1,
+	Suicide = 2
+};
```
Metadorius, line 157: 1. Could also add `Neutral` option.
2. Not sure on the name, it sounds broken a bit.

src/Utilities/Enum.h
```diff
@@ -149,6 +149,21 @@ enum class AreaFireTarget
...
+
+enum class HowToSuicide
```
Metadorius, line 159: Same deal with the name.

src/Ext/TechnoType/Body.h
```diff
@@ -214,9 +217,11 @@ class TechnoTypeExt
 			, DeployingAnim_KeepUnitVisible { false }
 			, DeployingAnim_ReverseForUndeploy { true }
 			, DeployingAnim_UseUnitDrawer { true }
+			, Death { HowToSuicide::Disabled }
 			, Death_NoAmmo { false }
 			, Death_Countdown { 0 }
-			, Death_Peaceful { false }
+			, Slaved_OwnerWhenMasterDead { SlavesGiveTo::Killer }
+			, SellSound{ -1 }
```
Metadorius, line 224:
```suggestion
			, SellSound { -1 }
```

Metadorius (changes requested):
Code is mostly good apart from a few issues.

src/Ext/BuildingType/Body.h
```diff
@@ -44,6 +44,8 @@ class BuildingTypeExt
 		CustomPalette PlacementPreview_Palette;
 		Nullable<int> PlacementPreview_TranslucentLevel;
 
+		ValueableIdx<VoxClass> EVA_Sold;
```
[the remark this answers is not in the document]
chaserli: I'm a little bit confused with the use of `ValueableIdx` and `NullableIdx`, Actually, I initialized this value in `Ext\TechnoType\Body.cpp`'s section `void BuildingTypeExt::ExtData::Initialize();` same as the SellSound stuff. I would like to know if this is correct

src/Ext/TechnoType/Body.h
chaserli: 1. Because vehicles can be sold on repair bays, if someone uses this eventually
2. same question mentioned above

src/Ext/BuildingType/Body.h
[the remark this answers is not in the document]
Metadorius: I don't think that's correct. Usually we don't duplicate the value like this if it's not set, I am pretty sure EVA sell sound already exists initialized somewhere. I guess you would need to change it to nullable and use get with a default value.

src/Utilities/Enum.h
chaserli: 1. I thought about this at the beginning, but it doesn't sound quite reasonable imo
2. A long enum name? like `SlaveOwnershipChangeType`?

src/Ext/BuildingType/Body.h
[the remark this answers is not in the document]
chaserli: 1.I was taking https://github.com/Phobos-developers/Antares/blob/7241a5ff20f4dbf7153cc77e16edca5c9db473d4/src/Ext/SWType/Body.cpp#L78
as a reference from ares. They initialized `EVA_InsufficientFunds` etc in the `::ExtData::Initialize` section too, which are also  initialized as -1 in the header file

2. I don't know if EVA_Sold is initialized or not, but in 0x449CC1 and 0x44AB22 (addresses where I hooked), the vanilla game seems to simply do `VoxClass::Play("EVA_StructureSold")`. What I intended to do instead was to get the vox index when initailizing, and use `VoxClass::PlayIndex(BuidlingTypeExt->EVA_Sold_idx)`.

chaserli:
New issue identified: if the object enters a transport, it will not get deleted.

src/Ext/Techno/Body.cpp
chaserli: changed to make the opentopped units coherent with other transports

src/Ext/BuildingType/Hooks.cpp
```diff
@@ -68,6 +68,29 @@ DEFINE_HOOK(0x458623, BuildingClass_KillOccupiers_Replace_MuzzleFix, 0x7)
...
+DEFINE_HOOK(0x44AB22, BuildingClass_Mission_Deconstruction_EVA_Sold_2, 0x6)
+{
+	GET(BuildingClass*, pThis, EBP);
+	const auto pTypeExt = BuildingTypeExt::ExtMap.Find(pThis->Type);
+
+	if (pThis->Owner->ControlledByPlayer())
+		VoxClass::PlayIndex(pTypeExt->EVA_Sold.Get());
+
+	return 0x44AB3B;
+}
```
Metadorius, line 92: I am pretty sure with how you still don't use `NullableIdx` and don't fallback to the default value that this would break all vanilla sell sounds.

src/Ext/Techno/Body.cpp
```diff
@@ -480,60 +480,83 @@ bool TechnoExt::CanFireNoAmmoWeapon(TechnoClass* pThis, int weaponIndex)
...
+				return;
 			}
 		}
+
+		Debug::Log("[Runtime Warning] %s can't be sold, killing it instead\n", pThis->get_ID());
+	}
+
+	default: //must be AutoDeathBehavior::Kill
+		pThis->ReceiveDamage(&pThis->Health, 0, RulesClass::Instance()->C4Warhead, nullptr,
+			true,// ignoreDefenses = false to let passengers escape, why?
```
Metadorius, line 514:
```suggestion
			true,  // ignoreDefenses = false to let passengers escape, why?
```

src/Ext/BuildingType/Hooks.cpp
chaserli: I'll have to hook inside the if statement in this case, I'm not sure how to do so

Metadorius: Um I don't think you need to, you just need to know what's the default index of the EVA sold sound, aka the one that is set globally, change type of `EVA_Sold` to `NullableIdx<VoxClass>` and then
```cpp
VoxClass::PlayIndex(pTypeExt->EVA_Sold.Get(/* default value here */));
```

chaserli: So like 
```cpp
VoxClass::PlayIndex(pTypeExt->EVA_Sold.Get(VoxClass::FindIndex((const char*)0x819030)));
```?

Comment (Metadorius) [reply src/Ext/BuildingType/Hooks.cpp:92]: You can also just check `.isset()` and `return 0` if the nullable is not set.
YRpp name seems to be incorrect. Offset 0x41A seems to be `IsOwnedByPlayer`.

File: src/Ext/Techno/Hooks.cpp [hunk 11]
```diff
@@ -481,3 +512,64 @@ DEFINE_HOOK(0x70E1A5, TechnoClass_GetTurretWeapon_LaserWeapon, 0x6)
...
+	return 0x0;
+}
+
+DEFINE_HOOK(0x4D9F8A, FootClass_Sell_Sellsound, 0x5)
+{
+	GET(FootClass*, pThis, ESI);
+
+	auto pTypeExt = TechnoTypeExt::ExtMap.Find(pThis->GetTechnoType());
+
+	VoxClass::PlayIndex(pTypeExt->EVA_Sold.Get(VoxClass::FindIndex((const char*)0x882630))); //EVA_UnitSold
```
chaserli, line 571: This is what I meant, very dirty

Metadorius: `return 0` isn't an option?

chaserli: Elsewhere, yes, but here…it seems too narrow to do so

src/Ext/Techno/Body.cpp
```diff
@@ -450,6 +449,7 @@ void TechnoExt::EatPassengers(TechnoClass* pThis)
 							}
 						}
 
+
```
Metadorius, line 452:
```suggestion
```

src/Ext/Techno/Body.cpp
Metadorius: Missed this :P

src/Ext/Techno/Body.cpp
```diff
@@ -482,64 +482,83 @@ bool TechnoExt::CanFireNoAmmoWeapon(TechnoClass* pThis, int weaponIndex)
...
+	{
 
-	const bool peacefulDeath = pTypeData->Death_Peaceful.Get();
-	// Death if no ammo
-	if (pTypeThis && pTypeData && pTypeData->Death_NoAmmo)
+	case AutoDeathBehavior::Vanish:
 	{
-		if (pTypeThis->Ammo > 0 && pThis->Ammo <= 0)
+		pThis->Limbo();
+		pThis->RegisterKill(pThis->Owner);
```
Metadorius, line 493: Forgot KillPassengers call here. May impact some edge cases like capturing enemy infantry and then killing them.

src/Ext/Techno/Hooks.cpp
```diff
@@ -481,3 +512,68 @@ DEFINE_HOOK(0x70E1A5, TechnoClass_GetTurretWeapon_LaserWeapon, 0x6)
 	R->EAX(pThis->GetTechnoType());
 	return Continue;
 }
+
+// SellSound and EVA dehardcode
+DEFINE_HOOK(0x449CC1, BuildingClass_Mission_Deconstruction_EVA_Sold_1, 0x6)
+{
+	GET(BuildingClass*, pThis, EBP);
+	enum { SkipVoxPlay = 0x449CEA };
```
Metadorius, line 520: Should keep enum as the first line.

src/Ext/Techno/Hooks.cpp
```diff
@@ -481,3 +512,68 @@ DEFINE_HOOK(0x70E1A5, TechnoClass_GetTurretWeapon_LaserWeapon, 0x6)
...
+		return SkipVoxPlay;
+	}
+
+	return 0x0;
+}
+
+DEFINE_HOOK(0x44AB22, BuildingClass_Mission_Deconstruction_EVA_Sold_2, 0x6)
+{
+	GET(BuildingClass*, pThis, EBP);
+	enum { SkipVoxPlay = 0x44AB3B };
```
Metadorius, line 537: Same here

src/Ext/Techno/Hooks.cpp
```diff
@@ -481,3 +512,68 @@ DEFINE_HOOK(0x70E1A5, TechnoClass_GetTurretWeapon_LaserWeapon, 0x6)
...
+	}
+
+	return 0x0;
+}
+
+
+DEFINE_HOOK(0x44A850, BuildingClass_Mission_Deconstruction_Sellsound, 0x6)
+{
+	GET(BuildingClass*, pThis, EBP);
+	enum { PlayVocLocally = 0x44A856 };
```
Metadorius, line 555: Ditto

src/Ext/Techno/Hooks.cpp
```diff
@@ -481,3 +512,68 @@ DEFINE_HOOK(0x70E1A5, TechnoClass_GetTurretWeapon_LaserWeapon, 0x6)
...
+		return PlayVocLocally;
+	}
+
+	return 0x0;
+}
+
+DEFINE_HOOK(0x4D9F8A, FootClass_Sell_Sellsound, 0x5)
+{
+	GET(FootClass*, pThis, ESI);
+	enum { EVA_UnitSold = 0x822630, SkipVoxVocPlay = 0x4D9FB5 };
```
Metadorius, line 570: Ditto

src/Ext/Techno/Hooks.cpp
```diff
@@ -481,3 +512,68 @@ DEFINE_HOOK(0x70E1A5, TechnoClass_GetTurretWeapon_LaserWeapon, 0x6)
...
+
+	auto pTypeExt = TechnoTypeExt::ExtMap.Find(pThis->GetTechnoType());
+
+	VoxClass::PlayIndex(pTypeExt->EVA_Sold.Get(VoxClass::FindIndex((const char*)EVA_UnitSold)));
```
Metadorius, line 574: Also strings like this should be in YRpp but I guess can be done in another PR, there are many of such cases in Phobos.

Metadorius (changes requested):
Good apart from a few minor nitpicks.

Metadorius (approved):

src/Ext/Techno/Body.cpp
```diff
@@ -482,64 +481,81 @@ bool TechnoExt::CanFireNoAmmoWeapon(TechnoClass* pThis, int weaponIndex)
...
+
+				return;
 			}
 		}
+
+		Debug::Log("[Runtime Warning] %s can't be sold, killing it instead\n", pThis->get_ID());
+	}
+
+	default: //must be AutoDeathBehavior::Kill
+		pThis->ReceiveDamage(&pThis->Health, 0, RulesClass::Instance()->C4Warhead, nullptr, true, false, pThis->Owner);
```
chaserli, line 515: Just a reminder : when `ignoreDefenses=yes`, Ares' `TechnoExt::SpawnSurvivors` will not eject survivors/passengers/hijacker

PR #635, opened 2022-05-11
Reimplemented animation damage / weapon /w improvements

Starkku:
This reimplements animation `Weapon` and `Damage.Delay` from Ares and adds ability to customize whether or not the Damage is dealt by the owner or invoker of the animation (does not affect which House is used to deal damage if `Warhead` is used, but does affect experience gain). `Weapon` is also automatically allocated and does not require explicit listing on `[WeaponTypes]`.

---------------------------

### Animation weapon and damage settings

- `Weapon` can be set to a WeaponType, to create a projectile and immediately detonate it instead of simply dealing `Damage` by `Warhead`. This allows weapon effects to be applied.
- `Damage.Delay` determines delay between two applications of `Damage`. Requires `Damage` to be set to 1.0 or above. Value of 0 disables the delay. Keep in mind that this is measured in animation frames, not game frames. Depending on `Rate`, animation may or may not advance animation frames on every game frame.
- `Damage.DealtByInvoker`, if set to true, makes any `Damage` dealt to be considered as coming from the animation's invoker (f.ex, firer of the weapon if it is Warhead `AnimList/SplashList` animation, the destroyed vehicle if it is `DestroyAnim` animation or the object the animation is attached to). Does not affect which house the `Damage` dealt by `Warhead` is dealt by.
- `Damage.ApplyOnce`, if set to true, makes `Damage` be dealt only once per animation loop instead of on every frame or intervals defined by `Damage.Delay`. The frame on which it is dealt is determined by `Damage.Delay`, defaulting to after the first animation frame.

In `artmd.ini`:
```ini
[SOMEANIM]                   ; AnimationType
Weapon=                      ; WeaponType
Damage.Delay=0               ; integer, animation frames
Damage.DealtByInvoker=false  ; boolean
Damage.ApplyOnce=false       ; boolean
```

*`Weapon` and `Damage.Delay`, beyond the other additions, should function similarly to the equivalent features introduced by Ares and take precedence over them if Phobos is used together with Ares.*

--------

**NOTE:** *Following replaces `DetachedFromOwner` on WeaponTypes which has now been **deprecated**.*

### Allowing damage dealt to firer

- You can now allow warhead to deal damage (and apply damage-adjacent effects such as `KillDriver` and `DisableWeapons/Sonar/Flash.Duration` *(Ares features)*) on the object that is considered as the firer of the Warhead even if it does not have `DamageSelf=true`.
  - Note that effect of `Psychedelic=true`, despite being tied to damage will still fail to apply on the firer as it does not affect any objects belonging to same house as the firer, including itself (**NOTE:** *This is because Ares replaces the game check in its entirety, otherwise would have also made this customizable.*)

In `rulesmd.ini`:
```ini
[SOMEWARHEAD]            ; WarheadType
AllowDamageOnSelf=false  ; boolean
```

### DIFF
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -387,6 +387,17 @@ DeployingAnim.UseUnitDrawer=true       ; boolean
 
 ## Warheads
 
+### Allowing damage dealt to firer
+
+- You can now allow warhead to deal damage (and apply damage-adjacent effects such as `KillDriver` and `DisableWeapons/Sonar/Flash.Duration` *(Ares features)*) on the object that is considered as the firer of the Warhead even if it does not have `DamageSelf=true`.
+  - Note that effect of `Psychedelic=true`, despite being tied to damage will still fail to apply on the firer as it does not affect any objects belonging to same house as the firer, including itself.
+
+In `rulesmd.ini`:
+```ini
+[SOMEWARHEAD]            ; WarheadType
+AllowDamageOnSelf=false  ; boolean
+```
+
 ### Customizing decloak on damaging targets
 
 - You can now specify whether or not the warhead decloaks objects that are damaged by the warhead.
@@ -419,18 +430,6 @@ In `rulesmd.ini`:
 [SOMEWEAPON]          ; WeaponType
 DiskLaser.Radius=38.2 ; floating point value
                       ; 38.2 is roughly the default saucer disk radius
-```
-
-### Detaching weapon from owner TechnoType
-
-- You can now control if weapon is detached from the TechnoType that fired it. This results in the weapon / warhead being able to damage the TechnoType itself even if it does not have `DamageSelf=true` set, but also treats it as if owned by no house or object, meaning any ownership-based checks like `AffectsAllies` do not function as expected and no experience is awarded.
-  - The effect of this is inherited through `AirburstWeapon` and `ShrapnelWeapon`.
-  - This does not affect projectile image or functionality or `FirersPalette` on initially fired weapon, but `FirersPalette` will not function for any weapons inheriting the effect.
-
-In `rulesmd.ini`:
-```ini
-[SOMEWEAPONTYPE]         ; WeaponType
-DetachedFromOwner=false  ; boolean
 ```
 
 ### Single-color lasers
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -66,6 +66,7 @@ This page describes all ingame logics that are fixed or improved in Phobos witho
 - Allow use of `Foundation=0x0` on TerrainTypes without crashing for similar results as for buildings.
 - Projectiles now remember the house of the firer even if the firer is destroyed before the projectile detonates. Does not currently apply to some Ares-introduced Warhead effects like EMP.
 - `OpenTopped` transports now take `OpenTransportWeapon` setting of passengers into consideration when determining weapon range used for threat scanning and approaching targets.
+- Trailer animations now inherit the owner of the object they are attached to.
 
 ## Animations
 
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -74,18 +74,20 @@ This page describes all ingame logics that are fixed or improved in Phobos witho
 
 - `Weapon` can be set to a WeaponType, to create a projectile and immediately detonate it instead of simply dealing `Damage` by `Warhead`. This allows weapon effects to be applied.
 - `Damage.Delay` determines delay between two applications of `Damage`. Requires `Damage` to be set to 1.0 or above. Value of 0 disables the delay. Keep in mind that this is measured in animation frames, not game frames. Depending on `Rate`, animation may or may not advance animation frames on every game frame.
-- `Damage.DealtByOwner`, if set to true, makes any `Damage` dealt to be considered as coming from the animation's owner. Owner in this case is the firer of the weapon if it is Warhead `AnimList/SplashList` animation, the destroyed vehicle if it is `DestroyAnim` animation or the object the animation is attached to. Does not affect which house the `Damage` dealt by `Warhead` is dealt by.
+- `Damage.DealtByInvoker`, if set to true, makes any `Damage` dealt to be considered as coming from the animation's invoker (f.ex, firer of the weapon if it is Warhead `AnimList/SplashList` animation, the destroyed vehicle if it is `DestroyAnim` animation or the object the animation is attached to). Does not affect which house the `Damage` dealt by `Warhead` is dealt by.
 - `Damage.ApplyOnce`, if set to true, makes `Damage` be dealt only once per animation loop instead of on every frame or intervals defined by `Damage.Delay`. The frame on which it is dealt is determined by `Damage.Delay`, defaulting to after the first animation frame.
-
-*`Weapon` and `Damage.Delay`, beyond the other additions, should function similarly to the equivalent features introduced by Ares and take precedence over them if Phobos is used together with Ares.*
 
 In `artmd.ini`:
 ```ini
-[SOMEANIM]                 ; AnimationType
-Weapon=                    ; WeaponType
-Damage.Delay=0             ; integer, animation frames
-Damage.DealtByOwner=false  ; boolean
-Damage.ApplyOnce=false     ; boolean
+[SOMEANIM]                   ; AnimationType
+Weapon=                      ; WeaponType
+Damage.Delay=0               ; integer, animation frames
+Damage.DealtByInvoker=false  ; boolean
+Damage.ApplyOnce=false       ; boolean
+```
+
+```{note}
+`Weapon` and `Damage.Delay`, beyond the other additions, should function similarly to the equivalent features introduced by Ares and take precedence over them if Phobos is used together with Ares.
 ```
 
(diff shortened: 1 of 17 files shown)


Starkku:
Misc. fixes

- Damage.Delay now starts the count from correct animation frame
- Total damage dealt by animation no longer has one extra point of damage
- Fixed a damage check that was incorrectly disallowing > 0 but < 1 damage

Starkku:
Fixed an issue with looping animations and Damage.Delay where it would 'reset' the damage delay counter when a new loop started. It now works independently from animation frame state (and thus works better with loops) and matches how it works in Ares.

Starkku:
Added `Damage.ApplyOnce` which allows for animation damage to be only dealt once per animation loop on a specific animation frame.

ayylmaoRotE:
Tested using a basic cannon weapon with AnimList as:

```ini
[DAMAGEDELAY]
Translucent=yes
Image=MININUKE
Damage=500
LoopCount=10
Damage.Delay=1
Damage.DealtByOwner=yes
Damage.ApplyOnce=yes
Weapon=NukePayload
```

With variations such as `Damage.ApplyOnce=no`, different `Damage.Delay` values and by removing `Weapon` tag and only using `Warhead` and `Damage`.

Confirming all working as described.

Observations:
1. If firer dies while animation is looping in damage, nothing happens, e.g. no crash 👍 
2. Firer gains veterancy as expected
3. TransactMoney part 1: Using only `Damage` and `Warhead` + `Damage.DealtByOwner=yes` will have the money displayed from the firer and the owner will receive the corresponding $, but `Damage.DealtByOwner=no` will have the money animation displayed at the detonation point. Will still give the owner $$ though
4. TransactMoney part 2: Using the same method but applying it via a `Weapon` with `Damage.DealtByOwner=yes` will have the same effect as it did in part 1. But `Damage.DealtByOwner=no` will not display any cash nor give any funds.
5. AttachEffects part 1: Will grant units veterancy via this method if enemies are killed but will also damage their firer while weapons/warheads triggering the same effect will not damage the firer. AffectsOwner=no fixes this, so I think the issue is actually that weapons _should_ damage firers by default but they do not. (e.g. triggered via AnimList)
6. AttachEffects part 2: Works the same as point 4 when combined with TransactMoney.
7. Works with effects such as Cluster and EMP but only on `Weapon` use, not `Warhead`. As expected
8. AttachEffects part3: Applying AE makes the targeted unit the owner, pretty neat, all other attributes are as described above.
9. Negative values did not work with damage neither in `Damage` values or `Verses`. TransactMoney negatives works.

Suggestions:
1. Probably Ares territory but is it possible to make it so if AE is applied to enemies, the animation owner is still the firer? If so maybe via a tag? Refer to point 8

Starkku:
> Suggestions:
> Probably Ares territory but is it possible to make it so if AE is applied to enemies, the animation owner is still the firer? If so maybe via a tag? Refer to point 8

Not doable under current circumstances.

Starkku:
Deprecated `DetachedFromOwner` weapon key in favour of adding `AllowDamageOnSelf` on warheads that can be applied for same effect (allow dealing damage and damage-adjacent effects on the object considered warhead's source/firer) in wider scenarios and does not have weird interactions with AirburstWeapon.

Starkku:
- Changed `Damage.DealtByOwner` to `Damage.DealtByInvoker`. This is to prevent confusion over the meaning of term 'owner' which often refers to a house.
- `Trailer` and `TrailerAnim` now have `Owner` and `Invoker` set appropriately based on the object they are attached to.

FS-21:
I tested these tags in a trailer animation and works as expected.

Weapon= works as expected
Damage.Delay= starts the damage from the selected animation frame as expected
Damage.DealtByInvoker=true works! Now units that use trailer anims for damaging other objects now can promote. Set as false is the vanilla behavior.
Damage.ApplyOnce=false/true works as expected

One of the final tests I did for my own project is and works as planned in **Trailers**.
```
; Weapon's Projectile
[WavesFakeProjectile]
Image=invisible_pixel
Trailer=TWSMOKEYWAVES     ; <---- **tested here**

[TWSMOKEYWAVES]
Warhead=TWSonicWarhead
Damage=10
Report=TWSonicWave
Flat=yes
Normalized=yes
Translucent=yes
Translucency=75
AltPalette=yes
ZAdjust=-150

Start=0
LoopStart=0
LoopEnd=19
LoopCount=1

**Damage.Delay=4** ; Starts applying damage at TWSMOKEYWAVES's frame 5
**Damage.DealtByInvoker=yes** ; Now the unit can promote when kills enemies
**Damage.ApplyOnce=no** ; Each TWSMOKEYWAVES's frame does 10 of damage
```

src/Ext/AnimType/Hooks.cpp
```diff
@@ -67,5 +68,109 @@ DEFINE_HOOK(0x424C49, AnimClass_AttachTo_BuildingCoords, 0x5)
...
+	if (adjustAccum)
+		pThis->Accum = damage - appliedDamage;
+	else
+		pThis->Accum = 0.0;
+
+	auto const pExt = AnimExt::ExtMap.Find(pThis);
+	TechnoClass* pInvoker = nullptr;
+
+	if (pTypeExt->Damage_DealtByInvoker)
+		pInvoker = pExt->Invoker ? pExt->Invoker : pThis->OwnerObject ? abstract_cast<TechnoClass*>(pThis->OwnerObject) : nullptr;
```
FS-21, line 141: Avoid nested ternary operators.

src/Ext/AnimType/Hooks.cpp
```diff
@@ -67,5 +68,109 @@ DEFINE_HOOK(0x424C49, AnimClass_AttachTo_BuildingCoords, 0x5)
...
+
+	if (pTypeExt->Weapon.isset())
+	{
+		WeaponTypeExt::DetonateAt(pTypeExt->Weapon.Get(), pThis->GetCoords(), pInvoker, appliedDamage);
+	}
+	else
+	{
+		auto pWarhead = pThis->Type->Warhead ? pThis->Type->Warhead :
```
FS-21, line 149: Avoid nested ternary operators.

src/Ext/AnimType/Hooks.cpp
```diff
@@ -67,5 +68,109 @@ DEFINE_HOOK(0x424C49, AnimClass_AttachTo_BuildingCoords, 0x5)
...
+			strcmp(pThis->Type->get_ID(), "INVISO") ? RulesClass::Instance->FlameDamage2 : RulesClass::Instance->C4Warhead;
+
+		auto pOwner = pInvoker ? pInvoker->Owner : pThis->Owner ? pThis->Owner :
```
FS-21, line 152: Avoid nested ternary operators.

FS-21 (changes requested):
Nested ternary operators are always bad, only one level of ternary operators should be used.

docs/Fixed-or-Improved-Logics.md
```diff
@@ -65,9 +65,30 @@ This page describes all ingame logics that are fixed or improved in Phobos witho
...
 
+### Animation weapon and damage settings
+
+- `Weapon` can be set to a WeaponType, to create a projectile and immediately detonate it instead of simply dealing `Damage` by `Warhead`. This allows weapon effects to be applied.
+- `Damage.Delay` determines delay between two applications of `Damage`. Requires `Damage` to be set to 1.0 or above. Value of 0 disables the delay. Keep in mind that this is measured in animation frames, not game frames. Depending on `Rate`, animation may or may not advance animation frames on every game frame.
+- `Damage.DealtByInvoker`, if set to true, makes any `Damage` dealt to be considered as coming from the animation's invoker (f.ex, firer of the weapon if it is Warhead `AnimList/SplashList` animation, the destroyed vehicle if it is `DestroyAnim` animation or the object the animation is attached to). Does not affect which house the `Damage` dealt by `Warhead` is dealt by.
+- `Damage.ApplyOnce`, if set to true, makes `Damage` be dealt only once per animation loop instead of on every frame or intervals defined by `Damage.Delay`. The frame on which it is dealt is determined by `Damage.Delay`, defaulting to after the first animation frame.
```
Metadorius, line 77: Confusing naming, probably should be amended to `Damage.ApplyOncePerLoop`.

src/Ext/Aircraft/Hooks.cpp
```diff
@@ -83,4 +84,21 @@ DEFINE_HOOK(0x418B1F, AircraftClass_Mission_Attack_FireAtTarget5Strafe_BurstFix,
...
+	{
+		if (auto const pAnimExt = AnimExt::ExtMap.Find(pAnim))
+		{
+			pAnim->Owner = pThis->Owner;
+			pAnimExt->Invoker = pThis;
+		}
+	}
+
+	return 0;
 }
```
Metadorius, line 104: Doesn't seem to be mentioned in docs?

Metadorius (changes requested):
Can be merged after addressing the comments. (Assuming MaxLenarc's feedback is sufficient.)

src/Ext/Aircraft/Hooks.cpp
Starkku: [But it is](https://github.com/Phobos-developers/Phobos/pull/635/files#diff-b650a8bcca29a01d5a69be07a793712810b9d1d400547d36d22e70677249c416R68)

Metadorius: I'm blind lol, that's why you don't review when it's 1AM kids

PR #639, opened 2022-05-15
Another implementation on turning jumpjet to its target

chaserli:
**This is the exact same thing as my previous "Jumpjet facing target" pull request, with a different implementation.
It won't check if the jumpjet is stationary or not.**

**Need comparison tests**

Description of the previous PR:
### Jumpjet facing target customization 

- Allows jumpjet units to face towards the target when firing from different directions. Set `[JumpjetControls] -> TurnToTarget=yes` to enable it for all jumpjet locomotor units. This behavior can be overriden by setting `[UnitType] -> JumpjetTurnToTarget` for specific units.
- This behavior does not apply to `TurretSpins=yes` units for obvious reasons.

In `rulesmd.ini`:
```ini
[JumpjetControls]
TurnToTarget=no        ; boolean

[SOMEUNITTYPE]         ; UnitType with jumpjet locomotor
JumpjetTurnToTarget=   ; boolean, override the tag in JumpjetControls
```

### DIFF
--- a/src/Ext/Techno/Hooks.VehicleToBuildingMCFix.cpp
+++ b/src/Ext/Techno/Hooks.VehicleToBuildingMCFix.cpp
@@ -97,7 +97,6 @@ DEFINE_HOOK(0x448460, BuildingClass_Captured_MuteSound, 0x6)
 		0x44848F : 0;
 }
 
-
 DEFINE_HOOK(0x54AEC0, JumpjetLocomotion_Process, 0x8)
 {
 	GET_STACK(ILocomotion*, iLoco, 0x4);
--- a/src/Ext/Techno/Hooks.VehicleToBuildingMCFix.cpp
+++ b/src/Ext/Techno/Hooks.VehicleToBuildingMCFix.cpp
@@ -1,7 +1,7 @@
 #include "Body.h"
 
 #include <Phobos.h>
-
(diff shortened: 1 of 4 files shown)


DOOMGUY2K:
[JumpjetControls]
TurnToTarget=yes

Works Globally

JumpjetTurnToTarget=no also overrides the global tag

SCHP can turn to face "target" even if one tile away, or if further will pursue and continue to face.

src/Ext/Techno/Hooks.VehicleToBuildingMCFix.cpp
```diff
@@ -96,3 +96,29 @@ DEFINE_HOOK(0x448460, BuildingClass_Captured_MuteSound, 0x6)
 	return MindControlFixTemp::isMindControlBeingTransferred ?
 		0x44848F : 0;
 }
+
```
FS-21, line 99: Delete 1 of the 2 empty lines.

src/Ext/Techno/Hooks.VehicleToBuildingMCFix.cpp
```diff
@@ -96,3 +96,29 @@ DEFINE_HOOK(0x448460, BuildingClass_Captured_MuteSound, 0x6)
...
+
```
Metadorius, line 100:
```suggestion
```

PR #641, opened 2022-05-16
Wrong picture in Chinese version document（中文版文档图片错了）

sovzzz:
Change it to the right one in English version.


sovzzz:
This mistake is caused by .PO files' shortcoming. They find strings by content not ID. GNU gettext was not designed to be used in Markdown files, making it worse. I also located some sentences was not tranlation, because they are not identified by it, they don't exist in .PO files. Such as New/Enhanced Logics->Warheads->generate credits on impact. Its title was translated, but main content remains untranslated.

Thrifinesma:
Yes, there are indeed something untranslated because of updates, I'll translate them later (maybe days later)

PR #645, opened 2022-05-22
Initial Strength For Cloned Infantry

FS-21:
- You can now specify how many hitpoints an Infantry Type starts with when leaves a Cloning Structure.
- If InitialStrength.Cloning.Min is declared then a random value between InitialStrength.Cloning.Min and InitialStrength.Cloning is used for the new unit.

In rulesmd.ini:

[SOMEBUILDING]                      ; BuildingType
InitialStrength.Cloning=           ; double / percentage
InitialStrength.Cloning.Min=    ; double / percentage


FS-21:
Example:
image

Metadorius:
I suggest you to check my SpawnsTiberium improvement tags for a better example on how to do ranged/single values.

src/Ext/Techno/Hooks.cpp
```diff
@@ -79,6 +79,49 @@ DEFINE_HOOK(0x414057, TechnoClass_Init_InitialStrength, 0x6)       // AircraftCl
...
 }
 
+DEFINE_HOOK(0x443C81, BuildingClass_ExitObject_InitialClonedHealth, 0x7)
+{
+	GET(BuildingClass*, pBuilding, ESI);
+	GET(FootClass*, pFoot, EDI);
+
+	bool isCloner = false;
+
+	if (pBuilding && pBuilding->Type->Cloning)
```
chaserli, line 89: Will any object coming out of the building be considered as clone?

FS-21: From what I tested the only objects that come out from Cloning=yes structures are cloned stuff.

FS-21:
Applied feedback so now is one 1 tag that can take 1 or 2 values of the type double/percentage.

For example:
imagen

MaxLenarc:
Tried out the logic. Took me a bit to figure out not to use the InitialStrengthCloning.Min at the top. Got it to work perfectly. Seems to hate Rise of the East though. I pasted the phobos logic with it into my test subject ROTE copy and it reverted the mod back to Vanilla RA2YR. Oh and here are the results of some of my tests. Hope it helps!
Only use InitialStrength.Cloning for it to work properly.
Works as intended.
Attempts to break the logic via using percentages 0% and 110% failed. Cloned units instead seem to come out with whatever HP in between.
Attempt to break logic with -1% and 200% failed. Same result as above.
Attempt to break logic with only -1% failed. Seems to not register negative numbers. Though Cloned units came out with lowest HP possible. Tested with Initiates.
No issues here from the looks of it.

FS-21:
Thanks for confirming it and yes, the tag for Min percentage was deprecated and now in the same tag you can put 1 value for a static percentage or 2 values separated by commas for a random percentage.

Starkku (approved):
Code looks fine to me as well.

src/Ext/Techno/Hooks.cpp
```diff
@@ -78,6 +78,38 @@ DEFINE_HOOK(0x414057, TechnoClass_Init_InitialStrength, 0x6)       // AircraftCl
...
+		isCloner = true;
+
+	if (isCloner && pFoot)
+	{
+		if (auto pTypeExt = TechnoTypeExt::ExtMap.Find(pBuilding->GetTechnoType()))
+		{
+			if (auto pTypeUnit = pFoot->GetTechnoType())
+			{
+				Vector2D<double> range = pTypeExt->InitialStrength_Cloning.Get();
+				double percentage = range.X >= range.Y ? range.X : (ScenarioClass::Instance->Random.RandomRanged(range.X * 100, range.Y * 100) / 100.0);
```
chaserli, line 98: I think it would be better to cast double to int to avoid compiler warnings

src/Ext/Techno/Hooks.cpp
chaserli: I think it would be better to add a sanity check to ensure infantrytype, in case some users attempt to clone vehicles with it.

PR #657, opened 2022-05-31
Implement C3 translucent SHP blitter fixes

Metadorius:
- Translucent SHPs are now drawn using a more precise (has no green tint in particular) and performant algorithm.
  - Currently only applies to Compression 3 SHPs.


Starkku (approved):
Code is fine. Also works as expected.
transparency
*(Transparency=75, left is old and right is new. The SHP used has a subtle blue hue)*

Due to the more accurate drawing, the new method can produce more opaque and/or lighter results. The old appearance is still available on uncompressed SHP's, at least for now, if it is needed for something specific.

Metadorius:
Looks more washed out, but I guess it's intended?

Metadorius:
I also wonder about the performance, are there any perf boost compared to the old one when there are, say, 50 of them?

Starkku:
> I also wonder about the performance, are there any perf boost compared to the old one when there are, say, 50 of them?

There isn't really easily measurable change in performance, FPS stays roughly the same with both on stress tests with dozens of simultaneous instances.

Belonit:
Perhaps it is worth making some kind of global flag, so that the moder would be able to enable/disable this fix?

Metadorius:
Unfortunately with this approach the global tag is impossible to do.

Belonit:
@Metadorius You can patch the memory after read the value from uimd.ini for example, but before the battle is loaded.

Starkku:
The code being patched here is extremely performance-critical though so adding any sort of additional runtime check there is probably a terrible idea though.

Metadorius:
I think the flag should be on by default. And in my opinion it would be better to control that from an artmd.ini tag. The code isn't run before reading the artmd anyways.

Metadorius:
Other than that I think this is a nice addition.

Belonit:
Replaced the command line key with the key from rulesmd.ini
`rulesmd.ini->[General]->FixTransparencyBlitters=yes`

PR #659, opened 2022-06-01
OpenTopped transport rangefinding / toggling firing if deactivated customizations

Starkku:
### Customizable OpenTopped properties

- `OpenTopped.IgnoreRangefinding` can be used to disable `OpenTopped` transport rangefinding behaviour where smallest weapon range between transport and all passengers is used when approaching targets that are out of range and when scanning for potential targets.
- `OpenTopped.AllowFiringIfDeactivated` can be used to customize whether or not passengers can fire out when the transport is deactivated (EMP, powered unit etc).

```ini
[SOMETECHNO]                              ; TechnoType
OpenTopped.IgnoreRangefinding=false       ; boolean
OpenTopped.AllowFiringIfDeactivated=true  ; boolean
```

Additionally, if the rangefinding is left enabled, `OpenTopped` transports now take `OpenTransportWeapon` setting of passengers into consideration when determining weapon range used for threat scanning and approaching targets instead of always using `Primary`.


FS-21 (commented):
The current code meets the code style guidelines.
I haven't noticed noticeable mistakes.

brsajo:
Works as intended, but perhaps needs a clearer description in the documentation (or maybe that's just me lol).

MaxLenarc:
Now that I understand what the logic does exactly, I can confirm that both toggles work exactly as intended.
OpenTopped.IgnoreRangefinding works as intended. Noticed the Battle Fortress always attacking from its weapon's range when the logic is on, instead of accounting for the shortest ranged infantry in the vehicle and getting very close when the logic is off.
OpenTopped.AllowFiringIfDeactivated=false works as intended. Tested by adding the Powered Unit tag to a battle fortress, then filling it with GGIs and parking it next to an enemy pillbox. Close enough for the GGIs to fire, then cut the power via selling a power plant and moving the fortress again to deactivate it. Once this is done the GGIs cease fire.
OpenTopped.AllowFiringIfDeactivated=true works as intended.

docs/New-or-Enhanced-Logics.md
```diff
@@ -477,20 +477,24 @@ Passengers.SyncOwner.RevertOnExit=true   ; boolean
...
 
 ```ini
-[SOMETECHNO]                       ; TechnoType
-OpenTopped.RangeBonus=1            ; integer
-OpenTopped.DamageMultiplier=1.3    ; float
-OpenTopped.WarpDistance=8          ; integer
+[SOMETECHNO]                              ; TechnoType
+OpenTopped.RangeBonus=1                   ; integer
+OpenTopped.DamageMultiplier=1.3           ; float
+OpenTopped.WarpDistance=8                 ; integer
```
Metadorius, line 495: Probably would be good to make those empty because they default to the value that you can modify globally.

Metadorius (approved):

PR #672, opened 2022-06-14
New trajectory called FullLine

FS-21:
- Is a straight-shot trajectory that doesn't end in the target.
- The weapon tag `ProjectileRange` must be set.
- Initial speed is determined by weapon's `Trajectory.Speed`.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -111,6 +111,7 @@ This page lists all the individual contributions to the project by their author.
   - AI Aircraft docks fix
   - Shared Ammo logic
   - Customizable FLH When Infantry Is Prone Or Deployed
+  - Full line projectile trajectory
 - **Starkku**:
   - Warhead shield penetration & breaking
   - Strafing aircraft weapon customization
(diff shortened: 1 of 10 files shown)


FS-21:
Visual example:
FullLine Trajectory

Metadorius:
What happens if the projectile goes out of physical map area?
Also `FullLine` doesn't sound intuitive? Maybe if the logic is similar to `Straight` then maybe make it a flag `Straight.PassThrough`?

Starkku:
I added a version that is part of `Trajectory=Straight` in my PR (#671) as `Trajectory.Straight.PassThrough` (name pending feedback). Sorry if I am stepping on your toes :pray:. My version takes into accord some things I fixed in that PR like warhead detonation/target snapping, PreImpactAnim etc. Only thing that's not really fixable is that the trajectory is still susceptible to the initial velocity vector, meaning that the firing angle is still determined by the original target so firing on ground right next to it will make the projectile hit the ground and not fly off to infinity.

FS-21:
Closing PR because #671 is more complete :-D

PR #683, opened 2022-06-30
Artillery projectile trajectory

FS-21:
- High altitude projctile capable to hit targets in high places.
- Height determined by `Trajectory.Artillery.MaxHeight`.
- Supported `Inaccurate=yes` and Ares costumization tags `BallisticScatter.Max` and `BallisticScatter.Min`.

In `rulesmd.ini`:
```ini
[SOMEPROJECTILE]                     ; Projectile
Trajectory=Artillery                 ; Trajectory type
Trajectory.Artillery.MaxHeight=2000  ; integer
```

Note: Re-used some code for Inaccurate projectiles from "arcing-fix" Phobos PR.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -111,6 +111,7 @@ This page lists all the individual contributions to the project by their author.
   - AI Aircraft docks fix
   - Shared Ammo logic
   - Customizable FLH When Infantry Is Prone Or Deployed
+  - Artillery projectile trajectory
 - **Starkku**:
   - Warhead shield penetration & breaking
   - Strafing aircraft weapon customization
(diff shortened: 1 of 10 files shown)


NetsuNegi:
VYZ}E_1O8ID2 @JLBHDB}E8

The impact point of bullet maybe have some problem, I don't know if it is caused by the vanilla game.

NetsuNegi:
The trail when bullet has been created is same as Trajectory=Straight, I think you should also adjust the Z velocity at the time of unlimbo.
The faster Trajectory.Speed means the more serious situation.

MortonPL:
What's the difference between this and `Lobber=yes` on a weapon?

Metadorius:
Eh more so, I would say maybe there's more point in continuing the arcing fix?

FS-21:
> What's the difference between this and `Lobber=yes` on a weapon?

Can you set a projectile height with lobber=yes? No... 
I still need to fix some cases  but I'm busy with another PR.

FS-21:
I want to remake it from scratch with a different approach so I close it and reopen it when I have free time for this idea.

PR #684, opened 2022-06-30
Allow detonating weapon warhead on all map objects

Starkku:
This is basically something to replace use of CellSpread=256 in a (potentially vain, CS=256 which is the current (artificial) cap is not enough to cover every cell on a large enough map) attempt to cover entire map for some effect, damage or whatever. It also has considerably better performance atleast if the warhead detonations do not generate copious amounts of animations and/or alpha lights, and possibly even then.

This is restricted to only working on warheads that are detonated using a weapon. I did at first attempt to also extend it to other use cases (which is why some bits of the code used here look like they could be reused elsewhere) but it is basically in vain. At the best you can make the damage apply to everything, anything that's handled on Ares side like AE or EMP are basically off limits in this case. So atleast for now this doesn't work on GenericWarhead SW or crit warheads and the like. The reason this is on a warhead instead of weapon is the old classic aka `AirburstWeapon` bullets not knowing about which WeaponType fired it off.

This feature effectively treats each chosen target as the actual target of the weapon/projectile. The consequence of this is that some special logics like mind control *will* work as expected on every target, allowing for what is essentially map-wide mind control. Other special logics may or may not work correctly, game makes an attempt to use them but some of them result in either crash (Temporal if self-targeting is allowed) or do next to nothing (Temporal otherwise, Airstrike). Modder discretion and common sense is advised.

 -----------

### Detonate Warhead on all objects on map

- Setting `DetonateOnAllMapObjects` to true allows a Warhead invoked by a weapon (and consequently any `Airburst/ShrapnelWeapon` that may follow) to detonate on each object currently alive and existing on the map regardless of its actual target, with optional filters. Note that this is done immediately prior Warhead detonation so after `PreImpactAnim` *(Ares feature)* has been displayed.
  - `DetonateOnAllMapObjects.AffectTargets` can be used to filter which types of targets (TechnoTypes) are considered valid. Only `all`, `aircraft`, `buildings`, `infantry` and `units` are valid values.
  - `DetonateOnAllMapObjects.AffectHouses` can be used to filter which houses targets can belong to be considered valid. Only applicable if the house that fired the projectile is known.
  - `DetonateOnAllMapObjects.AffectTypes` can be used to list specific TechnoTypes to be considered as valid targets. If any valid TechnoTypes are listed, then only matching objects will be targeted. Note that `DetonateOnAllMapObjects.AffectTargets` and `DetonateOnAllMapObjects.AffectHouses` take priority over this setting.
  - `DetonateOnAllMapObjects.IgnoreTypes` can be used to list specific TechnoTypes to be never considered as valid targets.
  - `DetonateOnAllMapObjects.RequireVerses`, if set to true, only considers targets whose armor type the warhead has non-zero `Verses` value against as valid. This is checked after all other filters listed above.
 
 In `rulesmd.ini`:
```ini
[SOMEWARHEAD]                                ; Warhead
DetonateOnAllMapObjects=false                ; boolean
DetonateOnAllMapObjects.AffectTargets=all    ; list of Affected Target Enumeration (aircraft|buildings|infantry|units|all)
DetonateOnAllMapObjects.AffectHouses=all     ; list of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
DetonateOnAllMapObjects.AffectTypes=         ; list of TechnoType names
DetonateOnAllMapObjects.IgnoreTypes=         ; list of TechnoType names
DetonateOnAllMapObjects.RequireVerses=false  ; boolean
```

**Warning:** While this feature can provide better performance than a large `CellSpread` value, it still has potential to slow down the game, especially if used in conjunction with things like animations, alpha lights etc. Modder discretion and use of the filter keys is advised.


src/Ext/Bullet/Hooks.cpp
```diff
@@ -222,5 +226,65 @@ DEFINE_HOOK(0x469A75, BulletClass_Logics_DamageHouse, 0x7)
...
+
+			if ((pWHExt->DetonateOnAllMapObjects_AffectTargets & AffectedTarget::Aircraft) != AffectedTarget::None)
+			{
+				for (int i = 0; i < AircraftClass::Array->Count; i++)
+					tryDetonate(AircraftClass::Array->GetItem(i));
+			}
+
+			if ((pWHExt->DetonateOnAllMapObjects_AffectTargets & AffectedTarget::Building) != AffectedTarget::None)
+			{
+				for (int i = 0; i < BuildingClass::Array->Count; i++)
```
chaserli, line 267: why not 
```cpp
	for(BuildingClass* pBld: *BuildingClass::Array)
		tryDetonate(pBld);
```

Metadorius:
How does it interact with ArmorTypes? As in if the versus is 0%, will it still apply?

Metadorius (commented):
I am not actually sure if this should be done on warhead level? Is it possible / was it considered to do this on weapon level?

Starkku:
> How does it interact with ArmorTypes? As in if the versus is 0%, will it still apply?

Does not check Verses when determining if to detonate on a specific object. Afterwards Verses are applied like normal on Warhead detonation. I could add an optional Verses check to the detonation as well.

> I am not actually sure if this should be done on warhead level? Is it possible / was it considered to do this on weapon level?

I did consider it (which was actually mentioned in the description too, along with the reason I did not do it). To add a further reason, I was actually mistaken about it not working with Crit warheads, it does since the Warhead detonation there uses a dummy bullet to properly detonate it instead of just calling `MapClass::DamageArea` or `ObjectClass::ReceiveDamage`. Might need to adjust the documentation slightly in light of this revelation

Metadorius:
Having a verse option would be good, yes.

Oh I missed that. Is there any usecase that benefits from this being a warhead tag though? Also is it possible to make airburst, shrapnel to have a weapon (specified in the same place) to allow customizing this and other params? I recall somebody on straight trajectory already asked about speed on splits etc.

Starkku:
> Oh I missed that. Is there any usecase that benefits from this being a warhead tag though? Also is it possible to make airburst, shrapnel to have a weapon (specified in the same place) to allow customizing this and other params? I recall somebody on straight trajectory already asked about speed on splits etc.

Maybe, not any that I personally thought up though. Reimplementing the entire Airburst code from Ares would be the only way, that or having it fixed on Ares' side. ShrapnelWeapon is already unaffected by the issue.

Starkku:
Added `DetonateOnAllMapObjects.RequireVerses`.

PR #677, opened 2022-06-21
RadSite with Ownership break certain Trigger Events Fix

Otamaa:
Close  #676

PR #686, opened 2022-07-01
Radiation improvements

### LINKED ISSUE #676: Unit killed by radiation with a tag of event 48 (Destroyed by anything) can‘t take effect
### First check

### Description

First, make a unit attaches a tag with event 48 (Destroyed by anything)  in map editor. Then train a desolater deployed to kill this unit in the game. Unit killed by radiation, but nothing happens. It should output “6” as the trigger action. The radiation of terrorist in FV or demolitions truck is the same. I tried it in YR1001. it won't happen.

### Phobos Version

build28

### Conditions to reproduce

_No response_

### INI code

_No response_

### Steps to reproduce

1.make a unit attaches a tag with event 48 (Destroyed by anything)  in map editor. 
2.train a desolater deployed to kill this unit in the game.
3.
...

### Expected behaviour

It should output “6” as the trigger action

### Actual behaviour

Nothing happens.

### Additional context

image

image

Starkku:
- `Rad.NoOwner` on weapons is deprecated. This has been replaced by `RadHasOwner` key on radiation types itself. It also defaults to no, so radiation once again has no owner house by default.
- Added `RadHasInvoker` which can be used to have radiation fields recognize the TechnoType that created them to allow them to gain experience from objects killed by radiation etc.
- Added `RadSiteWarhead.Detonate` can be used to make `RadSiteWarhead` detonate instead of being simply used to deal damage. This enables most warhead effects, animations etc.
- Fixed an issue that caused vehicles killed by damage dealt by a known house but without a known source TechnoType (f.ex animation warhead damage) to not be recorded as killed correctly and thus not spring map trigger events etc. This was the root cause of #676.

-----------------

### Custom Radiation Types

- There are several new properties available to all radiation types.
  - `RadSiteWarhead.Detonate` can be set to make `RadSiteWarhead` detonate on affected objects rather than only be used to dealt direct damage. This enables most Warhead effects, display of animations etc.
  - `RadHasOwner`, if set to true, makes damage dealt by the radiation count as having been dealt by the house that fired the projectile that created the radiation field. This means that Warhead controls such as `AffectsAllies` will be respected and any units killed will count towards that player's destroyed units count.
  - `RadHasInvoker`, if set to true, makes the damage dealt by the radiation count as having been dealt by the TechnoType (the 'invoker') that fired the projectile that created the radiation field. In addition to the effects of `RadHasOwner`, this will also grant experience from units killed by the radiation to the invoker.

In `rulesmd.ini`:
```ini
[SOMERADTYPE]                   ; RadType
RadSiteWarhead.Detonate=false   ; boolean
RadHasOwner=false               ; boolean
RadHasInvoker=false             ; boolean
```

-----------------

Closes #676
Closes #677


src/Ext/RadSite/Hooks.cpp
```diff
@@ -194,12 +207,24 @@ DEFINE_HOOK(0x4DA59F, FootClass_AI_Radiation, 0x5)
...
+				}
+				else
+				{
+					auto coords = CoordStruct::Empty;
+					coords = *pFoot->GetCoords(&coords);
+					WarheadTypeExt::DetonateAt(pWarhead, coords, pRadExt->RadInvoker, damage);
+
+					if (!pFoot->IsAlive)
+						break;
+				}
```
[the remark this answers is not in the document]
Starkku: Basically. Could probably be separated into a function.

MaxLenarc:
I've completed my testing of the Radiation Improvements logic, I think. RadApplicationDelay.Building, RadHasInvoker, and RadHasOwner all works as intended. RadSiteWarhead.Detonate, the Animation display part of it I got to work; not sure about warhead effects though.

src/Misc/Hooks.BugFixes.cpp
```diff
@@ -509,3 +509,8 @@ DEFINE_HOOK(0x73EFD8, UnitClass_Mission_Hunt_DeploysInto, 0x6)
 
 	return 0;
 }
+
+// Fixes an issue in TechnoClass::Record_The_Kill that prevents vehicle kills from being recorded
+// correctly if killed by damage that has owner house but no owner techno (animation warhead damage, radiation with owner etc.
+// Author: Starkku
+DEFINE_JUMP(LJMP, 0x7032BA, 0x7032C6);
```
chaserli, line 516: Just curious, I observed a similar issue in `TechnoClass::RegisterDestruction` as I've mentioned in #677 . Should `DEFINE_JUMP(LJMP,0x702DCA, 0x702DD6);` also be applied in this case? What makes unit type so special in both cases?

Starkku: > Just curious, I observed a similar issue in `TechnoClass::RegisterDestruction` as I've mentioned in #677 . Should `DEFINE_JUMP(LJMP,0x702DCA, 0x702DD6);` also be applied in this case? What makes unit type so special in both cases?

There are two variants of this function. They differ by the arg taken, one is HouseClass* (`0x703230`) and other is TechnoClass* (`0x702D40`). 

As far as I can ascertain, only the HouseClass variant actually suffers from this problem because the TechnoClass variant has override in UnitClass (`0x744720`) that springs the events. Unless you can confirm that the TechnoClass variant indeed actually suffers from issues related to this then it does not need to be patched.

I am unsure why UnitClass needs these checked separately, may be vestigial. I suspect that the HouseClass variant having the check against UnitClass technos is an oversight.

PR #693, opened 2022-07-05
New action 135 Unload from Transports

handama:
### `135` Unload from Transports

- A new unload action similar to x=8,n, but can avoid treating `VehicleTypes` with `OpenTopped=yes` and `Gunner=yes` as transports.
  - All `InitialPayload` will be regarded as passengers and added to team.
  - If `TransportsReturnOnUnload` is set to `yes` in this TeamType, and transports are lost, they will return after unloading.

In `aimd.ini`:
```ini
x=135,n
```

- The possible argument values are:

| *Argument* | *Description*                       |
| :------: | :-------------------------------------------: |
0         | Keep transports and units. |
1         | Keep transports and lose units. |
2         | Lose transports and keep units. |
3         | Lose transports and units. |
4         | Only `VehicleTypes` with largest `SizeLimit` will be regarded as transports. Keep transports and units. |
5         | Only `VehicleTypes` with largest `SizeLimit` will be regarded as transports. Keep transports and lose units. |
6         | Only `VehicleTypes` with largest `SizeLimit` will be regarded as transports. Lose transports and keep units. |
7         | Only `VehicleTypes` with largest `SizeLimit` will be regarded as transports. Lose transports and units. |

here is an example:
```
0=72,0
1=43,0
2=5,5
3=54,0
4=135,6
5=53,0
6=135,0
7=0,1
```
new unload script

### DIFF
--- a/docs/AI-Scripting-and-Mapping.md
+++ b/docs/AI-Scripting-and-Mapping.md
@@ -337,6 +337,7 @@ x=135,n
 6         | Only `VehicleTypes` with largest `SizeLimit` will be regarded as transports. Lose transports and keep units. |
 7         | Only `VehicleTypes` with largest `SizeLimit` will be regarded as transports. Lose transports and units. |
 
+
 ### `500 - 523` Edit Variable
 - Operate a variable's value
     - The variable's value type is int16 instead of int32 in trigger actions for some reason, which means it ranges from -2^15 to 2^15-1.
--- a/docs/AI-Scripting-and-Mapping.md
+++ b/docs/AI-Scripting-and-Mapping.md
@@ -314,6 +314,29 @@ In `aimd.ini`:
 x=126,n           ; integer n=0, in frames
 ```
 
+### `135` Unload from Transports
+
+- A new unload action similar to x=8,n, but can avoid treating `VehicleTypes` with `OpenTopped=yes` and `Gunner=yes` as transports.
+
+In `aimd.ini`:
+```ini
+[SOMESCRIPTTYPE]  ; ScriptType
+x=135,n
+```
+
+- The possible argument values are:
+
+| *Argument* | *Description*                       |
+| :------: | :-------------------------------------------: |
+0         | Keep transports and units. |
+1         | Keep transports and lose units. |
+2         | Lose transports and keep units. |
+3         | Lose transports and units. |
+4         | Only `VehicleTypes` with largest `SizeLimit` will be regarded as transports. Keep transports and units. |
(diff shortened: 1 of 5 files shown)


FS-21:
Why did you closed it? :-/

handama:
> Why did you closed it? :-/

I've found some small issues, and I'd like to commit it with some other new scripts together ;)

FS-21:
If they aren't related create multiple PRs so they get validated & merged faster than a PR with dozens of new actions... I have experience -_-'

Also, your version of this action can manage the allied Landing Craft/Soviet/Yuri hover transport issue when the unload command is executed in water?

handama:
> If they aren't related create multiple PRs so they get validated & merged faster than a PR with dozens of new actions... I have experience -_-'
> 
> Also, your version of this action can manage the allied Landing Craft/Soviet/Yuri hover transport issue when the unload command is executed in water? Like seeking for the nearest valid cell for unloading or similar approach.

No it cannot. I have no experience of dealing with pathfinding :(

Metadorius:
Actually it depends, it's better to have medium-size PRs I would say, because dealing with a lot of separate PRs is a headache on it's own. Both one gigantic PR and a lot of bite-sized PRs are eh for reviewing because you need to solve merge conflicts each time.

FS-21:
Also, can you try to add this case if possible?

A Transport shouldn't execute the unload command until it reach the destination (or CloseEnough distance if it doesn't work well with groups of transports).

Why? Because the action 8,x executes that action as soon as the game can do it so if your group of transports doesn't travel together each transport will unload in very different places. And adding another action before the 8,x for waiting the group isn't good if the first transports reached enemy territory.
So the suggestion is as soon as each one reach the destiny they should execute the unload, etc...

handama:
> Also, can you try to add this case if possible?
> 
> A Transport shouldn't execute the unload command until it reach the destination (or CloseEnough distance if it doesn't work well with groups of transports).
> 
> Why? Because the action 8,x executes that action as soon as the game can do it so if your group of transports doesn't travel together each transport will unload in very different places. And adding another action before the 8,x for waiting the group isn't good if the first transports reached enemy territory. So the suggestion is as soon as each one reach the destiny they should execute the unload, etc...


It sounds like a combination of Mission_Move and unload, I'll try to add this feature

FS-21:
More than a combo of 2 actions is like an individual unload command per-transport instead the "all at the same time" from vanilla.

This suggestion is as additional characteristic for the action (maybe the modder just wants an improved action 8,x and nothing more). If you can't I'll add it as a future action in my TO-DO list :-P

PR #694, opened 2022-07-06
`SW.AuxTechno`

ststl-s:
Like Ares entry `SW.AuxBuildings` and `SW.NegBuildings`, for all technotype.

In `rulesmd.ini`:
```ini

SW.AuxTechno=      ; TechnoTypes, like SW.AuxBuildings
SW.AuxTechno.Any=  ; boolean, one of or all of, default yes(one of)
SW.NegTechno=      ; TechnoTypes, like SW.NegBuildings
SW.NegTechno.Any=  ; boolean, one of or all of, default yes(one of)

```

Need test:

If don't abandon Ares hook, SW while exist but don't charge, so needs to test Ares SW available entries could work as use Ares hook.

### DIFF
--- a/Phobos.vcxproj
+++ b/Phobos.vcxproj
@@ -30,6 +30,7 @@
     <ClCompile Include="src\Ext\Anim\Body.cpp" />
     <ClCompile Include="src\Ext\SWType\FireSuperWeapon.cpp" />
     <ClCompile Include="src\Ext\SWType\Hooks.cpp" />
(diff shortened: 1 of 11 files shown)


gangster-ppm:
I would add **SW.AuxTechno.Owner** and **SW.NegTechno.Owner** to the list for cases like: "if enemy have scrambler your spy plane will be banned", "if ally have designator a missle strike will be granted", etc

ststl-s:
As test, Building's `SuperAnim` can't process like Ares3.0,  so mark as draft.

PR #701, opened 2022-07-14
Attempt to fix the residual display bug of flying strings

chaserli:


Metadorius:
Confirmed to be working by Dmitry Volkov

PR #702, opened 2022-07-16
Migration utility

Metadorius:
Non-DLL:
- Implemented a tool (sed wrapper) to semi-automatically upgrade INIs to use latest Phobos tags (by Kerbiter).

To use the utility, right-click on `utilities/MigrationUtility.ps1`, select `Run with PowerShell` and select the needed files.


Metadorius:
Porting to a separate repo.

PR #706, opened 2022-07-19
Trigger Action can use more than one string as parameter

ststl-s:
Why use WW action with only one text parameter.

Action 9940 as sample.

In `fadata.ini`
```ini
76=duration,0
77=InnerColor,8
78=OuterColor,9

9940=Draw Laser Between Two Waypoints,0,76,30,30,77,78,0,0,0,Draw laser between two waypoints color is hexadecimal of RGB24,0,1,9940,1
```

### DIFF
--- a/src/Ext/TAction/Body.cpp
+++ b/src/Ext/TAction/Body.cpp
@@ -21,7 +21,14 @@ TActionExt::ExtContainer TActionExt::ExtMap;
 template <typename T>
 void TActionExt::ExtData::Serialize(T& Stm)
 {
-	//Stm;
+	Stm
+		.Process(this->Value1)
+		.Process(this->Value2)
+		.Process(this->Parm3)
+		.Process(this->Parm4)
+		.Process(this->Parm5)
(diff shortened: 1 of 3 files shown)


src/Ext/TAction/Body.cpp
```diff
@@ -67,6 +74,8 @@ bool TActionExt::Execute(TActionClass* pThis, HouseClass* pHouse, ObjectClass* p
 		return TActionExt::RunSuperWeaponAtLocation(pThis, pHouse, pObject, pTrigger, location);
 	case PhobosTriggerAction::RunSuperWeaponAtWaypoint:
 		return TActionExt::RunSuperWeaponAtWaypoint(pThis, pHouse, pObject, pTrigger, location);
+	case PhobosTriggerAction::DrawLaserBetweenWeaypoints:
```
TaranDahl, line 77: It doesn't seem to be related to this PR.

TaranDahl (changes requested):

PR #707, opened 2022-07-21
Mobile Refinery Logic

NetsuNegi:
Techno can convert the ore under its feet into cash in real time, like GDI's EPIC unit MARV in Command & Conquer 3 Kane's Wrath. It also work on airforce unit.
This logic cannot work on buildings.

in `rulesmd.ini`
```
[SOMETECHNO]                          ; InfantryType, UnitType or AircraftType
MobileRefinery=no                     ; boolean
MobileRefinery.TransRate=30           ; unsigned int, how many frames are there between two conversions
MobileRefinery.CashMultiplier=1.0     ; float, it's multiplier of the final cash.
MobileRefinery.AmountPerCell=0        ; unsigned int, how many ore can be convert at one time in each cell
MobileRefinery.FrontOffset=           ; list of double, use cell as unit, multiple values mean that they are effective in multiple cells at the same time
MobileRefinery.LeftOffset=            ; list of double, use cell as unit, multiple values mean that they are effective in multiple cells at the same time
MobileRefinery.Display=yes            ; boolean, should it display translated cash ?
MobileRefinery.DisplayColor=57,197,187; RGB color
MobileRefinery.Anims=                 ; list of animations, if set 8 animations, it will select by techno's body facing, other amount will random select
MobileRefinery.AnimMove=yes           ; should the anim move as techno ?
```

8DVN 2X(GZX@49ZD}PJNN`3


mevitar:
TransRate= and MaxAmount= need more explanation on what exactly they do, i had to figure it out by trial and error (they didn't do what i expected). Same for Offsets - are they in cells or leptons? (answer: leptons, but i had to figure it out myself)
And i don't think there's any point in making it in leptons, since you can't clear part of a cell anyway.

Otherwise appears to work fine, although i think outright clearing the cell is too much. It would be better if it instead reduced the tiberium on the cell by a specified amount.

NetsuNegi:
> TransRate= and MaxAmount= need more explanation on what exactly they do, i had to figure it out by trial and error (they didn't do what i expected). Same for Offsets - are they in cells or leptons? (answer: leptons, but i had to figure it out myself) And i don't think there's any point in making it in leptons, since you can't clear part of a cell anyway.
> 
> Otherwise appears to work fine, although i think outright clearing the cell is too much. It would be better if it instead reduced the tiberium on the cell by a specified amount. Or perhaps a rate at which it reduced each cell of its tiberium (defaulting to instantly?).

I'll add more info about that, and the offset will use cell, no longer leptons. And are you want a tag to limit the total convert ore value in one time when set multi cells?

NetsuNegi:
> TransRate= and MaxAmount= need more explanation on what exactly they do, i had to figure it out by trial and error (they didn't do what i expected). Same for Offsets - are they in cells or leptons? (answer: leptons, but i had to figure it out myself) And i don't think there's any point in making it in leptons, since you can't clear part of a cell anyway.
> 
> Otherwise appears to work fine, although i think outright clearing the cell is too much. It would be better if it instead reduced the tiberium on the cell by a specified amount. Or perhaps a rate at which it reduced each cell of its tiberium (defaulting to instantly?).

Oh no, the ReduseTiberium Function always clear all ore in a cell, I cannot limit it...

NetsuNegi:
It's bad...

NetsuNegi:
> TransRate= and MaxAmount= need more explanation on what exactly they do, i had to figure it out by trial and error (they didn't do what i expected). Same for Offsets - are they in cells or leptons? (answer: leptons, but i had to figure it out myself) And i don't think there's any point in making it in leptons, since you can't clear part of a cell anyway.
> 
> Otherwise appears to work fine, although i think outright clearing the cell is too much. It would be better if it instead reduced the tiberium on the cell by a specified amount. Or perhaps a rate at which it reduced each cell of its tiberium (defaulting to instantly?).

I misunderstood the function. I will fix MaxAmount soon.

mevitar:
Tested this newer build and everything appears to work fine. However, since you changed how MobileRefinery.MaxAmount= works, the name might not be appropriate anymore. :P
Also, again, need better explanation what the tag reduces (cost value? tiberium stages? i assume it's tiberium stages as setting it to 1 reduced the patch 1 stage per each interval, up to 12 times on the richest cells).

I have an issue with how MobileRefinery.TransRate= works currently, though, as the timer appears to be counting all the time. That means, if the intervals are long enough. the unit can walk over a tiberium patch between the timer's intervals and nothing will be collected. IMO the timer should apply only while still occupying a cell (like standing still or moving very slowly), and once the unit moves to another cell it should start all over again.

But perhaps others think that changing it is unnecessary?

mevitar:
Why has this been closed? Was there not enough feedback on this?

NetsuNegi:
> Why has this been closed? Was there not enough feedback on this?

The code has not been updated for a long time. Now it is more troublesome to handle. Later, I will restart or open another pull request.

mevitar:
> Later, I will restart or open another pull request.
Good to hear that, i was afraid it was being abandoned.

chaserli:
1. Too many new entries, in this case do it like #787 
2. XXRate is in minutes, whereas XXDelay or XXDuration can be in frames, choose carefully
3. If you're doing it in TechnoClass::AI, make it obey the same format as others. In fact I think from now on `OnUpdate` like actions should be considered carefully for performance consideration
4. Please Implement AI logic, i.e. make it behave like harvesters for auto harvesting
5. The XXRate thing is perhaps interpreted improperly. IMO it should be something like "I gathered this amount of ore and it will take this amount of frames to monetize"
6. You should have just updated your branch and continued on that

NetsuNegi:
> > Later, I will restart or open another pull request.
> 
> Good to hear that, i was afraid it was being abandoned.

I had recreate a new PR#1111 for this logic

DeathFishAtEase:
Transfer to #1619

Issue #642, opened 2022-05-20
Allow PlacementPreview.Shape to support shps for different theaters


Cpj636:
### First check

- [X] The suggestion is **NOT** related to functionality of Ares, HAres, Kratos or any other YR engine extension.
- [X] The suggested feature / enhancement is **NOT** implemented in Ares/Phobos yet and the existing possibilites are not viable or good enough to achieve the wanted result.
- [X] I agree to elaborate the details if requested and provide thorough testing if the feature is implemented.
- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search and read the issue list to find a similar issue and didn't find it.
- [X] I thought out how the feature / enhancement should work and described it in the most detailed and clear way.

### Description

For different theaters, PlacementPreview will use different shp files by default like NAREFENMK's last frame for snow theater and NGREFENMK's last frame if other NXREFENMK missing. However, when it comes to customizations through PlacementPreview.Shape,  it cannot switch files according to prefix G/A/D etc. 
Hope there could be a way for PlacementPreview.Shape to load different shp files on different theaters. For example, allow PlacementPreview.Shape to load different files according to the second filename character. Or allow PlacementPreview.Shape to load different files according to the last filename character like seal.shp, seala.shp and seald.shp.

### INI code

```ini
No additional ini codes needed.
```

Belonit:
@Cpj636 Please check this PR #710

PR #710, opened 2022-07-23
Refactoring PlacementPreview logic

Belonit:
- General code refactoring
- Added ability to load theater-specific shapes - #642 
- Translucent tags now should be set to 0\25\50\70 instead of 0\1\2\3
- Renamed tags and fixed documentation errors:
  -  In AudioVisual
     - `BuildingPlacementPreview.DefaultTranslucentLeve` ---> `PlacementPreview.Translucency`
     - `BuildingPlacementGrid.TranslucentLevel` -------------> `PlacementGrid.Translucency`
  -  In BuildingType
     - `PlacementPreview.Show` ---------------------------------> `PlacementPreview`
     - `PlacementPreview.TranslucentLevel` -------------------> `PlacementPreview.Translucency`
  -  In ra2md.ini
     - `[Phobos]->ShowBuildingPlacementPreview` -------------> `[Phobos]->ShowPlacementPreview`

[Old](https://phobos.readthedocs.io/en/latest/User-Interface.html#placement-preview) Documentation vs [New](https://phobos--710.org.readthedocs.build/en/710/User-Interface.html#placement-preview)


Cpj636:
There is my testing feedback:
All three tags in [AudioVisual] work fine.
Theater-specific shapes can be specified by PlacementPreview.Shape.
Theater-specific palettes can be specified by PlacementPreview.Palette.
All the rest tags work as expected.

Belonit:
The `PlacementPreview.Shape` tag does not currently support save\load. Need to fix

docs/User-Interface.md
```diff
@@ -65,33 +65,40 @@ PrioritySelectionFiltering=true  ; boolean
...
-PlacementPreview.Shape=                             ; filename - including the .shp extension. If not set uses building's artmd.ini Buildup SHP (based on Building's Image)
-PlacementPreview.ShapeFrame=                        ; integer, zero-based frame index used for displaying the preview
-PlacementPreview.Offset=0,-15,1                     ; integer, expressed in X,Y,Z used to alter position preview
-PlacementPreview.Remap=true                         ; boolean, does this preview use player remap colors
-PlacementPreview.Palette=                           ; filename - including the .pal extension. This option is not used if PlacementPreview.Remap is set to true
-PlacementPreview.TranslucentLevel=                  ; integer, defaults to [AudioVisual]->BuildingPlacementPreview.DefaultTranslucentLevel
+[AudioVisual]
+PlacementPreview.Building=no               ; boolean
+PlacementPreview.Building.Translucent=75   ; integer, 0 25 50 75
+PlacementPreview.Grid.Translucent=0        ; integer, 0 25 50 75
```
Metadorius, line 79: I would call that `PlacementGrid.Translucency=0`.
```ini
PlacementGrid=yes
PlacementGrid.Translucency=0
PlacementGrid.WithPreview=yes
PlacementGrid.Translucency.WithPreview=0`

File: docs/User-Interface.md [hunk 2]
```diff
@@ -65,33 +65,40 @@ PrioritySelectionFiltering=true  ; boolean
...
+
+[SOMEBUILDING]
+PlacementPreview=yes                       ; boolean
+PlacementPreview.Shape=                    ; filename - including the .shp extension. If not set uses building's artmd.ini Buildup SHP (based on Building's Image)
+PlacementPreview.ShapeFrame=               ; integer, zero-based frame index used for displaying the preview
+PlacementPreview.Offset=0,-15,1            ; integer, expressed in X,Y,Z used to alter position preview
+PlacementPreview.Remap=yes                 ; boolean, does this preview use player remap colors
+PlacementPreview.Palette=                  ; filename - including the .pal extension.
+PlacementPreview.Translucent=              ; integer, defaults to [AudioVisual]->PlacementBuilding.Translucent
```
Metadorius, line 88: Not sure, is there really a need for this tag override? I don't see usecases for this, it only introduces inconsistencies.

docs/User-Interface.md
```diff
@@ -65,33 +65,40 @@ PrioritySelectionFiltering=true  ; boolean
...
+```
+```{note}
+The `PlacementPreview.Palett` option is not used if `PlacementPreview.Remap=yes`
+This may change in the future
 ```
```
Metadorius, line 93:
````suggestion
```{note}
The `PlacementPreview.Palette` option is not used if `PlacementPreview.Remap=yes`. This may change in future.
```
````

File: src/Ext/BuildingType/Body.cpp [hunk 4]
```diff
@@ -145,25 +145,39 @@ void BuildingTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
...
+					if (auto pImage = FileSystem::LoadSHPFile(pValue))
+						this->PlacementPreview_Shape = pImage;
+					else
+						Debug::Log("Failed to find file %s referenced by [%s]%s=%s\n", pValue, pSection, pKey, pValue);
+				}
+				else
+				{
+					Debug::Log("Cannot Load PlacementPreview.Shape for [%s]Art[%s] ! \n", pSection, pArtSection);
+				}
+			}
```
Metadorius, line 172: This looks like it should not be here but rather in `TemplateDef.h` or wherever else.

src/Phobos.cpp
```diff
@@ -237,6 +237,13 @@ DEFINE_HOOK(0x5FACDF, OptionsClass_LoadSettings_LoadPhobosSettings, 0x5)
 	if (pINI_RULESMD->ReadBool("General", "FixTransparencyBlitters", true))
 		BlittersFix::Apply();
 
+	// PlacementPreview
+	{
+		Phobos::Config::PlacementPreview_Enabled = pINI_RULESMD->ReadBool("AudioVisual", "PlacementPreview.Building", false);
```
Metadorius, line 242: We have RulesExt for this. Does not belong here.

src/Phobos.cpp
```diff
@@ -237,6 +237,13 @@ DEFINE_HOOK(0x5FACDF, OptionsClass_LoadSettings_LoadPhobosSettings, 0x5)
...
+		Phobos::Config::PlacementPreview_UserHasEnabled =
+			CCINIClass::INI_RA2MD->ReadBool("Phobos", "PlacementPreview.Building", Phobos::Config::PlacementPreview_Enabled);
```
Metadorius, line 244: `ShowPlacementPreview` should work fine as a variable name too.

src/Ext/BuildingType/Hooks.cpp
```diff
@@ -70,74 +70,87 @@ DEFINE_HOOK(0x458623, BuildingClass_KillOccupiers_Replace_MuzzleFix, 0x7)
...
+			blitFlags |= EnumFunctions::GetTranslucentLevel(
+				pTypeExt->PlacementPreview_Translucent.Get(
+					RulesExt::Global()->PlacementPreview_Building_Translucent
+				)
+			);
+		}
+
+		ConvertClass* pPalette = nullptr;
+		{
+			if(pTypeExt->PlacementPreview_Remap.Get())
```
Metadorius, line 133:
```suggestion
			if (pTypeExt->PlacementPreview_Remap.Get())
```

src/Ext/BuildingType/Hooks.cpp
```diff
@@ -70,74 +70,87 @@ DEFINE_HOOK(0x458623, BuildingClass_KillOccupiers_Replace_MuzzleFix, 0x7)
...
+			nPoint.Y += offset.Y;
 		}
+
+		BlitterFlags blitFlags = BlitterFlags::Centered | BlitterFlags::Nonzero | BlitterFlags::MultiPass;
+		{
+			blitFlags |= EnumFunctions::GetTranslucentLevel(
+				pTypeExt->PlacementPreview_Translucent.Get(
+					RulesExt::Global()->PlacementPreview_Building_Translucent
+				)
+			);
```
Metadorius, line 128: 1. Why not implement `TranslucencyLevel` as a type that could be converted to blitter flags instead? 
2. Default-to-global behavior could be done with a method.

src/Utilities/EnumFunctions.cpp
```diff
@@ -93,17 +93,16 @@ bool EnumFunctions::AreCellAndObjectsEligible(CellClass* const pCell, AffectedTa
...
-	case 2:
+	case 50:
 		return BlitterFlags::TransLucent50;
-	case 3:
+	case 75:
 		return BlitterFlags::TransLucent75;
 	}
-
-	return BlitterFlags::None;
 }
```
Metadorius, line 108: Should be moved to the `Read` specialization for the suggested `TranslucencyLevel` type.

src/Utilities/GeneralUtils.cpp
```diff
@@ -118,3 +119,18 @@ bool GeneralUtils::HasHealthRatioThresholdChanged(double oldRatio, double newRat
...
+
+bool GeneralUtils::ApplyTheaterSuffixToString(char* str)
+{
+	if (auto pSuffix = strstr(str, "~~~"))
+	{
+		auto theater = ScenarioClass::Instance->Theater;
+		auto pExtension = Theater::GetTheater(theater).Extension;
+		pSuffix[0] = pExtension[0];
+		pSuffix[1] = pExtension[1];
+		pSuffix[2] = pExtension[2];
```
Metadorius, line 131:
```suggestion

		pSuffix[0] = pExtension[0];
		pSuffix[1] = pExtension[1];
		pSuffix[2] = pExtension[2];

```
Also can't this be done via a function call? Also IMO should emit a warning if the amount of `~` symbols doesn't equate to 3, which is a developer error.

Metadorius (changes requested):
In general I am not a fan of period overusage in tags. Also not sure if the code blocks you made are really needed, could just put a comment explaining what is being done next instead of adding indentation.

docs/User-Interface.md
```diff
@@ -65,33 +65,40 @@ PrioritySelectionFiltering=true  ; boolean
...
-PlacementPreview.Show=                              ; boolean, defaults to [Phobos]->ShowBuildingPlacementPreview
-PlacementPreview.Shape=                             ; filename - including the .shp extension. If not set uses building's artmd.ini Buildup SHP (based on Building's Image)
-PlacementPreview.ShapeFrame=                        ; integer, zero-based frame index used for displaying the preview
-PlacementPreview.Offset=0,-15,1                     ; integer, expressed in X,Y,Z used to alter position preview
-PlacementPreview.Remap=true                         ; boolean, does this preview use player remap colors
-PlacementPreview.Palette=                           ; filename - including the .pal extension. This option is not used if PlacementPreview.Remap is set to true
-PlacementPreview.TranslucentLevel=                  ; integer, defaults to [AudioVisual]->BuildingPlacementPreview.DefaultTranslucentLevel
+[AudioVisual]
+PlacementPreview.Building=no               ; boolean
+PlacementPreview.Building.Translucent=75   ; integer, 0 25 50 75
```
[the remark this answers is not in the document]
Metadorius: I disagree, `PlacementPreview` is a preview of how building would look. for grid I recommend to name it `PlacementGrid`, it doesn't have anything to do with previewing the building itself, it only displays a grid.

docs/User-Interface.md
```diff
@@ -65,33 +65,40 @@ PrioritySelectionFiltering=true  ; boolean
...
 ```
 
+
 
 In `ra2md.ini`:
 ```ini
 [Phobos]
-ShowBuildingPlacementPreview=false  ; boolean
+PlacementPreview.Building=   ; boolean, defaults to rulesmd.ini->[AudioVisual]->PlacementPreview.Building
```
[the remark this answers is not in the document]
Belonit: If in the future someone needs to implement PlacementPreview for grid or some other entity, this can cause problems

src/Phobos.cpp
Belonit: As above, I'm all for better semantics

docs/User-Interface.md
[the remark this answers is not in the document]
Metadorius: There is literally nothing that can be placed down other than buildings, and as I said PlacementPreview has no relation to placement grid, it's two separate things.

src/Ext/BuildingType/Body.cpp
Belonit:
```cpp
// we cannot load same SHP file twice it may produce artifact , prevent it !
    if (_strcmpi(pValue, pSection) || _strcmpi(pValue, pArtSection))
```
This part applies only to this tag, and should not apply to others.

docs/User-Interface.md
Belonit: @Metadorius I don't understand what specific tag you are talking about? `PlacementPreview.Offset` ?

Metadorius: Hm it should've highlighted only the `PlacementPreview.Translucent`. Why would anyone want to do different transparency for only some of the placement previews?

src/Phobos.cpp
Belonit: This tag is used as the default value for RA2MD.ini. We should read it first

docs/User-Interface.md
Belonit: I'm guessing it might be useful for water-only buildings. In any case, we must keep this tag if you want to maintain pre-release compatibility.

src/Phobos.cpp
Metadorius: It makes no sense to use it as a default value in RA2MD.ini, those are controlled by the client. It should be a serializable tag that globally disables or enables the logic, and takes precedence over RA2MD.ini value.

Belonit: What should happen if the tag is not defined? This feature originally had false as the default value in ra2md.ini, which made it pretty pointless.

Metadorius: `true` of course. Yeah `false` by default is stupid. This should be an opt-out switch.

src/Utilities/GeneralUtils.cpp
Belonit: I'm not sure if this is really needed here

Belonit:
Reopen after accidental closure

docs/Whats-New.md
```diff
@@ -21,6 +21,7 @@ You can use the migration utility (can be found on [Phobos supplementaries repo]
 - `Trajectory.Speed` is now defined on projectile instead of weapon.
 - `Gravity=0` is not supported anymore as it will cause the projectile to fly backwards and be unable to hit the target which is not at the same height. Use `Straight` Trajectory instead. See [here](New-or-Enhanced-Logics.md#projectile-trajectories).
 - Automatic self-destruction logic logic has been reimplemented, `Death.NoAmmo`, `Death.Countdown` and `Death.Peaceful` tags have been remade/renamed and require adjustments to function.
+- [Placement Preview](User-Interface.md#placement-preview) logic has been reimplemented, `BuildingPlacementPreview.DefaultTranslucentLeve`, `BuildingPlacementGrid.TranslucentLevel`, `PlacementPreview.Show`, `PlacementPreview.TranslucentLevel` and `ShowBuildingPlacementPreview` tags have been remade/renamed and require adjustments to function.
```
Metadorius, line 24: 1. I think this is more of a refactor than a reimplementation. You didn't change the tag functions right? Might as well list what the changed names are.
2. Please also supply update script entries [here](https://github.com/Phobos-developers/PhobosSupplementaries/blob/develop/MigrationUtility/Scripts/UpdatePhobosTags.sed).
3. Typo in `DefaultTranslucentLeve`.

docs/User-Interface.md
```diff
@@ -65,33 +65,38 @@ PrioritySelectionFiltering=true  ; boolean
 ![placepreview](_static/images/placepreview.png)
 *Building placement preview using 50% translucency in [Rise of the East](https://www.moddb.com/mods/riseoftheeast)*
 
-- Building previews can now be enabled when placing a building for construction. This can be enabled on a global basis with `BuildingPlacementPreview.DefaultTranslucentLevel` and then further customized for each building with `PlacementPreview.TranslucentLevel`.
-- The building placement grid *(place.shp)* translucency setting can be adjusted via `BuildingPlacementGrid.TranslucentLevel`.
-- If using the building's appropriate `Buildup` is not desired, customizations allow for you to choose the exact SHP and frame you'd prefer to show as preview instead through `PlacementPreview.Shape` and `PlacementPreview.ShapeFrame`
+- Building previews can now be enabled when placing a building for construction. This can be enabled on a global basis with `[AudioVisual]->PlacementPreview` and then further customized for each building with `[SOMEBUILDING]->PlacementPreview`.
+- The building placement grid *(place.shp)* translucency setting can be adjusted via `PlacementGrid.Translucency`.
```
Metadorius, line 69:
```suggestion
- The building placement grid (`place.shp`) translucency setting can be adjusted via `PlacementGrid.Translucency`.
```

Also I would make this a separate section because it's not that related to this feature.

docs/User-Interface.md
```diff
@@ -65,33 +65,38 @@ PrioritySelectionFiltering=true  ; boolean
...
+- If using the building's appropriate `Buildup` is not desired, customizations allow for you to choose the exact SHP and frame you'd prefer to show as preview through `PlacementPreview.Shape`, `PlacementPreview.ShapeFrame` and `PlacementPreview.Palette`
```
Metadorius, line 70:
```suggestion
- If using the building's appropriate `Buildup` is not desired, customizations allow for you to choose the exact SHP and frame you'd prefer to show as preview through `PlacementPreview.Shape`, `PlacementPreview.ShapeFrame` and `PlacementPreview.Palette`.
```

Metadorius (changes requested):
Code's good, need a bit of doc changes, also missing a credits and changelog mentions of your refactor, also please PR to supplementaries repo.

docs/Whats-New.md
Belonit: There is also a problem in the old implementation. The `PlacementPreview.TranslucentLevel` tag is named so in the documentation, but it doesn't work. The `PlacementPreview.Translucent` tag is used instead.

Metadorius: It wasn't really reimplemented. The tag namess were adjusted. Also you still didn't address 2 and 3. Maybe then don't write the exact tags?

docs/User-Interface.md
```diff
@@ -65,33 +65,38 @@ PrioritySelectionFiltering=true  ; boolean
...
+PlacementPreview.Shape=              ; filename - including the .shp extension. If not set uses building's artmd.ini Buildup SHP (based on Building's Image)
+PlacementPreview.ShapeFrame=         ; integer, zero-based frame index used for displaying the preview
+PlacementPreview.Offset=0,-15,1      ; integer, expressed in X,Y,Z used to alter position preview
+PlacementPreview.Remap=yes           ; boolean, does this preview use player remap colors
+PlacementPreview.Palette=            ; filename - including the .pal extension.
+PlacementPreview.Translucency=       ; translucency level (0/25/50/75), defaults to [AudioVisual]->PlacementPreview.Translucency
+```
+```{note}
+The `PlacementPreview.Palett` option is not used if `PlacementPreview.Remap=yes` This may change in the future
 ```
```
Metadorius, line 92:
````suggestion

```{note}
The `PlacementPreview.Palette` option is not used when `PlacementPreview.Remap=yes` is set to yes. This may change in future.
```
````

**Review by Metadorius — CHANGES_REQUESTED** [review]
Still missing credits, update scripts.

File: docs/Whats-New.md (hunk 13 above)
Comment (Belonit) [reply docs/Whats-New.md:24]: This PR is broken. I can't add commits to it

File: src/Ext/BuildingType/Hooks.cpp [hunk 17]
```diff
@@ -70,74 +67,83 @@ DEFINE_HOOK(0x458623, BuildingClass_KillOccupiers_Replace_MuzzleFix, 0x7)
...
 			{
-				SHPStruct* Selected = nullptr;
-				bool bBuildupExist = false;
-
-				if (pType->LoadBuildup())
-				{
-					bBuildupExist = true;
-					Selected = pType->LoadBuildup();
-				}
+				if (pImage = pType->LoadBuildup())
```
chaserli, line 97: I recently discovered that when `LoadBuildUp` is called normally `ClearBuildUp` is called right after. Like 0x442CC4. Otherwise according to 0x4659D8 I'm afraid that `CCFileClass` or `SHPStruct` call only their CTOR, not DTOR.

Belonit: Good remark. I will check it

Belonit: image
I don't think this is a problem. Because if the Buildup is already loaded, then it will not be re-created.

PR #711, opened 2022-07-25
[Campaigns] make AI auto-repair base nodes

chaserli:
In singleplayer campaigns, base nodes cannot auto-repair. This PR aims to fix this:

In your `campaign.map`:
```ini
[Basic]
AIRepairBaseNodes=no  ; boolean
```

### DIFF
--- a/src/Ext/House/Hooks.cpp
+++ b/src/Ext/House/Hooks.cpp
@@ -63,3 +63,21 @@ DEFINE_HOOK(0x73E474, UnitClass_Unload_Storage, 0x6)
 
 	return 0;
 }
+
+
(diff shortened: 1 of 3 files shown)


chaserli:
The better way should be changing `pBuilding->ShouldRebuild` when the structure is built. However I didn't find out where the single-player mode involves yet, namely what makes skirmish AI different than campaigns'. 
Also, this enables the repairing globally, I'm not sure if micro-controlling it per house or even per buildingtype is necessary.

chaserli:
Buildings delivered by `Type=UnitDelivery` superweapon or trigger action 125 are not taken into account for the moment, unless someone proposed a better convention for distinguishing all these cases.

PR #717, opened 2022-08-01
hotfix IME Input in WWUI_NewEditCtrl

Belonit:


Thrifinesma:
image
I could prove this fix

src/Misc/TextInput.cpp
```diff
@@ -31,12 +31,10 @@ DEFINE_HOOK(0x5D46C7, MessageListClass_Input, 5)
 	return 0;
 }
 
-DEFINE_HOOK(0x61526C, WWUI_NewEditCtrl, 5)
+DEFINE_HOOK(0x61510E, WWUI_NewEditCtrl, 0)
```
secsome, line 34: I'd prefer to keep its hook size though:
`DEFINE_HOOK(0x61510E, WWUI_NewEditCtrl, 6)`
instead of
`DEFINE_HOOK(0x61510E, WWUI_NewEditCtrl, 0)`

Keep the size is not a bad thing I guess, though it works the same

Belonit: If I'm not mistaken, this is to force the syringe to allocate an additional 6 bytes of memory to store the old instructions. I try to keep the hook size zero, wherever it is not required to return 0

secsome: What I mean is that though we don't need to return 0 here, should it be better to leave the hook size here.
As we do that at other hooks

Belonit:
@Metadorius Could you check the docs and merge this ?

Metadorius (approved):

PR #724, opened 2022-08-04
Changed Forced Jump timer from frames to ingame seconds

FS-21:
Like Timed Area Guard (script action 71) it now use seconds in the argument.
Is more easy to use "15" for 15 seconds than "225" for the same seconds.

### DIFF
--- a/docs/AI-Scripting-and-Mapping.md
+++ b/docs/AI-Scripting-and-Mapping.md
@@ -301,7 +301,7 @@ x=124,0
 In `aimd.ini`:
 ```ini
 [SOMESCRIPTTYPE]  ; ScriptType
-x=125,n           ; integer n=0, in frames
+x=125,n           ; integer n=0, in ingame seconds
 ```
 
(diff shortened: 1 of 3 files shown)


Metadorius (commented):
You also have to create an upgrade script entry in Phobos Supplementaries repo.

FS-21:
After checking the PhobosSupplementaries repo I think there is a misunderstanding, These Forced jump actions 125 & 126 exist since 0.3 so no migration is needed. Last stable of Phobos only have 3 script actions if I'm not wrong (71-73).

Metadorius:
Migration means any manual adjustments to keep the game working as before. You'll have to add a comment to the lines with this script action that looks like this for example:
```ini
1234=125,15  ; FIXME Ensure the time is specified in ingame seconds (15 frames per second)
```

Metadorius:
And please pay attention to docs, we have inter-devbuild migration notes. Your statement would have held true had we released 0.3 half a year earlier, but we didn't, so to compensate we're doing this.

PR #726, opened 2022-08-10
New condition when some TechnoTypes exist/nonexist of automatic self-destruction logic

ststl-s:
```ini
AutoDeath.Nonexist=             ; list of TechnoType names
AutoDeath.Nonexist.House=owner  ; Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
AutoDeath.Exist=                ; list of TechnoType names
AutoDeath.Exist.House=owner     ; Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
```


docs/Whats-New.md
```diff
@@ -330,6 +330,7 @@ New:
 - Implemented support for PCX images for campaign loading screen (by FlyStar)
 - Implemented support for PCX images for observer loading screen (by Uranusian)
 - Animated (non-tiberium spawning) TerrainTypes (by Starkku)
+- New condition when some technotypes exist/nonexist of automatic self-destruction logic (by FlyStar)
```
Starkku, line 333: *New condition for killing objects automatically when TechnoType(s) exist/don't exist*

src/Ext/Techno/Body.cpp
```diff
@@ -109,6 +109,54 @@ void TechnoExt::ExtData::CheckDeathConditions()
...
+							pHouse->CountOwnedAndPresent(pType))
+							return true;
+					}
+
+					return false;
+				}
+			);
+
+			if (exist)
+				KillSelf(pThis, howToDie);
```
Starkku, line 158: I think this code could be simplified quite a lot. The lambda expression is essentially repeated for exist/does not exist conditions, when it could probably be reused by passing the TechnoType vector and AffectedHouses flags as params. Unless I am missing something that would crucially affect performance by doing it that way.

Starkku (changes requested):
Idea is good but the code needs some changes. Maybe others could chime in on their suggestions for the INI key names, I put forward my two cents on the matter in a comment as well since I don't think the current ones are descriptive enough.

In an ideal world, having a switch for the 'TechnoType exists' condition that makes it require *all* of the listed TechnoTypes to be present for the condition to be fulfilled would be nice but I believe under the current system where it can check for Technos under more than one House, this would be complicated to implement and could behave oddly so I think omitting it would probably be fine.

ststl-s:
> Idea is good but the code needs some changes. Maybe others could chime in on their suggestions for the INI key names, I put forward my two cents on the matter in a comment as well since I don't think the current ones are descriptive enough.
> 
> In an ideal world, having a switch for the 'TechnoType exists' condition that makes it require _all_ of the listed TechnoTypes to be present for the condition to be fulfilled would be nice but I believe under the current system where it can check for Technos under more than one House, this would be complicated to implement and could behave oddly so I think omitting it would probably be fine.

About the House setting temporarily reserved, is this dangerous? I haven't experimented a lot, just confirmed that it works.

docs/New-or-Enhanced-Logics.md
```diff
@@ -610,11 +610,17 @@ If the object enters transport, the countdown will continue, but it will not sel
...
+AutoDeath.Behavior=                     ; enumeration (kill | vanish | sell), default not set
+
+AutoDeath.OnAmmoDepletion=no            ; boolean
+AutoDeath.AfterDelay=0                  ; positive integer
+AutoDeath.TechnosDontExist=             ; list of TechnoType names
+AutoDeath.TechnoDontExist.Any=false     ; boolean
+AutoDeath.TechnoDontExist.Houses=owner  ; Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
+AutoDeath.TechnoExist=                  ; list of TechnoType names
+AutoDeath.TechnoExist.Any=true          ; boolean
+AutoDeath.TechnoExist.Houses=owner      ; Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
```
Starkku, line 623: Also missing the `s` here at end of Techno**s** too.

Starkku (approved):

chaserli:
I didn't manage to give my review on this in time.
Basically you give too many new arguments on this specific feature, which makes the code looks _extrêmement lourde_, 
in this case I would rather suggest you guys wrap those into a new entity, like the AttachEffectType or PrismForwarding in Ares,
and see if it can work like an AttachEffect.

Same suggestion for other similar tags like the interceptor one

PR #723, opened 2022-08-03
Fix Psychedelic warhead giving actual damage against shields, add "Shield is broken" trigger event

chaserli:
Sorry I accidentally deleted #703 
Linked with #721. Partially redirected to #728 

# Default setting of Psychedelic damage to shields

In previous builds, `Psychedelic=yes` warhead can damage shields without giving the actual berserk effect. 
From now on `Psychedelic=yes` warheads directly penetrate shields by default. An additional parameter can be given to shields for the immunity against berserk effect:

In `rulesmd.ini`:
```ini
[SOMESHIELDTYPE]
PsychedelicProof=no
```

Note that this doesn't prevent the unit from targeting at the shielded object even if the shield has `PsychedelicProof=yes`. `Versus.shieldArmor=0%` is still required in this case.

# Trigger Event 600: Shield of the attached object is broken

The trigger event is raised when the shield of the attached object is broken.
In `mycampaign.map`:
```ini
[Events]
...
ID=EventCount,[Event1],...,600,2,0,0,[EventX],...
...

PR #721, opened 2022-08-03
Add IgnoreDesignator option for LaunchSW, Refactor ParallelAIQueue & SWTypeExt;

chaserli:
I accidentally deleted #720, this PR restores it.

### Launch superweapons on impact
In `rulesmd.ini`:
```ini
[SOMEWARHEAD]                    ; Warhead
LaunchSW.IgnoreDesignators=true  ; boolean
```
Detached shield pointer invalidation part into #740 

### additional refactor
1. Fix the wrong hook size and stack offset for future usage of SW cell coordinates in `SuperClass::Launch` hook. Minor refactored limbo delivery part to improve the readability of #379 

2. **Moved all ForbidParalleAIlQueue related settings to `RulesExt`** #638 , serialized all related items as well as `[GlobalControls]->AllowParallelAIQueues` , fixed related code errors

PR #728, opened 2022-08-17
Refactor Harvester counter to improve performance

### LINKED ISSUE #603: [Phobos Bug] HarvesterCounter believes that the harvester in Chrono ready state is not working

chaserli:
Detached from #723 ,  restore #628 , closes #603 

* Consider non-selected guarding state as harvesting miner (can subject to change, Needs review from @Metadorius @Thrifinesma ). Returning chronominers will stop blinking the counter.

* Deprecate `TechnoTypeExt::ExtData::IsCountedAsHarvester` and merged the entire logic into `Harverster.Counted` in order to improve performance. Minor refactored related harvester-counter code.

* Add a static array of TechnoTypeClasses that are considered to be `Harvester.Counted=yes` to avoid loop through all technotypes every frame.

* Do not display harvester counter or power delta for defeated players

### DIFF
--- a/src/Ext/House/Body.cpp
+++ b/src/Ext/House/Body.cpp
@@ -13,32 +13,30 @@ HouseExt::ExtContainer HouseExt::ExtMap;
 
 int HouseExt::ActiveHarvesterCount(HouseClass* pThis)
 {
-	if (!pThis) return 0;
+	if (!pThis || pThis->Defeated) return 0;
 
 	int result = 0;
-	for (auto techno : *TechnoClass::Array) {
-		if (auto pTechnoExt = TechnoTypeExt::ExtMap.Find(techno->GetTechnoType())) {
-			if (pTechnoExt->IsCountedAsHarvester() && techno->Owner == pThis) {
-				if (auto pTechno = TechnoExt::ExtMap.Find(techno)) {
-					result += TechnoExt::IsHarvesting(techno);
-				}
-			}
+	for (auto techno : *TechnoClass::Array)
+	{
+		if (techno->Owner == pThis)
+		{
+			if (auto pTypeExt = TechnoTypeExt::ExtMap.Find(techno->GetTechnoType()))
+				result += pTypeExt->Harvester_Counted && TechnoExt::IsHarvesting(techno);
 		}
 	}
-
-	return result;
 }
 
 int HouseExt::TotalHarvesterCount(HouseClass* pThis)
 {
-	if (!pThis)	return 0;
+	if (!pThis || pThis->Defeated)	return 0;
 
 	int result = 0;
-	for (auto techno : *TechnoTypeClass::Array) {
-		if (auto pTechnoExt = TechnoTypeExt::ExtMap.Find(techno)) {
-			if (pTechnoExt->IsCountedAsHarvester()) {
+	for (auto techno : *TechnoTypeClass::Array)
+	{
+		if (auto pTypeExt = TechnoTypeExt::ExtMap.Find(techno))
+		{
+			if (pTypeExt->Harvester_Counted)
 				result += pThis->CountOwnedAndPresent(techno);
-			}
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


src/Ext/House/Body.cpp
```diff
@@ -13,34 +13,23 @@ HouseExt::ExtContainer HouseExt::ExtMap;
...
-
 	int result = 0;
-	for (auto techno : *TechnoTypeClass::Array) {
-		if (auto pTechnoExt = TechnoTypeExt::ExtMap.Find(techno)) {
-			if (pTechnoExt->IsCountedAsHarvester()) {
-				result += pThis->CountOwnedAndPresent(techno);
-			}
-		}
-	}
+	for (auto pType : TechnoTypeExt::HarvesterTypes)
```
chaserli, line 31: I got a problem here: When I tested with MO everything's fine, but when I test 
with my vanilla YR it gets an error here: either this `std::vector` or `pType` is some nonsense or `nullptr`

src/Ext/Techno/Body.cpp
```diff
@@ -351,11 +351,19 @@ bool TechnoExt::IsHarvesting(TechnoClass* pThis)
...
+		switch (pThis->GetCurrentMission())
+		{
+		case Mission::Harvest:
+		case Mission::Unload:
+		case Mission::Enter:
+			return true;
+		case Mission::Guard: // issue#603: idk how to do better
+			if (auto pUnit = abstract_cast<UnitClass*>(pThis))
+				return !pUnit->IsSelected && pUnit->Locomotor->Is_Really_Moving_Now();
+		default:break;
```
Metadorius, line 365:
```suggestion
		default: break;
```

src/Ext/TechnoType/Body.cpp
```diff
@@ -53,18 +53,19 @@ bool TechnoTypeExt::HasSelectionGroupID(ObjectTypeClass* pType, const char* pID)
 	return (_strcmpi(id, pID) == 0);
 }
 
+[[deprecated("Use Harvester_Counted directly.")]]
```
Metadorius, line 56: Why leave it in at all?

src/Ext/TechnoType/Body.cpp
```diff
@@ -476,3 +486,21 @@ DEFINE_HOOK(0x679CAF, RulesClass_LoadAfterTypeData_CompleteInitialization, 0x5)
...
+{
+	GET(UnitTypeClass*, pItem, ESI);
+
+	if (auto pTypeExt = TechnoTypeExt::ExtMap.Find(pItem))
+	{
+		if (!pTypeExt->Harvester_Counted.isset() && pItem->Harvester)
+		{
+			pTypeExt->Harvester_Counted = true;
+			auto& list = RulesExt::Global()->HarvesterTypes;
+			if(!list.Contains(pItem))
```
Metadorius, line 500:
```suggestion
			if (!list.Contains(pItem))
```

Metadorius (approved):

Metadorius (changes requested):
Forgot to say, add yourself to authors of the features you fixed and if it was in a stable build - append the fix to the changelog.

src/Ext/TechnoType/Body.cpp
chaserli: Removed. Ready to go with #721

PR #734, opened 2022-08-29
Attached particle system for animations

Starkku:
Basically `VoxelAnims` / projectile (Ares) `AttachedSystem` but for animations. Primary use case is to use for debris instead of animation trailers but there may be potential for other things too I suppose.

----

- It is now possible to attach a particle system to an animation. Only particle systems with `BehavesLike=Smoke` are supported. This works similarly to the identically named key on `VoxelAnims`.
  - On animations with `Next`, the particle system will be deleted when the next animation starts playing and new one created in its stead if the `Next` animation defines a different particle system.

In `artmd.ini`:
```ini
[SOMEANIM]       ; AnimationType
AttachedSystem=  ; ParticleSystem
```

### DIFF
--- a/src/Ext/Anim/Body.cpp
+++ b/src/Ext/Anim/Body.cpp
@@ -6,6 +6,29 @@
 template<> const DWORD Extension<AnimClass>::Canary = 0xAAAAAAAA;
 AnimExt::ExtContainer AnimExt::ExtMap;
 
+void AnimExt::ExtData::CreateAttachedSystem(ParticleSystemTypeClass* pSystemType)
+{
+	const auto pThis = this->OwnerObject();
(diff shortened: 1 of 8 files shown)


src/Ext/Anim/Body.cpp
```diff
@@ -76,6 +101,9 @@ DEFINE_HOOK(0x422967, AnimClass_DTOR, 0x6)
 {
 	GET(AnimClass*, pItem, ESI);
 
+	if (auto const pExt = AnimExt::ExtMap.Find(pItem))
```
secsome, line 104: Would it be better to put it in `AnimExt::~AnimExt`

Starkku: Probably. Fixed.

PR #735, opened 2022-08-30
[Hotfix] Fix slaves not always dying from suicide option when slave-miner gets killed

chaserli:
Fixed a parameter misuse from vanilla: when YR kills a slave it used `ignoreDefense=false` rather than true. Slaves may not always get killed

### DIFF
--- a/src/Ext/Techno/Hooks.cpp
+++ b/src/Ext/Techno/Hooks.cpp
@@ -432,7 +432,7 @@ DEFINE_HOOK(0x6B0B9C, SlaveManagerClass_Killed_DecideOwner, 0x6)
 	return 0x0;
 }
 
-//Fix slaves cannot always suicide due to armor multiplier or something
+// Fix slaves cannot always suicide due to armor multiplier or something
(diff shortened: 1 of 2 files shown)


src/Ext/Techno/Hooks.cpp
```diff
@@ -432,6 +432,11 @@ DEFINE_HOOK(0x6B0B9C, SlaveManagerClass_Killed_DecideOwner, 0x6)
 	return 0x0;
 }
 
+//Fix slaves cannot always suicide due to armor multiplier or something
```
Metadorius, line 435:
```suggestion
// Fix slaves cannot always suicide due to armor multiplier or something
```

Metadorius (changes requested):
What's new, docs, credits entries where?

chaserli:
> What's new, docs, credits entries where?

Actually no, it's just fixing #630 , an issue proposed by Akane from discord
https://discord.com/channels/338055131666972683/773636942775582720/1013827800597282947

PR #740, opened 2022-09-04
Performance refactor: Detach shield anim pointer invalidation from techno

chaserli:
Given the importance and the difficulty of this task, I detached this part from #721 

Prior to this PR, `ContainerMap.Find` and `InvalidatePointerHelper::Process<TechnoExt>` are the 2 major factors that greatly impact the performance, causing a significant FPS drop when objects number surpasses hundreds. 

This PR aims to detach the shield `IdleAnim` pointer invalidation from `TechnoExt` by nullifying the existing pointer invalidation virtual function of `TechnoExt::ExtData` and introducing a static one inside `ShieldClass`. So I used a global vector storing all shield class pointers. As for pointer invalidation, it loops through the vector to see if any `IdleAnim` matches.

### DIFF
--- a/src/New/Entity/ShieldClass.cpp
+++ b/src/New/Entity/ShieldClass.cpp
@@ -12,6 +12,11 @@
 #include <RadarEventClass.h>
 #include <TacticalClass.h>
 
+namespace NewEntities
+{
+	ValueableVector<ShieldClass*> Shields;
+}
+
 ShieldClass::ShieldClass() : Techno { nullptr }
 	, HP { 0 }
 	, Timers { }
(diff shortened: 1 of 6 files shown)


Starkku:
From what I can gather, serializing the shield instance vector is unnecessary as it is reconstructed upon loading the game anyway. You'd probably want to add `NewEntities::Shields.emplace_back(this);` to the other `ShieldClass` constructor as well, though.

src/New/Entity/ShieldClass.cpp
```diff
@@ -12,6 +12,11 @@
 #include <RadarEventClass.h>
 #include <TacticalClass.h>
 
+namespace NewEntities
+{
+	std::vector<ShieldClass*> Shields;
+}
+
```
Metadorius, line 19: Why not a public static array in ShieldClass? See how it's done in YRpp for arrays and in Techno Attachment branch for AttachmentClass.

src/New/Entity/ShieldClass.cpp
```diff
@@ -214,9 +258,9 @@ void ShieldClass::ResponseAttack()
 		if (pUnit->Type->Harvester)
 		{
 			const auto pos = pUnit->GetDestination(pUnit);
-
+			enum { EVA_OreMinerUnderAttack = 0x824784 };
```
Metadorius, line 261: Such stuff should be in YRpp I believe but that can be left for another PR.

Metadorius (changes requested):

src/New/Entity/ShieldClass.cpp
chaserli: I did it first and it couldn't compile, just turned out to be a type on my part LOL. Fixed and removed global serialization

src/New/Entity/ShieldClass.h
```diff
@@ -13,7 +13,9 @@ class ShieldClass
 	ShieldClass();
 	ShieldClass(TechnoClass* pTechno, bool isAttached);
 	ShieldClass(TechnoClass* pTechno) : ShieldClass(pTechno, false) {};
-	~ShieldClass() = default;
+	~ShieldClass();
+
+	static std::vector<ShieldClass*> Array;
```
Metadorius, line 18: Static variables like this one should be at the top I think.

Metadorius (changes requested):
One more change, otherwise LGTM.

src/New/Entity/ShieldClass.cpp
Metadorius: Will check later.

Metadorius (approved):
Hm should've pressed approve instead so you could merge after making a change without additional review.

chaserli:
Test done on my side: The only situation for shield anim pointer invalidation I can find is when the techno gets cloaked. Tested and the anim disappears as expected when techno gets cloaked, which is also confirmed by the debugger. `ShieldClass::Array`'s elements are inserted/removed as expected on save/load.

PR #738, opened 2022-09-03
Fix turreted jumpjet units always facing bottom-right direction when stop firing

chaserli:
Just as the title said

PR #744, opened 2022-09-07
Fix jumpjets' Sensors being unable to detect cloaked objects (temporarily)

chaserli:
Just as the title said.
BTW the bug of turreted jumpjet units always facing the bottom-right direction when not targeting anything is fixed too in #738, welcome to test them together. Only `Jumpjet=yes` units are considered.

Note:
My idb is very poor, and I can only assume at this moment that `JumpjetLocomotionClass::State` is an enum, where
* 0 - On the ground
* 1 - Taking off from the ground
* 2 - Hovering in the air
* 3 - Moving in air
* 4 - Deploying to land
* 5 - Crashing
* 6 - Unknown, probably related to crashing

It seems that `ObjectClass::UpdatePosition` is called in `4`, but not the others. I decided to add it in 3, but I'm not sure if it's also done elsewhere so I limited it to `Sensors=yes` types only for now.
Since there's something else processed in this function so there may be some additional (side) effects as well, which may be related with
* threat evaluation and targeting
*  trigger events evaluation (entered by sort of stuff)
*  planning waypoints processing

### DIFF
--- a/src/Ext/Unit/Hooks.Jumpjet.cpp
+++ b/src/Ext/Unit/Hooks.Jumpjet.cpp
@@ -79,15 +79,18 @@ DEFINE_HOOK(0x736BF3, UnitClass_UpdateRotation_TurretFacing, 0x6)
 // Bugfix: Jumpjet detect cloaked objects beneath
 DEFINE_HOOK(0x54C036, JumpjetLocomotionClass_State3_54BFF0_UpdateSensors, 0x7)
 {
-
 	GET(FootClass* const, pLinkedTo, ECX);
 	GET(CellStruct const, currentCell, EAX);
 
 	// Copied from FootClass::UpdatePosition
 	if (pLinkedTo->GetTechnoType()->SensorsSight)
 	{
-		pLinkedTo->RemoveSensorsAt(pLinkedTo->LastJumpjetMapCoords);
-		pLinkedTo->AddSensorsAt(currentCell);
+		CellStruct const lastCell = pLinkedTo->LastJumpjetMapCoords;
+		if (lastCell != currentCell)
+		{
+			pLinkedTo->RemoveSensorsAt(lastCell);
+			pLinkedTo->AddSensorsAt(currentCell);
+		}
 	}
 	// Something more may be missing
 
--- a/src/Ext/Unit/Hooks.Jumpjet.cpp
+++ b/src/Ext/Unit/Hooks.Jumpjet.cpp
@@ -3,17 +3,49 @@
 
 #include <Ext/TechnoType/Body.h>
 
-DEFINE_HOOK(0x54AEC0, JumpjetLocomotion_Process_TurnToTarget, 0x8)
+DEFINE_HOOK(0x54B8E9, JumpjetLocomotionClass_In_Which_Layer_Deviation, 0x6)
 {
-	GET_STACK(ILocomotion*, iLoco, 0x4);
-	const auto pLoco = static_cast<JumpjetLocomotionClass*>(iLoco);
-	const auto pThis = pLoco->Owner;
-	const auto pType = pThis->GetTechnoType();
-	const auto pTarget = pThis->Target;
+	GET(FootClass* const, pThis, EAX);
 
-	if (pTarget && pThis->IsInAir() && !pType->TurretSpins && pThis->WhatAmI() == AbstractType::Unit)
+	if (pThis->IsInAir())
 	{
-		const auto pTypeExt = TechnoTypeExt::ExtMap.Find(pType);
+		if (auto const pExt = TechnoTypeExt::ExtMap.Find(pThis->GetTechnoType()))
+		{
+			if (!pExt->JumpjetAllowLayerDeviation.Get(RulesExt::Global()->JumpjetAllowLayerDeviation.Get()))
+			{
+				R->EDX(INT32_MAX); // Override JumpjetHeight / CruiseHeight check so it always results in 3 / Layer::Air.
+				return 0x54B96B;
+			}
+		}
+	}
+
+	return 0;
+}
+
+// I think JumpjetLocomotionClass::State is probably an enum, where
+// 0 - On ground
+// 1 - Taking off from ground
+// 2 - Hovering in air
+// 3 - Moving in air
+// 4 - Deploying to land
+// 5 - Crashing
+// 6 - Invalid?
+
+// Bugfix: Jumpjet turn to target when attacking
+// Even though it's still not the best place to do this, given that 0x54BF5B has done the similar action, I'll do it here too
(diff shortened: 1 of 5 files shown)


Starkku (dismissed):
Code looks okay, approving for now so Github generates the builds for testers to use.

EDIT: Or not.

Starkku:
Appears to work though. Tested with both jumpjet (BalloonHover kind) infantry and vehicles. Pending further testing on any odd side effects though.

src/Ext/Unit/Hooks.Jumpjet.cpp
```diff
@@ -28,16 +30,39 @@ DEFINE_HOOK(0x54AEC0, JumpjetLocomotion_Process_TurnToTarget, 0x8)
...
+// 1 - Taking off from ground
+// 2 - Hovering in air
+// 3 - Moving in air
+// 4 - Deploying to land
+// 5 - Crashing
+DEFINE_HOOK(0x54C14B, JumpjetLocomotionClass_State3_54BFF0_UpdateMotion, 0x7)
+{
+	GET(FootClass*, pLinkedTo, EDI);
+
+	if (pLinkedTo->GetTechnoType()->Sensors) // this is probably not needed, as I think all jumpjets needs to UpdatePosition as well, or not?
```
Starkku, line 64: For sake of consistency I would probably remove this condition since calling this has implications beyond updating sensor sight.

chaserli: Alright, and I think if there are any, we will probably not be able to notice them when testing this PR only. I hope that `UpdatePosition` is idempotent

chaserli: I just tested it again, it appears that when a lot of jumpjets are moving at the same time the FPS drops significantly as well. The higher `Sight` value is, the slower the game becomes.

chaserli:
https://discord.com/channels/338055131666972683/737629486568177734/1018141764315512853 @Metadorius 
I mean, it's kind of strange. In `WalkLocomotionClass_ILocomotion_Process` it calls `UpdatePosition(2)` at 0x75BE3C, which should also be per-frame execution I suppose?. But since I don't find this locomotionclass in YRPP I'm not sure what happened before and after

Starkku:
If only the sensors stuff from `UpdatePosition()` matters here then you can just replace the hook with the following (with most up-to-date YRpp) instead of calling `UpdatePosition()` if it causes lag.

```cpp
DEFINE_HOOK(0x54C02D, JumpjetLocomotionClass_State3_54BFF0_UpdateSensors, 0x5)
{
	GET(FootClass* const, pLinkedTo, ECX);
	GET(CellStruct const, currentCell, EAX);

	auto const pType = pLinkedTo->GetTechnoType();

	if (pType->SensorsSight)
	{
		pLinkedTo->RemoveSensorsAt(pLinkedTo->LastJumpjetMapCoords);
		pLinkedTo->AddSensorsAt(currentCell);
	}

	return 0;
}
```

chaserli:
> If only the `SensorsSight` stuff from `UpdatePosition()` matters here then you can just replace the hook with the following (with most up-to-date YRpp) instead of calling `UpdatePosition()`.
> 
> ```c++
> DEFINE_HOOK(0x54C036, JumpjetLocomotionClass_State3_54BFF0_UpdateSensors, 0x7)
> {
> 	GET(FootClass* const, pLinkedTo, ECX);
> 	GET(CellStruct const, currentCell, EAX);
> 
> 	auto const pType = pLinkedTo->GetTechnoType();
> 
> 	if (pType->SensorsSight)
> 	{
> 		pLinkedTo->RemoveSensorsAt(pLinkedTo->LastJumpjetMapCoords);
> 		pLinkedTo->AddSensorsAt(currentCell);
> 	}
> 
> 	return 0;
> }
> ```
> 
> For what it is worth, `Sensors` (which is a separate logic from `SensorsSight` that reveals units in adjacent cells) does not seem to apply at all with jumpjet units (no matter if they are the one with `Sensors` or on receiving end) anyways, regardless of if `UpdatePosition()` is called or not.

That works. Well when I update YRPP I was thinking of doing the samething, but the unknown cell 0x8B3D88 from what 0xx4D8617 made me confused, and if I use that cell the sensor removal is not working.

Metadorius:
Actually considering sensors seem to be per-cell - can you check if the current cell is different compared to the cell which the previous sensor location was in? If it's different - only then move the sensor.

Metadorius (approved):

src/Ext/Unit/Hooks.Jumpjet.cpp
```diff
@@ -3,17 +3,49 @@
...
+// 1 - Taking off from ground
+// 2 - Hovering in air
+// 3 - Moving in air
+// 4 - Deploying to land
+// 5 - Crashing
+// 6 - Invalid?
+
+// Bugfix: Jumpjet turn to target when attacking
+// Even though it's still not the best place to do this, given that 0x54BF5B has done the similar action, I'll do it here too
+DEFINE_HOOK(0x54BD93, JumpjetLocomotionClass_State2_54BD30_TurnToTarget, 0x6)
```
chaserli, line 36: A reminder to myself: Units process the facing's rotation on firing at `UnitClass::sub_736DF0`, (I call it `UnitClass::UpdateFiring`), it only turns the PrimaryFacing and TurretFacing, yet jumpjet is an exception.
So the correct way should be like 
```cpp
DEFINE_HOOK_AGAIN(0x736FE8, UnitClass_UpdateFireError_736DF0_JumpjetFacing, 0x6)// No Turret and FireError ==FACING
DEFINE_HOOK(0x736EE9, UnitClass_UpdateFireError_736DF0_JumpjetFacing, 0x6) //FireError == OK
{
	GET(UnitClass* const, pThis, ESI);

	CLSID locoCLSID;
	if (SUCCEEDED(static_cast<LocomotionClass*>(pThis->Locomotor.get())->GetClassID(&locoCLSID)) && locoCLSID == LocomotionClass::CLSIDs::Jumpjet)
	{
		auto pLoco = static_cast<JumpjetLocomotionClass*>(pThis->Locomotor.get());
		CoordStruct& source = pThis->Location;
		CoordStruct target = pThis->Target->GetCoords();
		DirStruct tgtDir = DirStruct { Math::atan2(source.Y - target.Y, target.X - source.X) };

		if (pThis->GetRealFacing().GetFacing<32>() != tgtDir.GetFacing<32>())
			pLoco->LocomotionFacing.SetDesired(tgtDir);
	}
	return R->Origin() == 0x736EE9 ? 0 : 0x737021;
}
```
However in real game play I spotted almost no difference compared with the current one, except it exhibits a minor delay. So I'll just keep it the way it is for the moment

PR #745, opened 2022-09-07
Remove hardcoded AA/Gattling weapon selection restrictions

Starkku:
- If both `Primary` and `Secondary` weapons can fire at air targets (projectile has `AA=true`), `Primary` can now be picked instead of always forcing `Secondary`. Also applies to `IsGattling=true`, with odd-numbered and even-numbered `WeaponX` slots instead of `Primary` and `Secondary`, respectively.
- `IsGattling=true` can now fall back to ´secondary´ weapon slot (even-numbered `WeaponX` slots) if 'primary' one (odd-numbered) cannot fire at current target (armor type, `CanTarget(Houses)`, shield etc).

This allows for things like AA-only units with two different weapons and units using gattling logic with two sets of weapons that can both fire at ground targets.

As a bonus threw these fixes in:

- Fixed `LandTargeting=1` not preventing from targeting TerrainTypes (trees etc.) on land.
- Fixed `NavalTargeting=6` not preventing from targeting empty water cells or TerrainTypes (trees etc.) on water.
- Fixed `NavalTargeting=7` and/or `LandTargeting=2` resulting in still targeting TerrainTypes (trees etc.) on land with `Primary` weapon.

### DIFF
--- a/src/Ext/Techno/Hooks.Firing.cpp
+++ b/src/Ext/Techno/Hooks.Firing.cpp
@@ -1,6 +1,7 @@
 #include "Body.h"
 
 #include <ScenarioClass.h>
+#include <TerrainClass.h>
 
 #include <Ext/Bullet/Body.h>
 #include <Ext/WarheadType/Body.h>
@@ -239,12 +240,13 @@ DEFINE_HOOK(0x5218F3, InfantryClass_WhatWeaponShouldIUse_DeployFireWeapon, 0x6)
 
 DEFINE_HOOK(0x6FC339, TechnoClass_CanFire, 0x6)
 {
+	enum { CannotFire = 0x6FCB7E };
+
 	GET(TechnoClass*, pThis, ESI);
 	GET(WeaponTypeClass*, pWeapon, EDI);
 	GET_STACK(AbstractClass*, pTarget, STACK_OFFSET(0x20, 0x4));
 	// Checking for nullptr is not required here, since the game has already executed them before calling the hook  -- Belonit
 	const auto pWH = pWeapon->Warhead;
-	enum { CannotFire = 0x6FCB7E };
 
 	if (const auto pWHExt = WarheadTypeExt::ExtMap.Find(pWH))
 	{
@@ -299,6 +301,37 @@ DEFINE_HOOK(0x6FC587, TechnoClass_CanFire_OpenTopped, 0x6)
 			if (pTransport->Deactivated && !pExt->OpenTopped_AllowFiringIfDeactivated)
 				return DisallowFiring;
 		}
+	}
+
+	return 0;
+}
+
+DEFINE_HOOK(0x6FC689, TechnoClass_CanFire_LandNavalTarget, 0x6)
+{
+	enum { DisallowFiring = 0x6FC86A };
+
+	GET(TechnoClass*, pThis, ESI);
+	GET_STACK(AbstractClass*, pTarget, STACK_OFFS(0x20, -0x4));
+
+	const auto pType = pThis->GetTechnoType();
+	auto pCell = abstract_cast<CellClass*>(pTarget);
+
+	if (pCell)
+	{
+		if (pType->NavalTargeting == 6 &&
+			(pCell->LandType == LandType::Water || pCell->LandType == LandType::Beach))
+		{
+			return DisallowFiring;
+		}
+	}
+	else if (const auto pTerrain = abstract_cast<TerrainClass*>(pTarget))
+	{
+		pCell = pTerrain->GetCell();
+
+		if (pType->LandTargeting == 1 && pCell->LandType != LandType::Water && pCell->LandType != LandType::Beach)
+			return DisallowFiring;
+		else if (pType->NavalTargeting == 6 && (pCell->LandType == LandType::Water || pCell->LandType == LandType::Beach))
+			return DisallowFiring;
 	}
 
 	return 0;
--- a/src/Ext/Techno/Hooks.Firing.cpp
+++ b/src/Ext/Techno/Hooks.Firing.cpp
@@ -145,19 +145,16 @@ DEFINE_HOOK(0x6F36DB, TechnoClass_WhatWeaponShouldIUse, 0x8)
 
 DEFINE_HOOK(0x6F37EB, TechnoClass_WhatWeaponShouldIUse_AntiAir, 0x6)
 {
-	enum { ReturnValue = 0x6F37AF };
+	enum { Primary = 0x6F37AD, Secondary = 0x6F3807 };
 
 	GET(TechnoClass*, pTargetTechno, EBP);
 	GET_STACK(WeaponTypeClass*, pWeapon, STACK_OFFS(0x18, 0x4));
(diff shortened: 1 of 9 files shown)


src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -159,6 +128,89 @@ DEFINE_HOOK(0x6F36DB, TechnoClass_WhatWeaponShouldIUse, 0x8)
...
+DEFINE_HOOK(0x6F37EB, TechnoClass_WhatWeaponShouldIUse_AntiAir, 0x6)
+{
+	enum { ReturnValue = 0x6F37AF };
+
+	GET(TechnoClass*, pThis, ESI);
+	GET(TechnoClass*, pTargetTechno, EBP);
+
+	int returnValue = 0;
+
+	if (const auto pTypeExt = TechnoTypeExt::ExtMap.Find(pThis->GetTechnoType()))
```
chaserli, line 140: Where is this TypeExt used?

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -159,6 +128,89 @@ DEFINE_HOOK(0x6F36DB, TechnoClass_WhatWeaponShouldIUse, 0x8)
...
+	{
+		auto pWeapon = pThis->GetWeapon(0)->WeaponType;
+		auto pSecWeapon = pThis->GetWeapon(1)->WeaponType;
+
+		if (pWeapon->Projectile->AA && pTargetTechno && pTargetTechno->IsInAir())
+			returnValue = 0;
+		else if (pSecWeapon->Projectile->AA && pTargetTechno && pTargetTechno->IsInAir())
```
chaserli, line 147: `pTargetTechno` and inair checks seem duplicated, could be moved before?

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -159,6 +128,89 @@ DEFINE_HOOK(0x6F36DB, TechnoClass_WhatWeaponShouldIUse, 0x8)
...
+
+	GET(TechnoClass*, pThis, ESI);
+	GET(TechnoClass*, pTargetTechno, EBP);
+
+	int returnValue = 0;
+
+	if (const auto pTypeExt = TechnoTypeExt::ExtMap.Find(pThis->GetTechnoType()))
+	{
+		auto pWeapon = pThis->GetWeapon(0)->WeaponType;
+		auto pSecWeapon = pThis->GetWeapon(1)->WeaponType;
```
chaserli, line 143: seems to be `EAX` as well at this moment?

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -159,6 +128,89 @@ DEFINE_HOOK(0x6F36DB, TechnoClass_WhatWeaponShouldIUse, 0x8)
...
+		{
+			if (pWeapon && pWeapon->Projectile->AA)
+				returnValue = primaryIndex;
+			else if (pWeaponSec && pWeaponSec->Projectile->AA)
+				returnValue = secondaryIndex;
+		}
+	}
+
+	R->EAX(returnValue);
+	return ReturnValue;
```
chaserli, line 211: `returnValue` `ReturnValue` look confusing

Starkku:
Fixed the outstanding issues from feedback and simplified the code in `TechnoClass_WhatWeaponShouldIUse_Gattling` hook.

ststl-s:
I think if gattling can select weapon by versus....
Just a suggestion.

```
DEFINE_HOOK(0x6F3436, TechnoClass_SelectGattlingWeapon, 0x6)
{
	GET(TechnoClass*, pThis, ESI);
	GET(AbstractClass*, pTarget, EBP);

	enum { Odd = 0x6F346A, Even = 0x6F345C, Origin = 0 };
	int gattlingStage = pThis->CurrentGattlingStage;
	auto pTypeExt = TechnoTypeExt::ExtMap.Find(pThis->GetTechnoType());

	if (pTypeExt->Gattling_SelectWeaponByVersus)
	{
		WeaponTypeClass* pWeaponOdd = pThis->GetWeapon(gattlingStage << 1)->WeaponType;
		WeaponTypeClass* pWeaponEven = pThis->GetWeapon(gattlingStage << 1 | 1)->WeaponType;

		if (auto pTargetTechno = abstract_cast<TechnoClass*>(pTarget))
		{
			R->ESI(gattlingStage);

			auto pTargetExt = TechnoExt::ExtMap.Find(pTargetTechno);
			if (pTargetTechno->IsInAir())
			{
				if (!pWeaponEven->Projectile->AA)
					return Odd;

				if (GeneralUtils::GetWarheadVersusArmor(pWeaponEven->Warhead, pTargetExt->GetArmorIdx(pWeaponEven)) != 0.0)
					return Even;

				if (GeneralUtils::GetWarheadVersusArmor(pWeaponOdd->Warhead, pTargetExt->GetArmorIdx(pWeaponOdd)) == 0.0)
					return Even;

				return Odd;
			}
			else
			{
				if (!pWeaponOdd->Projectile->AG)
					return Even;

				if (GeneralUtils::GetWarheadVersusArmor(pWeaponOdd->Warhead, pTargetExt->GetArmorIdx(pWeaponOdd)) != 0.0)
					return Odd;

				if (GeneralUtils::GetWarheadVersusArmor(pWeaponEven->Warhead, pTargetExt->GetArmorIdx(pWeaponEven)) == 0.0)
					return Odd;

				return Even;
			}
		}
		else
		{
			if (auto pTargetObject = abstract_cast<ObjectClass*>(pTarget))
			{
				ObjectTypeClass* pTargetType = pTargetObject->GetType();

				if (GeneralUtils::GetWarheadVersusArmor(pWeaponOdd->Warhead, pTargetType->Armor) != 0.0)
					return Odd;

				if (GeneralUtils::GetWarheadVersusArmor(pWeaponEven->Warhead, pTargetType->Armor) == 0.0)
					return Odd;

				return Even;
			}
			else
			{
				return Odd;
			}
		}
	}

	return Origin;
}
```

Starkku:
Did some minor code tweaks and added `Land/NavalTargeting` support for `IsGattling` weapons in same manner as regular weapons.

> I think if gattling can select weapon by versus....
> Just a suggestion.

Current code already behaves largely the same as that, and not counting for Phobos-required checks and the Naval/LandTargeting things I just added, is also less verbose. Main difference is that the `AG=false` fallback from odd to even and armor type checks for non-techno targets does not exist, as they do not exist for regular weapons either. And the point was to make it work without separate toggle anyways, which I believe it currently does just fine.

Metadorius (approved):

NetsuNegi:
I found that if you ordered a boomber submarine (who has LandTargeting = 2) to attack a tree on land, it will try to use torpedo (primary weapon), I think it's incorrect, it should use missile(secondary weapon).

Starkku:
Fixed the vanilla bug of `LandTargeting=2` and/or `NavalTargeting=7` not changing weapon used against TerrainTypes on land to `Secondary`.

PR #747, opened 2022-09-09
Units controlled by human can auto target building with Non-zero `ThreatPosed`

ststl-s:
I think this is not good,  need suggestion.


chaserli:
I don't think it's good either, or I should say it's perhaps not the correct direction of extending the automatic target acquisition criterion.

ststl-s:
> I don't think it's good either, or I should say it's perhaps not the correct direction of extending the automatic target acquisition criterion.

This is just the most direct way for players' units to auto targetting buildings. Maybe has other way to achieve this. Now close this pr first.

brsajo:
I'm confused about what this pr is supposed to do but it would be cool if units could automatically target enemy structures with 0 threat when set to area guard mode (buildings could be excluded from this with a tag like ExcludeFromThreatScan).

PR #759, opened 2022-09-21
Custom SpyEffects

MortonPL:
### Spy Effects

- Additional espionage bonuses can be toggled with `SpyEffects.Custom`.
  - `SpyEffects.VictimSuperWeapon` instantly launches a Super Weapon for the owner of the infiltrated building at building's coordinates.
  - `SpyEffects.InfiltratorSuperWeapon` behaves the same as above, with the Super Weapon's owner being the owner of the spying unit.

In `rulesmd.ini`:
```ini
SpyEffects.Custom=false            ; boolean
SpyEffects.VictimSuperWeapon=      ; SuperWeaponType
SpyEffects.InfiltratorSuperWeapon= ; SuperWeaponType
```

```{warning}
The following Super Weapon types are supported:
`Type=LightningStorm`                     NO
`Type=MultiMissile`                       NO
`Type=PsychicDominator`                   NO
`Type=ChronoSphere`                       NO
`Type=ChronoWarp`                         NO
`Type=IronCurtain` and `Type=ForceShield` YES
`Type=GeneticConverter`                   YES
`Type=ParaDrop` and `Type=AmerParaDrop`   NO
`Type=SpyPlane`                           NO
`Type=PsychicReveal`                      YES
`Type=SonarPulse` (Ares)                  YES
`Type=GenericWarhead` (Ares)              YES
`Type=UnitDelivery` (Ares)                YES
`Type=Firestorm` (Ares)                   YES
`Type=HunterSeeker` (Ares)                YES
`Type=DropPod` (Ares)                     YES
`Type=EMPulse` (Ares)                     YES
`Type=Battery` (Ares)                     YES
```

### DIFF
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -275,6 +275,42 @@ PowerPlantEnhancer.Amount=0        ; integer
 PowerPlantEnhancer.Factor=1.0      ; floating point value
 ```
 
+### Spy Effects
+
+- Additional espionage bonuses can be toggled with `SpyEffects.Custom`.
+  - `SpyEffects.VictimSuperWeapon` instantly launches a Super Weapon for the owner of the infiltrated building at building's coordinates.
+  - `SpyEffects.InfiltratorSuperWeapon` behaves the same as above, with the Super Weapon's owner being the owner of the spying unit.
+
+In `rulesmd.ini`:
+```ini
+[SOMEBUILDING]                     ; BuildingType
+SpyEffects.Custom=false            ; boolean
+SpyEffects.VictimSuperWeapon=      ; SuperWeaponType
+SpyEffects.InfiltratorSuperWeapon= ; SuperWeaponType
+```
+
+```{warning}
+The following Super Weapon types are supported:
+`Type=LightningStorm`                     NO
+`Type=MultiMissile`                       NO
+`Type=PsychicDominator`                   NO
+`Type=ChronoSphere`                       NO
+`Type=ChronoWarp`                         NO
+`Type=IronCurtain` and `Type=ForceShield` YES
+`Type=GeneticConverter`                   YES
+`Type=ParaDrop` and `Type=AmerParaDrop`   NO
+`Type=SpyPlane`                           NO
+`Type=PsychicReveal`                      YES
+`Type=SonarPulse` (Ares)                  YES
+`Type=GenericWarhead` (Ares)              YES
+`Type=UnitDelivery` (Ares)                YES
+`Type=Firestorm` (Ares)                   YES
+`Type=HunterSeeker` (Ares)                YES
+`Type=DropPod` (Ares)                     YES
+`Type=EMPulse` (Ares)                     YES
+`Type=Battery` (Ares)                     YES
+```
+
 ## Infantry
 
 ### Customizable FLH When Infantry Is Prone Or Deployed
@@ -875,7 +911,28 @@ TransactMoney.Display.Offset=0,0     ; X,Y, pixels relative to default
 
 ```{note}
 For animation warheads/weapons to take effect, `Damage.DealtByInvoker` must be set.
-Also, due to the nature of some superweapon types, not all superweapons are suitable for launch.
+```
+
+```{warning}
+The following Super Weapon types are supported:
+`Type=LightningStorm`                     NO
+`Type=MultiMissile`                       NO
+`Type=PsychicDominator`                   NO
+`Type=ChronoSphere`                       NO
+`Type=ChronoWarp`                         NO
+`Type=IronCurtain` and `Type=ForceShield` YES
+`Type=GeneticConverter`                   YES
+`Type=ParaDrop` and `Type=AmerParaDrop`   NO
+`Type=SpyPlane`                           NO
+`Type=PsychicReveal`                      YES
+`Type=SonarPulse` (Ares)                  YES
+`Type=GenericWarhead` (Ares)              YES
+`Type=UnitDelivery` (Ares)                YES
+`Type=Firestorm` (Ares)                   YES
+`Type=HunterSeeker` (Ares)                YES
+`Type=DropPod` (Ares)                     YES
+`Type=EMPulse` (Ares)                     YES
+`Type=Battery` (Ares)                     YES
 ```
 
 In `rulesmd.ini`:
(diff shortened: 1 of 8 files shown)


src/Ext/Building/Hooks.cpp
```diff
@@ -256,3 +257,32 @@ DEFINE_HOOK(0x443CCA, BuildingClass_KickOutUnit_AircraftType_Phobos, 0xA)
 	HouseExt::ExtMap.Find(pHouse)->Factory_AircraftType = nullptr;
 	return 0;
 }
+
+// Note: We need another hook for when Phobos is run w/o Ares (or with Ares but with SpyEffect.Custom=no),
+// that will handle our logic earlier and disable the next hook by writing a very specific number to a free register.
+DEFINE_HOOK(0x45759D, BuildingClass_Infiltrate_Vanilla, 0x5)
+{
+	GET_STACK(HouseClass*, pInfiltratorHouse, STACK_OFFS(0x14, 0x4));
```
Starkku, line 265: Reminder to fix the deprecated `STACK_OFFS` macro when rebasing/merging `develop` before merging PR.

E.g `STACK_OFFS(0x14, 0x4)` -> `STACK_OFFSET(0x14, -0x4)`.

docs/New-or-Enhanced-Logics.md
```diff
@@ -274,6 +274,42 @@ PowerPlantEnhancer.Amount=0        ; integer
...
+`Type=SpyPlane`                           NO
+`Type=PsychicReveal`                      YES
+`Type=SonarPulse` (Ares)                  YES
+`Type=GenericWarhead` (Ares)              YES
+`Type=UnitDelivery` (Ares)                YES
+`Type=Firestorm` (Ares)                   YES
+`Type=HunterSeeker` (Ares)                YES
+`Type=DropPod` (Ares)                     YES
+`Type=EMPulse` (Ares)                     YES
+`Type=Battery` (Ares)                     YES
```
Starkku, line 310: Gonna have to rethink how to format this, line breaks are not currently preserved making it very hard to read.

image

docs/New-or-Enhanced-Logics.md
```diff
@@ -854,7 +890,28 @@ TransactMoney.Display.Offset=0,0     ; X,Y, pixels relative to default
...
```
Starkku, line 914: Ditto.

Starkku (changes requested):
Code looks okay to me. There's couple of small changes that should probably be done though.

docs/New-or-Enhanced-Logics.md
MortonPL: Would something like this be enough?
image

MortonPL: Update: the newest #760 commit makes all SW types supported, making this warning unnecessary. Removed.

src/Ext/Building/Hooks.cpp
```diff
@@ -257,3 +258,32 @@ DEFINE_HOOK(0x443CCA, BuildingClass_KickOutUnit_AircraftType_Phobos, 0xA)
...
+
+// Note: We need another hook for when Phobos is run w/o Ares (or with Ares but with SpyEffect.Custom=no),
+// that will handle our logic earlier and disable the next hook by writing a very specific number to a free register.
+DEFINE_HOOK(0x45759D, BuildingClass_Infiltrate_Vanilla, 0x5)
+{
+	GET_STACK(HouseClass*, pInfiltratorHouse, STACK_OFFSET(0x14, -0x4));
+	GET(BuildingClass*, pBuilding, EBP);
+
+	BuildingExt::HandleInfiltrate(pBuilding, pInfiltratorHouse);
+	R->EAX<int>(0x77777777);
```
Metadorius, line 270: What's that `0x77777777`?

MortonPL: See the comments for hooks.

Ares has a hook at `0x4571E0` (the beginning of BuildingClass::Infiltrate) and completely overwrites the function. Our logic has to be executed at the end (`0x4575A2`). The hook there assumes that registers have the exact content they had in the beginning (when Ares hook started, executed, and jumped) in order to work when Ares logic is used.

However, this will fail if Ares is *not* involved, because by the time we reach our hook, the registers will be different and we'll be reading garbage. That's why there is a second hook at `0x45759D`, which is only executed when Ares doesn't jump over this function. There, we execute our custom logic and then use EAX (which isn't used later, so it's safe to write to it) to "mark" that we're done with `0x77777777`. This way, when we reach the other hook, we check for this *very specific* value to prevent spy effects from happening twice.

The value itself doesn't matter, it just needs to be unique enough to not be accidentally produced by the game there.

Metadorius (approved):
Is there really a point in having a toggle? Otherwise LGTM.

docs/New-or-Enhanced-Logics.md
```diff
@@ -275,6 +275,20 @@ PowerPlantEnhancer.Amount=0        ; integer
...
+### Spy Effects
+
+- Additional espionage bonuses can be toggled with `SpyEffects.Custom`.
+  - `SpyEffects.VictimSuperWeapon` instantly launches a Super Weapon for the owner of the infiltrated building at building's coordinates.
+  - `SpyEffects.InfiltratorSuperWeapon` behaves the same as above, with the Super Weapon's owner being the owner of the spying unit.
+
+In `rulesmd.ini`:
+```ini
+[SOMEBUILDING]                     ; BuildingType
+SpyEffects.Custom=false            ; boolean
```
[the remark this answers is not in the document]
Metadorius: Disregard that, didn't know that's inherited from Ares...

src/Ext/Building/Body.cpp
```diff
@@ -301,6 +301,32 @@ void BuildingExt::ExtData::ApplyPoweredKillSpawns()
...
+	if (!pTypeExt->SpyEffect_Custom)
+		return false;
+
+	auto pVictimHouse = pBuilding->Owner;
+	if (pInfiltratorHouse != pVictimHouse)
+	{
+		if (pTypeExt->SpyEffect_VictimSuperWeapon)
+		{
+			const auto pSuper = pVictimHouse->Supers.GetItem(SuperWeaponTypeClass::Array->FindItemIndex(pTypeExt->SpyEffect_VictimSuperWeapon));
+			pSuper->Launch(CellClass::Coord2Cell(pBuilding->Location), pVictimHouse->IsControlledByHuman());
```
chaserli, line 317: Did you even test this? How could it work? Did you only test with charged SW?
I assume you were not launching for real, since otherwise normal click fire would suffice. So the SWs you wanted to fire were most likely virtually charged and not manually launchable.

chaserli (commented):
ehhhhhh

PR #760, opened 2022-09-21
Super Weapons launching Super Weapons, SWTypeExt improvements

MortonPL:
### Next

Superweapons can now launch other superweapons at the same target. Launched types can be additionally randomized using the same rules as with LimboDelivery (see above).
  - `SW.Next.RealLaunch` controls whether the owner who fired the initial superweapon must own all listed superweapons and sufficient funds to support `Money.Amout`. Otherwise they will be launched forcibly.
  - `SW.Next.IgnoreInhibitors` ignores `SW.Inhibitors`/`SW.AnyInhibitor` of each superweapon, otherwise only non-inhibited superweapons are launched.
  - `SW.Next.IgnoreDesignators` ignores `SW.Designators`/`SW.AnyDesignator` respectively.

In `rulesmd.ini`:
```ini
[SOMESW]                        ; Super Weapon
SW.Next=                        ; List of Super Weapons
SW.Next.RealLaunch=true         ; boolean
SW.Next.IgnoreInhibitors=false  ; boolean
SW.Next.IgnoreDesignators=true  ; boolean
SW.Next.RollChances=            ; List of percentages.
SW.Next.RandomWeightsN=         ; List of integers.
```

Additionally, all Phobos SW features should now support every SW `Type`.

### DIFF
--- a/src/Ext/SWType/Body.cpp
+++ b/src/Ext/SWType/Body.cpp
@@ -101,13 +101,9 @@ void SWTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 	if (weights.size())
 	{
 		if (this->LimboDelivery_RandomWeightsData.size())
-		{
 			this->LimboDelivery_RandomWeightsData[0] = weights;
-		}
 		else
-		{
 			this->LimboDelivery_RandomWeightsData.push_back(weights);
-		}
 	}
 
 	// SW.Next.RandomWeights
@@ -127,13 +123,9 @@ void SWTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 	if (weights2.size())
 	{
 		if (this->SW_Next_RandomWeightsData.size())
-		{
 			this->SW_Next_RandomWeightsData[0] = weights2;
-		}
 		else
-		{
 			this->SW_Next_RandomWeightsData.push_back(weights2);
-		}
 	}
 	this->Detonate_Warhead.Read(exINI, pSection, "Detonate.Warhead");
 	this->Detonate_Weapon.Read(exINI, pSection, "Detonate.Weapon", true);
--- a/src/Ext/SWType/Body.cpp
+++ b/src/Ext/SWType/Body.cpp
@@ -36,6 +36,10 @@ void SWTypeExt::ExtData::Serialize(T& Stm)
 		.Process(this->Detonate_Warhead)
 		.Process(this->Detonate_Weapon)
 		.Process(this->Detonate_Damage)
+		.Process(this->SW_Next)
+		.Process(this->SW_Next_IgnoreInhibitors)
+		.Process(this->SW_Next_RandomWeightsData)
+		.Process(this->SW_Next_RollChances)
 		;
 }
 
@@ -69,8 +73,14 @@ void SWTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 	this->LimboDelivery_Types.Read(exINI, pSection, "LimboDelivery.Types");
 	this->LimboDelivery_IDs.Read(exINI, pSection, "LimboDelivery.IDs");
 	this->LimboDelivery_RollChances.Read(exINI, pSection, "LimboDelivery.RollChances");
+	this->LimboKill_Affected.Read(exINI, pSection, "LimboKill.Affected");
+	this->LimboKill_IDs.Read(exINI, pSection, "LimboKill.IDs");
+	this->SW_Next.Read(exINI, pSection, "SW.Next");
+	this->SW_Next_IgnoreInhibitors.Read(exINI, pSection, "SW.Next.IgnoreInhibitors");
+	this->SW_Next_RollChances.Read(exINI, pSection, "SW.Next.RollChances");
 
 	char tempBuffer[32];
+	// LimboDelivery.RandomWeights
 	for (size_t i = 0; ; ++i)
 	{
 		ValueableVector<int> weights;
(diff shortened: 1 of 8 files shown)


src/Ext/SWType/Body.cpp
```diff
@@ -51,8 +57,16 @@ void SWTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI) {
 	this->LimboDelivery_Types.Read(exINI, pSection, "LimboDelivery.Types");
 	this->LimboDelivery_IDs.Read(exINI, pSection, "LimboDelivery.IDs");
 	this->LimboDelivery_RollChances.Read(exINI, pSection, "LimboDelivery.RollChances");
+	this->LimboKill_Affected.Read(exINI, pSection, "LimboKill.Affected");
+	this->LimboKill_IDs.Read(exINI, pSection, "LimboKill.IDs");
+	this->SW_Next.Read(exINI, pSection, "SW.Next");
+	this->SW_Next_RealLaunch.Read(exINI, pSection, "SW.Next.RealLaunch");
+	this->SW_Next_IgnoreInhibitors.Read(exINI, pSection, "SW.Next.IgnoreInhibitors");
```
chaserli, line 64: I don't see under which circumstances  should these be ignored

src/Ext/WarheadType/Detonate.cpp
```diff
@@ -80,14 +80,14 @@ void WarheadTypeExt::ExtData::Detonate(TechnoClass* pOwner, HouseClass* pHouse,
...
-				if ((pSWExt && pSuper->IsCharged && pHouse->CanTransactMoney(pSWExt->Money_Amount)) || !this->LaunchSW_RealLaunch)
+				if (!this->LaunchSW_RealLaunch || (pSWExt && pSuper->IsCharged && pHouse->CanTransactMoney(pSWExt->Money_Amount)))
 				{
 					if (this->LaunchSW_IgnoreInhibitors || !pSWExt->HasInhibitor(pHouse, cell)
 					&& (this->LaunchSW_IgnoreDesignators || pSWExt->HasDesignator(pHouse, cell)))
 					{
-						pSuper->SetReadiness(true);
 						pSuper->Launch(cell, true);
-						pSuper->Reset();
+						if (this->LaunchSW_RealLaunch)
```
chaserli, line 89: I'm not quite sure, is it safe?

src/Ext/SWType/FireSuperWeapon.cpp
```diff
@@ -115,60 +115,94 @@ void SWTypeExt::FireSuperWeaponExt(SuperClass* pSW, const CellStruct& cell)
...
+	return indices;
+}
 
-		for (size_t i = 0; i < rolls; i++)
-		{
-			this->RandomBuffer = ScenarioClass::Instance->Random.RandomDouble();
-			if (!rollOnce && this->RandomBuffer > this->LimboDelivery_RollChances[i])
-				continue;
+// SW.Next proper launching mechanic
+void Launch(HouseClass* pHouse, SWTypeExt::ExtData* pLauncherTypeExt, SuperWeaponTypeClass* pLaunchedType, const CellStruct& cell)
```
chaserli, line 159: The name confused me for a second at first glimpse

src/Ext/SWType/FireSuperWeapon.cpp
```diff
@@ -115,60 +115,94 @@ void SWTypeExt::FireSuperWeaponExt(SuperClass* pSW, const CellStruct& cell)
...
-		size_t ids = this->LimboDelivery_IDs.size();
-
+		size_t idsSize = this->LimboDelivery_IDs.size();
 		for (size_t i = 0; i < this->LimboDelivery_Types.size(); i++)
 		{
-			if (i < ids)
+			if (i < idsSize)
 				id = this->LimboDelivery_IDs[i];
 
 			LimboCreate(abstract_cast<BuildingTypeClass*>(this->LimboDelivery_Types[i]), pHouse, id);
```
chaserli, line 208: I don't think casting is needed here since you already changed the vector type

chaserli (changes requested):
Speaking of the discussion of SW handled by Ares/Phobos, I think maybe you can add a check to see if it's handled by Ares or not at 0x6CC390, since it will be executed after Ares' hook. And if Ares didn't handle it, you do `SWTypeExt::FireSuperWeaponExt(pSuper, *pCell)` at 0x6CC390, otherwise at 0x6CDE40. Will this work?

src/Ext/WarheadType/Detonate.cpp
MortonPL: `SuperClass::Reset()` only resets timer / charge and, according to the decompile, "animation", though I haven't observed anything about the latter.
If the user doesn't want a "real" launch, the charge imo should not be reset.

src/Ext/SWType/Body.cpp
MortonPL: Warhead `LaunchSW` also has `IgnoreInhibitors`. That, and we already allow the user to ignore designators, so inhibitors shouldn't be an exception.

src/Ext/SWType/FireSuperWeapon.cpp
MortonPL: True, this will be corrected.

src/Ext/SWType/FireSuperWeapon.cpp
MortonPL: Why? Both Ares and Phobos use "launch" and "fire" interchangeably when describing SW usage (i.e. already existing `LaunchSW`).

src/Ext/SWType/Hooks.cpp
```diff
@@ -2,8 +2,10 @@
 
 #include <SuperClass.h>
 
-//Ares hooked from 0x6CC390 and jumped to this offset
-DEFINE_HOOK(0x6CDE40, SuperClass_Place_FireExt, 0x3)
+//Ares hooked at 0x6CC390 and jumped to 0x6CDE40
+// If a super is not handled by Ares however, we do it at the original entry point
+DEFINE_HOOK_AGAIN(0x6CC390, SuperClass_Place_FireExt, 0x6)
```
chaserli, line 7:
```suggestion
bool SWLaunchHandledByAres=true;
DEFINE_HOOK(0x6CC390, SuperClass_Place_FireExt, 0x6)
{
    GET(SuperClass* const, pSuper, ECX);
	GET_STACK(CellStruct const* const, pCell, 0x4);
	SWLaunchHandledByAres=false;
	SWTypeExt::FireSuperWeaponExt(pSuper, *pCell);
	return 0;
}

DEFINE_HOOK(0x6CDE40, SuperClass_Place_FireExt, 0x4)
{
	GET(SuperClass* const, pSuper, ECX);
	GET_STACK(CellStruct const* const, pCell, 0x4);
	// GET_STACK(bool const, isPlayer, 0x8);
	if(!SWLaunchHandledByAres)
	    SWTypeExt::FireSuperWeaponExt(pSuper, *pCell);
	return 0;
}
```
I was thinking about this, otherwise you'll fire your ext twice, no?

MortonPL: SW is fired once (tested with all types), there is no need for an additional check. 
As for hook size, there's not much choice, as it's followed by a byte of padding and a jump table.

src/Ext/SWType/Body.cpp
chaserli: Well in that case it would be better to set both default value to flase imo

src/Ext/SWType/Body.h
```diff
@@ -37,12 +37,18 @@ class SWTypeExt
 		Valueable<AffectedHouse> LimboKill_Affected;
 		ValueableVector<int> LimboKill_IDs;
 		Valueable<double> RandomBuffer;
+		ValueableVector<SuperWeaponTypeClass*> SW_Next;
```
chaserli, line 40: I changed my `LaunchSW` to `ValueableIdxVector<SuperWeaponTypeClass>` since Ares used this one too for BuildingType's SuperWeapons, and I don't quite understand why Phobos still use DVC for that one.

It's up to you, I don't think it makes too much difference other than avoiding a search through all SWTypes

src/Ext/SWType/Hooks.cpp
chaserli: If it doesn't pose any problem then it should be fine, perhaps

src/Ext/SWType/FireSuperWeapon.cpp
chaserli: I mean, exactly because of that it made me believe this `Launch` function was a routine from vanilla code.

src/Ext/SWType/Body.h
```diff
@@ -37,12 +37,18 @@ class SWTypeExt
...
+		Valueable<bool> SW_Next_IgnoreInhibitors;
+		Valueable<bool> SW_Next_IgnoreDesignators;
+		ValueableVector<float> SW_Next_RollChances;
 
 		Nullable<WarheadTypeClass*> Detonate_Warhead;
 		Nullable<WeaponTypeClass*> Detonate_Weapon;
 		Nullable<int> Detonate_Damage;
 
 		ValueableVector<ValueableVector<int>> LimboDelivery_RandomWeightsData;
+		ValueableVector<ValueableVector<int>> SW_Next_RandomWeightsData;
```
chaserli, line 51: `std::vector<ValueableVector<`?

chaserli (changes requested):

PR #764, opened 2022-09-29
RadialIndicator visible to observer/allies & Different PowerDelta label when power sabotaged

chaserli:
In vanilla game, when the player selects a defense building, the radial indicator is shown only if it belongs to the player. Now the indicator is also drawn for observers. For other houses it can be specified by

In `rulesmd.ini`
```ini
[AudioVisual]
RadialIndicatorVisibility=allies  ; list of Affected House Enumeration (allies/ally | enemies/enemy | all)
```

When your power is sabotaged by a spy, the number given by the power delta counter may look confusing to players. Now I replaced it by ⚡❌, which can also be replaced by

In `uimd.ini`
```ini
[ToolTips]
PowerBlackoutLabel=  ; CSF entry key
```
If `[GlobalControls]->DebugKeysEnabled=yes` power delta is still displayed

### DIFF
--- a/src/Misc/Hooks.UI.cpp
+++ b/src/Misc/Hooks.UI.cpp
@@ -77,7 +77,7 @@ DEFINE_HOOK(0x641EE0, PreviewClass_ReadPreview, 0x6)
 DEFINE_HOOK(0x4A25E0, CreditsClass_GraphicLogic_HarvesterCounter, 0x7)
 {
 	auto const pPlayer = HouseClass::CurrentPlayer();
-	if (!pPlayer || pPlayer->Defeated)
+	if (pPlayer->Defeated)
 		return 0;
 
 	if (Phobos::UI::ShowHarvesterCounter)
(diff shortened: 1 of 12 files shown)


Starkku:
Code looks alright to me. The visibility control could also be implemented at per-building level with it defaulting to the global setting but I am unsure if there would be need for that degree of customization.

Starkku (approved):
Code looks alright to me. The visibility control could also be implemented at per-building level with it defaulting to the global setting but I am unsure if there would be need for that degree of customization.

FS-21:
> Code looks alright to me. The visibility control could also be implemented at per-building level with it defaulting to the global setting but I am unsure if there would be need for that degree of customization.

The per-building customization can be used for example for showing the sensors range in defenses capable to detect stealth diguised units.

src/Phobos.cpp
```diff
@@ -227,6 +228,9 @@ DEFINE_HOOK(0x5FACDF, OptionsClass_LoadSettings_LoadPhobosSettings, 0x5)
 		pINI_UIMD->ReadString(TOOLTIPS_SECTION, "PowerLabel", NONE_STR, Phobos::readBuffer);
 		Phobos::UI::PowerLabel = GeneralUtils::LoadStringOrDefault(Phobos::readBuffer, L"\u26a1"); // ⚡
 
+		pINI_UIMD->ReadString(TOOLTIPS_SECTION, "PowerBlackoutLabel", NONE_STR, Phobos::readBuffer);
+		Phobos::UI::PowerBlackoutLabel = GeneralUtils::LoadStringOrDefault(Phobos::readBuffer, L"\u26a1\u2704"); // ⚡✄
```
chaserli, line 232: I need a better default symbol for this one. My hand drawing is poor so I cannot add this one to the font yet

src/Misc/Hooks.UI.cpp
```diff
@@ -198,3 +201,16 @@ DEFINE_HOOK(0x6D4684, TacticalClass_Draw_FlyingStrings, 0x6)
...
 }
+
+DEFINE_HOOK(0x456776, BuildingClass_DrawRadialIndicator_Visibility, 0x6)
+{
+	enum { ContinueDraw = 0x456789, DoNotDraw = 0x456962 };
+	GET(BuildingClass* const, pThis, ESI);
+
+	if (HouseClass::IsCurrentPlayerObserver()
+		|| pThis->Owner->IsControlledByCurrentPlayer()
+		|| EnumFunctions::CanTargetHouse(RulesExt::Global()->RadialIndicatorVisibility, pThis->Owner, HouseClass::CurrentPlayer))
```
chaserli, line 212: I don't think per BuildingType config is necessary.  You can differentiate with `ConcentricRadialIndicator` for example

src/Phobos.cpp
chaserli: So how about disabling power delta display during power blackout?

Starkku: Don't know about hiding the power delta display. Color change would probably be better, defaults to something like gray.

PR #768, opened 2022-10-02
Delayed fire animation

FS-21:
Created for vehicles in mind but since it is a TechnoClass feature maybe works with other technos.

- Now a weapon attack can be delayed if an animation is set in `DelayedFire.Anim`.
- Optionally `DelayedFire.PostAnim` replaces `DelayedFire.Anim` when the weapon is fired.
- By default the animation will be placed in the weapon FLH location but the animation can be moved to the center of the object with `DelayedFire.Anim.UseFLH=false`.
- `DelayedFire.Duration` > 0 will interrupt the animation (and allow weapon fire) when the countdown ends.
- `DelayedFire.Suicide` will provoke the same effect of the original tag `Suicide` without the side effect explosion when the unit is killed by something.

In `rulesmd.ini`:

DelayedFire.Anim=            ; Animation
DelayedFire.PostAnim=        ; Animation
DelayedFire.Anim.UseFLH=true ; boolean - disabling this will create the animation at the center of the unit instead of its FLH coordinates
DelayedFire.Duration=0       ; integer - in-game frames.
DelayedFire.Suicide=false    ; boolean

## Summary by CodeRabbit

- **New Features**
  - Introduced a delayed fire animation feature, allowing weapon attacks to be delayed by setting an animation.
- **Bug Fixes**
  - Various fixes and improvements, including toggling Infantry/UnitsGainSelfHeal for MultiplayPassive=true houses, customizable straight trajectory detonation and snap distance with pass-through option, and adjustments to airstrike and spy plane spawn distance and height.
  - Fixed an issue that allowed AI to repair structures built from base nodes, trigger action 125, and SW delivery in single-player missions.
- **Documentation**
  - Updated documentation to include the new delayed fire animation feature and various fixes and improvements.


FS-21:
delayedfireweapon

; Example:
[Comet]
Damage=100
ROF=100
Range=10
Projectile=LargeCometP
Speed=40
Report=PrismTankAttack
Warhead=CometWH
Bright=yes
IsHouseColor=true
LaserOuterSpread= 0,0,0
LaserDuration = 15
IsLaser=true	; this flag tells the game to use the special laser draw effect
DelayedFire.Anim=PRISMCHARGE    ; put HERE your anim that will be used as delayer
DelayedFire.Anim.LoopCount=3

src/Ext/Techno/Hooks.cpp
```diff
@@ -45,6 +45,17 @@ DEFINE_HOOK(0x6F9E50, TechnoClass_AI, 0x5)
 
 	}
 
+	if (pExt->DelayedFire_Anim && !pThis->Target && pThis->GetCurrentMission() != Mission::Attack)
+	{
+		CDTimerClass* weaponReadyness = (CDTimerClass*)pThis->__DiskLaserTimer;
+		weaponReadyness->Start(pThis->DiskLaserTimer.GetTimeLeft() + 5);
```
MortonPL, line 51: What's that +5 for?

MortonPL (changes requested):
I left some fixes/suggestions for documentation and code.

src/Ext/Techno/Hooks.cpp
FS-21: Isn't the right question, the thing is for what is that "weaponReadyness" if it wasn't stored?
I think that declaration was a leftover of experiments :-S

FS-21:
With a bit of delay I applied the last feedback...

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -794,3 +794,112 @@ DEFINE_HOOK(0x5223B3, InfantryClass_Approach_Target_DeployFireWeapon, 0x6)
...
+
+	if (pWeaponTypeExt->DelayedFire_Anim_LoopCount <= 0)
+		return continueFireAt;
+
+	if (!pWeaponTypeExt->DelayedFire_Anim.isset())
+		return continueFireAt;
+
+	auto pExt = TechnoExt::ExtMap.Find(pThis);
+	if (!pExt)
+		return continueFireAt;
```
Starkku, line 820: These first three checks could probably be consolidated to a single `if` statement. Also there should be a space between other lines and `if`.

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -794,3 +794,112 @@ DEFINE_HOOK(0x5223B3, InfantryClass_Approach_Target_DeployFireWeapon, 0x6)
...
+	if (!pDelayedFireAnimType)
+	{
+		pExt->DelayedFire_Anim = nullptr;
+		pExt->DelayedFire_Anim_LoopCount = 0;
+		pExt->DelayedFire_Duration = -1;
+
+		return continueFireAt;
+	}
+
+	bool hasDeployAnimFinished = pExt->DelayedFire_Anim && (pExt->DelayedFire_Anim->Animation.Value >= pDelayedFireAnimType->End + pDelayedFireAnimType->Start - 1);
```
Starkku, line 836: Should probably be DelayAnim, not DeployAnim.

Starkku (changes requested):
There's a bunch of things that probably need looking into here.

1. The big one is the basic design of this system. Having it so that the delay is tied to the length of the animation with bespoke loop controls and an additional option to cap the delay is kind of backwards design. The delay and the animation should be separate, with the delay timer running exactly for the number of frames specified and animation duration, looping etc. left up to the modder.

2. The delay is currently only reset if the firer goes idle. If you swap weapons, which mind you could both have different `DelayedFire` settings without dropping attack mission the delay does not get reset. At the very least this should be an option that probably should default to true.

3. Another issue with the basic design, it doesn't currently play well with infantry at all. Due to being hooked at `FireAt`, the delay is only processed when that function is entered and on infantry there is already an delay determined by the firing sequence, which stacks up with this one. May be possible to salvage in the current implementation by adding some bespoke checks in infantry firing AI function that could also be used to make the sequence work nicely together with this, all of which could possibly also be looked at separately after this but figured it 

4. `DelayedFire.Anim.UseFLH` set to true does not currently produce accurate results. This is probably because you're using weapon index from `pThis->CurrentWeaponNumber` to fetch the FLH and that's only used for IFV/Gattling stuff. Get the actual weapon index from `FireAt` params from stack or something instead.

5. What exactly is the distinction between just using `Suicide` and `DelayedFire.Suicide`?, the documentation isn't very clear on this. In my tests the only thing I was able to figure out is that latter still fires off `FeedbackWeapon`.

6. Attaching the animation to the firer (`SetOwnerObject()`) should probably be toggleable, just in case. In majority of cases you'd want it on but there might be some cases where you do not.

Also some minor code nitpicks that I commented on separately.

FS-21:
> There's a bunch of things that probably need looking into here.
> 
> 1. The big one is the basic design of this system. Having it so that the delay is tied to the length of the animation with bespoke loop controls and an additional option to cap the delay is kind of backwards design. The delay and the animation should be separate, with the delay timer running exactly for the number of frames specified and animation duration, looping etc. left up to the modder.

Now the loops control is managed by artmd.ini
"DelayedFire.Duration" forces the end of the animation and fires the weapon. basically is how a vanilla YR tag works.

> 2. The delay is currently only reset if the firer goes idle. If you swap weapons, which mind you could both have different `DelayedFire` settings without dropping attack mission the delay does not get reset. At the very least this should be an option that probably should default to true.

Addressed in the new implementation :-)

> 3. Another issue with the basic design, it doesn't currently play well with infantry at all. Due to being hooked at `FireAt`, the delay is only processed when that function is entered and on infantry there is already an delay determined by the firing sequence, which stacks up with this one. May be possible to salvage in the current implementation by adding some bespoke checks in infantry firing AI function that could also be used to make the sequence work nicely together with this, all of which could possibly also be looked at separately after this but figured it should be mentioned anyway. And mentioned in docs too if it gets merged in this state.

I think currently works, but in this example the sequence is too short:
DelayedFire_charge_anim_test

> 4. `DelayedFire.Anim.UseFLH` set to true does not currently produce accurate results. This is probably because you're using weapon index from `pThis->CurrentWeaponNumber` to fetch the FLH and that's only used for IFV/Gattling stuff. Get the actual weapon index from `FireAt` params from stack or something instead.

Probably addressed in the new version.

> 5. What exactly is the distinction between just using `Suicide` and `DelayedFire.Suicide`? The documentation isn't very clear on this. In my tests the only thing I was able to figure out is that latter still fires off `FeedbackWeapon`.
> 6. Attaching the animation to the firer (`SetOwnerObject()`) should probably be toggleable, just in case. In majority of cases you'd want it on but there might be some cases where you do not.
> 
> Also some minor code nitpicks that I commented on separately.
If the unit  with "Suicide=yes" is killed it explodes and the new sub-tag not.

FS-21:
Temporally closed due to a desync. Probably I have to rethink this PR (feel free to take code from something else if is useful).

PR #769, opened 2022-10-03
Override Target under EMP attack behavior

FS-21:
Overrides a part of the vanilla YR logic for allowing units to use a different weapon if the target is under EMP effect.

In `rulesmd.ini`:

ForceWeapon.UnderEMP=-1  ; integer. 0 for primary weapon, 1 for secondary weapon, -1 to disable


FS-21:
forceweapon_emp

docs/Whats-New.md
```diff
@@ -339,6 +339,7 @@ New:
 - Implemented support for PCX images for observer loading screen (by Uranusian)
 - Animated (non-tiberium spawning) TerrainTypes (by Starkku)
 - Toggleable passenger killing for Explodes=true units (by Starkku)
+- Override target under EMP attack behavior (By FS-21)
```
MortonPL, line 342: Should be moved from 0.3 to 0.3.1.

MortonPL (commented):
Looks okay to me, nothing major. Out of curiosity - would this work on `Gunner` units and weapon indices past 1?

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -80,6 +80,17 @@ DEFINE_HOOK(0x6F3428, TechnoClass_WhatWeaponShouldIUse_ForceWeapon, 0x6)
 				R->EAX(pTechnoTypeExt->ForceWeapon_Naval_Decloaked);
 				return 0x6F37AF;
 			}
+
+			if (pTechnoTypeExt->ForceWeapon_UnderEMP >= 0
+				&& (pTarget->IsUnderEMP()
+					|| (pTechnoTypeExt->ImmuneToEMP.isset() && pTechnoTypeExt->ImmuneToEMP.Get() == true)
+					|| (!pTechnoTypeExt->ImmuneToEMP.isset() &&
+						((pTarget->WhatAmI() == AbstractType::Infantry && !static_cast<InfantryTypeClass*>(pTargetType)->Cyborg) ||
+							((pTarget->WhatAmI() == AbstractType::Unit || pTarget->WhatAmI() == AbstractType::Aircraft) && pTargetType->Organic)))))
```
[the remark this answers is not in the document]
FS-21: I never tried because that wasn't the scope of the feature sorry.

Coronia:
I'm not sure if we should also force weapon for targets that immune to EMP here. Maybe split that to another standalone weapon filter?

FS-21:
> I'm not sure if we should also force weapon for targets that immune to EMP here. Maybe split that to another standalone weapon filter?

I agree, I just refactorized the code a bit (is a small feature but is more readable now) and I feel the EMP immunity (and the other opposite conditions) can be a standalone case.

mevitar:
Is it really needed when AttachEffect.RequiredTypes= exists? Just put an AE on the EMPed target.

Still, i tested it, seems to be working fine. Didn't notice any conflicts with other targeting override mechanisms (like the mentioned AttachEffect.RequiredTypes= ), and they can even go in hand with them (like using NoAmmo= normally to make the unit switch to secondary when runs out of ammo, but it will still switch to secondary against the EMPed target even on full ammo if ForceWeapon.UnderEMP=1 is present).

docs/New-or-Enhanced-Logics.md
```diff
@@ -1059,6 +1059,19 @@ ForceWeapon.Cloaked=-1          ; integer. 0 for primary weapon, 1 for secondary
...
+*Enemy behavior against EMP targets in [C&C: Reloaded](https://www.moddb.com/mods/cncreloaded)*
+
+- Overrides a part of the vanilla YR logic for allowing units to use a different weapon if the target is under EMP effect.
+
+In `rulesmd.ini`:
+```ini
+[SOMETECHNO]             ; TechnoType
+ForceWeapon.UnderEMP=-1  ; integer. 0 for primary weapon, 1 for secondary weapon, -1 to disable
+```
+
```
Coronia, line 1074: should just move the document to the same place of other ForceWeapon tags

Coronia (changes requested):

TaranDahl (approved):
lgtm

CrimRecya (approved):
Simple and great.

PR #783, opened 2022-10-27
Allow UnitTypes to use `AlternateFLH` in opentopped transport

chaserli:
Units will use `AlternateFLH%d` as the FLH in opentopped transport.
Units will use infantries' FLH logic from now on
Therefore I Need testing of whether :
1. All original FLHs are correct (units not in opentopped transport, firing under all circumstances)
2. AlternateFLHs are respected in opentopped ones, under all circumstances ( land, hover, jumpjet...)
3. Any exception or restriction exist
4. Any performance issue exist (unlikely)

### DIFF
--- a/docs/Whats-New.md
+++ b/docs/Whats-New.md
@@ -253,6 +253,7 @@ New:
 - Forcing specific weapon against cloaked or disguised targets (by Starkku)
 - Customizable ROF random delay (by Starkku)
 - Animation with `Tiled=yes` now supports `CustomPalette` (by ststl)
+- Shields can inherit Techno ArmorType (by Starkku)
 
 Vanilla fixes:
 - Allow AI to repair structures built from base nodes/trigger action 125/SW delivery in single player missions (by Trsdy)
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -111,6 +111,7 @@ Pips.Shield.Building.Empty=0       ; integer, frame of pips.shp (zero-based) for
 Strength=0                           ; integer
 InitialStrength=0                    ; integer
 Armor=none                           ; ArmorType
+InheritArmorFromTechno=false         ; boolean
 Powered=false                        ; boolean
 AbsorbOverDamage=false               ; boolean
 SelfHealing=0.0                      ; double, percents or absolute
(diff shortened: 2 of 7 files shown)


CREDITS.md
```diff
@@ -189,6 +189,7 @@ This page lists all the individual contributions to the project by their author.
    - Object Self-destruction logic
    - Post-type-conversion update
    - Building EVA_StructureSold and SellSound dehardcode
+   - `AlternateFLH` of vehicles in opentopped transport.
```
Starkku, line 192: Capitalize `OpenTopped` since it refers to INI key.

docs/Fixed-or-Improved-Logics.md
```diff
@@ -26,7 +26,7 @@ This page describes all ingame logics that are fixed or improved in Phobos witho
 - Fixed interaction of `UnitAbsorb` & `InfantryAbsorb` with `Grinding` buildings. The keys will now make the building only accept appropriate types of objects.
 - Fixed missing 'no enter' cursor for VehicleTypes being unable to enter a `Grinding` building.
 - Fixed Engineers being able to enter `Grinding` buildings even when they shouldn't (such as ally building at full HP).
-
+- Allowed usage of `AlternateFLH%d` of vehicles in opentopped transport.
```
Starkku, line 29: Capitalize `OpenTopped` since it refers to INI key.

docs/Whats-New.md
```diff
@@ -257,6 +257,7 @@ New:
 
 Vanilla fixes:
 - Allow AI to repair structures built from base nodes/trigger action 125/SW delivery in single player missions (by Trsdy)
+- Allowed usage of `AlternateFLH%d` of vehicles in opentopped transport. (by Trsdy)
```
Starkku, line 260: Capitalize `OpenTopped` since it refers to INI key.

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -409,3 +409,10 @@ DEFINE_HOOK(0x6FF4CC, TechnoClass_FireAt_ToggleLaserWeaponIndex, 0x6)
 
 	return 0;
 }
+
+
+// Feature: Allow Units using AlternateFLHs - by Trsdy
+// I don't want to rewrite something new, so I use the Infantry one directly
+// afaik it didn't check infantry-specific stuff here
+// and neither Ares nor Phobos messed up with it so far, even that crawling flh one was in TechnoClass
+DEFINE_JUMP(VTABLE, 0x7F5D20, 0x523250); // Redirect UnitClass::GetFLH to InfantryClass::GetFLH (used to be TechnoClass::GetFLH)
```
Starkku, line 418: This could potentially backfire if anyone adds bespoke logic to InfantryClass regarding this in the future. A pretty big if so I guess it is fine for time being, alternate way would've been to do it like the health-related checks in `Techno\Hooks.Shield.cpp` with the wrappers, in which you'd do something similar to `InfantryClass::GetFLH()`.

Starkku (changes requested):
Some minor suggestions regarding spelling and a thought about the way this is implemented.

By the looks of this, this is just a simple vtable function call swap and doesn't interfere with #784 at all, and would in fact work in tandem with it (regardless of which one gets merged first) but that would probably have to be tested anyways in the future.

src/Ext/Techno/Hooks.Firing.cpp
chaserli: In case that happens in the future, we can revert this and do as you put it.
As for that AlternateFLH%d>5 PR, it was in `TechnoClass::GetFLH` anyway so it doesn't actually interfere with this one imo

PR #787, opened 2022-10-30
[Minor] InterceptorType

ststl-s:
As Trsdy's commet of #726

> I didn't manage to give my review on this in time. Basically you give too many new arguments on this specific feature, which makes the code looks _extrêmement lourde_, in this case I would rather suggest you guys wrap those into a new entity, like the AttachEffectType or PrismForwarding in Ares, and see if it can work like an AttachEffect. Moreover, it looks like a standalone trigger-action system already, so maybe not just limited to self-destruction.
> 
> Same suggestion for other similar tags like the interceptor one> I didn't manage to give my review on this in time. Basically you give too many new arguments on this specific feature, which makes the code looks _extrêmement lourde_, in this case I would rather suggest you guys wrap those into a new entity, like the AttachEffectType or PrismForwarding in Ares, and see if it can work like an AttachEffect. Moreover, it looks like a standalone trigger-action system already, so maybe not just limited to self-destruction.
> 
> Same suggestion for other similar tags like the interceptor one

### DIFF
--- a/src/Ext/TechnoType/Body.cpp
+++ b/src/Ext/TechnoType/Body.cpp
@@ -110,16 +110,21 @@ void TechnoTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 	this->UIDescription.Read(exINI, pSection, "UIDescription");
 	this->LowSelectionPriority.Read(exINI, pSection, "LowSelectionPriority");
 	this->MindControlRangeLimit.Read(exINI, pSection, "MindControlRangeLimit");
-	this->Interceptor.Read(exINI, pSection, "Interceptor");
-	this->Interceptor_CanTargetHouses.Read(exINI, pSection, "Interceptor.CanTargetHouses");
-	this->Interceptor_GuardRange.Read(exINI, pSection, "Interceptor.%sGuardRange");
-	this->Interceptor_MinimumGuardRange.Read(exINI, pSection, "Interceptor.%sMinimumGuardRange");
-	this->Interceptor_Weapon.Read(exINI, pSection, "Interceptor.Weapon");
-	this->Interceptor_DeleteOnIntercept.Read(exINI, pSection, "Interceptor.DeleteOnIntercept");
-	this->Interceptor_WeaponOverride.Read(exINI, pSection, "Interceptor.WeaponOverride", true);
-	this->Interceptor_WeaponReplaceProjectile.Read(exINI, pSection, "Interceptor.WeaponReplaceProjectile");
-	this->Interceptor_WeaponCumulativeDamage.Read(exINI, pSection, "Interceptor.WeaponCumulativeDamage");
-	this->Interceptor_KeepIntact.Read(exINI, pSection, "Interceptor.KeepIntact");
+
+	Nullable<bool> isInterceptor;
+	isInterceptor.Read(exINI, pSection, "Interceptor");
+
+	if (isInterceptor)
+	{
+		if (this->InterceptorType == nullptr)
+			this->InterceptorType = std::make_unique<InterceptorTypeClass>(this->OwnerObject());
+
+		this->InterceptorType->LoadFromINI(pINI, pSection);
+	}
+	else if (isInterceptor.isset())
+	{
+		this->InterceptorType.reset();
+	}
 
 	this->Spawner_LimitRange.Read(exINI, pSection, "Spawner.LimitRange");
 	this->Spawner_ExtraLimitRange.Read(exINI, pSection, "Spawner.ExtraLimitRange");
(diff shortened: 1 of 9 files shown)


src/Ext/TechnoType/Body.cpp
```diff
@@ -110,16 +110,17 @@ void TechnoTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
...
-	this->Interceptor_DeleteOnIntercept.Read(exINI, pSection, "Interceptor.DeleteOnIntercept");
-	this->Interceptor_WeaponOverride.Read(exINI, pSection, "Interceptor.WeaponOverride", true);
-	this->Interceptor_WeaponReplaceProjectile.Read(exINI, pSection, "Interceptor.WeaponReplaceProjectile");
-	this->Interceptor_WeaponCumulativeDamage.Read(exINI, pSection, "Interceptor.WeaponCumulativeDamage");
-	this->Interceptor_KeepIntact.Read(exINI, pSection, "Interceptor.KeepIntact");
+
+	if (this->Interceptor)
+	{
+		if (this->InterceptorType == nullptr)
+			this->InterceptorType.reset(new InterceptorTypeClass(this->OwnerObject()));
```
chaserli, line 119: std::make_unique

src/New/AnonymousType/InterceptorTypeClass.cpp
```diff
@@ -0,0 +1,63 @@
+#include "InterceptorTypeClass.h"
+
+#include <Utilities/SavegameDef.h>
+#include <Utilities/TemplateDef.h>
+
+InterceptorTypeClass::InterceptorTypeClass(TechnoTypeClass* pOwnedType)
+	: OwnedType(pOwnedType)
+	, Interceptor_CanTargetHouses { AffectedHouse::Enemies }
```
chaserli, line 8: You can remove those `Interceptor_` prefix since the name of the class has it already

ststl-s: Tested in singleplayer mode, including s/l work correctly. Has not test multiplayer mode.

src/Ext/TechnoType/Body.cpp
```diff
@@ -110,16 +110,17 @@ void TechnoTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
...
-	this->Interceptor_Weapon.Read(exINI, pSection, "Interceptor.Weapon");
-	this->Interceptor_DeleteOnIntercept.Read(exINI, pSection, "Interceptor.DeleteOnIntercept");
-	this->Interceptor_WeaponOverride.Read(exINI, pSection, "Interceptor.WeaponOverride", true);
-	this->Interceptor_WeaponReplaceProjectile.Read(exINI, pSection, "Interceptor.WeaponReplaceProjectile");
-	this->Interceptor_WeaponCumulativeDamage.Read(exINI, pSection, "Interceptor.WeaponCumulativeDamage");
-	this->Interceptor_KeepIntact.Read(exINI, pSection, "Interceptor.KeepIntact");
+
+	if (this->Interceptor)
+	{
+		if (this->InterceptorType == nullptr)
```
chaserli, line 118:
```suggestion
		if (!this->InterceptorType)
```
So if you are using `unique_ptr` for these, that means if `Interceptor=no` this `unique_ptr` should be null too, so the `Interceptor` member seems removable?

Metadorius (approved):
LGTM

chaserli:
> LGTM

The name of the subfolder, what should it called? I named it Parasite simply because it's not an independent type and requires a parent object. I know however it doesn't sound correct

Metadorius:
I say just leave it in the type folder.

chaserli:
> I say just leave it in the type folder.

I disagree, a distinction between enumerable (standalone) ones and parasitic (non-enumerable) ones is not unnecessary

Metadorius:
I see no point in doing so. Besides, it can be a good idea to create InterceptorClass.

chaserli:
> I see no point in doing so. Besides, it can be a good idea to create InterceptorClass.

I do. Besides, I think we need to set a criterion for such standalone types. Interceptor logic on BulletTypeClass currently takes 12 bytes per type, less than 4 new tags, I think it's fine not creating a standalone class. 

"More than 6 tags or bigger than 32 bytes" seems fair to me

Metadorius:
Who said type classes in `Types` folder have to also have a non-type class?

>  I think we need to set a criterion for such standalone types. Interceptor logic on BulletTypeClass currently takes 12 bytes per type, less than 4 new tags, I think it's fine not creating a standalone class.
>
>"More than 6 tags or bigger than 32 bytes" seems fair to me

Sounds like optimizing something that is very minor and isn't worth optimisation.

chaserli:
> Sounds like optimizing something that is very minor and isn't worth optimisation.

I don't think this structure even optimized anything. Smart pointers might even introduce some kind of overhead when I was dealing with the InvalidatePointer optimization for TechnoExt. It's for modularity and maintainability. Like if you have 8 bools I would also recommend wrapping them into a standalone class.

PR #791, opened 2022-11-08
Script action for repairing destroyed bridges

FS-21:
- Picks a Bridge Repair Hut from the map that is linked with a bridge with destroyed sections and is reachable by engineers and then send the Taskforce against it.
- The Taskforce must have any engineer and (Bridge Repair Huts in the map) or the script will end.
- Puts non-engineers into Area Guard mode when they arrive near the Bridge Repair Hut location.
- Only are evaluated Bridge Repair Huts that can be reached by the Engineers.

In aimd.ini:
x=10105,n                 ; integer, mode for selecting Bridge Repair Huts

- The possible argument values are:
0 -> Pick the closest
1 -> Pick the Farthest
-1 -> Pick Random


src/Ext/Script/Body.cpp
```diff
@@ -3088,3 +3092,465 @@ void ScriptExt::Stop_ForceJump_Countdown(TeamClass *pTeam)
...
+	if (pTeamData->MapPath_BridgeRepairHuts.Count == 0)
+	{
+		for (auto pTechno : *TechnoClass::Array)
+		{
+			if (pTechno->WhatAmI() == AbstractType::Building)
+			{
+				auto pBuilding = static_cast<BuildingClass*>(pTechno);
+				if (!pBuilding)
+					continue;
+
```
chaserli, line 3130:
```suggestion
		for (auto pBuilding : *BuildingClass::Array)
		{

```
Ahhh

FS-21:
Internal rewrite of the feature.

Thanks to Starkku I don't need to do a pathfinding routine, simplifying the code doing the same.

TaranDahl (approved):

Issue #792, opened 2022-11-09
Random crash with latest 0.3 branch


FS-21:
### First check

- [X] The issue happens on the **latest official** version of Phobos and wasn't fixed yet.
- [X] I agree to elaborate the details if requested and provide thorough testing if the bugfix is implemented.
- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search and read the issue list to find a similar issue and didn't find it.
- [X] I have attached as much information as possible *(screenshots, gifs, videos, debug and exception logs, etc)*.

### Description

I used the Latest commit in "release/v03" branch.

Is a random crash and I can't find if the cause and isn't the first time I get a crash :-/

If I use the build-31 I think that I can't replicate the crash under the this version.

Crashdump:
https://cncreloaded.com/phobos/snapshot-20221109-164900.zip

### Phobos Version

"release/v03" branch

### Conditions to reproduce

A dev-build of my mod, recently I started compiling the latests commits.

### Steps to reproduce

Random crash

### Expected behaviour

no IE

### Actual behaviour

Random crashes

PR #795, opened 2022-11-12
Fixes random crashes in some AI script actions

FS-21:
And tweaks the AI air transport behavior when infantry selects the transport.


Belonit:
Do I understand correctly that this PR fixes issues https://github.com/Phobos-developers/Phobos/issues/792?

FS-21:
Fix my own code that caused that crash, yes.

Looks that the issue is fixed.

PR #796, opened 2022-11-13
init

ststl-s:
- Now you can sort the timers of superweapons in ascending order from top to bottom

In `rulesmd.ini`
```ini
SW.Priority=0           ; integer
```


mevitar:
Seems to work fine, although sorting the opposite way that CameoPriority does might be confusing (CameoPriority puts higher value on top, while this puts higher value to the bottom).

ststl-s:
> Seems to work fine, although sorting the opposite way that CameoPriority does might be confusing (CameoPriority puts cameos with higher value on top, while this puts counters with higher value to the bottom).

That's because global mission timer is the bottom of all timers. I think it is the highest priority.

src/Ext/SWType/Hooks.cpp
```diff
@@ -15,3 +16,38 @@ DEFINE_HOOK(0x6CDE40, SuperClass_Place_FireExt, 0x4)
...
+		{
+			int otherPriority = SWTypeExt::ExtMap.Find(SuperClass::ShowTimers->GetItem(i)->Type)->SW_Priority;
+
+			if (priority > otherPriority)
+			{
+				std::swap(SuperClass::ShowTimers->Items[i], SuperClass::ShowTimers->Items[size - 1]);
+
+				for (int j = i + 1; j < size - 1; j++)
+				{
+					std::swap(SuperClass::ShowTimers->Items[j], SuperClass::ShowTimers->Items[size - 1]);
```
chaserli, line 44: Looks like a bubble sort? Well given that the number should be less than 10 in most cases it doesn't really matter. I'm just curious if a builtin sort method can be used

ststl-s: > Looks like a bubble sort? Well given that the number should be less than 10 in most cases it doesn't really matter. I'm just curious if a builtin sort method can be used

It's insertion sort.

src/Ext/SWType/Hooks.cpp
```diff
@@ -15,3 +16,38 @@ DEFINE_HOOK(0x6CDE40, SuperClass_Place_FireExt, 0x4)
...
 	return 0;
 }
+
+DEFINE_HOOK(0x6CB5D2, SuperClass_Grant_AddToShowTimer, 0x9)
+{
+	GET(SuperClass*, pThis, ESI);
+
+	enum { SkipGameCode = 0x6CB63E };
+
+	if (pThis->Type->ShowTimer && !pThis->Owner->Type->MultiplayPassive)
```
chaserli, line 26:
```suggestion
	if (pThis->Type->ShowTimer && !pThis->Owner->IsNeutral())
```

src/Ext/SWType/Hooks.cpp
chaserli: No I mean something like
```cpp
std::sort(SuperClass::ShowTimers->begin(), SuperClass::ShowTimers->end(), 
[](SuperClass* A, SuperClass* B) { return Aext->priority > Bext->priority; });
```

chaserli (changes requested):

src/Ext/SWType/Hooks.cpp
secsome: > No I mean something like
> 
> ```c++
> if(SuperClass::ShowTimers->AddItem(pThis))
>     std::sort(SuperClass::ShowTimers->begin(), SuperClass::ShowTimers->end(), 
>     [](SuperClass* A, SuperClass* B) { return Aext->priority > Bext->priority; });
> ```
> 
> I'm not sure if it's stable, i.e. behaves exactly like the vanilla one if no priority is set. And perhaps more complicated at runtime?

Use `stable_sort` if you need a stable sort

ststl-s: Every time you add a new SW, the existing ones are in order. Why use the `O(nlogn)` sorting method? It only needs `O(n)` to insert it.

chaserli:
Sorry for annihilating the commit records....

chaserli (approved):

PR #626, opened 2022-05-05
Jumpjet facing bugfix: Jumpjet units face towards target when firing from different directions

chaserli:
### Jumpjet facing target customization 

- Allows jumpjet units to face towards the target when firing from different directions. Set `[JumpjetControls] -> FacingTarget=yes` to enable it for all jumpjet locomotor units. This behavior can be overriden by setting `[UnitType] -> JumpjetFacingTarget` for specific units.
- This behavior does not apply to `TurretSpins=yes` units for obvious reasons.

In `rulesmd.ini`:
```ini
[JumpjetControls]
FacingTarget=no        ; boolean

[UnitType]
JumpjetFacingTarget=   ; boolean, override the tag in JumpjetControls
Locomotor=Jumpjet      ; Jumpjet only
TurretSpins=no         ; It's meaningless to use that for disks obviously
```

PR #608, opened 2022-04-25
Fix the jumpjet facing bug: Jumpjet helicopters can turn itself when firing

chaserli:
In vanilla, a jumpjet helicopter cannot turn itself when choosing a target within its range with different direction. 
This PR aims to solve this problem

In rulesmd.ini
```ini
[VehicleType]
JumpJet=yes           ; Necessary
Turret=no               ; Necessary
JumpjetFacing=no  ; Decide whether to face the target when firing.
```
If I did it correctly, the weapon would be able to not set `OmniFire=yes`.

Issue #443, opened 2021-11-18
[New Feature] Jumpjet aircraft/gunships facing targets as they fire


SaudiAramcat:
Description:
Its been quite bothersome and awkward to see aerial units, more specifically jumpjet aircrafts/gunships fire their weapon whilst facing any direction except the enemy that they are targeting. Tags such as OmniFire don't cut it in most cases. Although using a turret with an invisible body is used to mimic the behavior of flying units facing their targets while firing their weapons, this kind of technique is quite limiting, and in most cases not applicable in models such as helicopters.

[TechnoType]
TargetLocked=true

https://user-images.githubusercontent.com/84788793/142487319-6ea1d8e3-c714-4c0e-8371-1df8f9ecab1b.mp4

PR #799, opened 2022-11-15
JumpjetTurnToTarget: The correct fix, and OmniFire.TurnToTarget

chaserli:
# This is finally the **correct** fix for the jumpjet facing bug

Basically: 
###  If `OmniFire=no`, jumpjet units now **always** turn to the target as other locomotors do
**No additional tags needed** since it was actually a bug

besides, according to @Metadorius 's suggestion
### Make units try turning to target when firing with `OmniFire=yes`
Recommended for jumpjets' omnifiring weapons
In `rulesmd.ini`:
```ini
[SOMEWEAPON]                ; WeaponType
OmniFire=yes                      ; condition
OmniFire.TurnToTarget=no  ; boolean
```

This should finally close the concern starting from 
#215 
#443 
#608 
#626
#639 
#738 
#744 
There were 2 causes of the bug: 
1. When jumpjets arrived at their `FootClass::Destination`, they seem stuck at the Move mission, and this pointer is not nullified unlike other locomotors. This prevents the body facing from changing in most cases
2.  Jumpjets use the facing in its loco instead of techno, no synchronization was implemented

**Warning for DP-Kratos users**: You should better not using it together with its `JumpjetFacingTarget` related tags

Voxel units always **have and only have 32 facings**.

### DIFF
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -659,6 +659,17 @@ ForceWeapon.Cloaked=-1          ; integer. 0 for primary weapon, 1 for secondary
 ForceWeapon.Disguised=-1        ; integer. 0 for primary weapon, 1 for secondary weapon, -1 to disable
 ```
 
+### Make units try turning to target when firing with `OmniFire=yes`
+- The unit will try to turn the body to target even firing with `OmniFire=yes`
+  - Recommended for jumpjets if you want it to turn to target when firing.
+
+In `rulesmd.ini`:
+```ini
+[SOMEWEAPONTYPE]          ; WeaponType
+OmniFire=yes
+OmniFire.TurnToTarget=no  ; boolean
+```
+
 ### Initial strength for TechnoTypes and cloned infantry
 
 ![image](_static/images/initialstrength.cloning-01.png)
--- a/docs/Whats-New.md
+++ b/docs/Whats-New.md
@@ -16,6 +16,11 @@ You can use the migration utility (can be found on [Phobos supplementaries repo]
 
 ### From older Phobos versions
 
+### From 0.3
+
+- `[JumpjetControls]->TurnToTarget` and `JumpjetTurnToTarget` are obsolete. Jumpjet units who fire `OmniFire=no` weapons **always** turn to targets as other units do.
+  - `OmniFire.TurnToTarget` is recommended for jumpjet units' omnifiring weapons for facing turning.
+
 #### From pre-0.3 devbuilds
 
 - `Trajectory.Speed` is now defined on projectile instead of weapon.
--- a/src/Ext/Unit/Hooks.Jumpjet.cpp
+++ b/src/Ext/Unit/Hooks.Jumpjet.cpp
@@ -34,8 +34,11 @@ DEFINE_HOOK(0x736F78, UnitClass_UpdateFiring_FireErrorIsFACING, 0x6)
 	CoordStruct& source = pThis->Location;
 	CoordStruct target = pThis->Target->GetCoords(); // Target checked so it's not null here
 	DirStruct tgtDir { Math::atan2(source.Y - target.Y, target.X - source.X) };
+
 	if (pType->Turret && !pType->HasTurret) // 0x736F92
+	{
 		pThis->SecondaryFacing.SetDesired(tgtDir);
+	}
 	else // 0x736FB6
 	{
 		if (auto jjLoco = locomotion_cast<JumpjetLocomotionClass*>(pThis->Locomotor))
--- a/src/Ext/Unit/Hooks.Jumpjet.cpp
+++ b/src/Ext/Unit/Hooks.Jumpjet.cpp
@@ -42,7 +42,8 @@ DEFINE_HOOK(0x736F78, UnitClass_UpdateFiring_FireErrorIsFACING, 0x6)
 	else // 0x736FB6
 	{
 		if (auto jjLoco = locomotion_cast<JumpjetLocomotionClass*>(pThis->Locomotor))
-		{	//wrong destination check and wrong Is_Moving usage for jumpjets, should have used Is_Moving_Now
+		{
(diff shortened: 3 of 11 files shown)


src/Ext/WeaponType/Body.h
```diff
@@ -32,6 +32,7 @@ class WeaponTypeExt
 		Nullable<WeaponTypeClass*> FeedbackWeapon;
 		Valueable<bool> Laser_IsSingleColor;
 		Nullable<PartialVector2D<int>> ROF_RandomDelay;
+		Valueable<bool> OmniFireTurnToTarget;
```
Starkku, line 35: Underscore where the . is in INI name.

src/Ext/Unit/Hooks.Jumpjet.cpp
```diff
@@ -22,63 +23,76 @@ DEFINE_HOOK(0x54B8E9, JumpjetLocomotionClass_In_Which_Layer_Deviation, 0x6)
...
 
-	R->EAX(pTarget);
-	return EndFunction;
+	return 0;
 }
-#ifdef REMOVE_THIS_AFTER_V03_RELEASE
-// Bugfix: Align jumpjet turret's facing with body's
-DEFINE_HOOK(0x736BF3, UnitClass_UpdateRotation_TurretFacing, 0x6)
+
+DEFINE_HOOK(0x736EE9, UnitClass_UpdateFiring_Ferr_OK_OmniFire_Facing, 0x6)
```
Starkku, line 68: Try to avoid underscore spam in hook names and generally just stick to `Classname_Functionname_Purpose/Scope` where possible.

Starkku (changes requested):
The code looks mostly okay I think, haven't tested if it does what it is supposed to though. Couple of nitpicks I noticed and mentioned.

docs/New-or-Enhanced-Logics.md
```diff
@@ -659,6 +659,17 @@ ForceWeapon.Cloaked=-1          ; integer. 0 for primary weapon, 1 for secondary
...
 ```
 
+### Make units try turning to target when firing with `OmniFire=yes`
+- The unit will try to turn the body to target even firing with `OmniFire=yes`
+  - Recommended for jumpjets if you want it to turn to target when firing.
+
+In `rulesmd.ini`:
+```ini
+[SOMEWEAPONTYPE]          ; WeaponType
+OmniFire=yes
```
Metadorius, line 669: Do you think including vanilla tags should be done? I am not sure of that

docs/Whats-New.md
```diff
@@ -16,6 +16,11 @@ You can use the migration utility (can be found on [Phobos supplementaries repo]
 
 ### From older Phobos versions
 
+### From 0.3
+
+- `[JumpjetControls]->TurnToTarget` and `JumpjetTurnToTarget` are obsolete. Jumpjet units who fire `OmniFire=no` weapons **always** turn to targets as other units do.
```
Metadorius, line 21: may add a migration script line to comment that out

src/Ext/Unit/Hooks.Jumpjet.cpp
```diff
@@ -23,34 +24,73 @@ DEFINE_HOOK(0x54B8E9, JumpjetLocomotionClass_In_Which_Layer_Deviation, 0x6)
...
 
-	const auto pTarget = pLinkedTo->Target;
-	if (!pTarget)
-		return ContinueNoTarget;
+	auto pType = pThis->Type;
+	CoordStruct& source = pThis->Location;
+	CoordStruct target = pThis->Target->GetCoords(); // Target checked so it's not null here
+	DirStruct tgtDir { Math::atan2(source.Y - target.Y, target.X - source.X) };
+	if (pType->Turret && !pType->HasTurret) // 0x736F92
+		pThis->SecondaryFacing.SetDesired(tgtDir);
```
Metadorius, line 38:
```suggestion

	if (pType->Turret && !pType->HasTurret) // 0x736F92
	{
		pThis->SecondaryFacing.SetDesired(tgtDir);
	}
```

src/Ext/Unit/Hooks.Jumpjet.cpp
```diff
@@ -23,34 +24,73 @@ DEFINE_HOOK(0x54B8E9, JumpjetLocomotionClass_In_Which_Layer_Deviation, 0x6)
...
+	else // 0x736FB6
+	{
+		if (auto jjLoco = locomotion_cast<JumpjetLocomotionClass*>(pThis->Locomotor))
+		{	//wrong destination check and wrong Is_Moving usage for jumpjets, should have used Is_Moving_Now
+			if (jjLoco->State != JumpjetLocomotionClass::State::Cruising)
```
Metadorius, line 43:
```suggestion
		if (auto jjLoco = locomotion_cast<JumpjetLocomotionClass*>(pThis->Locomotor))
		{
			//wrong destination check and wrong Is_Moving usage for jumpjets, should have used Is_Moving_Now
			if (jjLoco->State != JumpjetLocomotionClass::State::Cruising)
```

Metadorius (approved):
Only some minor stuff to fix, LGTM otherwise.

Metadorius:
Also I assume it does work correctly with all cases of turreted/unturreted jumpjets, right?

docs/New-or-Enhanced-Logics.md
chaserli: It means that it only works if `OmniFire=yes`

docs/Whats-New.md
chaserli: ctrl-f JumpjetTurnToTarget, delete, done

chaserli:
> Also I assume it does work correctly with all cases of turreted/unturreted jumpjets, right?

Turreted jumpjets won't turn the primary facing as they shouldn't do it anyway, since ground units don't.

chaserli:
Also the entire FACING FireError code is rewritten, since no one test individually and I haven't observed any anormalies myself we'll have to test it in future versions.

PR #808, opened 2022-12-01
Allow `DieSound/VoiceDie` not being played when being grinded

chaserli:
When an infantry/unit enters a grinder, the `DieSound`/`VoiceDie` are always played.
You can now disable that per grinder

```ini
[SOMEBUILDING]
Grinding.PlayDieSound=yes
```

Can merge this only after v0.3 being merged into develop


Starkku (changes requested):
I propose changing the `UpdatePosition_Grinding_SkipDiesound` and `PerCellProcess_DoGrindingExtras` hook names to `UpdatePosition_GrindingBefore` and `UpdatePosition_GrindingAfter` or something to that effect, for consistency and as they do multiple different things within each hook.

Other than that it is fine I think.

PR #809, opened 2022-12-03
Crates generate fix

Super-StarX:
Fixed the bug that the crates has a higher chance of creating at the edge of the map.

### DIFF
--- a/src/Misc/Hooks.BugFixes.cpp
+++ b/src/Misc/Hooks.BugFixes.cpp
@@ -2,6 +2,7 @@
 #include <AnimClass.h>
 #include <BuildingClass.h>
 #include <TechnoClass.h>
+#include <TacticalClass.h>
 #include <FootClass.h>
 #include <UnitClass.h>
 #include <OverlayTypeClass.h>
@@ -608,4 +609,32 @@ DEFINE_HOOK(0x43D874, BuildingClass_Draw_BuildupBibShape, 0x6)
 		return DontDrawBib;
 
 	return 0;
+}
+
+DEFINE_HOOK(0x56BD8B, MapClass_PlaceCrate_InMapFix, 0x5)
+{
+	enum { CreateCrate = 0x56BE7B, DontCreate = 0x56BE91 };
+
+	int MapX = MapClass::Instance->MapRect.Width * 60;
+	int MapY = MapClass::Instance->MapRect.Height * 30;
+	int X = ScenarioClass::Instance->Random.RandomRanged(0, MapX) - MapX / 2;
+	int Y = ScenarioClass::Instance->Random.RandomRanged(0, MapY) + MapY / 2;
+	Vector3D<float> Location = { (float)X,(float)Y,0.0f };
+
+	auto result = Matrix3D::MatrixMultiply(TacticalClass::Instance->IsoTransformMatrix, Location);
+	CoordStruct MapLocation { (int)result.X,(int)result.Y,0 };
+
+	if (auto pCell = MapClass::Instance->TryGetCellAt(MapLocation))
+	{
+		REF_STACK(CellStruct, cell, STACK_OFFSET(0x28, 0x18));
+
+		auto SpeedType = pCell->LandType == LandType::Water ? SpeedType::Float : SpeedType::Track;
+		cell = MapClass::Instance->NearByLocation(pCell->MapCoords, SpeedType, -1, MovementZone::Normal, false, 1, 1, false,
+			false, false, true, CellStruct::Empty, false, false);
+
+		R->EAX(&cell);
+		return CreateCrate;
+	}
+
+	return DontCreate;
 }
--- a/src/Misc/Hooks.BugFixes.cpp
+++ b/src/Misc/Hooks.BugFixes.cpp
@@ -626,7 +626,7 @@ DEFINE_HOOK(0x56BD8B, MapClass_PlaceCrate_InMapFix, 0x5)
 
 	if (auto pCell = MapClass::Instance->TryGetCellAt(MapLocation))
 	{
-		REF_STACK(CellStruct, cell, STACK_OFFSET(0x28, 0x18));
+		REF_STACK(CellStruct, cell, STACK_OFFSET(0x28, -0x18));
 
 		auto SpeedType = pCell->LandType == LandType::Water ? SpeedType::Float : SpeedType::Track;
 		cell = MapClass::Instance->NearByLocation(pCell->MapCoords, SpeedType, -1, MovementZone::Normal, false, 1, 1, false,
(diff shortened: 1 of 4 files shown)


src/Misc/Hooks.BugFixes.cpp
```diff
@@ -609,3 +610,31 @@ DEFINE_HOOK(0x43D874, BuildingClass_Draw_BuildupBibShape, 0x6)
...
+{
+	enum { CreateCrate = 0x56BE7B, DontCreate = 0x56BE91 };
+
+	int MapX = MapClass::Instance->MapRect.Width * 60;
+	int MapY = MapClass::Instance->MapRect.Height * 30;
+	int X = ScenarioClass::Instance->Random.RandomRanged(0, MapX) - MapX / 2;
+	int Y = ScenarioClass::Instance->Random.RandomRanged(0, MapY) + MapY / 2;
+	Vector3D<float> Location = { (float)X,(float)Y,0.0f };
+
+	auto result = Matrix3D::MatrixMultiply(TacticalClass::Instance->IsoTransformMatrix, Location);
```
chaserli, line 624: you could have done the math manually here, this matrix seems to be 128/30*[1,2;-1,2]

src/Misc/Hooks.BugFixes.cpp
```diff
@@ -609,3 +610,31 @@ DEFINE_HOOK(0x43D874, BuildingClass_Draw_BuildupBibShape, 0x6)
...
+	CoordStruct MapLocation { (int)result.X,(int)result.Y,0 };
+
+	if (auto pCell = MapClass::Instance->TryGetCellAt(MapLocation))
+	{
+		REF_STACK(CellStruct, cell, STACK_OFFSET(0x28, -0x18));
+
+		auto SpeedType = pCell->LandType == LandType::Water ? SpeedType::Float : SpeedType::Track;
```
chaserli, line 631: can we have a new tag for land-only crate spawn?

chaserli (changes requested):
Monte-Carlo? Mind explaining why this gives better probability distribution?

Super-StarX:
> game considers the map as a horizontal rectangle parallel to the screen, but the map is inclined at 45°. So, the map will put some crates in the place without a cell (not in map), and then put them into the map through NearByLocation, which results in more crates generated at the edge of the map. This is obviously unreasonable.

src/Misc/Hooks.BugFixes.cpp
Belonit: @chaserli I think this is a topic for a separate pull request.

chaserli (changes requested):

chaserli (changes requested):
First of all, I appreciate your willingness to contribute on this topic and the identification of such an anomaly. However your basic assumption wasn't exactly correct in your explanation of non-uniform sampling when spawning crates:

In my opinion it's hard to assert that the problem was due to isometric projection in a probabilistic point of view. If you have some basic knowledge of probability, [rejection sampling](https://en.wikipedia.org/wiki/Rejection_sampling) is the way used in the vanilla routine, and proceed like that is the easiest way. The usage of `MapCoordBound` in the vanilla routine was therefore not wrong. The only missing part is checking whether the candidate cell is within the usable region.

If you no longer want to contribute to this PR, I can commit that directly and close this one.

Super-StarX:
> First of all, I appreciate your willingness to contribute on this topic and the identification of such an anomaly. However your basic assumption wasn't exactly correct in your explanation of non-uniform sampling when spawning crates:
> 
> In my opinion it's hard to assert that the problem was due to isometric projection in a probabilistic point of view. If you have some basic knowledge of probability, [rejection sampling](https://en.wikipedia.org/wiki/Rejection_sampling) is the way used in the vanilla routine, and proceed like that is the easiest way. The usage of `MapCoordBound` in the vanilla routine was therefore not wrong. The only missing part is checking whether the candidate cell is within the usable region.
> 
> If you no longer want to contribute to this PR, I can commit that directly and close this one.

I don't quite understand your meaning. You can use Chinese in your private chat

chaserli:
> 
You can check my implementation regarding this issue here #818 
I gave my explanation, you can check it out from here https://discord.com/channels/338055131666972683/773636942775582720/1060219669908308080

Super-StarX:
> > 
> 
> You can check my implementation regarding this issue here #818 I gave my explanation, you can check it out from here https://discord.com/channels/338055131666972683/773636942775582720/1060219669908308080
Sorry I still not understand. It seems that you want to solve the problem that crate will generate on water, which may be useful in some land battle maps with less water. 
What I try to solve:
image

chaserli:
> What I try to solve:

You were right just until this point, but you did nothing but the isometric projection and forgot to limit it to the actual **visible** area, which may eventually take cell height/slope/etc into account.
Let's say before the fix, if you create a large map with medium size visible area, and draw a histogram of spawned crates' locations, you'll get something like this: 
image
Your so-called "error generation by WW area" fall back to (0,0) after being projected by `NearByLocation`, and all other off-visible-map region fall back to the 4 boundaries.

So after your fix, those off-map regions will still be taken into account, which results in a higher probability at the boundaries. The issue is still not resolved.

So there's 2 way: 
1: Use `VisibleRect` rather than `MapRect` for the isometric projection, just as I said previously. And you can just do the matrix multiplication yourself since it's already known.
2. Keep the original random point generation method and take advantage of the `while(true)` loop with the simple math fact that I said here https://discord.com/channels/338055131666972683/773636942775582720/1060225075187568650
#818 uses such an idea and can ensure a uniform distribution over the desirable region
image


If you still didn't get it, maybe I should write down the demonstration for you

Super-StarX:
> > What I try to solve:
> 
> You were right just until this point, but you did nothing but the isometric projection and forgot to limit it to the actual **visible** area, which may eventually take cell height/slope/etc into account. Let's say before the fix, if you create a large map with medium size visible area, and draw a histogram of spawned crates' locations, you'll get something like this: image Your so-called "error generation by WW area" fall back to (0,0) after being projected by `NearByLocation`, and all other off-visible-map region fall back to the 4 boundaries.
> 
> So after your fix, those off-map regions will still be taken into account, which results in a higher probability at the boundaries. The issue is still not resolved.
> 
> So there's 2 way: 1: Use `VisibleRect` rather than `MapRect` for the isometric projection, just as I said previously. And you can just do the matrix multiplication yourself since it's already known. 2. Keep the original random point generation method and take advantage of the `while(true)` loop with the simple math fact that I said here https://discord.com/channels/338055131666972683/773636942775582720/1060225075187568650 #818 uses such an idea and can ensure a uniform distribution over the desirable region image
> 
> If you still didn't get it, maybe I should write down the demonstration for you

Oh, it's my fault. I have actually made changes according to the first tip you list, but I forgot to push. I'm not very proficient in the operation of Git XD

chaserli:
> Oh, it's my fault. I have actually made changes according to the first tip you list, but I forgot to push. I'm not very proficient in the operation of Git XD

Even if you did, it shouldn't be simply replacing MapRect with VisibleRect. The range is incorrect. You can check my implementation at #818 and continue the discussion there.

PR #816, opened 2022-12-30
Projectile obstacle logic additions

Starkku:
Additional obstacle logic for projectiles.

### Fixes

- Weapons with projectiles with `Level=true` now consider targets behind obstacles that cause the projectile to detonate (tiles belonging to non-water tilesets) as out of range and will attempt to reposition before firing.

### Projectiles blocked by land or water

- It is now possible to make projectiles consider either land or water as obstacles that block their path by setting `SubjectToLand/Water` to true, respectively. Weapons firing such projectiles will consider targets blocked by such obstacles as out of range and will attempt to reposition themselves so they can fire without being blocked by the said obstacles before firing and if `SubjectToLand/Water.Detonate` is set to true, the projectiles will detonate if they somehow manage to collide with the said obstacles.
  - In a special case, `Level=true` projectiles by default, if neither `SubjectToLand` or `SubjectToWater` are set, consider tiles belonging to non-water tilesets as obstacles, but this behaviour can be overridden by setting these keys.

In `rulesmd.ini`:
```ini
[SOMEPROJECTILE]              ; Projectile
SubjectToLand=                ; boolean
SubjectToLand.Detonate=true   ; boolean
SubjectToWater=               ; boolean
SubjectToWater.Detonate=true  ; boolean
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -176,6 +176,7 @@ This page lists all the individual contributions to the project by their author.
   - Custom warhead debris animations
   - Attached particle system for animations
   - Removal of hardcoded AA & Gattling weapon selection restrictions
+  - Projectile obstacle logic additions
 - **Morton (MortonPL)**:
   - `XDrawOffset`
   - Shield passthrough & absorption
--- a/Phobos.vcxproj
+++ b/Phobos.vcxproj
@@ -65,6 +65,7 @@
     <ClCompile Include="src\Ext\BulletType\Body.cpp" />
     <ClCompile Include="src\Ext\Bullet\Body.cpp" />
     <ClCompile Include="src\Ext\Bullet\Hooks.cpp" />
+    <ClCompile Include="src\Ext\Bullet\Hooks.Obstacles.cpp" />
     <ClCompile Include="src\Ext\House\Body.cpp" />
     <ClCompile Include="src\Ext\House\Hooks.cpp" />
     <ClCompile Include="src\Ext\RadSite\Body.cpp" />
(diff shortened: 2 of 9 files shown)


mevitar:
Appears to work fine, although the projectile detonates on cliffs even with SubjectToCliffs=no (i set all 4 above tags to "no"). Guess it needs SubjectToCliffs=yes so the firer will be forced to walk around cliffs.

Starkku:
> Appears to work fine, although the projectile detonates on cliffs even with SubjectToCliffs=no (i set all 4 above tags to "no"). Guess it needs SubjectToCliffs=yes so the firer will be forced to walk around cliffs.

Could you clarify what kind of projectile you are talking about?

mevitar:
This is the code for the projectile. Projectile is not a voxel.
```
AG=yes
Ranged=yes
Level=yes
Arm=2
ROT=12

SubjectToLand=no
SubjectToLand.Detonate=no
SubjectToWater=no
SubjectToWater.Detonate=no
```

Starkku:
`Level=true` restricts projectile's motion and always has. What you're experiencing is no different from `ROT > 0` projectile colliding with terrain because it can't path its way around it due to being too fast or not being able to turn fast enough.

mevitar:
This is pretty much the same code as the original [Torpedo] has (the difference is that works on ground and water). I had absolutely no clue how it would (or should) behave over cliffs.
If this is working as intended then i don't see an issue.

Metadorius (approved):
LGTM

PR #817, opened 2022-12-30
Separate AI naval vehicle production from regular vehicle production

Starkku:
By default, while game has separate FactoryClass instances and queues for regular and naval vehicles, game still keeps track of which VehicleType to build and which one was last built using only one reference each in HouseClass. This PR decouples this for ground and naval vehicles, allowing AI to produce both simultaneously.

Because Ares reimplements the HouseClass unit production update function (`0x4FEA60`) and this feature/fix is impossible to implement without touching that function, it has been reimplemented here again using [Ares 0.A version of the function](https://github.com/Ares-Developers/Ares/blob/4f1d929920aca31924c6cd4d3dfa849daa65252a/src/Ext/House/Hooks.100.cpp#L25) as a reference. However as it stands, I have no reliable way of telling if it has been altered to any significant degree in time since that and 3.0. It is possible that this PR uses less optimized code, or in worst-case, faulty or partial. **I would appreciate help of a more experienced reverse-engineer with this.**

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -186,6 +186,7 @@ This page lists all the individual contributions to the project by their author.
   - Option to center pause menu background
   - Disguise logic improvements
   - Custom insignias
+  - AI naval vehicle production fix
 - **Morton (MortonPL)**:
   - `XDrawOffset` for animations
   - Shield passthrough & absorption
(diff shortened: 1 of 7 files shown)


FS-21:
I tested it and works as expected, but now the Ares tag "AllowParallelAIQueues=no" for fixing the duplication of built unit in each available factory is broken (a classic vanilla bug).
The bug affects War factories and Naval Yards.

Starkku:
There is a known issue that crashes the game due to a seemingly corrupt `FactoryClass` pointer. This crash has EIP `0x4CA0F8`.

Starkku:
The crash mentioned in my previous comment should now be fixed.

DOOMGUY2K:
I would like to test this but it appears the link has expired.

DOOMGUY2K (approved):
I can confirm this is working as intended. 

I checked the previous notes

Tested on 

Vanilla 1.001
Ares 3.01p
And KratosDP

And combined inside my own Mod.

Metadorius (approved):

PR #818, opened 2023-01-03
The option to limit random crate generation only on land + sampling mechanism fix

chaserli:
This should fix the theoretical flaw of #809 
A histogram proving a more uniform distribution will be uploaded later

The game uses [rejection sampling](https://en.wikipedia.org/wiki/Rejection_sampling) for the random generation of crates.
- You can now limit the crates' spawn region to land only.

In `rulesmd.ini`:
```ini
[CrateRules]
CrateOnlyOnLand=no  ; boolean
```


src/Ext/Rules/Body.cpp
```diff
@@ -107,6 +107,8 @@ void RulesExt::ExtData::LoadBeforeTypeData(RulesClass* pThis, CCINIClass* pINI)
 	this->IronCurtain_EffectOnOrganics.Read(exINI, GameStrings::CombatDamage, "IronCurtain.EffectOnOrganics");
 	this->IronCurtain_KillOrganicsWarhead.Read(exINI, GameStrings::CombatDamage, "IronCurtain.KillOrganicsWarhead");
 
+	this->Crate_LandOnly.Read(exINI, reinterpret_cast<const char*>(0x839E9C), "Crate.LandOnly");
```
Metadorius, line 110: Should probably be YRpp'ed

PR #820, opened 2023-01-09
Allow jumpjet to crash without rotation

Super-StarX:
Do not rotate please.


NetsuNegi:
I tested it and ... it cause crash.

NetsuNegi:
> I tested it and ... it cause crash.

Tested for some times, it always crash in this hook.

NetsuNegi:
Tested it ,now can work fine.

Metadorius (approved):

PR #825, opened 2023-01-19
Fix weapon effects (beams/railguns/bolts etc) interaction with obstacles

Starkku:
Prevents weapon effects like lasers, railgun particles, rad beams, electric bolts and waves from being drawn beyond where the projectile detonates (walls, solid buildings etc).

In addition particles spawned by `BehavesLike=Railgun` particle system as well as `AmbientDamage` are no longer cut off by elevation changes along trajectory.

## Summary by CodeRabbit

- **New Features**
	- Added new options for customizing fire particle systems and ambient damage effects.
	- Introduced the ability to show a briefing screen at the start of singleplayer missions.
	- Enhanced missile targeting, turret rotation, and jumpjet stability under EMP conditions.
	- Added customization options for disk laser radius and cluster scatter distance in projectiles.

- **Bug Fixes**
	- Fixed graphical issues with weapon effects not interacting correctly with obstacles.
	- Resolved problems with railgun and fire particles being cut off by elevation changes.
	- Corrected behavior of `AmbientDamage` with `IsRailgun=yes` to prevent cutoff by elevation.

- **Documentation**
	- Updated documentation to reflect new features and bug fixes.
	- Renamed the `Detonate.Warhead` property to `Detonate.WarheadType` in configuration files for clarity.

- **Refactor**
	- Removed outdated laser drawing logic to streamline weapon effect processing.

- **Chores**
	- Integrated new source files for the `ParticleSystemType` module.

### DIFF
--- a/src/Ext/Techno/Hooks.WeaponEffects.cpp
+++ b/src/Ext/Techno/Hooks.WeaponEffects.cpp
@@ -68,11 +68,10 @@ DEFINE_HOOK(0x62FA20, ParticleSystemClass_FireAI_TargetCoords, 0x6)
 	return Continue;
 }
 
-//
-DEFINE_HOOK_AGAIN(0x62D685, ParticleSystemClass_Fire_Coords, 0x5)
-DEFINE_HOOK(0x62D596, ParticleSystemClass_Fire_Coords, 0x5)
+// Fix fire particles being disallowed from going upwards.
+DEFINE_HOOK(0x62D685, ParticleSystemClass_Fire_Coords, 0x5)
 {
-	enum { SkipGameCode1 = 0x62D5C8, SkipGameCode2 = 0x62D6B7 };
+	enum { SkipGameCode = 0x62D6B7 };
 
 	// Game checks if MapClass::GetCellFloorHeight() for currentCoords is larger than for previousCoords and sets the flags on ParticleClass to
 	// remove it if so. Below is an attempt to create a smarter check that allows upwards movement and does not needlessly collide with elevation
@@ -98,7 +97,7 @@ DEFINE_HOOK(0x62D596, ParticleSystemClass_Fire_Coords, 0x5)
 	}
 	*/
 
-	return R->Origin() == 0x62D596 ? SkipGameCode1 : SkipGameCode2;
+	return SkipGameCode;
 }
 
 // Fix railgun target coordinates potentially differing from actual target coords.
--- a/src/Ext/Techno/Hooks.WeaponEffects.cpp
+++ b/src/Ext/Techno/Hooks.WeaponEffects.cpp
@@ -35,6 +35,72 @@ DEFINE_HOOK(0x6FF15F, TechnoClass_FireAt_ObstacleCellSet, 0x6)
 	return 0;
 }
 
+// Apply obstacle logic to fire & spark particle system targets.
+DEFINE_HOOK_AGAIN(0x6FF1D7, TechnoClass_FireAt_SparkFireTargetSet, 0x5)
+DEFINE_HOOK(0x6FF189, TechnoClass_FireAt_SparkFireTargetSet, 0x5)
+{
+	if (FireAtTemp::pObstacleCell)
+	{
+		if (R->Origin() == 0x6FF189)
+			R->ECX(FireAtTemp::pObstacleCell);
+		else
+			R->EDX(FireAtTemp::pObstacleCell);
+	}
+
+	return 0;
+}
+
+// Fix fire particle target coordinates potentially differing from actual target coords.
+DEFINE_HOOK(0x62FA20, ParticleSystemClass_FireAI_TargetCoords, 0x6)
+{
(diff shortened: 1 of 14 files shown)


mevitar:
I managed to get this angle perfectly by accident and tested it with this build, and you can see here that the railgun is still drawing despite the projectile detonating on the edge of this solid building. Or, to be precise, on its corner (i know it can be hard to spot because of the animations i used. It was for IsDetachedRailgun=yes railgun, though, so don't know if it's supported.
railonedge

Or is it meant to work only for non-shp particles?

Starkku:
> you can see here that the railgun is still drawing despite the projectile detonating on the edge of this solid building

Just to eliminate the most obvious cause of issues like this, this is `Inviso=true` projectile yes? The parity is not guaranteed in cases where it isn't and is actually technically impossible to do without a *massive* rewrite of all of the weapon/projectile logic. What is probably up for debate if the obstacle logic should be disabled if `Inviso` is not true.

> It was for IsDetachedRailgun=yes railgun, though, so don't know if it's supported.

Should be.

mevitar:
> Just to eliminate the most obvious cause of issues like this, this is `Inviso=true` projectile yes?
it actually isn't. it's a straight trajectory projectile with an attached railgun. I didn't realize the fix will work only for Inviso=yes projectiles. Maybe that should be mentioned in the doc?

Retested it with an Inviso=yes projectile and that one does not even get blocked on this angle. It detonates on the target instead of the building, and the "beam" is also drawn from the firer to the target. From that, i can't say if the fix is working properly or not.

That pretty much means i'm out of ideas how to test this fix for solid buildings. Units will move around if they detect a building they can't fire through, and getting angles like this one is a matter of sheer luck.

Starkku:
> it actually isn't. it's a straight trajectory projectile with an attached railgun. I didn't realize the fix will work only for Inviso=yes projectiles. Maybe that should be mentioned in the doc?

Documentation does have this line, can probably clarify if needed though.
```Fixed weapon effects like railguns, fire particles, lasers, beams, bolts and waves drawing beyond where Warhead detonates if it hits an obstacle like wall. Note that this is not necessarily same as the actual projectile travel path if it is not Inviso=true projectile.```

> That pretty much means i'm out of ideas how to test this fix for solid buildings. Units will move around if they detect a building they can't fire through, and getting angles like this one is a matter of sheer luck.

This is pretty much how it is since the obstacle logic affects rangefinding. The cases where the projectile is allowed to fire but is blocked by an obstacle are very situational but they do happen. It is usually hitting corners like in the example you had.

**Comment by mevitar** [comment]
> Documentation does have this line, can probably clarify if needed though.

Nah, that's enough. I thought it's not there because i couldn't find it. Apparently i was looking in the incorrect doc file (clicked on the big "View Docs" button instead of on the small "View docs" hyperlink text).

File: src/Ext/Techno/Hooks.WeaponEffects.cpp [hunk 1]
```diff
@@ -0,0 +1,262 @@
...
+}
+
+// Do not adjust map coordinates for railgun or fire stream particles that are below cell coordinates.
+DEFINE_HOOK(0x62B8BC, ParticleClass_CTOR_CoordAdjust, 0x6)
+{
+	enum { SkipCoordAdjust = 0x62B8CB };
+
+	GET(ParticleClass*, pThis, ESI);
+
+	if (pThis->ParticleSystem && (pThis->ParticleSystem->Type->BehavesLike == BehavesLike::Railgun || pThis->ParticleSystem->Type->BehavesLike == BehavesLike::Fire))
```
Metadorius, line 184: Looks obnoxiously long.

Metadorius (commented):
LGTM apart from comments

src/Ext/Techno/Hooks.WeaponEffects.cpp
```diff
@@ -0,0 +1,262 @@
...
+	return 0;
+}
+
+// Restore original target values and unset obstacle cell.
+DEFINE_HOOK(0x6FF660, TechnoClass_FireAt_ObstacleCellUnset, 0x6)
+{
+	LEA_STACK(CoordStruct*, pTargetCoords, STACK_OFFSET(0xB0, -0x28));
+
+	auto coords = FireAtTemp::originalTargetCoords;
+	pTargetCoords = &coords;
```
chaserli, line 234: Not sure what you were intending to do but nothing is done here. Were you trying to say `*pTargetCoords = coords`?

src/Ext/Techno/Hooks.WeaponEffects.cpp
```diff
@@ -0,0 +1,262 @@
...
+	auto target = FireAtTemp::pOriginalTarget;
+
+	FireAtTemp::originalTargetCoords = CoordStruct::Empty;
+	FireAtTemp::pOriginalTarget = nullptr;
```
chaserli, line 238: Some of these might not be indispensable,  those on ebp/esp should have remained intact

PR #826, opened 2023-01-22
AttachEffect / Custom tints / Revenge weapons / Warhead effect Verses requirement change

Starkku:
# [Link to the bot comment with download link](https://github.com/Phobos-developers/Phobos/pull/826#issuecomment-1636739508)

Phobos-introduced Warhead effects like shield modifiers, critical hits, disguise & mind control removal now require Warhead `Verses` to affect target to apply unless `EffectsRequireVerses` is set to false. Shield armor type is used if target has an active shield that cannot be penetrated by the Warhead.

----------------------------

### Attached effects

- Similar (but not identical) to Ares' AttachEffect, but with some differences and new features. The largest difference is that here attached effects are explicitly defined types.
  - `Duration` determines how long the effect lasts for. It can be overriden by `DurationOverrides` on TechnoTypes and Warheads.
  - `Cumulative`, if set to true, allows the same type of effect to be applied on same object multiple times, up to `Cumulative.MaxCount` number or with no limit if `Cumulative.MaxCount` is a negative number.
  - `Powered` controls whether or not the effect is rendered inactive if the object it is attached to is deactivated (`PoweredUnit` or affected by EMP) or on low power. What happens to animation is controlled by `Animation.OfflineAction`.
  - `DiscardOn` accepts a list of values corresponding to conditions where the attached effect should be discarded. Defaults to `none`, meaning it is never discarded.
    - `entry`: Discard on exiting the map when entering transports or buildings etc.
    - `move`: Discard when the object the effect is attached on moves. Ignored if the object is a building.
    - `stationary`: Discard when the object the effect is attached on stops moving. Ignored if the object is a building.
    - `drain`: Discard when the object is being affected by a weapon with `DrainWeapon=true`.
  - If `PenetratesIronCurtain` is not set to true, the effect is not applied on currently invulnerable objects (Iron Curtain / Force Shield).
  - `Animation` defines animation to play in an indefinite loop for as long as the effect is active on the object it is attached to.
    - If `Animation.ResetOnReapply` is set to true, the animation playback is reset every time the effect is applied if `Cumulative=false`.
    - `Animation.OfflineAction` determines what happens to the animation when the attached object is deactivated or not powered. Only applies if `Powered=true`.
    - `Animation.TemporalAction` determines what happens to the animation when the attached object is under effect of `Temporal=true` Warhead.
    - `Animation.UseInvokerAsOwner` can be used to set the house and TechnoType that created the effect (e.g firer of the weapon that applied it) as the animation's owner & invoker instead of the object the effect is attached to.
  - `CumulativeAnimations` can be used to declare a list of animations used for `Cumulative=true` types instead of `Animation`. An animation is picked from the list in order matching the number of active instances of the type on the object, with last listed animation used if number is higher than the number of listed animations. This animation is only displayed once, on the first active instance of the effect found attached and is updated and restarted if the number of active instances changed.
  - Attached effect can fire off a weapon when expired / removed / object dies by setting `ExpireWeapon`.
    - `ExpireWeapon.TriggerOn` determines the exact conditions upon which the weapon is fired, defaults to `expire` which means only if the effect naturally expires.
    - `ExpireWeapon.CumulativeOnlyOnce`, if set to true, makes it so that `Cumulative=true` attached effects only detonate the weapon once period, instead of once per active instance. On `remove` and `expire` condition this means it will only detonate after last instance has expired or been removed.
  - `Tint.Color` & `Tint.Intensity` can be used to set a color tint effect and additive lighting increase/decrease on the object the effect is attached to, respectively.
    - `Tint.VisibleToHouses` can be used to control which houses can see the tint effect.
  - `FirepowerMultiplier`, `ArmorMultiplier`, `SpeedMultiplier` and `ROFMultiplier` can be used to modify the object's firepower, armor strength, movement speed and weapon reload rate, respectively.
    - If `ROFMultiplier.ApplyOnCurrentTimer` is set to true, `ROFMultiplier` is applied on currently running reload timer (if any) when the effect is first applied.
  - If `Cloakable` is set to true, the object the effect is attached to is granted ability to cloak itself for duration of the effect.
  - `ForceDecloak`, if set to true, will uncloak and make the object the effect is attached to unable to cloak itself for duration of the effect.
  - `WeaponRange.Multiplier` and `WeaponRange.ExtraRange` can be used to multiply the weapon firing range of the object the effect is attached to, or give it an increase / decrease (measured in cells), respectively. `ExtraRange` is cumulatively applied from all attached effects after all `Multiplier` values have been applied.
    - `WeaponRange.AllowWeapons` can be used to list only weapons that can benefit from this range bonus and `WeaponRange.DisallowWeapons` weapons that are not allowed to, respectively.
    - On TechnoTypes with `OpenTopped=true`, `OpenTopped.UseTransportRangeModifiers` can be set to true to make passengers firing out use the transport's active range bonuses instead.
  - `Crit.Multiplier` and `Crit.ExtraChance` can be used to multiply the [critical hit](#chance-based-extra-damage-or-warhead-detonation--critical-hits) chance or grant a fixed bonus to it for the object the effect is attached to, respectively.
    - `Crit.AllowWarheads` can be used to list only Warheads that can benefit from this critical hit chance multiplier and `Crit.DisallowWarheads` weapons that are not allowed to, respectively.
  - `RevengeWeapon` can be used to temporarily grant the specified weapon as a [revenge weapon](#revenge-weapon) for the attached object.
    - `RevengeWeapon.AffectsHouses` customizes which houses can trigger the revenge weapon.
  - `DisableWeapons` can be used to disable ability to fire any and all weapons.
  - It is possible to set groups for attach effect types by defining strings in `Groups`.
    - Groups can be used instead of types for removing effects and weapon filters.

- AttachEffectTypes can be attached to TechnoTypes using `AttachEffect.AttachTypes`.
  - `AttachEffect.DurationOverrides` can be used to override the default durations. Duration matching the position in `AttachTypes` is used for that type, or the last listed duration if not available.
  - `AttachEffect.Delays` can be used to set the delays for recreating the effects on the TechnoType after they expire. Defaults to 0 (immediately), negative values mean the effects are not recreated. Delay matching the position in `AttachTypes` is used for that type, or the last listed delay if not available.
  - `AttachEffect.InitialDelays` can be used to set the delays before first creating the effects on TechnoType. Defaults to 0 (immediately). Delay matching the position in `AttachTypes` is used for that type, or the last listed delay if not available.
  - `AttachEffect.RecreationDelays` is used to determine if the effect can be recreated if it is removed completely (e.g `AttachEffect.RemoveTypes`), and if yes, how long this takes. Defaults to -1, meaning no recreation. Delay matching the position in `AttachTypes` is used for that type, or the last listed delay if not available.

- AttachEffectTypes can be attached to objects via Warheads using `AttachEffect.AttachTypes`.
  - `AttachEffect.DurationOverrides` can be used to override the default durations. Duration matching the position in `AttachTypes` is used for that type, or the last listed duration if not available.
  - Attached effects can be removed from objects by Warheads using `AttachEffect.RemoveTypes` or `AttachEffect.RemoveGroups`.
    - `AttachEffect.CumulativeRemoveMinCounts` sets minimum number of active instaces per `RemoveTypes`/`RemoveGroups` required for `Cumulative=true` types to be removed.
    - `AttachEffect.CumulativeRemoveMaxCounts` sets maximum number of active instaces per `RemoveTypes`/`RemoveGroups` for `Cumulative=true` that are removed at once by this Warhead.

- Weapons can require attached effects on target to fire, or be prevented by firing if specific attached effects are applied.
  - `AttachEffect.RequiredTypes` can be used to list attached effects required to be on target to fire, all listed effect types must be present to allow firing.
  - `AttachEffect.DisallowedTypes` can be used to list attached effects that when present prevent the weapon from firing, any of the listed effect types will prevent firing if present.
  - `AttachEffect.Required/DisallowedGroups` have the same effect except applied with/to all types that have one of the listed groups in their `Groups` listing.
  - `AttachEffect.(Required|Disallowed)MinCounts & (Required|Disallowed)MaxCounts` can be used to set the minimum and maximum number of instances required / disallowed to be on the Techno for `Cumulative=true` types (ignored for other types) respectively.
  - `AttachEffect.IgnoreFromSameSource` can be set to true to ignore effects that have been attached by the firer of the weapon and its Warhead.

In `rulesmd.ini`:
```ini
[AttachEffectTypes]
0=SOMEATTACHEFFECT

[SOMEATTACHEFFECT]                           ; AttachEffectType
Duration=0                                   ; integer - game frames or negative value for indefinite duration
Cumulative=false                             ; boolean
Cumulative.MaxCount=-1                       ; integer
Powered=false                                ; boolean
DiscardOn=none                               ; list of discard condition enumeration (none|entry|move|stationary|drain)
PenetratesIronCurtain=false                  ; boolean
Animation=                                   ; Animation
Animation.ResetOnReapply=false               ; boolean
Animation.OfflineAction=Hides                ; AttachedAnimFlag (None, Hides, Temporal, Paused or PausedTemporal)
Animation.TemporalAction=None                ; AttachedAnimFlag (None, Hides, Temporal, Paused or PausedTemporal)
Animation.UseInvokerAsOwner=false            ; boolean
CumulativeAnimations=                        ; list of animations
ExpireWeapon=
ExpireWeapon.TriggerOn=expire                ; List of expire weapon trigger condition enumeration (none|expire|remove|death|all)
ExpireWeapon.CumulativeOnlyOnce=false        ; boolean
Tint.Color=                                  ; integer - R,G,B
Tint.Intensity=                              ; floating point value
Tint.VisibleToHouses=all                     ; list of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
FirepowerMultiplier=1.0                      ; floating point value
ArmorMultiplier=1.0                          ; floating point value
SpeedMultiplier=1.0                          ; floating point value
ROFMultiplier=1.0                            ; floating point value
ROFMultiplier.ApplyOnCurrentTimer=true       ; boolean
Cloakable=false                              ; boolean
ForceDecloak=false                           ; boolean
WeaponRange.Multiplier=1.0                   ; floating point value
WeaponRange.ExtraRange=0.0                   ; floating point value
WeaponRange.AllowWeapons=                    ; list of WeaponTypes
WeaponRange.DisallowWeapons=                 ; list of WeaponTypes
Crit.Multiplier=1.0                          ; floating point value
Crit.ExtraChance=0.0                         ; floating point value
Crit.AllowWarheads=                          ; list of WarheadTypes
Crit.DisallowWarheads=                       ; list of WarheadTypes
RevengeWeapon=                               ; WeaponType
RevengeWeapon.AffectsHouses=all              ; list of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
DisableWeapons=false                         ; boolean
Groups=                                      ; comma-separated list of strings (group IDs)

[SOMETECHNO]                                 ; TechnoType
AttachEffect.AttachTypes=                    ; List of AttachEffectTypes
AttachEffect.DurationOverrides=              ; integer - duration overrides (comma-separated) for AttachTypes in order from first to last.
AttachEffect.Delays=                         ; integer - delays (comma-separated) for AttachTypes in order from first to last.
AttachEffect.InitialDelays=                  ; integer - initial delays (comma-separated) for AttachTypes in order from first to last.
AttachEffect.RecreationDelays=               ; integer - recreation delays (comma-separated) for AttachTypes in order from first to last.
OpenTopped.UseTransportRangeModifiers=false  ; boolean

[SOMEWEAPON]                                 ; WeaponType
AttachEffect.RequiredTypes=                  ; List of AttachEffectTypes
AttachEffect.DisallowedTypes=                ; List of AttachEffectTypes
AttachEffect.RequiredGroups=                 ; comma-separated list of strings (group IDs)
AttachEffect.DisallowedGroups=               ; comma-separated list of strings (group IDs)
AttachEffect.RequiredMinCounts=              ; integer - minimum required instance count (comma-separated) for cumulative types in order from first to last.
AttachEffect.RequiredMaxCounts=              ; integer - maximum required instance count (comma-separated) for cumulative types in order from first to last.
AttachEffect.DisallowedMinCounts=            ; integer - minimum disallowed instance count (comma-separated) for cumulative types in order from first to last.
AttachEffect.DisallowedMaxCounts=            ; integer - maximum disallowed instance count (comma-separated) for cumulative types in order from first to last.
AttachEffect.IgnoreFromSameSource=false      ; boolean

[SOMEWARHEAD]
AttachEffect.AttachTypes=                    ; List of AttachEffectTypes
AttachEffect.RemoveTypes=                    ; List of AttachEffectTypes
AttachEffect.RemoveGroups=                   ; comma-separated list of strings (group IDs)
AttachEffect.CumulativeRemoveMinCounts=      ; integer - minimum required instance count (comma-separated) for cumulative types in order from first to last.
AttachEffect.CumulativeRemoveMaxCounts=      ; integer - maximum removed instance count (comma-separated) for cumulative types in order from first to last.
AttachEffect.DurationOverrides=              ; integer - duration overrides (comma-separated) for AttachTypes in order from first to last.
```

### Custom tint on TechnoTypes

- A tint effect similar to that used by Iron Curtain / Force Shield or `Psychedelic=true` Warheads can be applied to TechnoTypes naturally by setting `Tint.Color` and/or `Tint Intensity`.
  - `Tint.Intensity` is additive lighting increase/decrease - 1.0 is the default object lighting.
  - `Tint.VisibleToHouses` can be used to customize which houses can see the tint effect.

In `rulesmd.ini`:
```ini
Tint.Color=               ; integer - R,G,B
Tint.Intensity=0.0        ; floating point value
Tint.VisibleToHouses=all  ; list of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
```
Theres also a couple of tinting-related bugfixes:

- Tint effects are now correctly applied to SHP vehicles and all types of aircraft as well as building animations regardless of their position.
- Iron Curtained / Force Shielded objects now always use the correct tint color.

-----------------

### Revenge weapon

- Similar to `DeathWeapon` in that it is fired after a TechnoType is killed, but with the difference that it will be fired on whoever dealt the damage that killed the TechnoType. If TechnoType died of sources other than direct damage dealt by another TechnoType, `RevengeWeapon` will not be fired.
  - `RevengeWeapon.AffectsHouses` can be used to filter which houses the damage that killed the TechnoType is allowed to come from to fire the weapon.

In `rulesmd.ini`:
```ini
RevengeWeapon=                  ; WeaponType
RevengeWeapon.AffectsHouses=all ; list of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
```

### DIFF
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -14,6 +14,7 @@ This page describes all the engine features that are either new and introduced b
     - `entry`: Discard on exiting the map when entering transports or buildings etc.
     - `move`: Discard when the object the effect is attached on moves. Ignored if the object is a building.
     - `stationary`: Discard when the object the effect is attached on stops moving. Ignored if the object is a building.
+    - `drain`: Discard when the object is being affected by a weapon with `DrainWeapon=true`.
   - If `PenetratesIronCurtain` is not set to true, the effect is not applied on currently invulnerable objects (Iron Curtain / Force Shield).
   - `Animation` defines animation to play in an indefinite loop for as long as the effect is active on the object it is attached to.
     - If `Animation.ResetOnReapply` is set to true, the animation playback is reset every time the effect is applied if `Cumulative=false`.
@@ -60,7 +61,7 @@ Duration=0                                   ; integer - game frames or negative
 Cumulative=false                             ; boolean
 Cumulative.MaxCount=-1                       ; integer
 Powered=false                                ; boolean
-DiscardOn=none                               ; list of discard condition enumeration (none|entry|move|stationary)
+DiscardOn=none                               ; list of discard condition enumeration (none|entry|move|stationary|drain)
 PenetratesIronCurtain=false                  ; boolean
 Animation=                                   ; Animation
 Animation.ResetOnReapply=false               ; boolean
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -26,8 +26,8 @@ This page describes all the engine features that are either new and introduced b
   - `WeaponRange.Multiplier` and `WeaponRange.ExtraRange` can be used to multiply the weapon firing range of the object the effect is attached to, or give it an increase / decrease (measured in cells), respectively. `ExtraRange` is cumulatively applied from all attached effects after all `Multiplier` values have been applied.
     - `WeaponRange.AllowWeapons` can be used to list only weapons that can benefit from this range bonus and `WeaponRange.DisallowWeapons` weapons that are not allowed to, respectively.
     - On TechnoTypes with `OpenTopped=true`, `OpenTopped.UseTransportRangeModifiers` can be set to true to make passengers firing out use the transport's active range bonuses instead.
-  - `CritMultiplier` can be used to multiply the [critical hit](#chance-based-extra-damage-or-warhead-detonation--critical-hits) chance of the object the effect is attached to.
-    - `CritMultiplier.AllowWarheads` can be used to list only Warheads that can benefit from this critical hit chance multiplier and `CritMultiplier.DisallowWarheads` weapons that are not allowed to, respectively.
+  - `Crit.Multiplier` and `Crit.ExtraChance` can be used to multiply the [critical hit](#chance-based-extra-damage-or-warhead-detonation--critical-hits) chance or grant a fixed bonus to it for the object the effect is attached to, respectively.
+    - `Crit.AllowWarheads` can be used to list only Warheads that can benefit from this critical hit chance multiplier and `Crit.DisallowWarheads` weapons that are not allowed to, respectively.
   - `RevengeWeapon` can be used to temporarily grant the specified weapon as a [revenge weapon](#revenge-weapon) for the attached object.
     - `RevengeWeapon.AffectsHouses` customizes which houses can trigger the revenge weapon.
 
@@ -76,9 +76,10 @@ WeaponRange.Multiplier=1.0                   ; floating point value
 WeaponRange.ExtraRange=0.0                   ; floating point value
 WeaponRange.AllowWeapons=                    ; list of WeaponTypes
 WeaponRange.DisallowWeapons=                 ; list of WeaponTypes
-CritMultiplier=1.0                           ; floating point value
-CritMultiplier.AllowWarheads=                ; list of WarheadTypes
-CritMultiplier.DisallowWarheads=             ; list of WarheadTypes
+Crit.Multiplier=1.0                          ; floating point value
+Crit.ExtraChance=0.0                         ; floating point value
+Crit.AllowWarheads=                          ; list of WarheadTypes
+Crit.DisallowWarheads=                       ; list of WarheadTypes
 RevengeWeapon=                               ; WeaponType
 RevengeWeapon.AffectsHouses=all              ; list of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
 
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -33,6 +33,7 @@ This page describes all the engine features that are either new and introduced b
     - `Crit.AllowWarheads` can be used to list only Warheads that can benefit from this critical hit chance multiplier and `Crit.DisallowWarheads` weapons that are not allowed to, respectively.
   - `RevengeWeapon` can be used to temporarily grant the specified weapon as a [revenge weapon](#revenge-weapon) for the attached object.
     - `RevengeWeapon.AffectsHouses` customizes which houses can trigger the revenge weapon.
+  - `DisableWeapons` can be used to disable ability to fire any and all weapons.
 
 - AttachEffectTypes can be attached to TechnoTypes using `AttachEffect.AttachTypes`.
   - `AttachEffect.DurationOverrides` can be used to override the default durations. Duration matching the position in `AttachTypes` is used for that type, or the last listed duration if not available.
@@ -86,6 +87,7 @@ Crit.AllowWarheads=                          ; list of WarheadTypes
 Crit.DisallowWarheads=                       ; list of WarheadTypes
 RevengeWeapon=                               ; WeaponType
 RevengeWeapon.AffectsHouses=all              ; list of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
+DisableWeapons=false                         ; boolean
 
 [SOMETECHNO]                                 ; TechnoType
 AttachEffect.AttachTypes=                    ; List of AttachEffectTypes
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -23,8 +23,8 @@ This page describes all the engine features that are either new and introduced b
     - If `ROFMultiplier.ApplyOnCurrentTimer` is set to true, `ROFMultiplier` is applied on currently running reload timer (if any) when the effect is first applied.
   - If `Cloakable` is set to true, the object the effect is attached to is granted ability to cloak itself for duration of the effect.
   - `ForceDecloak`, if set to true, will uncloak and make the object the effect is attached to unable to cloak itself for duration of the effect.
-  - `WeaponRangeBonus` can be used to give the object the effect is attached an increase or decrease to weapon range, defined in cells.
-    - `WeaponRangeBonus.AllowWeapons` can be used to list only weapons that can benefit from this range bonus and `WeaponRangeBonus.DisallowWeapons` weapons that are not allowed to, respectively.
+  - `WeaponRange.Multiplier` and `WeaponRange.ExtraRange` can be used to multiply the weapon firing range of the object the effect is attached to, or give it an increase / decrease (measured in cells), respectively. `ExtraRange` is cumulatively applied from all attached effects after all `Multiplier` values have been applied.
+    - `WeaponRange.AllowWeapons` can be used to list only weapons that can benefit from this range bonus and `WeaponRange.DisallowWeapons` weapons that are not allowed to, respectively.
     - On TechnoTypes with `OpenTopped=true`, `OpenTopped.UseTransportRangeModifiers` can be set to true to make passengers firing out use the transport's active range bonuses instead.
   - `RevengeWeapon` can be used to temporarily grant the specified weapon as a [revenge weapon](#revenge-weapon) for the attached object.
     - `RevengeWeapon.AffectsHouses` customizes which houses can trigger the revenge weapon.
@@ -70,9 +70,10 @@ ROFMultiplier=1.0                            ; floating point value
 ROFMultiplier.ApplyOnCurrentTimer=true       ; boolean
 Cloakable=false                              ; boolean
 ForceDecloak=false                           ; boolean
-WeaponRangeBonus=0.0                         ; floating point value
-WeaponRangeBonus.AllowWeapons=               ; list of WeaponTypes
-WeaponRangeBonus.DisallowWeapons=            ; list of WeaponTypes
+WeaponRange.Multiplier=1.0                   ; floating point value
+WeaponRange.ExtraRange=0.0                   ; floating point value
+WeaponRange.AllowWeapons=                    ; list of WeaponTypes
+WeaponRange.DisallowWeapons=                 ; list of WeaponTypes
 RevengeWeapon=                               ; WeaponType
 RevengeWeapon.AffectsHouses=all              ; list of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
 
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -44,6 +44,7 @@ This page describes all the engine features that are either new and introduced b
 - Weapons can require attached effects on target to fire, or be prevented by firing if specific attached effects are applied.
   - `AttachEffect.RequiredTypes` can be used to list attached effects required to be on target to fire, all listed effect types must be present to allow firing.
   - `AttachEffect.DisallowedTypes` can be used to list attached effects that when present prevent the weapon from firing, any of the listed effect types will prevent firing if present.
+  - `AttachEffect.IgnoreFromSameSource` can be set to true to ignore effects that have been attached by the firer of the weapon and its Warhead.
 
 In `rulesmd.ini`:
 ```ini
@@ -94,6 +95,7 @@ OpenTopped.UseTransportRangeModifiers=false  ; boolean
 [SOMEWEAPON]                                 ; WeaponType
 AttachEffect.RequiredTypes=                  ; List of AttachEffectTypes
 AttachEffect.DisallowedTypes=                ; List of AttachEffectTypes
+AttachEffect.IgnoreFromSameSource=false      ; boolean
 
 [SOMEWARHEAD]
 AttachEffect.AttachTypes=                    ; List of AttachEffectTypes
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -10,7 +10,10 @@ This page describes all the engine features that are either new and introduced b
   - `Duration` determines how long the effect lasts for. It can be overriden by `DurationOverrides` on TechnoTypes and Warheads.
   - `Cumulative`, if set to true, allows the same type of effect to be applied on same object multiple times, up to `Cumulative.MaxCount` number or with no limit if `Cumulative.MaxCount` is a negative number.
   - `Powered` controls whether or not the effect is rendered inactive if the object it is attached to is deactivated (`PoweredUnit` or affected by EMP) or on low power. What happens to animation is controlled by `Animation.OfflineAction`.
-  - If `DiscardOnEntry` is set to true, the effect is removed when the object it is attached to exits map (enters transport or a building, etc).
+  - `DiscardOn` accepts a list of values corresponding to conditions where the attached effect should be discarded. Defaults to `none`, meaning it is never discarded.
+    - `entry`: Discard on exiting the map when entering transports or buildings etc.
+    - `move`: Discard when the object the effect is attached on moves. Ignored if the object is a building.
+    - `stationary`: Discard when the object the effect is attached on stops moving. Ignored if the object is a building.
   - If `PenetratesIronCurtain` is not set to true, the effect is not applied on currently invulnerable objects (Iron Curtain / Force Shield).
   - `Animation` defines animation to play in an indefinite loop for as long as the effect is active on the object it is attached to.
     - If `Animation.ResetOnReapply` is set to true, the animation playback is reset every time the effect is applied if `Cumulative=false`.
@@ -56,7 +59,7 @@ Duration=0                                   ; integer - game frames or negative
 Cumulative=false                             ; boolean
 Cumulative.MaxCount=-1                       ; integer
 Powered=false                                ; boolean
-DiscardOnEntry=false                         ; boolean
+DiscardOn=none                               ; list of discard condition enumeration (none|entry|move|stationary)
 PenetratesIronCurtain=false                  ; boolean
 Animation=                                   ; Animation
 Animation.ResetOnReapply=false               ; boolean
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -35,6 +35,8 @@ This page describes all the engine features that are either new and introduced b
   - `RevengeWeapon` can be used to temporarily grant the specified weapon as a [revenge weapon](#revenge-weapon) for the attached object.
     - `RevengeWeapon.AffectsHouses` customizes which houses can trigger the revenge weapon.
   - `DisableWeapons` can be used to disable ability to fire any and all weapons.
+  - It is possible to set groups for attach effect types by defining strings in `Groups`.
+    - Groups can be used instead of types for removing effects and weapon filters.
 
 - AttachEffectTypes can be attached to TechnoTypes using `AttachEffect.AttachTypes`.
   - `AttachEffect.DurationOverrides` can be used to override the default durations. Duration matching the position in `AttachTypes` is used for that type, or the last listed duration if not available.
@@ -44,11 +46,12 @@ This page describes all the engine features that are either new and introduced b
 
 - AttachEffectTypes can be attached to objects via Warheads using `AttachEffect.AttachTypes`.
   - `AttachEffect.DurationOverrides` can be used to override the default durations. Duration matching the position in `AttachTypes` is used for that type, or the last listed duration if not available.
-  - Attached effects can be removed from objects by Warheads using `AttachEffect.RemoveTypes`.
+  - Attached effects can be removed from objects by Warheads using `AttachEffect.RemoveTypes` or `AttachEffect.RemoveGroups`.
 
 - Weapons can require attached effects on target to fire, or be prevented by firing if specific attached effects are applied.
   - `AttachEffect.RequiredTypes` can be used to list attached effects required to be on target to fire, all listed effect types must be present to allow firing.
   - `AttachEffect.DisallowedTypes` can be used to list attached effects that when present prevent the weapon from firing, any of the listed effect types will prevent firing if present.
+  - `AttachEffect.Required/DisallowedGroups` have the same effect except applied with/to all types that have one of the listed groups in their `Groups` listing.
   - `AttachEffect.IgnoreFromSameSource` can be set to true to ignore effects that have been attached by the firer of the weapon and its Warhead.
 
 In `rulesmd.ini`:
@@ -89,6 +92,7 @@ Crit.DisallowWarheads=                       ; list of WarheadTypes
 RevengeWeapon=                               ; WeaponType
 RevengeWeapon.AffectsHouses=all              ; list of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
 DisableWeapons=false                         ; boolean
+Groups=                                      ; comma-separated list of strings (group IDs)
 
 [SOMETECHNO]                                 ; TechnoType
 AttachEffect.AttachTypes=                    ; List of AttachEffectTypes
@@ -101,11 +105,14 @@ OpenTopped.UseTransportRangeModifiers=false  ; boolean
 [SOMEWEAPON]                                 ; WeaponType
 AttachEffect.RequiredTypes=                  ; List of AttachEffectTypes
 AttachEffect.DisallowedTypes=                ; List of AttachEffectTypes
+AttachEffect.RequiredGroups=                 ; comma-separated list of strings (group IDs)
+AttachEffect.DisallowedGroups=               ; comma-separated list of strings (group IDs)
 AttachEffect.IgnoreFromSameSource=false      ; boolean
 
 [SOMEWARHEAD]
 AttachEffect.AttachTypes=                    ; List of AttachEffectTypes
 AttachEffect.RemoveTypes=                    ; List of AttachEffectTypes
+AttachEffect.RemoveGroups=                   ; comma-separated list of strings (group IDs)
 AttachEffect.DurationOverrides=              ; integer - duration overrides (comma-separated) for AttachTypes in order from first to last.
 ```
 
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -22,6 +22,9 @@ This page describes all the engine features that are either new and introduced b
     - `Animation.TemporalAction` determines what happens to the animation when the attached object is under effect of `Temporal=true` Warhead.
     - `Animation.UseInvokerAsOwner` can be used to set the house and TechnoType that created the effect (e.g firer of the weapon that applied it) as the animation's owner & invoker instead of the object the effect is attached to.
   - `CumulativeAnimations` can be used to declare a list of animations used for `Cumulative=true` types instead of `Animation`. An animation is picked from the list in order matching the number of active instances of the type on the object, with last listed animation used if number is higher than the number of listed animations. This animation is only displayed once, on the first active instance of the effect found attached and is updated and restarted if the number of active instances changed.
+  - Attached effect can fire off a weapon when expired / removed / object dies by setting `ExpireWeapon`.
+    - `ExpireWeapon.TriggerOn` determines the exact conditions upon which the weapon is fired, defaults to `expire` which means only if the effect naturally expires.
+    - `ExpireWeapon.CumulativeOnlyOnce`, if set to true, makes it so that `Cumulative=true` attached effects only detonate the weapon once period, instead of once per active instance. On `remove` and `expire` condition this means it will only detonate after last instance has expired or been removed.
   - `Tint.Color` & `Tint.Intensity` can be used to set a color tint effect and additive lighting increase/decrease on the object the effect is attached to, respectively.
     - `Tint.VisibleToHouses` can be used to control which houses can see the tint effect.
   - `FirepowerMultiplier`, `ArmorMultiplier`, `SpeedMultiplier` and `ROFMultiplier` can be used to modify the object's firepower, armor strength, movement speed and weapon reload rate, respectively.
@@ -76,6 +79,9 @@ Animation.OfflineAction=Hides                ; AttachedAnimFlag (None, Hides, Te
 Animation.TemporalAction=None                ; AttachedAnimFlag (None, Hides, Temporal, Paused or PausedTemporal)
 Animation.UseInvokerAsOwner=false            ; boolean
 CumulativeAnimations=                        ; list of animations
+ExpireWeapon=
+ExpireWeapon.TriggerOn=expire                ; List of expire weapon trigger condition enumeration (none|expire|remove|death|all)
+ExpireWeapon.CumulativeOnlyOnce=false        ; boolean
 Tint.Color=                                  ; integer - R,G,B
 Tint.Intensity=                              ; floating point value
 Tint.VisibleToHouses=all                     ; list of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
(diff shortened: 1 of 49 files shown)


src/Ext/Techno/Body.h
```diff
@@ -118,4 +124,6 @@ class TechnoExt
 	static void SyncIronCurtainStatus(TechnoClass* pFrom, TechnoClass* pTo);
 	static CoordStruct PassengerKickOutLocation(TechnoClass* pThis, FootClass* pPassenger, int maxAttempts);
 	static WeaponTypeClass* GetDeployFireWeapon(TechnoClass* pThis, int& weaponIndex);
+	static int GetTintColor(TechnoClass* pThis, bool invulnerability, bool airstrike, bool berserk);
```
secsome, line 127: Call it `GetVanillaTintColor` should be better?
And `GetTintColor` should be the wrapper calling `GetVanillaTintColor` and `GetCustomTintColor`?

src/Utilities/GeneralUtils.cpp
```diff
@@ -127,3 +127,29 @@ bool GeneralUtils::ApplyTheaterSuffixToString(char* str)
...
+
+	if (colorIndex < 0 || colorIndex >= (sizeof(colorAdd) / sizeof(ColorStruct)))
+		return colorValue;
+
+	auto const& color = colorAdd[colorIndex];
+	int red = color.R;
+	int green = color.G;
+	int blue = color.B;
+
+	if (Drawing::ColorMode() == RGBMode::RGB565)
```
secsome, line 146: Just call `Drawing::RGB_To_Int` I guess

Starkku: So what I think `Drawing::RGB_To_Int` actually does is convert RGB888 color to whatever 16-bit format is currently in use by the game, practically always RGB565. The thing is, `[ColorAdd]` values are already in 16-bit RGB format (I think it is specifically RGB565 but not certain), but I didn't exactly have the conversion algo to shove the values into an integer at hand so I just took what game does with them.

src/Ext/Techno/Body.h
Starkku: Perhaps.

Falaka21:
Do you remember that,  years ago, "Morale,Bravery,Panic" system request by someone that I can't remember the nickname but , "EvilRenegade" himself was very proud of the idea when Ares (AlexB) active... With the 'negative'verses on WHs, this could be awesome for "killer" units ,as well.... (while killer gets temporal AE improvements, dying units'  specific   'DeathWeapon'  can cause its comrades some negative AEs !    "DP" already has this ,but I am proud of Phobos has its own tags now...

src/Ext/Building/Hooks.cpp
```diff
@@ -341,3 +342,19 @@ DEFINE_HOOK(0x4575A2, BuildingClass_Infiltrate_AfterAres, 0xE)
 	return 0;
 }
 #undef INFILTRATE_HOOK_MAGIC
+
+DEFINE_HOOK(0x4519A2, BuildingClass_UpdateAnim_SetParentBuilding, 0x6)
```
chaserli, line 346: We can merge this and the laser trail thing before the entire PR

ZYJYGE:
I tested the feature of "tint effects applied by Warheads". It worked correctly in normal situation but didn't work in the warhead of radiation like the default "RadSite" and other custom warheads for radiation.

Starkku:
Massively reworked the PR by basically reimplementing AttachEffect as explicit types.

- This system should be able to coexist with Ares' AttachEffect without a hitch, there might be some edge cases I have not caught though.
- `ROFMultiplier` is by default also applied on currently running reload timer here unlike with Ares' version. This can be customized via `ROFMultplier.ApplyOnCurrentTimer`.
- Note that `ForceDecloak` works differently here (it keeps the objects uncloaked for duration of the effect instead of uncloaking once when applied), Ares' implementation is not particularly useful.
- `WeaponRangeBonus` and weapons requiring or being blocked by target's AttachEffects are new logic I decided to throw in, should allow for some interesting things.

Due to the large amount of `TechnoExt::ExtMap.Find` calls necessiated by the custom tint logic, weapon range bonus as well AttachEffect as whole, this PR will likely run at less than optimal performance until that system is reworked.

chaserli:
Holy shit, the entire AutoDeath logic should be merged into this framework eventually

src/Ext/Techno/Body.h
```diff
@@ -33,13 +34,23 @@ class TechnoExt
 		OptionalStruct<int, false> DamageNumberOffset;
 		OptionalStruct<int, true> CurrentLaserWeaponIndex;
 		bool IsInTunnel;
+		bool IsBurrowed;
 		CDTimerClass DeployFireTimer;
 		bool ForceFullRearmDelay;
+		std::vector<std::unique_ptr<AttachEffectClass>> AttachedEffects;
```
chaserli, line 40: why a vector of pointers instead of instances directly, like Ares?

Starkku:
Rebased on develop, as well as a couple of other changes.

- The weapon range modifier keys have been renamed to accomodate addition of a multiplier alongside the 'fixed' extra range bonus.
- Added critical hit chance multiplier.

Starkku:
Rebased on develop again, as well as introduced fixed critical hit chance bonus effect. Note that rest of the critical hit modifier keys have been renamed as well.

Metadorius:
Before I forget, a thought for the future (not necessarily this PR). To unify the attachability of different things what I think maybe the best approach would be splitting the concept of an AttachEffect into two entities:
- the component that can be attached (in this case the effect, the shield as well, maybe stuff like LaserTrails, interceptors etc.), without the lifetime/attachment/detachment conditions etc., just the component itself which won't have any self-management logic
- the attacher itself, which could manage multiple components at once using timers, conditions for attachment/detachment/activation/deactivation/etc.

Up to further discussion.

reedom114514:
There is a bug, when using these AEs on mobile buildings which is packing into vehicles, the game freezes.

Starkku:
> There is a bug, when using these AEs on mobile buildings which is packing into vehicles, the game freezes.

Good catch, should be fixed now.

reedom114514:
> > There is a bug, when using these AEs on mobile buildings which is packing into vehicles, the game freezes.
> 
> Good catch, should be fixed now.

After bug fix, there is another one about it: when building affected by new AE turns (in my test it is a SODAR) to vehicle, and adjust game speed in this step, after 1-2 sec the game freezes,  EIP:5B767260

Starkku:
> > > There is a bug, when using these AEs on mobile buildings which is packing into vehicles, the game freezes.
> > 
> > 
> > Good catch, should be fixed now.
> 
> After bug fix, there is another one about it: when building affected by new AE turns (in my test it is a SODAR) to vehicle, and adjust game speed in this step, after 1-2 sec the game freezes, EIP:5B767260

Is this on the most up to date build? For some reason the github-actions bot isn't updating the links correctly right now.

Here's the link https://github.com/Phobos-developers/Phobos/suites/13202812824/artifacts/718933906

reedom114514:
> > > > There is a bug, when using these AEs on mobile buildings which is packing into vehicles, the game freezes.
> > > 
> > > 
> > > Good catch, should be fixed now.
> > 
> > 
> > After bug fix, there is another one about it: when building affected by new AE turns (in my test it is a SODAR) to vehicle, and adjust game speed in this step, after 1-2 sec the game freezes, EIP:5B767260
> 
> Is this on the most up to date build? For some reason the github-actions bot isn't updating the links correctly right now.
> 
> Here's the link https://github.com/Phobos-developers/Phobos/suites/13202812824/artifacts/718933906

Oh, not the newest one, the one I downloaded is about 9 hours earlier
Ok, I shall try

reedom114514:
> > > > There is a bug, when using these AEs on mobile buildings which is packing into vehicles, the game freezes.
> > > 
> > > 
> > > Good catch, should be fixed now.
> > 
> > 
> > After bug fix, there is another one about it: when building affected by new AE turns (in my test it is a SODAR) to vehicle, and adjust game speed in this step, after 1-2 sec the game freezes, EIP:5B767260
> 
> Is this on the most up to date build? For some reason the github-actions bot isn't updating the links correctly right now.
> 
> Here's the link https://github.com/Phobos-developers/Phobos/suites/13202812824/artifacts/718933906

Oh, not the newest one, the one I downloaded is about 9 hours earlier
Ok, I shall try

reedom114514:
> > > > There is a bug, when using these AEs on mobile buildings which is packing into vehicles, the game freezes.
> > > 
> > > 
> > > Good catch, should be fixed now.
> > 
> > 
> > After bug fix, there is another one about it: when building affected by new AE turns (in my test it is a SODAR) to vehicle, and adjust game speed in this step, after 1-2 sec the game freezes, EIP:5B767260
> 
> Is this on the most up to date build? For some reason the github-actions bot isn't updating the links correctly right now.
> 
> Here's the link https://github.com/Phobos-developers/Phobos/suites/13202812824/artifacts/718933906

Oh, not the newest one, the one I downloaded is about 9 hours earlier
Ok, I shall try

reedom114514:
> > > > There is a bug, when using these AEs on mobile buildings which is packing into vehicles, the game freezes.
> > > 
> > > 
> > > Good catch, should be fixed now.
> > 
> > 
> > After bug fix, there is another one about it: when building affected by new AE turns (in my test it is a SODAR) to vehicle, and adjust game speed in this step, after 1-2 sec the game freezes, EIP:5B767260
> 
> Is this on the most up to date build? For some reason the github-actions bot isn't updating the links correctly right now.
> 
> Here's the link https://github.com/Phobos-developers/Phobos/suites/13202812824/artifacts/718933906

With the newest one there is still a FE, EIP:5B33B669, randomly during packing/unpacking the SODAR, Firepower/Armor/Speed/Range

Starkku:
> With the newest one there is still a FE, EIP:5B33B669, randomly during packing/unpacking the SODAR, Firepower/Armor/Speed/Range

Could you provide the exact AttachEffect code used in this case? Mostly the duration and anything related is of interest to me because while I've been able to reproduce this crash, it isn't reliable and it occurs really rarely for me.

Starkku:
https://github.com/Phobos-developers/Phobos/suites/13280697552/artifacts/724636536

Refactored attaching/detaching code and changed how transfering works, I was unable to reproduce the crash afterwards.

reedom114514:
> https://github.com/Phobos-developers/Phobos/suites/13280697552/artifacts/724636536
> 
> Refactored attaching/detaching code and changed how transfering works, I was unable to reproduce the crash afterwards.

After testing I cannot reproduce this too, but in multi-player the RE happens (using this PR branch), shall I post that on the other feedback?

Starkku:
`DiscardOnEntry` has been deprecated. Use the new key `DiscardOn` (f.ex `DiscardOn=entry`) instead, it also accepts to different values for discarding when object moves or becomes stationary, respectively.

Additionally fixed an issue with discarded attacheffects not being recreated even if they should after exiting from limbo.

Metadorius:
Means they are exclusive now? Maybe different tags or enum flags would work better for combination of different conditions?

Starkku:
> Means they are exclusive now? Maybe different tags or enum flags would work better for combination of different conditions?

I worded it badly, it takes a list of values

src/New/Entity/AttachEffectClass.cpp
```diff
@@ -0,0 +1,680 @@
...
+	Type { pType }, Techno { pTechno }, InvokerHouse { pInvokerHouse }, Invoker { pInvoker }, Source { pSource },
+	DurationOverride { durationOverride }, Delay { delay }, InitialDelay { initialDelay }, RecreationDelay { recreationDelay }
+	, Duration { 0 }
+	, CurrentDelay { 0 }
+	, Animation { nullptr }
+	, IsAnimHidden { false }
+	, IsUnderTemporal { false }
+	, IsOnline { true }
+	, IsCloaked { false }
+	, NeedsDurationRefresh { false }
```
Metadorius, line 32: Same question here. Also I think `:` should be on newline, it's easy to lose it visually at the end of the line.

src/New/Entity/AttachEffectClass.cpp
```diff
@@ -0,0 +1,680 @@
...
+}
+
+void AttachEffectClass::CloakCheck()
+{
+	const auto cloakState = this->Techno->CloakState;
+
+	if (cloakState == CloakState::Cloaked || cloakState == CloakState::Cloaking)
+	{
+		this->IsCloaked = true;
+		KillAnim();
```
Metadorius, line 250:
```suggestion
		this->KillAnim();
```

src/New/Entity/AttachEffectClass.cpp
```diff
@@ -0,0 +1,680 @@
...
+}
+
+bool AttachEffectClass::IsSelfOwned() const
+{
+	return this->Source == this->Techno;
+}
+
+bool AttachEffectClass::HasExpired() const
+{
+	return IsSelfOwned() && this->Delay >= 0 ? false : !this->Duration;
```
Metadorius, line 333:
```suggestion
	return this->IsSelfOwned() && this->Delay >= 0 ? false : !this->Duration;
```

src/New/Entity/AttachEffectClass.cpp
```diff
@@ -0,0 +1,680 @@
...
+
+bool AttachEffectClass::Attach(AttachEffectTypeClass* pType, TechnoClass* pTarget, HouseClass* pInvokerHouse, TechnoClass* pInvoker,
+	AbstractClass* pSource, int durationOverride, int delay, int initialDelay, int recreationDelay)
+{
+	if (!pType || !pTarget)
+		return false;
+
+	auto const pTargetExt = TechnoExt::ExtMap.Find(pTarget);
+
+	if (auto const pAE = CreateAndAttach(pType, pTarget, pTargetExt->AttachedEffects, pInvokerHouse, pInvoker, pSource, durationOverride, delay, initialDelay, recreationDelay))
```
Metadorius, line 381: Is CreateAndAttach a static function in this class or what? IMO all class function calls should have a qualifier for clarity.

src/New/Entity/AttachEffectClass.cpp
```diff
@@ -0,0 +1,680 @@
...
+		if (delays.size() > 0)
+			delay = delays[delays.size() > i ? i : delays.size() - 1];
+
+		if (initialDelays.size() > 0)
+			initialDelay = initialDelays[initialDelays.size() > i ? i : initialDelays.size() - 1];
+
+		if (recreationDelays.size() > 0)
+			recreationDelay = recreationDelays[recreationDelays.size() > i ? i : recreationDelays.size() - 1];
+
+		if (auto const pAE = CreateAndAttach(pType, pTarget, pTargetExt->AttachedEffects, pInvokerHouse, pInvoker, pSource, durationOverride, delay, initialDelay, recreationDelay))
```
Metadorius, line 435: Same here

src/New/Entity/AttachEffectClass.cpp
```diff
@@ -0,0 +1,680 @@
...
+}
+
+bool AttachEffectClass::IsFromSource(TechnoClass* pInvoker, AbstractClass* pSource) const
+{
+	return pInvoker == this->Invoker && pSource == this->Source;
+}
+
+AttachEffectTypeClass* AttachEffectClass::GetType() const
+{
+	return Type;
```
Metadorius, line 370:
```suggestion
	return this->Type;
```

src/New/Entity/AttachEffectClass.cpp
```diff
@@ -0,0 +1,680 @@
...
+}
+
+bool AttachEffectClass::Attach(AttachEffectTypeClass* pType, TechnoClass* pTarget, HouseClass* pInvokerHouse, TechnoClass* pInvoker,
```
Metadorius, line 373: I'd add a separate section (pragma region?) for statics and other types of members for clarity.

src/New/Entity/AttachEffectClass.cpp
```diff
@@ -0,0 +1,680 @@
...
+		{
+			sourceMatch->Duration = Math::max(sourceMatch->Duration, attachEffect->Duration);
+		}
+		else if (!type->Cumulative && currentTypeCount > 0 && match)
+		{
+			match->Duration = Math::max(sourceMatch->Duration, attachEffect->Duration);
+		}
+		else
+		{
+			if (auto const pAE = CreateAndAttach(type, pTarget, pTargetExt->AttachedEffects, attachEffect->InvokerHouse, attachEffect->Invoker, attachEffect->Source, attachEffect->DurationOverride))
```
Metadorius, line 636: Also lacks qualifier.

src/New/Entity/ShieldClass.cpp
```diff
@@ -436,6 +436,9 @@ void ShieldClass::OnlineCheck()
 
 	if (!isActive)
 	{
+		if (this->Online)
+			UpdateTint();
```
Metadorius, line 440:
```suggestion
			this->UpdateTint();
```

src/New/Entity/ShieldClass.cpp
```diff
@@ -464,6 +467,9 @@ void ShieldClass::OnlineCheck()
 	}
 	else
 	{
+		if (!this->Online)
+			UpdateTint();
```
Metadorius, line 471:
```suggestion
			this->UpdateTint();
```

src/New/Entity/ShieldClass.cpp
```diff
@@ -511,6 +517,7 @@ bool ShieldClass::ConvertCheck()
 		this->KillAnim();
 		pTechnoExt->CurrentShieldType = ShieldTypeClass::FindOrAllocate(NONE_STR);
 		pTechnoExt->Shield = nullptr;
+		UpdateTint();
```
Metadorius, line 520:
```suggestion
		this->UpdateTint();
```

src/New/Entity/ShieldClass.cpp
```diff
@@ -553,6 +560,7 @@ bool ShieldClass::ConvertCheck()
 	}
 
 	this->TechnoID = newID;
+	UpdateTint();
```
Metadorius, line 563:
```suggestion
	this->UpdateTint();
```

src/New/Entity/ShieldClass.cpp
```diff
@@ -649,6 +657,8 @@ void ShieldClass::BreakShield(AnimTypeClass* pBreakAnim, WeaponTypeClass* pBreak
 
 	this->LastBreakFrame = Unsorted::CurrentFrame;
 
+	UpdateTint();
```
Metadorius, line 660:
```suggestion
	this->UpdateTint();
```

src/New/Entity/ShieldClass.cpp
```diff
@@ -663,6 +673,7 @@ void ShieldClass::RespawnShield()
 		timer->Stop();
 		double amount = timerWHModifier->InProgress() ? Respawn_Warhead : this->Type->Respawn;
 		this->HP = this->GetPercentageAmount(amount);
+		UpdateTint();
```
Metadorius, line 676:
```suggestion
		this->UpdateTint();
```

src/New/Type/AttachEffectTypeClass.h
```diff
@@ -0,0 +1,90 @@
...
+	Valueable<bool> DisableWeapons;
+
+	AttachEffectTypeClass(const char* const pTitle) : Enumerable<AttachEffectTypeClass>(pTitle)
+		, Duration { 0 }
+		, Cumulative { false }
+		, Cumulative_MaxCount { -1 }
+		, Powered { false }
+		, DiscardOn { DiscardCondition::None }
+		, PenetratesIronCurtain { false }
+		, Animation {}
```
Metadorius, line 50: Can't remember for sure if documented, did we have any consensus regarding `{}` vs `{ }`?

Metadorius (changes requested):

src/New/Entity/AttachEffectClass.cpp
```diff
@@ -0,0 +1,680 @@
...
+
+#include <Ext/Anim/Body.h>
+#include <Ext/Techno/Body.h>
+
+std::vector<AttachEffectClass*> AttachEffectClass::Array;
+
+AttachEffectClass::AttachEffectClass() : Type { nullptr }, Techno { nullptr }, InvokerHouse { nullptr }, Invoker { nullptr }, Source { nullptr }, DurationOverride { 0 }, Delay { 0 }, InitialDelay { 0 }, RecreationDelay { -1 }
+, Duration { 0 }
+, CurrentDelay { 0 }
+, NeedsDurationRefresh { false }
```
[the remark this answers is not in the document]
Starkku: VS document formatting on current settings does this.

src/New/Type/AttachEffectTypeClass.h
Starkku: Not in style guide and I cannot recall it being discussed.

docs/New-or-Enhanced-Logics.md
```diff
@@ -4,6 +4,111 @@ This page describes all the engine features that are either new and introduced b
...
+  - `AttachEffect.InitialDelays` can be used to set the delays before first creating the effects on TechnoType. Defaults to 0 (immediately). Delay matching the position in `AttachTypes` is used for that type, or the last listed delay if not available.
+  - `AttachEffect.RecreationDelays` is used to determine if the effect can be recreated if it is removed completely (e.g `AttachEffect.RemoveTypes`), and if yes, how long this takes. Defaults to -1, meaning no recreation. Delay matching the position in `AttachTypes` is used for that type, or the last listed delay if not available.
+
+- AttachEffectTypes can be attached to objects via Warheads using `AttachEffect.AttachTypes`.
+  - `AttachEffect.DurationOverrides` can be used to override the default durations. Duration matching the position in `AttachTypes` is used for that type, or the last listed duration if not available.
+  - Attached effects can be removed from objects by Warheads using `AttachEffect.RemoveTypes`.
```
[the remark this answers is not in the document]
Starkku: Maybe. It would require more lines in INI but might be clearer.

src/Ext/Techno/Body.h
```diff
@@ -34,15 +35,26 @@ class TechnoExt
 		AnimTypeClass* MindControlRingAnimType;
 		OptionalStruct<int, false> DamageNumberOffset;
 		bool IsInTunnel;
+		bool IsBurrowed;
 		bool HasBeenPlacedOnMap; // Set to true on first Unlimbo() call.
 		CDTimerClass DeployFireTimer;
 		bool ForceFullRearmDelay;
 		int WHAnimRemainingCreationInterval;
+		std::vector<std::unique_ptr<AttachEffectClass>> AttachedEffects;
```
chaserli, line 43:
```suggestion
		std::vector<AttachEffectClass> AttachedEffects;
```
[a container of instances](https://github.com/Ares-Developers/Ares/blob/master/src/Ext/Techno/Body.h#L64) of instances should suffice (

src/New/Type/AttachEffectTypeClass.cpp
```diff
@@ -0,0 +1,113 @@
+#include "AttachEffectTypeClass.h"
+
+Enumerable<AttachEffectTypeClass>::container_t Enumerable<AttachEffectTypeClass>::Array;
```
chaserli, line 3:
```suggestion
```

src/New/Type/AttachEffectTypeClass.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+bool AttachEffectTypeClass::HasTint()
+{
+	return this->Tint_Color.isset() || this->Tint_Intensity != 0.0;
+}
+
+const char* Enumerable<AttachEffectTypeClass>::GetMainSection()
```
chaserli, line 10:
```suggestion
template<>
const char* Enumerable<AttachEffectTypeClass>::GetMainSection()
```

chaserli (changes requested):

docs/New-or-Enhanced-Logics.md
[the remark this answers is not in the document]
Metadorius: Let's roll with that change then.

Metadorius (commented):
Considering it's a replacement for Ares' AttachEffect - it may be good to add a note stating what changes are required to migrate to Phobos one.

Starkku:
> Considering it's a replacement for Ares' AttachEffect - it may be good to add a note stating what changes are required to migrate to Phobos one.

It doesn't stop any of the Ares' AttachEffect features from working though.

Metadorius:
> It doesn't stop any of the Ares' AttachEffect features from working though. It simply offers new system alongside the old one.

Correct, I didn't mean this in fact. I mean it as just an upgrade scenario, if the user would want to migrate from Ares to Phobos AE and/or back, and we could point them to the note.

src/Ext/Techno/Body.h
Starkku: Not implementing for now due to amount of code refactoring required beyond the suggested change.

Starkku:
Added some additional features to AttachEffect, mostly things that have been requested from me (grouping & group filters, Cumulative stack filters/animations, expire weapon). I will not expand the current functionality any further in this PR and consider this PR finished, save for potential bugfixes.

Starkku:
> `AttachEffect.RequiredTypes=;` on weapons make the unit not able to attack the target until the AE is on it (as intended), but the weapon can still be fired on the ground, so in a way it's still able to damage said target if it moves to the ground cell being attacked. This can be "fixed" with `CanTarget=` but doesn't feel like it interacts well.

Noted but not the easiest 'problem' to solve, `CanTarget` itself also technically can have same issue depending on values.

> Having an AttachEffectType with an `ExpireWeapon=` that triggers another AttachEffectType crashes the game.

This needs a crash dump.

> `ExpireWeapon=` does not damage the owner/invoker, but does damage friendly units, would be good to control this. Effects such as (Ares AE) do get applied to the owner/invoker though.

Use `AllowDamageOnSelf` on Warhead, technos don't take damage from weapons/warheads that are considered to be fired by themselves normally.

> `Powered=true` does not work on low power as described, but it does work when the unit is disabled e.g. EMP

Will investigate.

> `Crit.ExtraChance` does not work if the Warhead has 0.0 crit to begin with, which it should, e.g.

Noted.

ayylmaoRotE:
>This needs a crash dump.
>
Attached.
https://1drv.ms/u/s!Ah86cVoN9dYikZJA5vSbP6yWYCRwJA?e=1uQJh9
>Cannot reproduce. This is technically not a 'low base power' check and it is not advertised as such in the documentation either, but rather simply if the techno it is attached to is 'offline' or not. For buildings that require power it does work as intended.
>
Oh yep that makes sense, didn't look into buildings

Will do MP testing later too

Starkku:
I could change it to use vector of instances like @chaserli suggested earlier, but besides having to go through the effort to refactor a bunch of code are there any other actual concerns I should take into account if I do that? Is it even a feasible approach for something like this, I know shields use `unique_ptr` so that is why I picked it in the first place. If I nuke the vector (e.g techno dies) will it automagically delete the objects contained as well even if I don't use `unique_ptr`?

Metadorius:
I personally think that for consistency sake it's better to use pointer vector for now, after all everything uses that. If there are benefits for instance vectors - the refactoring should be separate.

src/Ext/Techno/Hooks.Tint.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+	GET(BuildingClass*, pThis, ESI);
+
+	RulesClass* rules = RulesClass::Instance;
+
+	R->ECX(rules);
+	R->EAX(pThis->ForceShielded ? rules->ForceShieldColor : rules->IronCurtainColor);
+
+	return SkipGameCode;
+}
```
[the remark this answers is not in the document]
Metadorius: Okay nevermind, I missed a part of the changes. Would be good to have this and the next hooks not mixed with the rest but left in a separate block so they don't mix with unrelated changes.

PR #836, opened 2023-02-04
New TransferTypes for Warhead to transfer and convert resources between units

### LINKED ISSUE #670: Reload certain amount of ammo by being attacked by certain warheads
### First check

- [X] The suggestion is **NOT** related to functionality of Ares, HAres, Kratos or any other YR engine extension.
- [X] The suggested feature / enhancement is **NOT** implemented in Ares/Phobos yet and the existing possibilites are not viable or good enough to achieve the wanted result.
- [X] I agree to elaborate the details if requested and provide thorough testing if the feature is implemented.
- [X] I thought out how the feature / enhancement should work and described it in the most detailed and clear way.

### Description

yes

### INI code

```ini
[SomeSection]
SomeKey=SomeValue
```

### Additional context

_No response_

### LINKED ISSUE #692: New feature : Status Transfer
### First check

- [X] The suggestion is **NOT** related to functionality of Ares, HAres, Kratos or any other YR engine extension.
- [X] The suggested feature / enhancement is **NOT** implemented in Ares/Phobos yet and the existing possibilites are not viable or good enough to achieve the wanted result.
- [X] I agree to elaborate the details if requested and provide thorough testing if the feature is implemented.
- [X] I thought out how the feature / enhancement should work and described it in the most detailed and clear way.

### Description

Status will be transfer from a unit to another unit that fires the weapon.See example below:

### INI code

```ini
[TechnoTypes]
ImmuneToTransfer=no
Transfer.Overflows=yes                  ;For example:"Unit A" is at full health:750, its weapon has 
;"StatusTransfer.StrengthMultiplier=0.5" ;"Unit B" has a strength of 1000,and when it is hit by this "Unit A"'s warhead,it loses 
;500(half of Unit B's Strength), the loss strength will transfer to "Unit A" and "Unit A"'s Strength changes to 1250.

[SomeWarhead]
StatusTransfers=yes                      ;the units hit by this warhead lose listed abilities below
StatusTransfer.StrengthMultiplier=1     ;available for units and buildings
StatusTransfer.SpeedMultiplier=1        ;available for units
StatusTransfer.SightMultiplier=1        ;available for units and buildings
StatusTransfer.PowerMultiplier=1        ;available for buildings
StatusTransfer.AmmoCount=1              ;available for buildings
StatusTransfer.ROTMultiplier=1          ;available for units and buildings
StatusTransfer.Abilities=STRONGER,FIREPOWER,SCATTER,ROF,SIGHT,TIBERIUM_PROOF,VEIN_PROOF,SELF_HEAL,RADAR_INVISIBLE
,SENSORS,FEARLESS,TIBERIUM_HEAL,GUARD_AREA,FASTER,CLOAK,CRUSHER,C4,EXPLODES
                                        ;the units hit by this warhead lose listed abilities above
```

### Additional context

_No response_

Envoidener:
Resolves #57, resolve #168, resolve #670, resolve #692 by introducing TransferTypes


MortonPL (commented):
I believe that for a so complex feature, documentation should be more extensive than just an explanation of tags. There are many, *many* possible scenarios and tags affect each other a lot. Some examples or more images, descriptions more focusing on parts of the feature (i.e. minimums/maximums, value type conversions, splash) rather than just what each tag does. I know that other new features aren't explained like that, but I think it's necessary, so that the end users can understand it by reading.

I can think of several questions which aren't answered:
- What if you specify only sender or receiver?
- What if you use negative values?
- What if there is no target techno (i.e. by using force fire)?
- What if there is no source techno (i.e. by using superweapons, animation weapons, other sources of warheads)?
- What if the source techno is also one of the targets (i.e. detonating a splash warhead on self)?
- What is the relation between `Target.Spread.Distribution` and `Receive.Value.ProportionalToSent` for multiple targets?
- Why is `Receive.Value.ProportionalToSent` affected by `Receive.Value.PercentOfTotal`?
- How does conversion between value types respect `Send.RequireRealResource` and `Receive.RefuseOverflow`?
- Will the effect be triggered on a killing blow?
- Will sending all of sender's health cause death?

Other issues:
- TransferType documentation should be in "New types / ingame entities", not "Warheads"
- `Target.Warhead` isn't mentioned in text, only in tag listing
- Comments in tag listing should be indented

Envoidener:
Wanted to add more extensive documentation before, but seeing how it got this big with only this, decided to stop. Need some time then.

Envoidener:
> What if you use negative values?
It would take absolute value before, but after some thought this will be scrapped and some tags are changed to match this
> Will sending all of sender's health cause death?
It would leave minimal health before, but this too will be controlled by specific tag(s)
And regarding most of the other questions, will be restructuring this by only increasing control over special interactions

Envoidener:
"_I am finished with the building._".
Needs general review of code and documentation. Will add more GIFs.

src/Ext/Building/Hooks.IncomeAndSelling.cpp
```diff
@@ -57,7 +57,7 @@ DEFINE_HOOK(0x522E4F, InfantryClass_SlaveGiveMoney_CheckBalanceAfter, 0x6)
 	else if (auto pBldTypeExt = BuildingTypeExt::ExtMap.Find(slaveMiner->GetTechnoType()->DeploysInto))
 	{
 		if (pBldTypeExt->DisplayIncome.Get(RulesExt::Global()->DisplayIncome.Get()))
-			FlyingStrings::AddMoneyString(money, slaveMiner->Owner, RulesExt::Global()->DisplayIncome_Houses.Get(), slaveMiner->Location);
+			FlyingStrings::AddNumberString(money, slaveMiner->Owner, RulesExt::Global()->DisplayIncome_Houses.Get(), slaveMiner->Location);
```
chaserli, line 60: No. Please. Leave my money string alone and write your own debug string helpers. Check or modify `TechnoExt::DisplayDamageNumberString`

Envoidener: It's just rename

chaserli: Do you realize that positive and negative cash amounts use different colors? Even if they didn't, I insisted so because they are for different purposes.

Envoidener:
Closing this, because it was/is being rewritten for special reasons.

PR #839, opened 2023-02-08
Real time timers

MortonPL:
### Real time timers

- Timers can now display values in real time, taking game speed into account. This can be enabled with `RealTimeTimers=yes`.
- By default, time is calculated relative to desired framerate. Enabling `RealTimeTimers.Adaptive=yes` (always on for unlimited FPS) will calculate time relative to *current* FPS, accounting for lag.
  - When playing with unlimited FPS (or custom speed above 60 FPS), the timers might constantly change value because of the unstable nature.
- This option respects custom game speeds.

In `ra2md.ini`:
```ini
[Phobos]
RealTimeTimers=no           ; boolean
RealTimeTimers.Adaptive=no  ; boolean
```

### DIFF
--- a/src/Misc/Hooks.Timers.cpp
+++ b/src/Misc/Hooks.Timers.cpp
@@ -2,16 +2,20 @@
 #include <GameOptionsClass.h>
 #include <FPSCounter.h>
 #include <SessionClass.h>
+#include <Phobos.h>
 
 int oldValue;
 DEFINE_HOOK(0x6D4B50, Print_Timer_On_Tactical_Start, 0x6)
 {
+	if (!Phobos::Config::RealTimeTimers)
+		return 0;
+
 	REF_STACK(int, value, STACK_OFFSET(0, 0x4));
 	oldValue = value;
 
 	if (false && !SessionClass::IsMultiplayer()) // TODO change when custom game speed gets merged
 	{
-		value = std::max(value / (int)(std::max((double)FPSCounter::CurrentFrameRate, 1.0) / 15.0), 1);
+		value = (int)((double)value / (std::max((double)FPSCounter::CurrentFrameRate, 1.0) / 15.0));
 		return 0;
 	}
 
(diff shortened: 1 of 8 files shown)


src/Misc/Hooks.Timers.cpp
```diff
@@ -0,0 +1,58 @@
+#include <Utilities/Macro.h>
+#include <GameOptionsClass.h>
+#include <FPSCounter.h>
+#include <SessionClass.h>
+#include <Phobos.h>
+
+int oldValue;
```
secsome, line 7: NEVER pollute the namespace

MortonPL: Then, is this okay?

```
namespace TimerValueTemp
{
	static int oldValue;
};
```

src/Misc/PhobosToolTip.cpp
```diff
@@ -93,6 +94,19 @@ void PhobosToolTip::HelpText(BuildType& cameo)
...
 
+inline int tickTimeToSeconds(int tickTime)
+{
+	if (!Phobos::Config::RealTimeTimers)
+		return tickTime / 15;
+
+	if (Phobos::Config::RealTimeTimers_Adaptive
+		|| GameOptionsClass::Instance->GameSpeed == 0
+		|| (false && !SessionClass::IsMultiplayer())) // TODO change when custom game speed gets merged
+		return tickTime / std::max((int)FPSCounter::CurrentFrameRate, 1);
```
Metadorius, line 105:
```suggestion
	{
		return tickTime / std::max((int)FPSCounter::CurrentFrameRate, 1);
	}
```

Metadorius (commented):
How many frames does the FPS timer take into account? I think for adaptive game speed it would be beneficial to sample the framerate for a second or a few and then take the average frame rate for the last few seconds, otherwise the time will jitter *a lot*.

src/Misc/PhobosToolTip.cpp
MortonPL: `FPSCounter::CurrentFrameRate` comment says "The number of frames processed in the last second", which seems to be correct (?) when looking at decompiled code.

PR #369, opened 2021-09-18
TechnoType conversion warhead

MortonPL:
### Convert TechnoType

- Superweapons can now change TechnoTypes of affected units to other Types in the same category (infantry to infantry, vehicles to vehicles, aircraft to aircraft). This feature has the same limitations as [Ares' Type Conversion](https://ares-developers.github.io/Ares-docs/new/typeconversion.html). ***This feature does not support BuildingTypes.***
  - `Convert.From` specifies which TechnoTypes are valid for conversion. This entry can have many types listed, meanging that many types will be converted at once. When no types are included, conversion will affect all valid targets.
  - `Convert.To` specifies which TechnoTypes are the result of converion. This entry should have the same length as `Convert.From` (when the latter is not empty), because entries are grouped in pairs.
  - `Convert.AffectedHouses` specifies whose units can be converted.
  - Conversion affects *all* existing units of set TechnoTypes, this includes units in: transports, occupied buildings, buildings with `InfantryAbsorb=yes` or `UnitAbsorb=yes`, buildings with `Bunker=yes`.

In example, this warhead would convert all owned and friendly SOLDIERA to SOLDIERB and TANKA to TANKB.
```ini
[SOMESW]
Convert.From=SOLDIERA,TANKA
Convert.To=SOLDIERB,TANKB
Convert.AffectedHouses=team
```

```{warning}
This feature requires Ares 3.0 or higher to function! If you're not using Ares 3.0+, nothing will happen.
```

In `rulesmd.ini`:
```ini
[SOMEWARHEAD]
Convert.From=                   ; list of tehnoTypes
Convert.To=                     ; list of TechnoTypes
Convert.AffectedHouses=owner    ; list of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
```

### Convert TechnoType on impact

- Warheads can now change TechnoTypes of affected units to other Types in the same category (infantry to infantry, vehicles to vehicles, aircraft to aircraft). This feature has the same limitations as [Ares' Type Conversion](https://ares-developers.github.io/Ares-docs/new/typeconversion.html). ***This feature does not support BuildingTypes.***
  - `Convert.From` specifies which TechnoTypes are valid for conversion. This entry can have many types listed, meanging that many types will be converted at once. When no types are included, conversion will affect all valid targets.
  - `Convert.To` specifies which TechnoTypes are the result of converion. This entry should have the same length as `Convert.From` (when the latter is not empty), because entries are grouped in pairs.

[...]

PR #840, opened 2023-02-09
[Minor] Add Ares integration mechanism

MortonPL:
This PR is purely technical and doesn't offer any new modding features. Moved relevant AresData code from #369.


src/Misc/AresData.cpp
```diff
@@ -0,0 +1,109 @@
...
+				Debug::LogDeferred("Module %s base address : 0x%p.\n", modEntry.szModule, modEntry.modBaseAddr);
+				if (!_strcmpi(modEntry.szModule, "Ares.dll"))
+				{
+					AresData::AresBaseAddress = (uintptr_t)modEntry.modBaseAddr;
+				}
+				else if(!_strcmpi(modEntry.szModule, PHOBOS_DLL))
+				{
+					AresData::PhobosBaseAddress = (uintptr_t)modEntry.modBaseAddr;
+				}
+			} while (Module32Next(hSnap, &modEntry));
```
Metadorius, line 46:
```suggestion
			}
			while (Module32Next(hSnap, &modEntry));
```

src/Misc/AresData.cpp
```diff
@@ -0,0 +1,109 @@
...
+			{
+				Debug::LogDeferred("Module %s base address : 0x%p.\n", modEntry.szModule, modEntry.modBaseAddr);
+				if (!_strcmpi(modEntry.szModule, "Ares.dll"))
+				{
+					AresData::AresBaseAddress = (uintptr_t)modEntry.modBaseAddr;
+				}
+				else if(!_strcmpi(modEntry.szModule, PHOBOS_DLL))
+				{
+					AresData::PhobosBaseAddress = (uintptr_t)modEntry.modBaseAddr;
+				}
```
Metadorius, line 45:
```suggestion
				if (!_strcmpi(modEntry.szModule, "Ares.dll"))
					AresData::AresBaseAddress = (uintptr_t)modEntry.modBaseAddr;
				else if(!_strcmpi(modEntry.szModule, PHOBOS_DLL))
					AresData::PhobosBaseAddress = (uintptr_t)modEntry.modBaseAddr;
```

Does this use the filename which it was compiled with or the actual filename of the DLL?

src/Misc/AresData.cpp
```diff
@@ -0,0 +1,109 @@
...
+	case AresTimestampBytes[Version::Ares30]:
+		AresVersionId = Version::Ares30;
+		CanUseAres = true;
+		Debug::LogDeferred("[Phobos] Detected Ares 3.0.\n");
+		break;
+	case AresTimestampBytes[Version::Ares30p]:
+		AresVersionId = Version::Ares30p;
+		CanUseAres = true;
+		Debug::LogDeferred("[Phobos] Detected Ares 3.0p1.\n");
+		break;
```
Metadorius, line 78: Note for future: this would probably need to be unhardcoded if things would change / HAres folks would also like to be supported.

src/Misc/AresData.h
Metadorius, line 1: I am not sure on the usability of this tbh, especially using enums to do a pseudo-map, specifying addresses that way is not going to be good in the long run.

Metadorius: Have no idea on the alternatives yet, will get back to that if I get some thoughts.

src/Misc/AresData.cpp
chaserli: Compiled with ($(ProjectName)), cuz I didn't know how to get the filename at runtime

src/Misc/AresData.h
chaserli: I agree, but in order to support multiple versions it should consider different addresses per function name.
BTW it would be better to move those Aresxxxcall dummy structs into the cpp file and reduce the header file size a bit, since they shouldn't be used anywhere else

src/Misc/AresData.h
```diff
@@ -0,0 +1,143 @@
...
+		decltype(auto) operator()(TArgs... args) const
+		{
+			reinterpret_cast<fp_type>(AresFunctionOffsetsFinal[idx])(args...);
+		}
+	};
+
+	template<int idx, typename Tret, typename TThis, typename... TArgs>
+	struct AresThiscall
+	{
+		using fp_type = Tret(__fastcall*)(TThis, void*, TArgs...);
```
chaserli, line 127: I think you can just use __thiscall* here

chaserli (dismissed):

src/Misc/AresData.cpp
Metadorius: No I mean if it's uses the internal name (`Original filename` in metadata) then it's fine, otherwise what if someone renames it?

chaserli: Je ne sais pas.
I also tried to print all submodules names in log file within cncnet5.dll or DP at later game, but Phobos wasn't listed. I don't know why

MortonPL: Modified to use `Original filename`.
Still, I wouldn't ever advise renaming Ares.dll and Phobos.dll, as Syringe loads hooks in alphabetical order and having APhobos.dll and ZAres.dll (or the other way around) will mess things up.

src/Misc/AresData.cpp
MortonPL: I don't believe HAres support is possible, as it's mutually exclusive with Ares (on which we more or less depend in some places) and I would guess with Phobos as well.

Metadorius: Just treat it as another version of Ares?

src/Misc/AresData.cpp
```diff
@@ -0,0 +1,212 @@
...
+	}
+}
+
+void AresData::UnInit()
+{
+	if (!AresBaseAddress)
+		return;
+
+	if (CanUseAres)
+		FreeLibrary(AresDllHmodule);
```
chaserli, line 111: > If you obtain a module handle using GetModuleHandleExW at the beginning of your program and do not call LoadLibrary to load the module, it is not necessary to call FreeLibrary on that handle at the end of your program.

> GetModuleHandleExW retrieves a handle to a module that is already loaded into the process. The handle returned by GetModuleHandleExW is not a reference count to the module, but rather a pointer to the module's base address in memory.

> Since your code does not load the module using LoadLibrary, there is no reference count associated with the module handle obtained from GetModuleHandleExW. Therefore, you do not need to call FreeLibrary on that handle when your program ends.

> However, if you do load the module using LoadLibrary, you should call FreeLibrary when you are done with the module to decrement the reference count and unload the module from memory if the reference count drops to zero.

said ChatGPT.

MortonPL: I agree with the first point, I disagree with the second point. We want to obtain the *original* library name from file metadata, so that a user can rename Ares.dll to something else and still enjoy integration (see one of the resolved conversations above). Current code works like that, generated does not.

src/Misc/AresData.h
```diff
@@ -0,0 +1,86 @@
...
+
+	// timestamp bytes for each version
+	static constexpr DWORD AresTimestampBytes[AresData::AresVersionCount] =
+	{
+		0x5fc37ef6,	// 3.0
+		0x61daa114, // 3.0p
+	};
+
+	// offsets of function addresses for each version
+	static constexpr DWORD AresFunctionOffsets[AresData::AresFunctionCount][AresData::AresVersionCount] =
```
Metadorius, line 48: Instead of this I propose to have implicit specification of the first dimension size (keep second dimension reference for second dword array) and calculate length using constexpr.

https://stackoverflow.com/questions/49454005/how-to-get-an-array-size-at-compile-time

src/Misc/AresData.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+int AresData::HasFactory(int buffer, HouseClass* pOwner, TechnoTypeClass* pType, bool skipAircraft, bool requirePower, bool checkCanBuild, bool unknown)
+{
+	return AresStdcall<HasFactoryID, int, int, HouseClass*, TechnoTypeClass*, bool, bool, bool, bool>()(buffer, pOwner, pType, skipAircraft, requirePower, checkCanBuild, unknown);
+}
+
+bool AresData::CanBeBuiltAt(DWORD pTechnoTypeExt, BuildingTypeClass* pBuildingType)
+{
+	return AresThiscall<CanBeBuiltAtID, bool, DWORD, BuildingTypeClass*>()(pTechnoTypeExt, pBuildingType);
+}
```
Metadorius, line 203: 1. Think the functions and the calling mechanism itself should be separated in separate files as Trsdy said.
2. Not sure, maybe explicit template params can be avoided where unneeded?

Metadorius (approved):
Although I am not a fan of the way it's currently implemented (had something closer to how YRpp is defined in mind) I guess we could still merge it to not block the depending PRs. Few minor comments and let's merge it.

src/Misc/AresData.cpp
```diff
@@ -15,9 +13,47 @@ class TechnoTypeClass;
...
+			0x0217C0,
+		},
+	},
+	{
+		"CanBeBuiltAt",     // TechnoTypeExt
+		{
+			0x03E3B0,
+			0x03E3B0,
+		},
+	},
```
Metadorius, line 55: What do you think about
1. a map per version?
2. nameof(function)? https://github.com/Neargye/nameof

src/Misc/AresFunctions.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
+#include "AresData.h"
+
+#include <ASMMacros.h>
+
+namespace AresFunctions
+{
+#define JMP_ARES(name) const auto x = AresData::AresFunctionOffsetsFinal[name]; JMP(x)
+#define JMP_THIS_ARES(name) const auto x = AresData::AresFunctionOffsetsFinal[name]; JMP_THIS(x)
```
Metadorius, line 9: With `nameof` or an alternative also could make the jmp macro automatically use function name, essentially lessening the burden of keeping the code up. Rename the function via IDE - the reference to it in nameof is also gonna get renamed.

Metadorius (approved):

src/Misc/AresFunctions.h
chaserli: I still don't think it's a good idea. You manipulate strings just to decide which function to call? This creates unnecessary overhead at runtime. I think we might just declare function pointers and assign them when detecting ares' base addresses. No need for maps. Simple and fast

PR #843, opened 2023-02-10
ExtMap optimization

Starkku:
This is mostly @Otamaa's work, I just ported it over and did some tweaks.

Basically replaces the existing `ExtMap / Container` implementation that relies on `std::map` with storing the extension class pointers in unused data in game classes. The offset at which the pointers are stored is determined by `ExtPointerOffset` field in the extension classes, or if not present, a constant set to `0x18` is used which corresponds to an unused field in `AbstractClass`.

There's bunch of changes to existing code outside `Container.h` to accomodate these changes, mostly to do with pointer invalidation which is now mostly called from type-specific hooks in game's `Detach()` functions and fixing incompatibilities with the way `ExtMap.Find()` now works, as calling it with corrupted pointers now crashes instead of returning nullptr.

`TypeExtData` field in some of the extension classes is also now deprecated in favour of calling `ExtMap.Find()` to get the type on demand to eliminate any possibility of conflicts with type conversion logics.


src/Ext/Building/Body.cpp
```diff
@@ -382,8 +403,10 @@ DEFINE_HOOK(0x43BCBD, BuildingClass_CTOR, 0x6)
 {
 	GET(BuildingClass*, pItem, ESI);
 
-	auto const pExt = BuildingExt::ExtMap.FindOrAllocate(pItem);
-	pExt->TypeExtData = BuildingTypeExt::ExtMap.Find(pItem->Type);
+	if (pItem && strcmp(pItem->get_ID(), "YAEXPMIND") == 0)
+		Debug::Log("Test\n");
```
MortonPL, line 407:
```suggestion
```
Marking for later clean-up

src/Ext/WarheadType/Detonate.cpp
```diff
@@ -307,12 +309,16 @@ void WarheadTypeExt::ExtData::InterceptBullets(TechnoClass* pOwner, WeaponTypeCl
...
-			auto const pTypeExt = pExt->TypeExtData;
+			if (pBullet->Location.DistanceFrom(coords) > cellSpread * Unsorted::LeptonsPerCell)
+				continue;
+
+			auto const pExt = BulletExt::ExtMap.Find(pBullet);
+			auto const pTypeExt = BulletTypeExt::ExtMap.Find(pBullet->Type);
 
 			// Cells don't know about bullets that may or may not be located on them so it has to be this way.
-			if (pTypeExt && pTypeExt->Interceptable && pBullet->Location.DistanceFrom(coords) <= cellSpread * Unsorted::LeptonsPerCell)
+			if (pTypeExt && pTypeExt->Interceptable)
```
MortonPL, line 321: What's the reason for this?

MortonPL (commented):
Out of curiosity, have you measured performance before and after?

src/Ext/WarheadType/Detonate.cpp
Starkku: Just a really minor optimization, moved the check before all the other code since it can be done earlier.

src/Ext/Building/Body.cpp
Starkku: 😔

Metadorius: *You have been marked for termination*

Starkku: 🙃🔫

Starkku:
> Out of curiosity, have you measured performance before and after?

I did do some tests but I don't know how conclusive they were, I got around 15-20% improvement in average FPS in those tests.

chaserli:
Didn't look into the very details yet given the quantity, a few overall personal opinions
1.
> `TypeExtData` field in some of the extension classes is also now deprecated in favour of calling `ExtMap.Find()` to get the type on demand to eliminate any possibility of conflicts with type conversion logics.
Don't think that's even necessary. 
Also, iterating ExtMap should be as fast as (or even faster) than your current one.
2. Just as Morton said, profile first using tools like verysleepy and see which ones matter most, afaik primarily `TechnoExt`,`TechnoTypeExt` ,`BuildingExt`, `BuildingTypeExt` and `AnimExt`. 
3. For base types having `ArrayIndex`(namely `SWTypeExt`, `HouseExt`, `SideExt`,etc...) you can create a static array and fetch the ext based on the `ArrayIndex`.

Starkku:
Reverted the pointer invalidation changes because it seems doing it with specific hooks vs. mass actions on a single hook is actually slower. I guess it makes sense since there's less hooks in latter.

src/Ext/Anim/Body.cpp
```diff
@@ -226,3 +224,28 @@ DEFINE_HOOK(0x4253FF, AnimClass_Save_Suffix, 0x5)
...
+
+DEFINE_JUMP(LJMP, 0x42543A, 0x425448)
+
+DEFINE_HOOK_AGAIN(0x421EF4, AnimClass_CTOR_setD0, 0x6)
+DEFINE_HOOK(0x42276D, AnimClass_CTOR_setD0, 0x6)
+{
+	GET(AnimClass*, pThis, ESI);
+	pThis->unknown_D0 = 0;
+	return R->Origin() + 0x6;
+}
```
Metadorius, line 251: What does this do? Probably an explanation comment or a better hook name wouldn't hurt.

src/Ext/Building/Body.cpp
```diff
@@ -421,3 +423,20 @@ DEFINE_HOOK(0x454244, BuildingClass_Save_Suffix, 0x7)
 
 	return 0;
 }
+
+// Removes setting otherwise unused field (0x6FC) when building has airstrike applied on it.
+DEFINE_JUMP(LJMP, 0x41D9FB, 0x41DA05);
```
Metadorius, line 428: Does any other ext. use this field?

src/Ext/Anim/Body.cpp
```diff
@@ -226,3 +224,28 @@ DEFINE_HOOK(0x4253FF, AnimClass_Save_Suffix, 0x5)
...
+	GET(void*, target, EDI);
+	GET_STACK(bool, all, STACK_OFFSET(0xC, 0x8));
+
+	if (auto pAnimExt = AnimExt::ExtMap.Find(pThis))
+		pAnimExt->InvalidatePointer(target, all);
+
+	R->EBX(0);
+	return pThis->OwnerObject == target && target ? 0x425174 : 0x4251A3;
+}
+*/
```
Metadorius, line 241: Why is the code commented out?

Metadorius (changes requested):
ExtMap wise I don't see much issues, but all the other seemingly random changes lack context and are unobvious, and it also would be better if they were split.

src/Ext/Anim/Body.cpp
Metadorius: Likewise with other code blocks like this one

src/Ext/Anim/Body.cpp
Starkku: Makes sure the otherwise unused field isn't set in constructor because it is used for the `ExtPointerOffset` of `AnimExt`.

https://github.com/Phobos-developers/Phobos/pull/843/files#diff-a74a98d60066e2d7da1fefb6c90efd397d66a7cbbb5551f45d68a4f9cdd8e034R16

But yeah will clarify I guess.

src/Ext/Building/Body.cpp
Starkku: > Does any other ext. use this field?

Ext as in class or another extension DLL? No to former, no clue about the latter.

Same case as with the first comment, this is used for `BuildingExt` `ExtPointerOffset`.

src/Ext/Anim/Body.cpp
Starkku: Mass-action based pointer invalidation turned out to have better performance so obsolete code. Can probably clear it out before merging.

Metadorius (approved):

PR #845, opened 2023-02-13
WIP Reimplement INI inheritance

MortonPL:
### Include files
- INI files can now include other files (merge them into self) using `[$Include]` section.
  - `[$Include]` section contains a list of files to read and include. Files can be directly in the Red Alert 2 directory or in a loaded MIX file.
  - Files will be added in the same order they are defined. Index of each file **must be unique among all included files**.
  - Inclusion can be nested recursively (included files can include files further). Recursion is depth-first (before including next file, check if the current one includes anything).
  - When the same entry exists in two files, then the one read later will overwrite the value.
  - This feature can be used in *any* INI file, be it `rulesmd.ini`, `artmd.ini`, `soundmd.ini` or others.

```{warning}
When Phobos is present, the [Ares equivalent of $Include](https://ares-developers.github.io/Ares-docs/new/misc/include.html) is disabled!
```

In any file:
```ini
[$Include]
0=somefile.ini	; file name
```

### Section inheritance
- You can now make sections (children) inherit entries from other sections (parents) with `$Inherits`.
  - When a section has no value set for an entry (or an entry is missing), the game will attempt to use parent's value. If no value is found, only then the default will be used.
  - When multiple parents are specified, the order of inheritance is "first come, first served", looking up comma separated parents from left to right.
  - Inheritance can be nested recursively (parent sections can have their own parents). Recursion is depth-first (before inheriting from the next parent, check if the current parent has parents).
  - This feature can be used in *any* INI file, be it `rulesmd.ini`, `artmd.ini`, `soundmd.ini` or others.

```{warning}
When Phobos is present, the Ares equivalent of $Inherits (undocumented) is disabled!
```

In any file:
```ini
[PARENT1SECTION]

[PARENT2SECTION]

[CHILDSECTION]
$Inherits=PARENT1SECTION,PARENT2SECTION...  ; section names
```

### DIFF
--- a/docs/Miscellanous.md
+++ b/docs/Miscellanous.md
@@ -123,14 +123,33 @@ function onInput() {
 
 ## INI
 
-### Section inheritance
-- You can now make sections (children) inherit entries from other sections (parents) with `$Inherits`.
-  - Every parent entry will be inherited verbatim, without checking if it is valid.
-  - When there are multiple parents, they will be inherited in the order they were given.
-  - In the child entry, entries before `$Inherits` can be overwritten if they are defined inside parents. **It is advised to always keep `$Inherits` as the first entry in a section.**
+### Include files
+- INI files can now include other files (merge them into self) using `[$Include]` section.
+  - `[$Include]` section contains a list of files to read and include. Files can be directly in the Red Alert 2 directory or in a loaded MIX file.
+  - Files will be added in the same order they are defined. Index of each file **must be unique among all included files**.
+  - Inclusion can be nested recursively (included files can include files further). Recursion is depth-first (before including next file, check if the current one includes anything).
+  - When the same entry exists in two files, then the one read later will overwrite the value.
+  - This feature can be used in *any* INI file, be it `rulesmd.ini`, `artmd.ini`, `soundmd.ini` or others.
 
 ```{warning}
-Currently, this feature only works in files included by [Ares #include feature](https://ares-developers.github.io/Ares-docs/new/misc/include.html).
+When Phobos is present, the [Ares equivalent of $Include](https://ares-developers.github.io/Ares-docs/new/misc/include.html) is disabled!
+```
+
+In any file:
+```ini
+[$Include]
+0=somefile.ini	; file name
+```
+
+### Section inheritance
+- You can now make sections (children) inherit entries from other sections (parents) with `$Inherits`.
+  - When a section has no value set for an entry (or an entry is missing), the game will attempt to use parent's value. If no value is found, only then the default will be used.
+  - When multiple parents are specified, the order of inheritance is "first come, first served", looking up comma separated parents from left to right.
+  - Inheritance can be nested recursively (parent sections can have their own parents). Recursion is depth-first (before inheriting from the next parent, check if the current parent has parents).
+  - This feature can be used in *any* INI file, be it `rulesmd.ini`, `artmd.ini`, `soundmd.ini` or others.
+
+```{warning}
+When Phobos is present, the Ares equivalent of $Inherits (undocumented) is disabled!
 ```
 
 In any file:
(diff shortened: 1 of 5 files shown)


docs/Miscellanous.md
```diff
@@ -120,3 +120,25 @@ function onInput() {
 	}
 }
 </script>
+
+## INI
+
+### Section inheritance
+- You can now make sections (children) inherit entries from other sections (parents) with `$Inherits`.
+  - Every parent entry will be inherited verbatim, without checking if it is valid.
+  - When there are multiple parents, they will be inherited in the order they were given.
```
Metadorius, line 129: Not clear what's meant

docs/Miscellanous.md
```diff
@@ -120,3 +120,25 @@ function onInput() {
...
+  - In the child entry, entries before `$Inherits` can be overwritten if they are defined inside parents. **It is advised to always keep `$Inherits` as the first entry in a section.**
```
Metadorius, line 130: 🤨

docs/Miscellanous.md
```diff
@@ -120,3 +120,44 @@ function onInput() {
...
+
+## INI
+
+### Include files
+- INI files can now include other files (merge them into self) using `[$Include]` section.
+  - `[$Include]` section contains a list of files to read and include. Files can be directly in the Red Alert 2 directory or in a loaded MIX file.
+  - Files will be added in the same order they are defined. Index of each file **must be unique among all included files**.
+  - Inclusion can be nested recursively (included files can include files further). Recursion is depth-first (before including next file, check if the current one includes anything).
+  - When the same entry exists in two files, then the one read later will overwrite the value.
+  - This feature can be used in *any* INI file, be it `rulesmd.ini`, `artmd.ini`, `soundmd.ini` or others.
```
Metadorius, line 132:
```suggestion
  - This feature can be used in *any* INI file, be it `rulesmd.ini`, `artmd.ini`, `soundmd.ini`, map file or anything else.
```

docs/Miscellanous.md
```diff
@@ -120,3 +120,44 @@ function onInput() {
...
+[$Include]
+0=somefile.ini	; file name
+```
+
+### Section inheritance
+- You can now make sections (children) inherit entries from other sections (parents) with `$Inherits`.
+  - When a section has no value set for an entry (or an entry is missing), the game will attempt to use parent's value. If no value is found, only then the default will be used.
+  - When multiple parents are specified, the order of inheritance is "first come, first served", looking up comma separated parents from left to right.
+  - Inheritance can be nested recursively (parent sections can have their own parents). Recursion is depth-first (before inheriting from the next parent, check if the current parent has parents).
+  - This feature can be used in *any* INI file, be it `rulesmd.ini`, `artmd.ini`, `soundmd.ini` or others.
```
Metadorius, line 149:
```suggestion
  - This feature can be used in *any* INI file, be it `rulesmd.ini`, `artmd.ini`, `soundmd.ini`, map file or anything else.
```

Metadorius (approved):
I think would also be good to do a sed script amendments in migration tool scripts to switch from Ares to Phobos. Should be two simple regexes.

Belonit:
Why did you choose to disable `#include` provided by ares?

Metadorius:
> Why did you choose to disable `#include` provided by ares?

Because it's impossible to reimplement it properly without disabling it completely. It should be working the same way, just change the # (which is not valid in some INI parsers) to $.

src/Misc/Hooks.INIInheritance.cpp
```diff
@@ -0,0 +1,202 @@
...
+{
+	char* entry = INIInheritance::SavedEntries.back();
+	if (entry)
+		free(entry);
+	INIInheritance::SavedEntries.pop_back();
+
+	char* section = INIInheritance::SavedSections.back();
+	if (section)
+		free(section);
+	INIInheritance::SavedSections.pop_back();
```
Metadorius, line 130: Shouldn't we use `GameDelete` instead?

MortonPL: These are char* that we _strdup()'d earlier though, not game resources. See line 112.

Metadorius (approved):

PR #847, opened 2023-02-16
Straight trajectory additions/improvements

Starkku:
Reworked version of #671 

  - `Trajectory.Straight.DetonationDistance` controls the maximum distance in cells from intended target (checked at start of each game frame, before the projectile moves) at which the projectile will be forced to detonate. Set to 0 to disable forced detonation (note that this can cause the projectile to overshoot the target).
  - `Trajectory.Straight.PassThrough` enables special case logic where the projectile does not detonate in contact with the target but ínstead travels up to a distance defined by `Trajectory.Straight.DetonationDistance`. Note that the firing angle of the projectile is adjusted with this in mind, making it fire straight ahead if the target is on same elevation.

In `rulesmd.ini`:
```ini
[SOMEPROJECTILE]                            ; Projectile
Trajectory.Straight.DetonationDistance=0.4  ; floating point value
Trajectory.Straight.TargetSnapDistance=0.5  ; floating point value
Trajectory.Straight.PassThrough=false       ; boolean
```
------

`DetonationDistance` functionality already exists in `develop` branch, it is just hardcoded to 100 leptons (0.4 cells = 102 leptons, so it has been slightly changed but in practice this makes no difference).

### DIFF
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -493,7 +493,7 @@ In `rulesmd.ini`:
 [SOMEPROJECTILE]                            ; Projectile
 Trajectory=Straight                         ; Trajectory type
 Trajectory.Straight.DetonationDistance=0.4  ; floating point value
-Trajectory.Straight.TargetSnapDistance=0.0  ; floating point value
+Trajectory.Straight.TargetSnapDistance=0.5  ; floating point value
 Trajectory.Straight.PassThrough=false       ; boolean
 ```
 
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -484,11 +484,15 @@ Trajectory.Speed=100.0  ; floating point value
 *Straight trajectory used to make blasters in a private mod by @brsajo#9745*
 
 - Self-explanatory, is a straight-shot trajectory.
+  - `Trajectory.Straight.DetonationDistance` controls the maximum distance in cells from intended target (checked at start of each game frame, before the projectile moves) at which the projectile will be forced to detonate. Set to 0 to disable forced detonation (note that this can cause the projectile to overshoot the target).
+  - `Trajectory.Straight.TargetSnapDistance` controls the maximum distance in cells from intended target the projectile can be at moment of detonation to make the projectile 'snap' on the intended target. Set to 0 to disable snapping.
 
 In `rulesmd.ini`:
 ```ini
-[SOMEPROJECTILE]     ; Projectile
-Trajectory=Straight  ; Trajectory type
+[SOMEPROJECTILE]                            ; Projectile
+Trajectory=Straight                         ; Trajectory type
+Trajectory.Straight.DetonationDistance=0.4  ; floating point value
+Trajectory.Straight.TargetSnapDistance=0.0  ; floating point value
 ```
 
 #### Bombard trajectory
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -486,6 +486,7 @@ Trajectory.Speed=100.0  ; floating point value
 - Self-explanatory, is a straight-shot trajectory.
   - `Trajectory.Straight.DetonationDistance` controls the maximum distance in cells from intended target (checked at start of each game frame, before the projectile moves) at which the projectile will be forced to detonate. Set to 0 to disable forced detonation (note that this can cause the projectile to overshoot the target).
   - `Trajectory.Straight.TargetSnapDistance` controls the maximum distance in cells from intended target the projectile can be at moment of detonation to make the projectile 'snap' on the intended target. Set to 0 to disable snapping.
+  - `Trajectory.Straight.PassThrough` enables special case logic where the projectile does not detonate in contact with the target but ínstead travels up to a distance defined by `Trajectory.Straight.DetonationDistance`. Note that the firing angle of the projectile is adjusted with this in mind, making it fire straight ahead if the target is on same elevation.
 
 In `rulesmd.ini`:
 ```ini
@@ -493,6 +494,7 @@ In `rulesmd.ini`:
 Trajectory=Straight                         ; Trajectory type
 Trajectory.Straight.DetonationDistance=0.4  ; floating point value
 Trajectory.Straight.TargetSnapDistance=0.0  ; floating point value
+Trajectory.Straight.PassThrough=false       ; boolean
 ```
 
 #### Bombard trajectory
--- a/src/Ext/Bullet/Trajectories/StraightTrajectory.cpp
+++ b/src/Ext/Bullet/Trajectories/StraightTrajectory.cpp
@@ -126,11 +126,13 @@ TrajectoryCheckReturnType StraightTrajectory::OnAITargetCoordCheck(BulletClass* 
 	}
 	else
 	{
+		/*
 		auto const bulletCell = CellClass::Coord2Cell(pBullet->Location);
 		auto const targetCell = CellClass::Coord2Cell(pBullet->TargetCoords);
 
 		if (bulletCell == targetCell && pBullet->GetHeight() < 2 * Unsorted::LevelHeight)
 			return TrajectoryCheckReturnType::Detonate; // Detonate projectile.
+		*/
 
 		if (this->FiredFromAboveTarget && pBullet->Location.Z < pBullet->TargetCoords.Z)
 			return TrajectoryCheckReturnType::Detonate; // Detonate projectile.
--- a/src/Ext/Bullet/Trajectories/StraightTrajectory.cpp
+++ b/src/Ext/Bullet/Trajectories/StraightTrajectory.cpp
@@ -4,12 +4,26 @@
 bool StraightTrajectoryType::Load(PhobosStreamReader& Stm, bool RegisterForChange)
 {
 	this->PhobosTrajectoryType::Load(Stm, false);
+
+	Stm
+		.Process(this->DetonationDistance, false)
+		.Process(this->TargetSnapDistance, false)
+		.Process(this->PassThrough, false)
+		;
+
 	return true;
 }
 
 bool StraightTrajectoryType::Save(PhobosStreamWriter& Stm) const
 {
 	this->PhobosTrajectoryType::Save(Stm);
+
+	Stm
+		.Process(this->DetonationDistance)
+		.Process(this->TargetSnapDistance)
+		.Process(this->PassThrough)
+		;
+
 	return true;
 }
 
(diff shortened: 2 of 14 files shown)


src/Ext/Bullet/Trajectories/StraightTrajectory.cpp
```diff
@@ -30,32 +35,97 @@ bool StraightTrajectory::Save(PhobosStreamWriter& Stm) const
...
+			return TrajectoryCheckReturnType::Detonate; // Detonate projectile.
+
+		if (pBullet->Location.Z < pBullet->TargetCoords.Z)
+			return TrajectoryCheckReturnType::Detonate; // Detonate projectile.
+	}
+	else
+	{
+		bool isAboveTarget = false;
+
+		if ((pBullet->Owner && pBullet->Owner->Location.Z > pBullet->TargetCoords.Z) || (!pBullet->Owner && pBullet->SourceCoords.Z > pBullet->TargetCoords.Z))
```
FS-21, line 105: Or split it in 2 lines or follow the guidelines:
image

FS-21 (commented):
Good code writing. I have to test the feature in-game if it is considered ready by the author :-D

mevitar:
Straight projectiles with default values have problems hitting even immobile targets (including empty cells), especially at exact Range= and MinimumRange= distances. It seems to happen 100% on these edge values at + firing angles (ingame NW, NE, SE, SW). I think TargetSnapDistance should get a higher value as a default, as this didn't happen if i set it to values as high as .75 or 1.

Also, i was expecting DetonationDistance=0 and TargetSnapDistance=0 to make the projectile detonate right in the middle of the targeted cell, but instead, it seems to detonate exactly on the edge of that cell. Tested with default Trajectory.Speed (100) and with 50, with same results, so it's clearly not a matter of speed being too high.

mevitar:
I forgot to mention that, with default values, i've had the projectile occasionally detonate right after being fired from the unit instead of flying towards it, while all other shots detonated somewhere around the edge of the cell that the target was on. This was on flat ground, neither the target or the firer were moving, there was no obstacles, and the target was on the very edge of the firer's range, on a straight line to the right (ingame NE).

Unfortunately i didn't get to record it. If i will, i will post it.

mevitar:
Now i understand what's the issue - straight projectiles now detonate on the firer when fired at an airborne target. You basically can't use straight projectiles for anti-air units anymore, and anti-ground straight projectiles have a chance to explode prematurely if fired at hovering units (when they are too high). It "fixes" itself when i set ROT on the projectile, but then it also ignores the straight behavior.

This didn't happen in earlier builds, and doesn't happen in 0.3 either. I retested the same anti-air projectile with 0.3 and it works fine there (even if inaccurate, which is to be expected), but with this build here it detonates immediately on the firer.

Starkku:
Made a couple of changes:
- Fix issues with save / load
- Fix projectiles vs. targets at higher elevation detonating instantly
- Attempt to fix issues with hover units and PassThrough=true projectiles

> Also, i was expecting DetonationDistance=0 and TargetSnapDistance=0 to make the projectile detonate right in the middle of the targeted cell, but instead, it seems to detonate exactly on the edge of that cell, the very moment it "touches" it. Tested with default Trajectory.Speed (100) and with 50, with same results, so it's clearly not a matter of speed being too high or too low.

This is because of a check (that exists in vanilla game too) that detonates the projectile at its current position when it is in same cell as target AND at height below 208 leptons. It seems to usually occur at edge of cell rather than at dead center, likely for some math reasons. I have disabled this check atleast for time being, to see if the behaviour is more desirable now (other checks should prevent the projectile from going below the ground).

SnapDistance default could be changes although I am not sure what would be sensible default.

mevitar:
DetonationDistance=0 and TargetSnapDistance=0 projectiles hit the center of the cell without issues now, at least for speed 100 or less. With higher speeds it becomes less reliable, but i guess that's to be expected.

That is, as long as they're not on a building. If the same projectile is fired by a building (at least one with a turret) then the projectile will forget to detonate and fly infinitely. This happens even with PassThrough=no. Same projectile detonates properly in the previous build.

As for air and hover units, projectiles still detonate instantly when fired at them, so nothing changed here.

Starkku:
Fixed a regression that broke the vs. elevated targets fix.

I can't reproduce issues with buildings firing straight trajectory weapons, might need more info on that.

I changed default value for `TargetSnapDistance` from 0 to 0.5.

mevitar:
With the newest build, If i launch a testmap, the game freezes on start, with an empty GUI and no cameos. If i launch a skirmish game, the game crashes before it finishes loading the map.

Starkku:
> With the newest build, If i launch a testmap, the game freezes on start, with an empty GUI and no cameos. If i launch a skirmish game, the game crashes before it finishes loading the map.

Does it also do the same on newest nightly build for develop branch?

mevitar:
> Does it also do the same on newest nightly build for develop branch? And related to that, do you make any use of Ares' undocumented INI section inheritance feature?

Yes to both questions.
And the newest build right before the inheritance reimplementation works fine.

mevitar:
Tested it on unmodded YR (+Ares and Phobos) with the projectiles copy+pasted into it, the projectiles no longer detonate on the firer if fired at a hover on airborne unit. However, i noticed that there are still some problems.

First, if fired at a hover unit, the projectile might still detonate prematurely (the moment at which it happens depends on the hover unit's current position in the hover wobble cycle - if it's on the lower end, the projectile will hit, but if it's on the upper end, the projectile will detonate somewhere on its path). Both of these projectiles have this issue.
```
[Shell]
Image=DRAGON
Shadow=no
SubjectToCliffs=yes
SubjectToElevation=yes
SubjectToWalls=yes
SubjectToBuildings=yes
Trajectory=Straight
Trajectory.Straight.DetonationDistance=.75
Trajectory.Straight.TargetSnapDistance=.75
Trajectory.Straight.PassThrough=no
Trajectory.Speed=150

[Shell2]
Image=MEDUSA
Shadow=no
SubjectToCliffs=yes
SubjectToElevation=yes
SubjectToWalls=yes
SubjectToBuildings=yes
Trajectory=Straight
Trajectory.Straight.PassThrough=no
Trajectory.Speed=150
```
Funnily enough, issue seems to completely disappear if the firer is also hovering.

Second, if fired at a moving airborne target, the projectile might occasionally fail to detonate and keep flying forever (it's as if it failed to snap to the target and the game didn't kill the projectile despite a miss).
This is the code for the projectile:
```
Image=TORPEDO
Shadow=no
AA=yes
SubjectToCliffs=no
SubjectToElevation=no
SubjectToWalls=no
SubjectToBuildings=no
Trajectory=Straight
Trajectory.Straight.TargetSnapDistance=.5
Trajectory.Straight.PassThrough=no
Trajectory.Speed=300
```
BUT if i add Ranged=yes OR lower Trajectory.Speed= to like 100, it won't fly forever anymore.

Otherwise, everything seems to be working fine.

> I can't reproduce issues with buildings firing straight trajectory weapons, might need more info on that.

It was a building that deploys from a vehicle (like the Tick Tank, except this one was an SHP walker, deploying into a building with an SHP turret). I was trying to reproduce it on unmodded YR but i can't. If i'll

mevitar:
I checked the newest build and the two issues from my newest post (premature detonation vs hover units and infinite flight vs flying units) still exist.

Starkku:
I can't reproduce the issue against hover units and the infinite flight is something that may occur if the proximity detonation checks fail due to the projectile for one or other reason moving beyond the range within single frame. I would recommend either increasing `DetonationDistance` and `SnapDistance` or using `Ranged=yes` and `ProjectileRange` with fast projectiles. There's no automatic killswitch for straight trajectory projectiles as of current, hence why it is up to modder to use `ProjectileRange` if needed.

mevitar:
Indeed it doesn't happen for voxel hover units anymore. Only shp hover units have this problem now. This happens even if i use the same projectile for both units - the voxel one will hit the hovering target properly, but the shp one can miss when attacking a hover (doesn't matter if it's a voxel or shp) unit.
There definitely wasn't such distinction at first, because i found this issue when testing with a regular Robot Tank and not a single shp unit being used.

Hover settings i used to test now:
```
HoverHeight=120
HoverDampen=75%
HoverBob=.4
HoverBoost=150%
HoverAcceleration=.04
HoverBrake=.04
```

As for infinite flight projectile, i tried with increasing DetonationDistance and it stops happening even with TrajectorySpeed=300 and lack of any Ranged=yes (i used Trajectory.Straight.DetonationDistance=1 and Trajectory.Straight.TargetSnapDistance=1). Lowering DetonationDistance will bring this issue back and only then it will need Ranged=yes.
Guess it's indeed just a matter of adjusting projectile stats.

MortonPL (approved):

mevitar:
I rechecked the hover units, turns out it's not an shp thing, but an FLH thing. If a hover unit has FLH=0,0,0, then there's a chance the projectile will detonate prematurely when firing on another hover unit. Might be specific to 0,0,0, as i tried with positive and negative Front and Height, and all combinations worked fine.

Metadorius:
> I rechecked the hover units, turns out it's not an shp thing, but an FLH thing. If a hover unit has FLH=0,0,0, then there's a chance the projectile will detonate prematurely when firing on another hover unit. Might be specific to 0,0,0, as i tried with positive and negative Front and Height, and all combinations worked fine.

@mevitar @Starkku was this ever addressed and retested?

Starkku:
> @mevitar @Starkku was this ever addressed and retested?

Maybe now. Might have also broken something else, who knows

mevitar:
No issues with 0,0,0 FLH now. Didn't see any new issues pop up either. Default tag values seem to be working fine for Trajectory.Speed=150, higher speed might need adjustments but that's understandable.

docs/New-or-Enhanced-Logics.md
```diff
@@ -484,11 +484,17 @@ Trajectory.Speed=100.0  ; floating point value
...
+  - `Trajectory.Straight.PassThrough` enables special case logic where the projectile does not detonate in contact with the target but ínstead travels up to a distance defined by `Trajectory.Straight.DetonationDistance`. Note that the firing angle of the projectile is adjusted with this in mind, making it fire straight ahead if the target is on same elevation.
 
 In `rulesmd.ini`:
 ```ini
-[SOMEPROJECTILE]     ; Projectile
-Trajectory=Straight  ; Trajectory type
+[SOMEPROJECTILE]                            ; Projectile
+Trajectory=Straight                         ; Trajectory type
+Trajectory.Straight.DetonationDistance=0.4  ; floating point value
+Trajectory.Straight.TargetSnapDistance=0.5  ; floating point value
```
Metadorius, line 496: I just noticed, is there any reason it's not in leptons but in cells instead?

mevitar: Most likely because all ranges for weapons and warheads are in cells, not leptons.

Metadorius: Oh, fair point, I've just never ever touched weapon code, mostly dealt with FLH so it looked alien to me.

Starkku:
Merging as there are no major issues after this has been tested by several people including myself and mevitar.

PR #848, opened 2023-02-18
Fix some details in the transport script action code

FS-21:
- If there are no transports in the vector it shouldn't continue and jump to the next mission.
- Removed the increment of current script line execution because it skips the next script line and the game will run the "next+1" instead.


Starkku:
Don't think the size check is technically necessary, the iterator loop shouldn't do anything if the transports vector is empty. It may or may not save some instructions though so might not be all bad but dunno.

PR #849, opened 2023-02-21
Transfer Types for transfering, converting, refilling, losing resources like experience, ammo and such

### LINKED ISSUE #401: [New Feature] Gain EXP when Healing/Repairing
### Description:
Healing/Repairing an unit/infantry won't get Experience, thus it will be good if we can make it to get experience. (To get Veteran/EliteAbilities, EliteWeapons)

### Possible Implementation:
```ini
[FV]
Experience.FromEachHealing=10% (percentage, def 0%)
; If we can get exp based on the **each destroyed** target's cost, then the repair/heal exp will be obtained by **each healing** based on target's cost.
```

### LINKED ISSUE #670: Reload certain amount of ammo by being attacked by certain warheads

Envoidener:
Touches next issues:
- "continues" and closes #57 
- improves and closes #168 
- can be used in the same way, closes #401 
- closes #670 

By attaching Transfer types to warheads one can make their warheads to:
- transfer resource from a unit to another
- convert one resource to another
- give/take resources in fixed amount or percentage basis

and many more by different combinations of the above.

Currently available resources for Transfers are:
- money/credits
- experience
- health
- ammo
- gatling rate

While there are many tags to control end result of the transfer, only several of them are needed for basic tasks.
- `Direction` determines the sender and receiver side of the transfer. Senders use `Send.` tags and receivers use `Receive.`.
  - `sourcetotarget` [`s2t` for short] makes warhead firer (source) sender and target(s) within `CellSpread` of the warhead receiver(s). Requires source to be techno (the warhead is fired by weapon of specific techno) unless resource to send is money, otherwise transfer fails (nothing happens). Default value.
  - `sourcetosource` [`s2s`] makes warhead firer both sender and receiver. Requires source to be techno unless resources to send and receive are money, otherwise transfer fails.
  - `targettosource` [`t2s`] makes target(s) within `CellSpread` of the warhead sender(s) and warhead firer receiver. Requires source to be techno unless resource to receive is money, otherwise transfer fails.
  - `targettoextra` [`t2e`] makes target(s) within `CellSpread` of warhead sender(s) and units within `CellSpread` of `Extra.Warhead` receivers. Extra warhead's epicenter can be initial target or source and requires source to be techno if epicenter is set to be source.
  - `targettotarget` [`t2t`] makes target(s) within `CellSpread` of warhead both sender(s) and receiver(s).
  - `extratotarget` [`e2t`] makes units within `CellSpread` of `Extra.Warhead` senders and target(s) within `CellSpread` of warhead receiver(s). Extra warhead's epicenter can be initial target or source and requires source to be techno if epicenter is set to be source.
- `Target.` tags are used only if `Direction` contains `..target..`. Note that PercentAtMax of the source warhead applies to transfer values as well.
  - `Target.ConsiderArmor` modifies targets' transfer values according to warhead Verses. Units with armor that have 0% in Verses do not participate in the transfer. Defaults to `no`.
  - `Target.VersusWarhead` uses specified warhead's Verses to calculate `Target.ConsiderArmor` modifications instead of initial warhead. Does not have other uses. Defaults to initial warhead.
  - `Target.AffectHouses` determines houses to be affected by transfer. Others do not participate in the transfer. Defaults to `all`.
  - Enabled `Target.Spread.IgnoreSelf` removes firer of warhead from target list of the transfer if they were caught in CellSpread radius. Defaults to `no`
- `Extra.` tags are used only if `Direction` contains `..extra..` and manipulate transfer values for extra targets. Note that extra targets can overlap with initial warhead targets.
  - `Extra.Warhead`'s CellSpread and PercentAtMax values are used to determine units to be included as extra targets which are used for `..extra..` `Direction`. Defaults to initial warhead.
  - `Extra.ConsiderArmor` and `Extra.AffectHouses` work in the same manner as `Target.ConsiderArmor` and `Target.AffectHouses`, respectively, but they make changes for extra targets instead. Defaults to `no`.
  - `Extra.EpicenterIsSource` changes epicenter of extra warhead to be source (warhead firer) instead of the same epicenter as initial warhead. As stated above, requires source to be techno for transfer to work. Defaults to `no`.
  - `Extra.IgnoreEpicenter` removes epicenter unit from extra targets list. Epicenter unit can be initial target or source depending on `Extra.EpicenterIsSource`. Defaults to `no`.
- `VeterancyMultiplier.` tags read two floating point numbers or percentages for veteran (first) and elite (second) ranks and applies to `Over...` depending on `...Over`'s veterancy rank.
  - `VeterancyMultiplier.SourceOverSender` applies multiplier to senders' transfer values depending on veterancy rank of the source.
  - `VeterancyMultiplier.SourceOverReceiver` applies multiplier to receivers' transfer values depending on veterancy rank of the source.
  - `VeterancyMultiplier.TargetOverTarget` applies multiplier to each target's transfer values depending on veterancy rank of the same target.
  - `VeterancyMultiplier.TargetOverTarget` applies multiplier to each extra target's transfer values depending on veterancy rank of the same extra target.
- `Send.` tags are used by senders (`...to` side of `Direction`) and are used to determine how much resource value sender has to "send".
  - `Send.Resource` determines which resource to send, i.e. get a change in.
    - `money` [`$`] to give or take money/credits.
    - `experience` [`exp`] to gain or lose experience, i.e. kill cost.
    - `health` [`hp`] to heal or damage health/strength.
    - `ammo` to refill or lose ammo.
    - `gatlingrate` [`rate`] to up or down gatling rate.
  - `Send.Value` reads floating number or percentage to determine the amount of resource change sender gets (also referenced as transfer value of the unit). Positive values mean increase (e.g. heal), while negative values mean decrease (e.g. damage).
  - `Send.Value.Type` determine whether to use resource value of sender unit for `Send.Value`.
    - `fixed` means `Send.Value` is perceived as is. Default value.
    - `current` multiplies `Send.Value` by current resource value of the unit.
      - Current money value is current amount of money owner of the unit holds.
      - Current experience value is accumulated experience by the unit, i.e. total cost of units killed by this unit.
      - Current health and ammo are self-explanatory.
      - Current gatling rate is currently accumulated rate.
    - `total` multiplies `Send.Value` by maximum resource value of the unit.
      - "Maximum" money is Cost tag of the unit.
      - Maximum experience is total amount of experience needed for the rookie rank of unit to reach elite.
      - Maximum health is Strength tag of the unit.
      - Maximum ammo is Ammo tag of the unit.
      - Maximum gatling rate is the rate needed for last stage of gatling. Uses EliteStage if the unit is elite.
    - `missing` multiplies `Send.Value` by resource value needed for the unit to reach maximum, i.e. (`total`-`current`)
      - Missing money is the exception and is equal Soylent tag of the unit.
  - `Send.Value.FlatLimits` reads two integers: minimum and maximum, limits to the resulting transfer value of senders. Has inverted effect when dealing with negative multipliers, e.g. negative Verses with `ConsiderArmor`, negative `VeterancyMultiplier` and etc.: minimum becomes negated maximum and vice versa (e.g. 50,60 => -60,-50). Set to 0 for disabling specific limit. Defaults to 0,0.
  - `Send.PreventUnderflow` and `Send.PreventOverflow` prevent sending nonexistent and excess resource. Has no effect while `Receive.SentFactor` is `none`. There is no overflow for money. Default to `no`.
- `Receive.` tags are used by receivers (`to...` side of `Direction`) and are used to determine how much resource value receiver receives.
  - `Receive.Resource`, `Receive.Value`, `Receive.Value.Type`, `Receive.Value.FlatLimits` work in the same way as their `Send.` counterparts but for receivers.
  - `Receive.SentFactor` determines whether receiver's transfer value is influenced by sent transfer value. The sole reason to be different than `Send.` tags.
    - `none` means no effect from senders. Default value.
    - `highest` multiplies the transfer value of receiver by highest transfer value among sent values. On negative `Send.Value` searches for lowest value instead.
    - `sum` multiplies the transfer value of receiver by sum of sent tranfer values.
    - `average` multiplies the transfer value of receiver by sum of sent transfer values divided by senders' count.
    - `count` multiplies the transfer value of receiver by senders' count.
  - `Receive.SentSplit` set to `yes` divides each receivers transfer value by receivers' count. Defaults to `no`.
    - There is a special interaction when `Direction=targettotarget`, `Receive.SentFactor=average` and `Receive.SentSplit=yes`, multiplying each target's receiving transfer value by only their own sent value instead of `average`.
- `Experience.PreventDemote` prevents demotion on losing experience. Works with `Send.PreventUnderflow` preventing sender from sending any amount on elite rank or only until minimum of veteran rank is reached on veteran. Works for both sides. Defaults to `no`.
- `Health.PreventKill` prevents death from damage by transfer. Can still die to other means, i.e. warhead detonating weapon itself and such. "Works" with `Send.PreventUnderflow` changing lower limit for health from 0 to 1.
- Non-negative `GatlingRate.LimitStageChange` limits stage change caused by rate transfer. Cannot be larger or equal to WeaponStages, as it would just come full circle. `gatlingrate` changes can restart cycle on units with `Gattling.Cycle=yes`. Does not work with `Send.PreventUnderflow`. Defaults to -1.
- `Money.Display` tags determine whether to show money changes in flying texts. Do not work if the corresponding side's transfer resource is not money.
  - `Money.Display.Sender` or/and `Money.Display.Receiver` set to `yes` show the change in money (green is gain, red is loss, yellow is no change). Default to `no`.
  - `Money.Display.Sender.Houses` and `Money.Display.Receiver.Houses` show the texts only to specific houses. Default to `all`.
  - `Money.Display.Sender.Offset` and `Money.Display.Receiver.Offset` determine the offset of flying texts relative to the position of transfer unit.


MetaSorcerer:
Thanks, these function are insanely useful.

Envoidener:
Closing this because the implementation was bad, reworking to a new solution with less (convoluted) tags and without extra class.

PR #853, opened 2023-03-07
[Minor] Ext/Techno folder misc clean up

chaserli:
files in Ext/Techno subfolder are becoming more and more painful to read, a minor cleanup/ sort is perhaps needed
* Moved all firing related hooks, (`TechnoClass::SelectWeapon, TechnoClass::GetFireError, TechnoClass::Fire, TechnoClass::GetFLH` etc from `Hooks.cpp` to `Hooks.Firing.cpp`
* Moved methods of `TechnoExt` and `TechnoExt::ExtData` that are either directly related to `TechnoClass::Update` hooks or virtually updates something from `Body.cpp` to `Body.Ext.Update.cpp`. Other methods that are used only once are moved to `Body.Ext.cpp`
* Hooks that aren't directly related to technos are moved to `Hooks.Misc.cpp`


chaserli:
If no further opinion on this one I'll merge this in 60 hours

Metadorius:
Why `Ext.cpp`? We don't use that name anywhere for files.

chaserli:
> Why `Ext.cpp`? We don't use that name anywhere for files.

Any suggestions?

PR #854, opened 2023-03-11
[Minor] Fix incomplete inheritance for some value types

MortonPL:
Booleans, floating point numbers, number pairs, and locomotors will now be inherited properly.


Metadorius (approved):

Starkku:
Seems to work now.

Starkku:
Actually on further testing, inheritance for Point3D seems to be missing still.

Metadorius:
@MortonPL just thought of something, doesn't GetInt also have an inlined GetString? Thus, we could just plug in the reimpl instead of implementing the inheritance logic twice.

PR #1068, opened 2023-05-28
Chrono sparkle display customization & improvements

Starkku:
### Chrono sparkle animation customization & improvements

- It is now possible to customize the frame delay between instances of `[General]` -> `ChronoSparkle1` animations created on objects being warped by setting `[General]` -> `ChronoSparkleDisplayDelay`.
- By default on buildings with `MaxOccupants` higher than 0, chrono sparkle animation would be shown at each of the `MuzzleFlashN` coordinates. This behaviour is now customizable, and supports `MuzzleFlashN` indices higher than 10.
  - `[General]` -> `ChronoSparkleBuildingDisplayPositions` can be set to show the sparkle animation on the building (`building`), muzzle flash coordinates of current occupants (`occupants`), muzzle flash coordinates of all occupant slots (`occupantslots`) or any combination of these.
    - If `occupants` or `occupantslots` is listed without `building`, a single chrono sparkle animation is still displayed on building if it doesn't have any occupants or it has `MaxOccupants` value less than 1, respectively. 
- The chrono sparkle animation that is displayed on building itself is also now displayed at the center of it rather than at center of its topmost cell.

In `rulesmd.ini`:
```ini
[General]
ChronoSparkleDisplayDelay=24                         ; integer, game frames
ChronoSparkleBuildingDisplayPositions=occupantslots  ; list of chrono sparkle position enum (building | occupants | occupantslots | all)
```

### DIFF
--- a/src/Ext/Building/Hooks.cpp
+++ b/src/Ext/Building/Hooks.cpp
@@ -6,6 +6,7 @@
 #include <GameOptionsClass.h>
 #include <Ext/House/Body.h>
 #include <Ext/WarheadType/Body.h>
+#include <TacticalClass.h>
 
 //After TechnoClass_AI?
 DEFINE_HOOK(0x43FE69, BuildingClass_AI, 0xA)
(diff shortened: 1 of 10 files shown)


src/Ext/Building/Hooks.cpp
```diff
@@ -30,6 +31,53 @@ DEFINE_HOOK(0x43FE69, BuildingClass_AI, 0xA)
 	return 0;
 }
 
+DEFINE_HOOK(0x4403D4, BuildingClass_AI_ChronoSparkle, 0x6)
+{
+	enum { SkipGameCode = 0x44055D };
```
MortonPL, line 36: Other hooks in this PR don't have named addresses.

MortonPL (approved):

PR #1071, opened 2023-05-29
Map event 603: There are no technos of the specified houses list

FS-21:
- Returns `True` if there are no technos of the specified houses in the map.
- The second parameter is a 0-based index for the `AIHousesList` section that specifies the list of possible `HouseTypes` that can be evaluated. The new `AIHousesList` section must be declared in `rulesmd.ini` for making this script work:

In `mycampaign.map`:
```ini
[Events]
...
ID=EventCount,...,601,0,[Index],...
...
```

In `rulesmd.ini`:
```ini
[AIHousesList]  ; List of HouseType lists
0=SOMEHOUSETYPE,SOMEOTHERHOUSETYPE,SAMPLEHOUSETYPE
1=ANOTHERHOUSETYPE,YETANOTHERHOUSETYPE
; ...
```

For example:

If you want to trigger the event if there are no units/structures of British, French and Germans houses (not related to players) then:

[Events]
01000000=1,601,0,1

[AIHousesList]
0=YuriCountry
1=British,French,Germans  ; <---------- This line will be used in the Map event
2=Germans
3=...


src/Ext/TEvent/Body.cpp
```diff
@@ -159,6 +163,42 @@ bool TEventExt::VariableCheckBinary(TEventClass* pThis)
...
+	{
+		if (ScriptExt::IsUnitAvailable(pTechno, false, true))
+		{
+			for (auto pHouse : housesList)
+			{
+				if (pTechno->Owner->Type == pHouse)
+					return false;
+			}
+		}
+	}
```
MortonPL, line 197:
```suggestion
    for (const auto pTechno : *TechnoClass::Array)
    {
		if (!ScriptExt::IsUnitAvailable(pTechno, false, true))
		    continue;
		for (const auto pHouse : housesList)
		{
			if (pTechno->Owner->Type == pHouse)
				return false;
		}
    }
```

Avoid nesting blocks if possible. Use `const` when possible.

docs/AI-Scripting-and-Mapping.md
```diff
@@ -583,3 +583,23 @@ ID=EventCount,...,600,2,0,0,...
 ...
 ```
 
+### `601` There are no technos of the specified houses list
+
+- Returns `True` if there are no technos of the specified houses in the map.
+- The second parameter is a 0-based index for the `AIHousesList` section that specifies the list of possible `HouseTypes` that can be evaluated. The new `AIHousesList` section must be declared in `rulesmd.ini` for making this script work:
```
MortonPL, line 589: I would call them `Countries` instead of `HouseTypes`, because the former term is used in rules and the latter inside the game engine.

MortonPL (changes requested):
Looks good except for a few small style changes and one terminology nitpick.

FS-21:
What requested changes are missed here?

PR #1073, opened 2023-06-01
Burst projectile retargeting

FS-21:
### Weapon random target
- Weapons with this logic will randomly pick targets within range upon firing or launching spawners.
- Works with missiles (no splits, airbursts, etc), cannons, lasers & spawners.
- A valid techno is required for triggering the logic.
- `OmniFire=yes` will make selectable any targets around the firer, limited by the weapon range.
- `OmniFire=no` will force the weapon to pick targets in an area composed by the firer's weapon range around the original target intersected with the firer's weapon range around the firer.
- `RandomTarget.Spawners.MultipleTargets=true` gives each spawner aircraft its own target.
- `RandomTarget.Spawners.RememberTargets=true` controls whether spawned aircraft remember and finish off their previous target upon returning/reloading if it is still alive (if `false`, a new random target is chosen every wave).
- `RandomTarget.MissChance` controls the probability of a shot missing completely and targeting a random ground cell in the area.
- `RandomTarget.FriendlyFireChance` controls the probability of a shot inverting its target house filter to target friendly/allied objects.
- This logic should be used only in one weapon of the object.

In `rulesmd.ini`:
```ini
[SOMEWEAPON]                                ; WeaponType
RandomTarget=0.0                            ; double or percentage
RandomTarget.Spawners.MultipleTargets=false ; boolean
RandomTarget.Spawners.RememberTargets=true  ; boolean
RandomTarget.MissChance=0.0                  ; double or percentage
RandomTarget.FriendlyFireChance=0.0          ; double or percentage
```


FS-21:
Regarding how affects the Omnifire tag.
imagen

mevitar:
Appears to work fine (even tried it for Trajectory=Straight projectiles), except for a case where a jumpjet without a turret and with OmniFire=no retargeted its burst fire into a landed AircraftType that was on a straight line behind. Same thing happened multuple times.

The jumpjet also had a secondary weapon that it switches to after running out of ammo, and that one has OmniFire=yes but it doesn't retarget.

FS-21:
> Appears to work fine (even tried it for Trajectory=Straight projectiles), except for a case where a jumpjet without a turret and with OmniFire=no retargeted its burst fire into a landed AircraftType that was on a straight line behind. Same thing happened multuple times.
> 
> The jumpjet also had a secondary weapon that it switches to after running out of ammo, and that one has OmniFire=yes but it doesn't retarget.

What kind of weapons were used in the jumpjet case? I specified they should be of the type missile (but I noticed that invisible bullets works too).

mevitar:
I know you said that it was designed for regular missiles, but it also appears to works fine for Trajectory=Straight (also tried with Arcing=yes, and it doesn't work there at all).
I've encountered this problem by accident while testing Trajectory=Straight, but apparently it's not the projectile, because there's an issue with some jumpjets that they might attack an incorrect target, regardless of what weapon it has. Probably a combination of firing angle + the jumpjet trying to reposition itself and face the target is what caused such behavior, but i'm unable to reproduce it anymore.

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -794,3 +795,89 @@ DEFINE_HOOK(0x5223B3, InfantryClass_Approach_Target_DeployFireWeapon, 0x6)
...
+	GET(BulletClass*, pBullet, EAX);
+
+	if (!pBullet || pWeapon->Burst < 2)
+		return 0;
+
+	const auto pWeaponExt = WeaponTypeExt::ExtMap.Find(pWeapon);
+	if (!pWeaponExt || pWeaponExt->Burst_Retarget <= 0.0)
+		return 0;
+
+	int retargetProbability = pWeaponExt->Burst_Retarget > 1.0 ? 100 : (int)round(pWeaponExt->Burst_Retarget * 100);
```
MortonPL, line 812:
```suggestion
	int retargetProbability = min((int)round(pWeaponExt->Burst_Retarget * 100), 100);
```

Perhaps a min() function would be more readable.

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -794,3 +795,89 @@ DEFINE_HOOK(0x5223B3, InfantryClass_Approach_Target_DeployFireWeapon, 0x6)
...
+	int dice = ScenarioClass::Instance->Random.RandomRanged(1,100);
```
MortonPL, line 813:
```suggestion
	int dice = ScenarioClass::Instance->Random.RandomRanged(1, 100);
```

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -794,3 +795,89 @@ DEFINE_HOOK(0x5223B3, InfantryClass_Approach_Target_DeployFireWeapon, 0x6)
...
+			|| pThisType->Immune
+			|| !EnumFunctions::IsTechnoEligible(pTarget, pWeaponExt->CanTarget, true)
+			|| (!pWeapon->Projectile->AA && pTarget->IsInAir())
+			|| (!pWeapon->Projectile->AG && !pTarget->IsInAir())
+			|| (!friendlyFire && (pThis->Owner->IsAlliedWith(pTarget) || ScriptExt:: IsUnitMindControlledFriendly(pThis->Owner, pTarget)))
+			|| pTarget->GetTechnoType()->Underwater && pTarget->GetTechnoType()->NavalTargeting == NavalTargetingType::Underwater_Never
+			|| pTarget->GetTechnoType()->Naval && pTarget->GetTechnoType()->NavalTargeting == NavalTargetingType::Naval_None)
+		{
+			continue;
+		}
```
MortonPL, line 839: Is there a better way to handle targeting other than manually checking all conditions? What about warhead versus, or CLEG'ed units, or shields, or cloaked units?

MortonPL (commented):
A few minor things and one question.

FS-21:
I updated with the rewrite of the feature.
Now isn't required a Burst > 1 to work and it supports lasers and spawners (I don't know if there are more supported cases).

Now the tag is called "RandomTarget" (docs updated and the first post has been updated)
With the new tag _RandomTarget.Spawners.MultipleTargets=true_ the firer gives each spawner it's own random target.

mevitar:
I found a potential issue with spawners on the Sept 22nd build, but can't test it if it still happens later because the newer build didn't build properly.

src/Ext/Techno/Body.cpp
```diff
@@ -637,6 +639,213 @@ bool TechnoExt::IsTypeImmune(TechnoClass* pThis, TechnoClass* pSource)
...
+		}
+		else
+		{
+			int distanceFromOriginalTargetToCandidate = pCandidate->DistanceFrom(pThis->Target);
+			int distanceFromOriginalTarget = pThis->DistanceFrom(pThis->Target);
+
+			if (pCandidate->IsInAir())
+				range = airRange;
+
+			if (distanceFromAttacker <= range && distanceFromOriginalTargetToCandidate <= distanceFromOriginalTarget)
```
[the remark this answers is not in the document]
Metadorius: I am also not sure what does this function do and why is it in ScriptExt: `ScriptExt::IsUnitMindControlledFriendly(pThis->Owner, pCandidate)`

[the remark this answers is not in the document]
Starkku: Function `ScriptExt::IsUnitMindControlledFriendly()` was originally placed in `ScriptExt` because that is the only place where it was called. If it is now being called elsewhere it should likely be relocated to `TechnoExt` since that would be more appropriate for function that checks `TechnoClass` state.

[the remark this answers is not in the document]
Metadorius: We usually place static extension functions as if they were the non-static functions of the class that is being extended (think of C# extension methods), so if it accepts the TechnoClass, it should be in TechnoExt. I am not sure if the description you gave is accurate, I would say that it needs a clarification from @FS-21.

mevitar:
Something with weapon firing restrictions is terribly wrong, because GI-like infantry is be able to fire while their deploy sequence is playing. This is not the appropriate behavior, nor does it happen in the dev build or without Phobos - they should wait with firing their weapon until their deploy sequence finishes playing.

Retargetting doesn't work properly for Aircraft Carrier type spawners. Spawned aircrafts can get bugged, and, instead of following the proper target, they will try to fly outside of the map.
retargetspawn2x

docs/New-or-Enhanced-Logics.md
```diff
@@ -1245,6 +1245,22 @@ ExtraWarheads=                  ; list of WarheadTypes
...
+- `RandomTarget.Spawners.MultipleTargets=true` gives each spawner it's own target.
+- This logic should be used only in one weapon of the object.
+
+In `rulesmd.ini`:
+```ini
+[SOMEWEAPON]                                ; WeaponType
+RandomTarget=0.0                            ; double or percentage
+RandomTarget.Spawners.MultipleTargets=false ; boolean
+```
+
```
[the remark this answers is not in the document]
FS-21: It was explained in that part, omit this documentation part.

src/Ext/Techno/Body.Update.cpp
```diff
@@ -6,6 +6,7 @@
 #include <Ext/Anim/Body.h>
 #include <Ext/Bullet/Body.h>
 #include <Ext/House/Body.h>
+#include <Ext/Script/Body.h>
```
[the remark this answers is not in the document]
FS-21: All headers here looks used in some functions so skip this suggestion.

src/Ext/Techno/Body.Update.cpp
```diff
@@ -719,3 +720,49 @@ void TechnoExt::UpdateSharedAmmo(TechnoClass* pThis)
...
+					pSpawn->Unit->Target = pSpawnExt->CurrentRandomTarget;
+				}
+				else if (pSpawn->Status == SpawnNodeStatus::Preparing && pSpawn->Unit->IsInAir())
+				{
+					if (!pSpawn->Unit->Target && pSpawnExt->CurrentRandomTarget)
+						pSpawn->Unit->Target = pSpawnExt->CurrentRandomTarget;
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
FS-21: You are right!

src/Ext/Techno/Body.Update.cpp
```diff
@@ -719,3 +720,49 @@ void TechnoExt::UpdateSharedAmmo(TechnoClass* pThis)
...
+
+	if (pExt->OriginalTarget && !pThis->Target && ScriptExt::IsUnitAvailable(static_cast<TechnoClass*>(pExt->OriginalTarget), true) && !pThis->IsInAir())
+	{
+		if (pExt->CurrentRandomTarget && ScriptExt::IsUnitAvailable(pExt->CurrentRandomTarget, true))
+			pThis->SetTarget(pExt->CurrentRandomTarget);
+		else
+			pThis->SetTarget(pExt->OriginalTarget);
+	}
```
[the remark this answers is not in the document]
FS-21: IsUnitAvailable() uses 2 different parameter values so you are wrong in this case.

src/Ext/Techno/Body.cpp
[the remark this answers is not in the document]
FS-21: Do you mean ScriptExt::IsUnitMindControlledFriendly()? I don't touch what other people do, like that ScriptExt::IsUnitAvailable().
I'll move them to TechnoExt then.

FS-21:
> Something with weapon firing restrictions is terribly wrong, because GI-like infantry is be able to fire while their deploy sequence is playing. This is not the appropriate behavior, nor does it happen in the dev build or without Phobos - they should wait with firing their weapon until their deploy sequence finishes playing.

Then I need to add additional checks in that case.

> Retargetting doesn't work properly for Aircraft Carrier type spawners. Spawned aircrafts can get bugged, and, instead of following the proper target, they will try to fly outside of the map. retargetspawn2x [ retargetspawn2x ](https://user-images.githubusercontent.com/1313344/282047376-60a36d94-1951-4202-88d6-2959c61c8e6b.gif) [ ](https://user-images.githubusercontent.com/1313344/282047376-60a36d94-1951-4202-88d6-2959c61c8e6b.gif) This is with RandomTarget=100% and RandomTarget.Spawners.MultipleTargets=yes.

How is that Carrier weapon? The original Allied Carrier works fine with this:
```
[HornetLauncher]
Damage=1
ROF=150
Range=25
Spawner=yes
Projectile=Invisible
Speed=10
Warhead=Super
OmniFire=yes
RandomTarget=1.0 ; double or percentage
RandomTarget.Spawners.MultipleTargets=true ; boolean
```

mevitar:
It has nothing to do with the weapon.

The retargeting apparently ignores whether an object is actually targetable by the spawner weapon, which means it will still consider cloaked (and invisible) objects, objects with Immune=yes, and objects with 0% verses as legal targets. Thus, this behavior, with spawned aircraft flying away because it is unable to attack the thing it was ordered to target.

Don't know if this happens to non-spawner weapons.

FS-21:
Yes, currently the method for picking the target doesn't ask for the spawned weapons capabilities so I'll address the deploy & this issue in the weekend.

mevitar:
> The retargeting apparently ignores whether an object is actually targetable by the spawner weapon, which means it will still consider cloaked (and invisible) objects, objects with Immune=yes, and objects with 0% verses armor as legal targets. Thus, this behavior, with spawned aircraft flying away because it is unable to attack the thing it was ordered to target.

I don't know if this was supposed to be fixed or not, but i got to test it in the newest build available, and the issue persists. Cloaked units still cause spawners to fly away.

FS-21:
I forgot about that case x'D

When I arrive home I would like to update the code.

mevitar:
No need to hurry.

On a side note:
> Something with weapon firing restrictions is terribly wrong, because GI-like infantry is be able to fire while their deploy sequence is playing. This is not the appropriate behavior, nor does it happen in the dev build or without Phobos - they should wait with firing their weapon until their deploy sequence finishes playing.

There's something else going with this, it seems. I don't know how did i manage to make it not happen without Phobos before, but i retested it now and it does happen, and i didn't really change anything for the unit in question (at least i don't think i did). It appears to happen randomly, and the more infantry units are in a group, the more noticeable it becomes.

Either way, no need to fix anything with this one, I'll probably report that as a separate issue.

src/Ext/Event/Body.cpp
```diff
@@ -13,8 +14,8 @@ void EventExt::RespondEvent()
 {
 	switch (this->Type)
 	{
-	case EventTypeExt::Sample:
-		// Place the handler here
+	case EventTypeExt::SyncStopRandomTargetTarNav:
```
TaranDahl: This is incorrect. You should hook into the process of Event_Idle and add your process there, rather than add a new event.

src/Ext/Techno/Body.h
```diff
@@ -265,6 +278,10 @@ class TechnoExt
 	static Point2D GetBuildingSelectBracketPosition(TechnoClass* pThis, BuildingSelectBracketPosition bracketPosition);
 	static void DrawSelectBox(TechnoClass* pThis, const Point2D* pLocation, const RectangleStruct* pBounds, bool drawBefore = false);
 	static void ProcessDigitalDisplays(TechnoClass* pThis);
+	static void NewRandomTarget(TechnoClass* pThis = nullptr);
+	static TechnoClass* FindRandomTarget(TechnoClass* pThis = nullptr);
```
TaranDahl: Why declared as static?

docs/New-or-Enhanced-Logics.md
```diff
@@ -2558,6 +2558,22 @@ ExtraWarheads.DetonationChances=  ; List of floating-point values (percentage or
...
 ```
 
+### Projectile's random target
+- The firer will pick targets randomly.
+- Works with missiles (no splits, airbusts, etc), cannons, lasers & spawners.
+- A valid techno is required for trigger the logic.
+- `OmniFire=yes` will make selectable any targets around the firer, limited by the weapon range.
+- `OmniFire=no` will force the firer to pick targets in an area composed by the firer's weapon range around the original target intersected with the firer's weapon range around the firer.
+- `RandomTarget.Spawners.MultipleTargets=true` gives each spawner it's own target.
+- This logic should be used only in one weapon of the object.
```
TaranDahl, line 3570: Then why not put it on TechnoType?

docs/New-or-Enhanced-Logics.md
```diff
@@ -2558,6 +2558,22 @@ ExtraWarheads.DetonationChances=  ; List of floating-point values (percentage or
 ExtraWarheads.FullDetonation=     ; List of booleans
 ```
 
+### Projectile's random target
+- The firer will pick targets randomly.
```
TaranDahl: The title and the description are inconsistent.
Firer and proj are two completely different things. Your title says proj, when your description says firer.

TaranDahl (dismissed):
There are too many redundant checks, and I can't bear to look at them.
I imagine that the implementation of this logic should be very concise and doesn't require that much code. I think you're trying to implement the function you want in a completely wrong way, and according to feedback from others, it's not working the way you hoped.
Please rewrite the whole using the following logic, or delete unnecessary codes, briefly summarize the operational logic of your program in words and convince me why not to use the following logic:
For the case of ordinary projectiles: When a unit is about to fire, call `CanAutoTargetObject` and `GetFireError` on all units within range. If the former returns false, or the latter returns a code other than Facing and OK, exclude that unit. Randomly select one from the available units and fire at it (should hook into Fire func and change the FireError check).
For the case of the spawner: Declare a target list for SpawnerExt, and replace all reference to the vanilla property Target with the target list. When spawning spawners, find a valid target and add it to the target list. The spawners will act according to the target corresponding to themselves in the target list.

docs/New-or-Enhanced-Logics.md
FS-21: Why all weapons should be affected at the same time by this logic instead of per weapon?

FS-21:
All the code has been rewritten.
Updated the first message of this PR.

src/Ext/Techno/Body.RandomTarget.cpp
```diff
@@ -0,0 +1,202 @@
...
+	if (pOriginalTarget)
+	{
+		if (auto const pOrigTechno = abstract_cast<TechnoClass*>(pOriginalTarget))
+		{
+			if (!targetAllies || (pFirer->Owner && (pFirer->Owner->IsAlliedWith(pOrigTechno->Owner) || pOrigTechno->Owner == pFirer->Owner)))
+				candidates.push_back(pOrigTechno);
+		}
+	}
+
+	for (const auto pCandidate : TechnoClass::Array)
```
TaranDahl: Use the helper

src/Ext/Techno/Body.RandomTarget.cpp
```diff
@@ -0,0 +1,202 @@
...
+			continue;
+
+		candidates.push_back(pCandidate);
+	}
+
+	if (candidates.empty())
+		return pOriginalTarget;
+
+	const int pick = ScenarioClass::Instance->Random.RandomRanged(0, static_cast<int>(candidates.size()) - 1);
+	return candidates[pick];
```
TaranDahl: Why reinvent the wheel for GetFireError?

src/Ext/WeaponType/Body.h
```diff
@@ -113,6 +112,13 @@ class WeaponTypeExt final : public AbstractTypeExt
 	Nullable<bool> AttackCursorOnFriendlies;
 	Nullable<bool> AttackNoThreatBuildings;
 
+	Valueable<double> RandomTarget;
+	Valueable<bool> RandomTarget_Spawners_MultipleTargets;
+	Valueable<bool> RandomTarget_Spawners_RememberTargets;
+	Valueable<double> RandomTarget_MissChance;
+	Valueable<double> RandomTarget_FriendlyFireChance;
+	ValueableVector<TechnoTypeClass*> OnlyTargetTechnos;
```
TaranDahl, line 120: This is not in the doc.  For what reason?

src/Ext/Techno/Body.RandomTarget.cpp
```diff
@@ -0,0 +1,202 @@
...
+		if (!pWeaponExt->CanOnlyTargetTheseTechnos(pCandidateType))
+			continue;
+
+		if (!pWeaponExt->IsHealthInThreshold(pCandidate))
+			continue;
+
+		if (!pWeaponExt->HasRequiredAttachedEffects(pCandidate, pFirer))
+			continue;
+
+		candidates.push_back(pCandidate);
```
TaranDahl, line 170: Why reinvent the wheel for GetFireError?

TaranDahl (changes requested):

src/Ext/WeaponType/Body.h
FS-21: What tag?

src/Ext/Techno/Body.RandomTarget.cpp
FS-21: Reused old code from the first implementations.

PR #1075, opened 2023-06-09
Warhead animation improvements

Starkku:
- Reorganized `BulletClass::Logics()` hooks and moved them to separate files, and moved hooks not directly related to `CreateUnit` away from `Hooks.AnimCreateUnit.cpp`
- `SplashList.PickRandom` now applies to the anims from `[CombatDamage]` -> `SplashList` if Warhead doesn't have `SplashList` animations defined.
- `AnimList.ShowOnZeroDamage` has been renamed to `CreateAnimsOnZeroDamage` to make it more clear it applies to both `AnimList` and splash animations.
- Moved all AnimList etc. related documentation to a single location.
-  Added `AnimList/SplashList.CreateAll` which makes Warhead create all listed animations instead of only one.
- Added `Conventional.IgnoreUnits` which makes Warhead treat all non-underwater VehicleTypes on water tiles as if it was detonated on water instead of land.
------------------------------

### Customizable Warhead animation behaviour

- It is possible to make game play random animation from `AnimList` by setting `AnimList.PickRandom` to true. The result is similar to what `EMEffect=true` produces, however it comes with no side-effects (`EMEffect=true` prevents `Inviso=true` projectiles from snapping on targets, making them miss moving targets).
- If `AnimList.CreateAll` is set to true, all animations from `AnimList` are created, instead of a single anim based on damage or random if  `AnimList.PickRandom` is set to true.
- `SplashList` can be used to override animations displayed if the Warhead has `Conventional=true` and it hits water, by default animations from `[CombatDamage]` -> `SplashList` are used.
  - `SplashList.PickRandom` and `SplashList.CreateAll` apply to these animations in same manner as the `AnimList` equivalents.
- `CreateAnimsOnZeroDamage`, if set to true, makes it so that `AnimList` or `SplashList` animations are created even if the weapon that fired the Warhead deals zero damage.
- Setting `Conventional.IgnoreUnits` to true on Warhead with `Conventional=true` will make the Warhead detonate on non-underwater VehicleTypes on water tiles as if they are water tiles, instead of treating it as land.

In `rulesmd.ini`:
```ini
[SOMEWARHEAD]                   ; WarheadType
AnimList.PickRandom=false       ; boolean
AnimList.CreateAll=false        ; boolean
SplashList=                     ; List of animations
SplashList.PickRandom=false     ; boolean
SplashList.CreateAll=false      ; boolean
CreateAnimsOnZeroDamage=false   ; boolean
Conventional.IgnoreUnits=false  ; boolean
```


mevitar:
I can't get RemapAnim=yes to work for anything. I tried for AircraftType trailers, projectile trailers, animation trailers, and also shield animations, nothing was changed. Animations were still drawn in animation palette, and if i set AltPalette=yes, they were remapping to orange (which indicates no owner set).

Phobos documentation says that trailer anims now inherit the owner of the object they are attached to, but do they really?

(actually i shouldn't be saying "for anything", because RemapAnim=yes does work for warhead anims, but that's also the only thing i tested that worked)

Starkku:
There is a distinction between having a known owner and having the animation's drawer set to the house's unit palette drawer. Some of the Phobos animation things may be missing latter, I'll go over it later and add where needed.

Starkku:
Added support for trailer animation and shield animation remapping

mevitar:
Perhaps the first thing i should have done is to ask which animations have an owner, because, since i didn't exactly know, i tested everything i could think of, and RemapAnim=yes seems to work only for very few specific cases.

Shields and trailers do remap now, but i noticed they don't change their remap color when changing owner (like through mind control). They always remap to their original owner.
Don't even know if changing this is feasible, since the anim would have to check for its ownership each frame, and that might get too costy performance-wise.
Perhaps it should stay like this and a note about this behavior should be made in the doc. Or perhaps asking to make them remapable was a mistake...

Speaking of mind control, RemapAnim=yes does not work for mind control animations. I saw this before, but completely forgot to mention it. Same with Ares' version of AE, but i suppose that might be difficult to change due to it being an Ares thing.
Didn't check the Phobos AE, but i assume it also doesn't work.

What surprised me was that RemapAnim=yes also doesn't work for PreImpactAnim=. I assumed it would be handled the same way as regular AnimList=animations, but apparently i was wrong.

So, either animations from each of these tags require their drawer to be set separately, or i'm testing things RemapAnim=yes wasn't meant to support in the first place...
And i'm afraid to ask if i should test SuperWeapons too...

Starkku:
Problem is that it needs to be implemented separately for every type of animation created so doing full coverage is kind of difficult. It is technically separate from animation owner, as I stated before. Game also needs to set the animation's drawer to the one that has team colors.

Starkku:
An option that I realized would be feasible is replacing `RemapAnim` with fixing `AltPalette` to use animation owner's remap color (which would also fix it drawing over shroud). I am unsure if this would be preferrable though, not sure if people somewhere are relying on these animations drawing over shroud.

mevitar:
Personally, i think it should be fixed. Even if people rely on AltPalette=yes for over-the-shroud visibility, the first thing that modders expect from this tag is that it makes their anims able to remap just like units do, instead of remapping into orange.

However, there is an alternative option of making RemapAnim=yes replicate what AltPalette=yes does, except "properly", allowing the animations to be remapped (and without the drawing-over-shroud thing). Sure, it means redundancy, but AltPalette=yes could be left untouched.

Starkku:
Removed `RemapAnim` as it is now obsolete with the `AltPalette` improvements.

docs/Whats-New.md
```diff
@@ -20,6 +20,7 @@ You can use the migration utility (can be found on [Phobos supplementaries repo]
 
 #### From post-0.3 devbuilds
 
+- `AnimList.ShowOnZeroDamage` has been renamed to `CreateAnimsOnZeroDamage` to make it more clear it applies to both `AnimList` and splash animations.
```
MortonPL, line 23: This should be reflected in the sed script in the supplementary repo.

MortonPL (approved):

docs/Whats-New.md
Starkku: Done

PR #1078, opened 2023-06-15
Hit bright

Super-StarX:
same as iron curtain or force shield be hit
now only on shield, enhance if needed.

### DIFF
--- a/src/New/Type/ShieldTypeClass.h
+++ b/src/New/Type/ShieldTypeClass.h
@@ -28,7 +28,7 @@ public:
 	Nullable<AnimTypeClass*> BreakAnim;
 	Nullable<AnimTypeClass*> HitAnim;
 	Nullable<WeaponTypeClass*> BreakWeapon;
-	Valueable<CoordStruct> HitBright;
+	Valueable<ColorStruct> HitBright;
 	Valueable<double> AbsorbPercent;
 	Valueable<double> PassPercent;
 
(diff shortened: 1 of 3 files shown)


src/New/Type/ShieldTypeClass.h
```diff
@@ -28,6 +28,7 @@ class ShieldTypeClass final : public Enumerable<ShieldTypeClass>
 	Nullable<AnimTypeClass*> BreakAnim;
 	Nullable<AnimTypeClass*> HitAnim;
 	Nullable<WeaponTypeClass*> BreakWeapon;
+	Valueable<CoordStruct> HitBright;
```
MortonPL, line 31: Use `ColorStruct` instead of `CoordStruct` to store color information.

src/New/Type/ShieldTypeClass.h
```diff
@@ -60,6 +61,7 @@ class ShieldTypeClass final : public Enumerable<ShieldTypeClass>
 		, BreakAnim { }
 		, HitAnim { }
 		, BreakWeapon { }
+		, HitBright { {0,0,0} }
```
MortonPL, line 64: Use empty initializer list unless you provide a non-zero default value.

MortonPL (changes requested):
Considering that this is effect is referred to as ["Flash" in Ares](https://ares-developers.github.io/Ares-docs/new/warheads/ironcurtain.html#iron-curtain-flash), this should be renamed to remain consistent.

Add an entry in credits, documentation and "What's New".

Starkku:
Implemented in more comprehensive way in 7cbffda, so closing this.

PR #1084, opened 2023-06-24
Jumpjet initial facing fix

Metadorius:
- Fixed the bug when jumpjets were snapping into facing bottom-right when starting movement (observable when the starting unit is a jumpjet and is ordered to move).

How to test: make any jumpjet unit as a starting unit. It should be facing top right when spawned in. When you order it to move, without the fix it will snap to bottom right, with the fix it will keep the initial facing.


NetsuNegi:
It works fine

MortonPL (approved):
LGTM.

mevitar:
The fix makes it so pre-placed jumpjets will turn to ingame-north immediately after lifting up, no matter what direction they were facing on ground.

PR #1092, opened 2023-06-30
[Minor] Re-fix jumpjet initial facing

Metadorius:
https://github.com/Phobos-developers/Phobos/issues/1084#issuecomment-1613371056

How to test:
- pre-place jumpjet units on a map and move them, they should keep their direction and not snap into any other dir;
- use jumpjets as starting units, they also should keep the initial direction

@mevitar


Starkku:
Seems to work.

PR #1098, opened 2023-07-07
Additional desync logging

Starkku:
This adds following things, including relevant parameters, function caller/return address and current frame to be appended to `SYNC#.txt` log files upon desynchronization occuring:

- Last 4096 calls to game's synced random number generator function (the one that is `ScenarioClass` member) 
- Last 1024 `FacingClass::Set()` calls
- Last 1024 `TechnoClass::AssignTarget()` calls (including overrides)
- Last 1024 `TechnoClass::AssignDestination()` calls (including overrides)
- Last 256 `TechnoClass::OverrideMission()` calls
- Last 512 `AnimClass` constructor calls

These numbers can be amended later if needed. This feature is also currently enabled by default in multiplayer games only (the hooks that log this information are disabled in singleplayer/skirmish), and can incur a performance decrease. A toggle can be added later if needed.

### DIFF
--- a/src/Misc/SyncLogging.h
+++ b/src/Misc/SyncLogging.h
@@ -9,6 +9,7 @@
 static constexpr unsigned int RNGCalls_Size = 4096;
 static constexpr unsigned int FacingChanges_Size = 1024;
 static constexpr unsigned int TargetChanges_Size = 1024;
+static constexpr unsigned int DestinationChanges_Size = 1024;
 
 template <typename T, unsigned int size>
 class SyncLogEventTracker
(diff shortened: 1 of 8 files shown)


NetsuNegi:
I think it's better to add a command line param switch, if this cause performence problem.

src/Misc/SyncLogging.h
```diff
@@ -0,0 +1,173 @@
...
+static constexpr unsigned int TargetChanges_Size = 1024;
+static constexpr unsigned int DestinationChanges_Size = 1024;
+static constexpr unsigned int MissionOverrides_Size = 256;
+static constexpr unsigned int AnimCreations_Size = 512;
+
+template <typename T, unsigned int size>
+class SyncLogEventBuffer
+{
+private:
+	std::vector<T> Data;
```
Metadorius, line 21: Maybe it is better to use a queue approach here instead of a vector? Not that it should change much but eh, better usage experience.

Metadorius (approved):

PR #1100, opened 2023-07-09
Convert.Anim tag

FS-21:
"Convert.Anim" tag for the Ares Conversion logic.

All successful conversions should attach the animation if specified.


FS-21:
convert_anim_01

src/Ext/SWType/FireSuperWeapon.cpp
```diff
@@ -307,6 +307,9 @@ void SWTypeExt::ExtData::ApplyTypeConversion(SuperClass* pSW)
 	if (this->Convert_Pairs.size() == 0)
 		return;
 
+	const auto pTypeExt = SWTypeExt::ExtMap.Find(pSW->Type);
```
MortonPL, line 310: This is unnecessary, because what you end up with is the same as `this`.

MortonPL (changes requested):
Found one unnecessary thing, rest should be okay.

brsajo:
Is it possible to have an animation play prior to the conversion? E.g. to mimic the genetic mutator brute anim where unit A is hidden while the anim plays and unit B appears once the anim is over.

src/Ext/SWType/FireSuperWeapon.cpp
FS-21: Makes sense!

chaserli:
pretty sure we have plenty methods for attaching animations on wh effects, idk why this is even needed

FS-21:
> pretty sure we have plenty methods for attaching animations on wh effects, idk why this is even needed

Show me if you can easily show what units were converted with superweapons or Warheads, (I'm curious as modder). :-/
The only tag I find I miss here is the animation offset-Y because you could reuse the same animation for units/structures with different heights...

chaserli:
> Show me if you can easily show what units were converted with superweapons or Warheads,

Convert sw includes units in limbo, that's the only exception from a generic full map detonation. If you want to create an anim the unit should be in logic, in that case detonating 1 or multiple wh to attach an animation should suffice 

In case a wh couldn't apply the effect, the change should be generalized to that aspect, not this specific issue

PR #1101, opened 2023-07-09
Added "VoiceCreated" for units as alternative to CreateSound

Fryone:
Plays unit's "VoiceCreated" sound entry at it position after being trained/built.


brsajo:
What's the advantage of using this instead of CreateSound? Or is it just an alias?

Metadorius:
Think it's played only for the player who built it.

docs/New-or-Enhanced-Logics.md
```diff
@@ -928,6 +928,16 @@ EVA.Sold=       ; EVA entry
 SellSound=      ; sound entry
 ```
 
+### Sound entry on unit's creation
+
+- When unit is created, at it's position `VoiceCreated` will be played.
```
MortonPL, line 933:
```suggestion
- When a unit is created, sound specified in `VoiceCreated` will be played for the unit owner.
```

MortonPL (commented):
Looks fine as it is, but I think it'd be a good idea to either play the sound globally (but only to the player) by default or add a toggle for that, because this is how it works in most other RTS games, including later C&C titles.

docs/New-or-Enhanced-Logics.md
Fryone: Well, I can add toggle to make it global, but with global `VoiceCreated` sound I need to disable EVA's `UnitReady` to prevent 
mess

mevitar:
I'd like to test the newest build here but can't since it didn't build. :D

Flactine:
That's great! I like listen "Kirov Reporting" when I have Kirov while my enemies can't heard it.

Fryone:
> I'd like to test the newest build here but can't since it didn't build. :D

Needs YRPP update with "EVA_UnitReady", that's why it's not building.

Metadorius:
> > I'd like to test the newest build here but can't since it didn't build. :D
> 
> Needs YRPP update with "EVA_UnitReady", that's why it's not building.

You have to push your YRpp update to a branch too and PR it to YRpp, then it'll build.

Fryone:
> > > I'd like to test the newest build here but can't since it didn't build. :D
> > 
> > 
> > Needs YRPP update with "EVA_UnitReady", that's why it's not building.
> 
> You have to push your YRpp update to a branch too and PR it to YRpp, then it'll build.

https://github.com/Phobos-developers/YRpp/pull/16

mevitar:
Seems to work fine.
However, is IsVoiceCreatedGlobal=yes really necessary? Putting Type=global on sound's entry in SoundMD.ini does exactly the same thing.

Fryone:
> Seems to work fine. However, is IsVoiceCreatedGlobal=yes really necessary? Putting Type=global on sound's entry in SoundMD.ini does exactly the same thing.

When it does not set as global, it also plays "EVA_UnitReady", after creating, like in original

mevitar:
Ah, right, the doc says that it plays instead of that one.

MortonPL (changes requested):

MortonPL (approved):

src/Ext/House/Hooks.UnitFromFactory.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include <Ext/TechnoType/Body.h>
+#include <Utilities/EnumFunctions.h>
+#include <Utilities/GeneralUtils.h>
+#include <Utilities/Macro.h>
+
+DEFINE_HOOK(0x4FB63A, HouseClass_UnitFromFactory_DisablingEVAUnitReady, 0xF)
+{
+	if(!RulesExt::Global()->IsVoiceCreatedGlobal.Get())
+		VoxClass::Play(GameStrings::EVA_UnitReady);
+	return 0x4FB649;
```
Metadorius, line 13:
```suggestion

	return 0x4FB649;
```

src/Ext/House/Hooks.UnitFromFactory.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include <FactoryClass.h>
+
+#include <Ext/TechnoType/Body.h>
+#include <Utilities/EnumFunctions.h>
+#include <Utilities/GeneralUtils.h>
+#include <Utilities/Macro.h>
+
+DEFINE_HOOK(0x4FB63A, HouseClass_UnitFromFactory_DisablingEVAUnitReady, 0xF)
+{
+	if(!RulesExt::Global()->IsVoiceCreatedGlobal.Get())
```
Metadorius, line 11:
```suggestion
	if (!RulesExt::Global()->IsVoiceCreatedGlobal.Get())
```

PR #1105, opened 2023-07-17
Fixed a bug that computer player record cannot be log normally in non English mode

NetsuNegi:
Fixed the bug that computer player record cannot be log normally in non English mode


MortonPL (changes requested):
You should describe what the bug exactly is and how to reproduce it / when it happens.

Coronia (approved):

PR #1118, opened 2023-08-01
Change target Owner on warhead impact

Fryone:
### Change target Owner on impact

- Warheads can now change targets owner to warhead's owner.
- `ChangeOwner.SetAsMindControl` makes the effect work like permanent mind control, which respects `ImmuneToPsionics`.
  - `ChangeOwner.MindControlAnim` determines the mind control anim of this effect, which respects `MindControlRingOffset`.

In `rulesmd.ini`:
```ini
ChangeOwner=false                           ; boolean
ChangeOwner.SetAsMindControl=false          ; boolean
ChangeOwner.MindControlAnim=                ; Animation
```


brsajo:
The `IsOwnerChanger.ControlAnim` anim doesn't follow the unit around like the normal mindcontrol animation as I would have expected. If not a bug, could this be added as an option?

image

Fryone:
> The `IsOwnerChanger.ControlAnim` anim doesn't follow the unit around like the normal mindcontrol animation as I would have expected. If not a bug, could this be added as an option?
> 
weird, can you please provide settings for owner changer? As I tested, my anim is following techno as should

chaserli:
It's a bit messed up.
If this is supposed to be a generic house change, fine, but do not mess up with existing MindControl or permanent mind control logic.
Even if it's for a generic house change, a simple ownership change vtbl call is not sufficient. I would suggest you look at how it's done for permanent mind control in YR and Ares in the first place.
Same issue with all your other pr.
btw it seems the bot is triggered upon every commit, so plz don't just merge from upstream constantly

Fryone:
I'm not touching MC, instead I'm using PermaMC as it's much simpler. The only issue is that ownerchanged units without PermaMC set, allowed to enter in transports, makes their attached animation (if its looping) dissapear on leaving transport. So I'll just drop it for ares' attached anim instead.

brsajo:
> weird, can you please provide settings for owner changer? As I tested, my anim is following techno as should

The nightly build seems to be failing so I can't test your latest changes but here's what I have
```
[Controller]
AnimList=YURICNTL
; MindControl=yes
Verses=100%,100%,100%,100%,100%,100%,0%,0%,0%,100%,100%
IsOwnerChanger=yes
IsOwnerChanger.SetAsMindControl=yes
IsOwnerChanger.ControlAnim=MINDANIM
IsOwnerChanger.AffectElites=yes
IsOwnerChanger.HealthThreshold=1.0
```
```
[MINDANIM]
LoopCount=-1
Rate=300
```
I also tested with EMP_FX01 and MINDANIMR (unchanged from vanilla).

Fryone:
Yep, as I thought, now 'ControlAnim' changed to 'MindControlAnim' and will follow target as it should.

src/Ext/WarheadType/Detonate.cpp
```diff
@@ -250,6 +253,53 @@ void WarheadTypeExt::ExtData::ApplyRemoveDisguiseToInf(HouseClass* pHouse, Techn
...
+	const auto pOwnerAnimType = this->ChangeOwner_MindAnim.Get(nullptr);
+	const bool doesAffectElites = this->ChangeOwner_AffectElites;
+	const bool targetImmuneToPsionics = pTarget->GetTechnoType()->ImmuneToPsionics;
+	const bool isBld = pTarget->What_Am_I() == AbstractType::Building;
+
+	bool isImmune = (isMindControl && targetImmuneToPsionics) || (pTarget->GetHealthPercentage() > ownerChangeHealthThreshold) || (!doesAffectElites && pTarget->Veterancy.IsElite());
+
+	if (!isImmune)
+	{
+		if ((GeneralUtils::GetWarheadVersusArmor(this->OwnerObject(), armorType) != 0.0) && (!pTarget->IsMindControlled()))
```
TaranDahl: No need to check for versus, since `CanAffectTarget` will check it.

src/Ext/WarheadType/Detonate.cpp
```diff
@@ -250,6 +253,53 @@ void WarheadTypeExt::ExtData::ApplyRemoveDisguiseToInf(HouseClass* pHouse, Techn
...
+				{
+					CoordStruct OwningAnimLocation = pTarget->Location;
+					if (isBld)
+						OwningAnimLocation.Z += specific_cast<BuildingClass*>(pTarget)->Type->Height * Unsorted::LevelHeight;
+					else
+						OwningAnimLocation.Z += pTarget->GetTechnoType()->MindControlRingOffset;
+
+					if (auto pOwnerAnim = GameCreate<AnimClass>(pOwnerAnimType, OwningAnimLocation, 0, 1))
+					{
+						pOwnerAnim->Owner = pHouse;
```
TaranDahl: Not sure if this is correct. Vanilla MC does not set this.

src/Ext/WarheadType/Body.cpp
```diff
@@ -127,6 +127,11 @@ void WarheadTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 	this->DecloakDamagedTargets.Read(exINI, pSection, "DecloakDamagedTargets");
 	this->ShakeIsLocal.Read(exINI, pSection, "ShakeIsLocal");
 	this->ApplyModifiersOnNegativeDamage.Read(exINI, pSection, "ApplyModifiersOnNegativeDamage");
+	this->ChangeOwner.Read(exINI, pSection, "IsOwnerChanger");
+	this->ChangeOwner_MindControl.Read(exINI, pSection, "IsOwnerChanger.SetAsMindControl");
+	this->ChangeOwner_MindAnim.Read(exINI, pSection, "IsOwnerChanger.MindControlAnim");
+	this->ChangeOwner_Threshold.Read(exINI, pSection, "IsOwnerChanger.HealthThreshold");
```
TaranDahl: `AffectsBelowPercent` and `AffectsAbovePercent` have already done that.

src/Ext/WarheadType/Body.cpp
```diff
@@ -127,6 +127,11 @@ void WarheadTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
...
+	this->ChangeOwner_AffectElites.Read(exINI, pSection, "IsOwnerChanger.AffectElites");
```
TaranDahl: This should be another general checker in `WarheadTypeExt::ExtData::CanAffectTarget`.

TaranDahl (changes requested):
Fix the following things, then this can be merged.

Coronia:
changed tag names to `ChangeOwner` and removed repetitive versus and health check

I'm not sure about whether the elite check should be removed tho. There're only a few warhead effects which really need it, so maybe it don't have to be a general check

Fryone:
maybe `ChangeOwner.MindControlAnim` should be renamed to `ChangeOwner.ControlAnim` too

Coronia (approved):
removed ChangeOwner.AffectElites due to the introduction of generic veterancy filter: https://github.com/Phobos-developers/Phobos/pull/2008

Fryone:
Is it worth to add timed control, similar to Dune's Deviator?

Coronia:
I think it could be useful, but we should also think about how to deal with the interaction of multiple ChangeOwner warheads and other owner changer effects (mind control, capture etc). Maybe it's an easier way to just make the timer be interrupted by another change house event (there's existing hook for that too)

Fryone:
I thought of something like levels, basic integer that orders different changers (change if new one is higher)

Coronia:
ChangeOwner.Duration is removed for now cuz it's difficult to handle the scenarios when multiple house changing events are involved. Make it a one time change again until we figure out a better interaction

PR #1120, opened 2023-08-06
Fix issues with AutoDeath and limbo interactions

Starkku:
- All AutoDeath conditions are now considered for LimboDelivery objects.
- Technos(Dont)Exist.AllowLimboed now works as expected and doesn't count units still being built.
  - This required adding a new tracker and bunch of bullshit, I tested most cases but I wouldn't be surprised if I missed something.

### DIFF
--- a/src/Ext/House/Body.cpp
+++ b/src/Ext/House/Body.cpp
@@ -388,22 +388,100 @@ HouseClass* HouseExt::GetHouseKind(OwnerHouseKind const kind, bool const allowRa
 
 void HouseExt::ExtData::UpdateAutoDeathObjectsInLimbo()
 {
-	for (auto pExt : this->OwnedTimedAutoDeathObjects)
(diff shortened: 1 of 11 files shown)


src/Ext/House/Body.cpp
```diff
@@ -366,24 +366,102 @@ HouseClass* HouseExt::GetHouseKind(OwnerHouseKind const kind, bool const allowRa
...
+		if (!pBuilding->IsInLogic && pBuilding->IsAlive)
 		{
-			auto const pBuilding = abstract_cast<BuildingClass*>(pItem);
+			if (pExt->TechnoExtData->CheckDeathConditions(true))
+			{
+				this->OwnedLimboDeliveredBuildings.erase(it++);
+				this->RemoveFromLimboTracking(pBuilding->Type);
+			}
+			else
+				it++;
```
Metadorius, line 382:
```suggestion
			else
			{
				it++;
			}
```

Metadorius (approved):

PR #269, opened 2021-07-04
Feature/arcing fix

Thrifinesma:
Fix `Arcing` projectiles about their inaccuracy due to Z calculation fault (?)

PR #1121, opened 2023-08-06
Fix arcing elevation bug for most part

Starkku:
- `Arcing=true` projectile elevation inaccuracy can now be fixed by setting `Arcing.AllowElevationInaccuracy=false`
- `EMPulseCannon=yes` building weapons now respect `Floater` and Phobos-added `Gravity`setting.

It is not a perfect solution as the projectiles will still miss by like 1 cell with large elevation difference. Presumably this entire problem is caused by some faulty math in `0x48A8D0` or the functions it calls as changing the fourth passed param (target height - firing height) to make it so that it never goes above zero is how I achieved these results. Figuring out the root cause is beyond me, however.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -205,6 +205,7 @@ This page lists all the individual contributions to the project by their author.
   - PipScale pip size & ammo pip frame customization
   - Extension class optimization
   - Additional sync logging
+  - `Arcing` elevation inaccuracy fix
 - **Morton (MortonPL)**:
   - `XDrawOffset` for animations
   - Shield passthrough & absorption
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -206,6 +206,7 @@ This page lists all the individual contributions to the project by their author.
   - Extension class optimization
   - Additional sync logging
   - `Arcing` elevation inaccuracy fix
+  - `EMPulseCannon` projectile gravity fix
 - **Morton (MortonPL)**:
   - `XDrawOffset` for animations
   - Shield passthrough & absorption
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -132,8 +132,8 @@ This page describes all ingame logics that are fixed or improved in Phobos witho
   - They can also have map lighting apply on them if `AltPalette.ApplyLighting` is set to true.
 - Fixed `DeployToFire` not considering building placement rules for `DeploysInto` buildings and as a result not working properly with `WaterBound` buildings.
 - Fixed `DeployToFire` not recalculating firer's position on land if it cannot currently deploy.
-- `Arcing=true` projectiles no longer massively overshoot targets at higher elevation. Old behaviour can be enabled by setting `Arcing.AllowElevationInaccuracy=true` on the projectile.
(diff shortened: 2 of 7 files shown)


Metadorius:
What's the relation between this and #269?

Starkku:
> What's the relation between this and #269?

Nada, completely different approach.

docs/Fixed-or-Improved-Logics.md
```diff
@@ -132,6 +132,8 @@ This page describes all ingame logics that are fixed or improved in Phobos witho
   - They can also have map lighting apply on them if `AltPalette.ApplyLighting` is set to true.
 - Fixed `DeployToFire` not considering building placement rules for `DeploysInto` buildings and as a result not working properly with `WaterBound` buildings.
 - Fixed `DeployToFire` not recalculating firer's position on land if it cannot currently deploy.
+- `Arcing=true` projectiles no longer massively overshoot targets at higher elevation. Old behaviour can be enabled by setting `Arcing.AllowElevationInaccuracy=true` on the projectile.
+- `EMPulseCannon=yes` building weapons now respect `Floater` and Phobos-added `Gravity`setting
```
Metadorius, line 136:
```suggestion
- `EMPulseCannon=yes` building weapons now respect `Floater` and Phobos-added `Gravity` setting.
```

src/Ext/Bullet/Hooks.cpp
```diff
@@ -188,6 +188,36 @@ DEFINE_HOOK(0x6FECB2, TechnoClass_FireAt_ApplyGravity, 0x6)
...
+
+DEFINE_HOOK(0x44D2AE, BuildingClass_Mission_Missile_ApplyGravity3, 0x6)
+{
+	GET(WeaponTypeClass* const, pWeapon, EBP);
+
+	auto const nGravity = BulletTypeExt::GetAdjustedGravity(pWeapon->Projectile);
+	__asm { fld nGravity };
+
+	return 0x44D2B4;
+}
```
Metadorius, line 219: Could probably merge the hooks code somehow (maybe using a macro).

src/Ext/BulletType/Body.h
```diff
@@ -67,6 +68,7 @@ class BulletTypeExt
 			, SubjectToWater {}
 			, SubjectToWater_Detonate { true }
 			, AAOnly { false }
+			, Arcing_AllowElevationInaccuracy { false }
```
Metadorius, line 71: Do we know that people won't be pissed at us for changing the defaults again?

CREDITS.md
```diff
@@ -205,6 +205,7 @@ This page lists all the individual contributions to the project by their author.
   - PipScale pip size & ammo pip frame customization
   - Extension class optimization
   - Additional sync logging
+  - `Arcing` elevation inaccuracy fix
```
Metadorius, line 208: Forgot EMPulse probably

Metadorius (approved):
LGTM, not sure if much testing should be done, I gather can merge after addressing the stuff.

src/Ext/BulletType/Body.h
Starkku: Nope, so I set it to true by default now.

PR #1134, opened 2023-08-17
MP houses are named and allowed to house TechnoTypes owned by them on the map

ZivDero:
This PR renames the houses spaned on position 0-7 in MP/Skirmish to <Player @ X> and allows them to own TechnoTypes on the map, allowing for pre-placed buildings/units/etc.

### DIFF
--- a/Phobos.vcxproj
+++ b/Phobos.vcxproj
@@ -50,6 +50,7 @@
     <ClCompile Include="src\Ext\Trigger\Hooks.cpp" />
     <ClCompile Include="src\Ext\Unit\Hooks.Crushing.cpp" />
     <ClCompile Include="src\Locomotion\TestLocomotionClass.cpp" />
+    <ClCompile Include="src\Misc\Hooks.AssignHouses.cpp" />
     <ClCompile Include="src\Misc\Hooks.Gamespeed.cpp" />
     <ClCompile Include="src\Misc\Hooks.Ares.cpp" />
     <ClCompile Include="src\Misc\PhobosToolTip.cpp" />
--- a/src/Misc/Hooks.AssignHouses.cpp
+++ b/src/Misc/Hooks.AssignHouses.cpp
@@ -1,0 +1,29 @@
+#include <Utilities/Macro.h>
+#include <Utilities/Debug.h>
+#include <HouseClass.h>
+#include <Ext/House/Body.h>
+
+DEFINE_HOOK(0x68804A, AssignHouses_PlayerHouses, 0x5)
+{
+	Debug::Log("Assinging player houses' names\n");
+
+	GET(HouseClass*, pPlayerHouse, EBP);
+
+	HouseExt::SetSkirmishHouseName(pPlayerHouse);
+
+	return 0x68808E;
+}
+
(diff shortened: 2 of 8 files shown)


src/Misc/Hooks.AssignHouses.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+	strcpy(pHouse->PlainName, name.c_str());
+
+	Debug::Log("%s, %ls, position %d\n", pHouse->PlainName, pHouse->UIName, spawn_position);
+}
+
+void LogCountries()
+{
+	Debug::Log("--------------- Countries ---------------\n");
+	int count = std::min(HouseTypeClass::Array->Count, 20);
```
Metadorius, line 68: Not sure why is this needed? Also what's the 20 magic number?

src/Misc/Hooks.AssignHouses.cpp
```diff
@@ -0,0 +1,74 @@
...
+		else
+			name = "Computer";
+	}
+	else
+	{
+		char letters[9] = "ABCDEFGH";
+
+		name = "<Player @ ";
+		name.push_back(letters[spawn_position]);
+		name.push_back('>');
```
Metadorius, line 57: Formatted print would be better here IMO

src/Misc/Hooks.AssignHouses.cpp
```diff
@@ -0,0 +1,74 @@
...
+	if (spawn_position < 0 || spawn_position > 7)
+	{
+		if (pHouse->IsHumanPlayer)
+			name = "<human player>";
+		else
+			name = "Computer";
+	}
+	else
+	{
+		char letters[9] = "ABCDEFGH";
```
Metadorius, line 53: throw a const on it

src/Misc/Hooks.AssignHouses.cpp
```diff
@@ -0,0 +1,74 @@
...
+void LogCountries();
+
+DEFINE_HOOK(0x68804A, AssignHouses_PlayerHouses, 0x5)
+{
+	Debug::Log("Assinging player houses' names\n");
+
+	GET(HouseClass*, pPlayerHouse, EBP);
+
+	WriteHouseName(pPlayerHouse);
+	//LogCountries();
```
Metadorius, line 17: Could be useful to still log countries in debug.log somewhere?

Phobos.vcxproj
```diff
@@ -44,6 +44,7 @@
     <ClCompile Include="src\Ext\Trigger\Hooks.cpp" />
     <ClCompile Include="src\Ext\Unit\Hooks.Jumpjet.cpp" />
     <ClCompile Include="src\Misc\CaptureManager.cpp" />
+	<ClCompile Include="src\Misc\Hooks.AssignHouses.cpp" />
```
Metadorius, line 47: Indentation issue

Metadorius (changes requested):
Rebase onto develop for the workflows to run then adjust according to workflow run results and to the review, should be good otherwise.

Metadorius (approved):

PR #1144, opened 2023-08-31
Save variables to file on scenario end / hotkey

MortonPL:
- There's a new hotkey to write all local variables to `locals.ini` and all global variables to `globals.ini`. Available only if `DebugKeysEnabled` under `[GlobalControls]` is set to true in `rulesmd.ini`.
- Variables will be also automatically saved to file on scenario end if `[General]SaveVariablesOnScenarioEnd=true` is set in `rulesmd.ini`.
- Variable section will use the same name as the mission file name in capital letters, i.e. `[MYCAMPAIGN.MAP]`.
  - Variables will be written as key-value pairs, i.e. `MyVariable=1`
- If an INI file with the same name (`locals.ini`/`globals.ini`) doesn't exist, it will be created. If it exists, all sections will be preserved.

In `rulesmd.ini`:
```ini
[General]
SaveVariablesOnScenarioEnd=false    ; boolean
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -226,6 +226,7 @@ This page lists all the individual contributions to the project by their author.
   - TechnoType conversion warhead & superweapon
   - Unlimited skirmish colors
   - Show designator & inhibitor range
+  - Dump variables to file trigger action
   - Help with docs
 - **ChrisLv_CN** (work relicensed under [following permission](https://github.com/Phobos-developers/Phobos/blob/develop/images/ChrisLv-relicense.png)):
    - General assistance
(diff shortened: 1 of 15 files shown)


a851903106:
Well, that's good.

Is it possible in the future to make a trigger action that reads the variables in locals.ini/globals.ini and assigns them to the game?

MortonPL:
This logic as a whole has been remade to work automatically on scenario end + on hotkey press because of developer feedback.

Metadorius (approved):

chaserli:
idk about the intention of doing so but this breaks loading game especially from retry dialog

PR #1148, opened 2023-09-05
[User Interface] Display huge bar on screen

ststl-s:
130WC{TFZ%L_NNZRT5CXE{D

Now you can configure a huge bar on screen to display the health and shield of critical units. Huge bar will be diplayed on battle screen until this unit disappear or be destroyed.


src/Ext/Techno/Body.cpp
```diff
@@ -10,9 +10,12 @@
 
 #include <Utilities/AresFunctions.h>
 #include <Utilities/EnumFunctions.h>
+#include <Utilities/ShapeTextPrinter.h>
 
 TechnoExt::ExtContainer TechnoExt::ExtMap;
 
+std::vector<std::pair<int, TechnoClass*>> TechnoExt::Techno_HugeBar;
+
```
Starkku, line 18: Does not look like the first value from the pair is currently actually read anywhere?

src/Ext/Techno/Body.cpp
```diff
@@ -628,6 +631,360 @@ bool TechnoExt::IsTypeImmune(TechnoClass* pThis, TechnoClass* pSource)
...
+			0,
+			0,
+			ZGradient::Ground,
+			1000,
+			0,
+			nullptr,
+			0,
+			0,
+			0
+		);
```
Starkku, line 783: As good as this may be for readabilty, `Ext/Techno/Body.cpp` has enough stuff in it to not want to waste vertical space like this so might want to consider condensing this and the other call a bit.

src/Ext/Techno/Body.h
```diff
@@ -108,8 +108,29 @@ class TechnoExt
...
+				{
+					removedIterator = it;
+					break;
+				}
+			}
+
+			Techno_HugeBar.erase(removedIterator);
+		}
+	}
+
```
Starkku, line 129: Are you sure this is actually being called? Currently the container class has no `InvalidatePointer` virtual nor is it being called. They exist in `Container.h` but are commented out. That said this could also fairly safely be done in `TechnoExt` destructor.

src/Ext/Techno/Hooks.cpp
```diff
@@ -83,6 +85,22 @@ DEFINE_HOOK(0x6F42F7, TechnoClass_Init, 0x2)
 	return 0;
 }
 
+DEFINE_HOOK(0x6F6F20, TechnoClass_Unlimbo, 0x6)
+{
+	GET(TechnoClass*, pThis, ESI);
+
+	TechnoExt::InitializeHugeBar(pThis);
```
Starkku, line 92: Why is this called again in `Unlimbo()` when it is called once in `TechnoClass::Init()`?

src/Ext/Techno/Hooks.cpp
```diff
@@ -83,6 +85,22 @@ DEFINE_HOOK(0x6F42F7, TechnoClass_Init, 0x2)
...
+
+	return 0;
+}
+
+DEFINE_HOOK(0x4F4583, TechnoClass_ProcessHugeBar, 0x6)
```
Starkku, line 97: The hook name should contain the name of the class & function it is hooked at, not just the use case e.g `GScreenClass_Render_ProcessHugeBar`.

Starkku (changes requested):
Bunch of smaller things I've noticed for now, there might be more but that'd require a more comprehensive review than I can do right now.

src/Ext/Techno/Hooks.cpp
ststl-s: Some units generated by trigger actions seems do not go through `TechnoClass::Init`, such as passengers.

src/Ext/Techno/Body.h
ststl-s: This called by `PhobosTypeRegistry::InvalidatePointer` in `Phobos.Ext.cpp`.

PR #1149, opened 2023-09-05
Allow customizing aircraft landing/docking direction

Starkku:
- Aircraft docking on buildings now respect `[AudioVisual]`->`PoseDir` as the default setting and do not always land facing north or in case of pre-placed buildings, the building's direction.

----------------------

### Landing direction

- By default aircraft land facing the direction specified by `[AudioVisual]`->`PoseDir`. This can now be customized per AircraftType via `LandingDir`, defaults to `[AudioVisual]`->`PoseDir`. If the building the aircraft is docking to has aircraft docking direction set, that setting takes priority over this.
  - Negative values are allowed as a special case for `AirportBound=false` aircraft which makes them land facing their current direction.

In `rulesmd.ini`:
```ini
[SOMEAIRCRAFT]  ; AircraftType
LandingDir=         ; Direction type (integers from 0-255). Accepts negative values as a special case.
```

### Aircraft docking direction

- It is now possible to customize the landing direction for docking aircraft via `AircraftDockingDir(N)` (`N` optionally replaced by 0-based index for each `DockingOffset` separately, `AircraftDockingDir` and `AircraftDockingDir0` are synonymous and will be used if direction is not set for a specific offset) on the dock building. This overrides the aircraft's own landing direction setting and defaults to `[AudioVisual]` -> `PoseDir`.

In `rulesmd.ini`:
```ini
[SOMEBUILDING]          ; BuildingType
AircraftDockingDir(N)=  ; Direction type (integers from 0-255)
```

## Summary by CodeRabbit

- **New Features**
	- Added functionality for aircraft landing/docking direction in the project.
	- Introduced customization options for aircraft landing direction per aircraft or per dock.
	- Implemented fire particle rotation coordinate adjust toggle.
	- Added `AmbientDamage` warhead & main target ignore customization.
	- Included projectile return weapon feature.
- **Enhancements**
	- Updated `CreateUnit` declaration to specify `Facing` parameter as a "Direction type".
- **Documentation**
	- Included detailed guides on new aircraft landing direction features and logic enhancements.


MortonPL (approved):

McPwny:
it would be nice to be able to set this on a per-dock basis rather than a per-building one.

Starkku:
> it would be nice to be able to set this on a per-dock basis rather than a per-building one.

This has now been implemented.

src/Ext/Aircraft/Body.cpp
```diff
@@ -50,3 +54,50 @@ bool AircraftExt::PlaceReinforcementAircraft(AircraftClass* pThis, CellStruct ed
...
 	return result;
 }
+
+DirType AircraftExt::GetLandingDir(AircraftClass* pThis, BuildingClass* pDock)
+{
+	auto const poseDir = static_cast<DirType>(RulesClass::Instance->PoseDir);
+
+	if (!pThis)
+		return poseDir;
+
```
chaserli, line 64: LGTM except for a minor improvement/bugfix:
In vanilla game when spawned aircrafts returns to the owner their facing are not aligned with the unit, it can be fixed right over here:
```suggestion

if (auto pOwner = pAir->SpawnOwner)
	return pOwner->PrimaryFacing.Current().GetDir();

```
if (auto pOwner = pAir->SpawnOwner)
			return pOwner->PrimaryFacing.Current().GetDir();

chaserli (changes requested):

chaserli (approved):

PR #1160, opened 2023-09-14
Display "Banners" on screen via trigger actions（pr #563)

ststl-s:
I am very interested in pr#563, so I have recoded this. @MortonPL 
Source: [Display "Banners" on screen via trigger actions](https://github.com/Phobos-developers/Phobos/pull/563)
Author: [MortonPL](https://github.com/MortonPL)

- Display a 'banner' at a fixed location that is relative to the screen.
  - Action `800` will create a new banner or replace the banner with the same Banner ID if it exists. Using a local variable's value when displaying a text banner.
  - Action `801` will create a new banner or replace the banner with the same Banner ID if it exists. Using a global variable's value when displaying a text banner.
  - Action `802` will delete the banner corresponding to the set Banner ID.
- To make use of this, you need to set the properties of a `BannerType` in your ini file. The banner can either be a `PCX` file, a Shape (`SHP`) file or a `CSF` text. If multiple are set the first one in the above listed order takes effect.
  - `SHP.Palette` controls the palette that'll be used when drawing a banner for Shape file.
  - `CSF.Color` controls the color of the text that'll be used when drawing a text banner.
  - `CSF.Background` controls whether a black background will be displayed below the text banner.
  - `CSF.VariableFormat` controls the way to print a variable together with the text banner.
    - `none` will make the text banner not display the variable.
    - `variable` will make the text banner display the variable alone and will ignore the text in `CSF`.
    - `prefix`/`prefixed` will make the text banner display the variable before any other text.
    - `surfix`/`surfixed` will make the text banner display the variable after any other text.

In `rulesmd.ini`:
```ini
[BannerTypes]
0=SOMEBANNER

[SOMEBANNER]                ; BannerType
PCX=                        ; filename - excluding the .pcx extension
SHP=                        ; filename - excluding the .shp extension
SHP.Palette=palette.pal     ; filename - excluding the .pal extension
CSF=                        ; CSF entry key
CSF.Color=                  ; integer - R,G,B, defaults to MessageTextColor of the owner Side
CSF.Background=false        ; boolean
CSF.VariableFormat=none     ; List of Variable Format Enumeration (none|variable|prefix/prefixed|surfix/surfixed)
```

In `mycampaign.map`:
```ini
[Actions]
...
ID=ActionCount,[Action1],800,[Banner ID],0,[BannerType],[Horizontal position],[Vertical position],[VariableIndex],A,[ActionX]
ID=ActionCount,[Action1],801,[Banner ID],0,[BannerType],[Horizontal position],[Vertical position],[VariableIndex],A,[ActionX]
ID=ActionCount,[Action1],802,[Banner ID],0,0,0,0,0,A,[ActionX]
...
```


src/Utilities/TemplateDef.h
```diff
@@ -326,6 +326,38 @@ namespace detail
 		return false;
 	}
 
+	template<>
+	inline bool read<BSurface*>(BSurface*& value, INI_EX& parser, const char* pSection, const char* pKey)
```
CrimRecya: Looks like this part hasn't been used?

CrimRecya (approved):
LGTM.
In addition, I think it would be better to create a new entity class template (similar to the new type class template - `Enumerable<T>`), but maybe it's not necessary yet.

PR #1164, opened 2023-09-23
Trigger Action 511 for Buildings undeploy to specified waypoint

a851903106:
This is a first attempt and there may be some unreasonable errors.

In FAData.ini

```
[ParamTypes]
69=OnlyAttached,10
```

```
[ActionsRA2]
511=Buildings undeploy to specified waypoint (Phobos),-4,8,69,2,30,0,0,0,0, If the building meets the conditions to be undeployed, have the building undeploy and move to the specified waypoint, There is no problem writing the numeric serial number or ID of the building. If the building ID is <All>, all buildings can be undeployed. OnlyAttached=1 then only the attached building can be undeployed. Houes=-1 then use the triggered house. House=-2 allows all houses.,0,1,511
```


Metadorius:
Why was this and the other one closed?

PR #1176, opened 2023-11-02
Custom isometric tile palette

ststl-s:
Base on [pr#158](https://github.com/Phobos-developers/Phobos/pull/158).

- You can now use custom palette for isometric tile.

In `SomeTheater.ini`
```ini
[IsometricTileType]
CustomPalette=        ; palette
```

Test by [FlyStar](https://github.com/a851903106).
No errors on save/load.
FA2 or FA2sp can't see tile use custom palette normally now.


McPwny:
it works but it also discolors overlays, terrain objects, and smudges that sit on the tile
image

havent tested animated tiles yet but presumably they also draw from the tile pal (which is a non-issue)

thjkhdf12:
> it works but it also discolors overlays, terrain objects, and smudges that sit on the tile image
> 
> havent tested animated tiles yet but presumably they also draw from the tile pal (which is a non-issue)

Terrains and objects use lightconvert of cell, which is palette of isometric.

ststl-s:
> it works but it also discolors overlays, terrain objects, and smudges that sit on the tile image
> 
> havent tested animated tiles yet but presumably they also draw from the tile pal (which is a non-issue)

Terrains and objects use `LightConvert` of cell, based on palette of isometric.

ZivDero:
Note:
IsometricTileTypeClasses are not saved/loaded conventionally - instead, when you load a saved game, if it's the same theater as your previous game, it just clears references, and if it's a different theater - it loads it from scratch. However, this loading takes place much earlier than the type lists' - so the Read_INI function won't effectively work (because nothing exists yet).
This means that you need to place a hook after the Swizzle Manager is run to call the Read_INI function on all the extensions again.

docs/New-or-Enhanced-Logics.md
```diff
@@ -763,6 +763,18 @@ In `artmd.ini`:
 OnlyUseLandSequences=false  ; boolean
 ```
 
+## Isometric Tile
+
+### Custom palette
+
+- You can now use custom palette for isometric tile.
```
TaranDahl, line 770: You are currently modifying the cell drawer. As far as I know, it is used in many places, including animations, smudge and terrain. Considering your description:

> You can now use custom palette for ***isometric tile***.

I am not sure if this is what you intended.
At the very least, a distinction should be made between the isometric palette and the cell drawer.
I believe CustomPalette should modify the call at position 480494.

TaranDahl (changes requested):

docs/New-or-Enhanced-Logics.md
Fly-Star-him: > You are currently modifying the cell drawer. As far as I know, it is used in many places, including animations, smudge and terrain. Considering your description:
> 
> > You can now use custom palette for _**isometric tile**_.
> 
> I am not sure if this is what you intended. At the very least, a distinction should be made between the isometric palette and the cell drawer. I believe CustomPalette should modify the call at position 480494.

Please let me know your suggestions. 
How should I revise it to make it look better?

Metadorius: looping in @zivdero

TaranDahl: > Please let me know your suggestions. How should I revise it to make it look better?

I find it difficult to provide clear advice because I cannot confirm whether the issue mentioned here is intentional by design or a technical limitation.

> > it works but it also discolors overlays, terrain objects, and smudges that sit on the tile image
> > havent tested animated tiles yet but presumably they also draw from the tile pal (which is a non-issue)
> 
> Terrains and objects use `LightConvert` of cell, based on palette of isometric.

If it is a technical limitation, the current hook should be repositioned to pass the custom palette from the location I mentioned above instead of the current position.

ZivDero:
My 2 cents on this:
whatever is going on here with *tiles* owning converters is bonkers.
There should be s ingle map of "palette name -> converter". Same as with custom palettes for other stuff.
There should be no need for any tint shenanigans. At `87F698` there is `DVC<LightConvertClass*> TileDrawers` that contains all the tile drawers, at `53AD2C` all of those get adjusted when the global lighting is changed. No need for *new* converters for tints.

All the tile should do is cache a pointer to the converter that was created from the requested palette. This PR is wildly overcomplicated IMO.

Flactine:
a7860a3635d0db60f5acdf7413785084
20daf9057d2c388caaf42321ee890fa4

TerrainType seems to have been fixed, but SmudgeTypes has not fixed yet.

ZivDero:
So I took a moment to skim through this again.
I've got 2 questions:
1) What in the world is BytePalette? I cannot find its definition anywhere. Is it PaletteClass? And why does there need to be a map of raw palettes? It should be sufficient to keep just LightConvertClass'es around.
2) Why is the map still a double-map with tints? Is there a reason you cannot have it work like the rest of the game where it just retints the converters? Or am I wrong about something/missing some info?

Starkku:
>  Is it PaletteClass

Yes

EDIT: Or at least it wants to be

PR #1199, opened 2024-02-08
Linked superweapons

FS-21:
- Superweapons can now set a list of linked superweapons with `SW.Link`, which will be granted or set timers when the original superweapon is launched.
- `SW.Link.Grant` allow the linked superweapons to be added 1-time to the firer like the nuke crate if it's not presented.
- `SW.Link.Ready` specifies if superweapons timers should be set to readiness.
- `SW.Link.Reset` specifies if superweapons timers should be reset. Takes precedence over `SW.Link.Ready`.
  - For a granted superweapon, the other will be: check `SW.Link.Reset` to see if it needs to be reset. If false, check `SW.Link.Ready` to see if it needs to be set to readiness. If false, whether it'll be ready or not is decided by the granted superweapon's `SW.InitialReady`.
- `Message.LinkedSWAcquired` will be displayed to the firer when at least 1 linked superweapon is acquired or has timer set.
- `EVA.LinkedSWAcquired` will be played to the firer when at least 1 linked superweapon is acquired or has timer set.
- These superweapons can be made random with these optional tags. The game will randomly choose only a single superweapon from the list for each roll chance provided.
  - `SW.Link.RollChances` lists chances of each "dice roll" happening. Valid values range from 0% (never happens) to 100% (always happens). Defaults to a single sure roll.
  - `SW.Link.RandomWeightsN` lists the weights for each "dice roll" that increase the probability of picking a specific superweapon. Valid values are 0 (don't pick) and above (the higher value, the bigger the likelyhood). `RandomWeights` are a valid alias for `RandomWeights0`. If a roll attempt doesn't have weights specified, the last weights will be used.

In `rulesmd.ini`:
```ini
[SOMESW]                     ; SuperWeaponType
SW.Link=                     ; List of SuperWeaponTypes
SW.Link.Grant=false          ; boolean
SW.Link.Ready=               ; boolean, default to `SW.InitialReady` for granted superweapons, false otherwise
SW.Link.Reset=false          ; boolean
SW.Link.RollChances=         ; List of percentages.
SW.Link.RandomWeightsN=      ; List of integers.
Message.LinkedSWAcquired=    ; CSF entry key
EVA.LinkedSWAcquired=        ; EVA entry
```

- Tested in skirmish & in online and no desyncs were found.
- This grants a list of super weapons of only 1 use. Is like if you picked a Nuke SW from a crate so if you remember it doesn't follow the standard checks of the super weapons like aux./neg buildings etc.

## Summary by CodeRabbit

- **New Features**
	- Enhanced gameplay with the ability to grant one-time superweapons, adding strategic depth to the game.
- **Documentation**
	- Updated documentation to explain the new one-time superweapon granting system and its impact on gameplay.
- **Enhancements**
	- Improved superweapon mechanics to support one-time use superweapons, including readiness management and in-game messaging for a better player experience.


docs/New-or-Enhanced-Logics.md
```diff
@@ -678,6 +678,24 @@ Detonate.Damage=        ; integer
 Detonate.AtFirer=false  ; boolean
 ```
 
+### Grant new superweapons in superweapons
+
+- Superweapons can add 1-time superweapons to the firer like the nuke crate. Granted types can be additionally randomized using the same rules as with LimboDelivery (see above).
+- `SW.GrantOneTime.InitialReady` specifies if all new granted superweapons will be ready for launch. If not set this behaviour will be managed by `SW.InitialReady` of the granted superweapon.
+- `Message.GrantOneTimeLaunched` will be displayed to the firer when the main superweapon is launched.
+- `EVA.GrantOneTimeLaunched` will be played to the firer when the main superweapon is launched.
+
```
Coronia, line 687: `SW.GrantOneTime.RollChances` and `SW.GrantOneTime.RandomWeightsN` should be listed in the document as well

src/Ext/SWType/Body.cpp
```diff
@@ -153,6 +159,34 @@ void SWTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 	this->UseWeeds_Amount.Read(exINI, pSection, "UseWeeds.Amount");
 	this->UseWeeds_StorageTimer.Read(exINI, pSection, "UseWeeds.StorageTimer");
 	this->UseWeeds_ReadinessAnimationPercentage.Read(exINI, pSection, "UseWeeds.ReadinessAnimationPercentage");
+
```
Coronia, line 162: I would suggest to put this ini parsing section under `LimboDelivery` and `SW_Next`. They use the same RandomWeight reading pattern and tempBuffer, so it'll be easier to read and maintain if they're in the same place

src/Ext/SWType/FireSuperWeapon.cpp
```diff
@@ -282,3 +286,75 @@ void SWTypeExt::ExtData::ApplyTypeConversion(SuperClass* pSW)
...
+				}
+			}
+
+			return granted;
+		}
+
+		return false;
+	};
+
+	bool grantedAnySW = false;
```
Coronia, line 331: this variable hasn't been used anywhere, oversight or leftover?

Coronia (commented):
other than the issues in comment it looks fine

Coronia:
Honestly I'm not sure how'll SuperClass::Grant handle a superweapon that's already owned by the house. But I would suggest to add extra options for these already-existed SWs for more extendibility. An example below

In rulesmd.ini:
```ini
SW.GrantOneTime.ReadyIfExist= ; boolean, when the granted superweapon is already existed, its cooldown will be cleared if set to true
SW.GrantOneTime.ResetIfExist= ; boolean, when the granted superweapon is already existed, its cooldown will be reset if set to true
```

Technically you can even set superweapon's timer to any given value or percentage here

docs/New-or-Enhanced-Logics.md
FS-21: image
As "well" documented as in SW.Next tags ...

src/Ext/SWType/Body.cpp
FS-21: Because they were together but newly merged features gets in the way and looks as you mentioned :-D

src/Ext/SWType/FireSuperWeapon.cpp
FS-21: I believe it was a leftover.

docs/New-or-Enhanced-Logics.md
Coronia: should be explained in the above text description too, instead of only in ini section

FS-21: Updated the docs.

FS-21:
> Honestly I'm not sure how'll SuperClass::Grant handle a superweapon that's already owned by the house. But I would suggest to add extra options for these already-existed SWs for more extendibility. An example below
> 
> In rulesmd.ini:
> 
> ```ini
> SW.GrantOneTime.ReadyIfExist= ; boolean, when the granted superweapon is already existed, its cooldown will be cleared if set to true
> SW.GrantOneTime.ResetIfExist= ; boolean, when the granted superweapon is already existed, its cooldown will be reset if set to true
> ```
> 
> Technically you can even set superweapon's timer to any given value or percentage here

I don't know why I can't reply directly to the original revioew message :-S

The current behaviour does (I had to re-test it because it was done some time ago and I didn't remember how it worked):
- If is granted a SW the user doesn't have it will be fired 1 time like when you get a Nuke SW crate.
- If the used has the structure that provides a SW (for example the Soviet Nuclear Silo) and is granted the Nuke SW it does nothing because the user already has this SW, no timer reset.

Because I don't remember if changing the timer on the fly could produce desyncs I prefer not touching it in this PR and leave what you suggested as new addition in a new PR...

FS-21:
> Honestly I'm not sure how'll SuperClass::Grant handle a superweapon that's already owned by the house. But I would suggest to add extra options for these already-existed SWs for more extendibility. An example below
> 
> In rulesmd.ini:
> 
> ```ini
> SW.GrantOneTime.ReadyIfExist= ; boolean, when the granted superweapon is already existed, its cooldown will be cleared if set to true
> SW.GrantOneTime.ResetIfExist= ; boolean, when the granted superweapon is already existed, its cooldown will be reset if set to true
> ```
> 
> Technically you can even set superweapon's timer to any given value or percentage here

Added & tested in skirmish:
SW.GrantOneTime.ReadyIfExists
SW.GrantOneTime.ResetIfExists

Takes precedence this last one if requisites are met.

Coronia (approved):

CrimRecya (approved):
LGTM. And haven't encountered any problems yet.

Metadorius (changes requested):
Is there anything that is stopping this logic from appearing on a warhead instead of SW? Like, IMO if you do it on a warhead instead of an SW you could just launch a GenericWarhead/GenericWeapon SW to achieve the same effect, but you could *also* do a lot more with warhead-granted SWs.

Coronia:
one phenomenon in our testing: if the house trigger a GrantOneTime super weapon, then being defeated without using it, it'll still stay in the sidebar and can be used. This might be a thing that always existed for such one time super weapons

Coronia:
can confirm that 1-time-SW granted by any source could be used when defeated

Coronia:
gonna change this a bit to use a pattern that's more compatible with the planned transfer type for SW Timer: https://github.com/Phobos-developers/Phobos/discussions/1251#discussioncomment-12879192, the proposed new pattern would be like this:

```ini
SW.Link=                     ; List of super weapons
SW.Link.RollChances=         ; List of percentages
SW.Link.RandomWeightsN=      ; List of integers
SW.Link.Grant=false          ; boolean, grant the supeweapon if not exists
SW.Link.Ready=false          ; boolean, set the superweapon to readiness
SW.Link.Reset=false          ; boolean, reset the superweapon
; if neither of these 2 are set, it'll follow the superweapon's InitialReady
Message.LinkedSuperWeaponLaunched=        ; CSF entry key
EVA.LinkedSuperWeaponLaunched=            ; EVA entry
```

in this case, the grant and set readiness processes are decoupled, and also leave rooms for future modification on SW timers, such as:

```ini
SW.Link.ReduceTimer=
SW.Link.ReduceTimerType=flat | current | missing | total
```

Any other suggestions are welcome.

> Is there anything that is stopping this logic from appearing on a warhead instead of SW? Like, IMO if you do it on a warhead instead of an SW you could just launch a GenericWarhead/GenericWeapon SW to achieve the same effect, but you could _also_ do a lot more with warhead-granted SWs.

If anything this should be a part of transfer type as said above, which is a bit different from simply grant or set readiness of an SW, so I think the current design is still legit. You can see my intention for it in the above link

FS-21:
Coronia, you didn't credit yourself for the latest changes.

PR #1202, opened 2024-02-11
Fix map tint not affecting unit palette objects and tiberiums

MortonPL:
- Fixed units, buildings and tiberiums not being affected by map lighting color tint (can be disabled with `FixLightingTint=false` under `[General]` in `rulesmd.ini`).


MortonPL:
TODO: Fix a new bug which breaks all lamp posts until a map retint action is fired.

Sovietianqi (approved):

src/Misc/Hooks.MapTint.cpp
```diff
@@ -0,0 +1,82 @@
+#include <Utilities/Macro.h>
+#include <Helpers/Macro.h>
+#include <ScenarioClass.h>
+#include <Utilities/TemplateDef.h>
+
+namespace MapTintFix
+{
+	std::unique_ptr<LightConvertClass> TiberiumLightDrawer;
```
chaserli, line 8: `LightConvertClass` has a virtual destructor, you might end up allocating with dll new but deallocating with exe delete. Besides unique_ptr does not seem to increase safety in this case.
```suggestion
	LightConvertClass* TiberiumLightDrawer;
```

src/Misc/Hooks.MapTint.cpp
```diff
@@ -0,0 +1,82 @@
...
+		return 0;
+
+	auto tint = ScenarioClass::Instance->NormalLighting.Tint;
+	ScenarioClass::RecalcLighting(tint.Red * 10, tint.Green * 10, tint.Blue * 10, true);
+	return 0;
+}
+
+DEFINE_HOOK(0x52BE3B, InitGame_CreateTiberiumDrawer, 0x5)
+{
+	MapTintFix::TiberiumLightDrawer = std::make_unique<LightConvertClass>(LightConvertClass(
```
chaserli, line 33:
```suggestion
	MapTintFix::TiberiumLightDrawer = GameCreate<LightConvertClass>(
```

src/Misc/Hooks.MapTint.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+DEFINE_HOOK(0x5FE5F9, OverlayTypeClass_DrawIt_ReplaceTiberiumDrawer, 6)
+{
+	R->EDX(MapTintFix::TiberiumLightDrawer.get());
+	return 0x5FE5FF;
+}
+
+DEFINE_HOOK(0x6BE468, Prog_End_DeleteTiberiumDrawer, 6)
+{
+	MapTintFix::TiberiumLightDrawer.release();
```
chaserli, line 80:
```suggestion
	GameDelete(MapTintFix::TiberiumLightDrawer);
```

chaserli (changes requested):
Is there a way to incorporate the save/load fix #529?

PR #1204, opened 2024-02-12
Buildings as destroyable pathfinding obstacles

Starkku:
### Destroyable pathfinding obstacles

- It is possible to make buildings be considered pathfinding obstacles that can be destroyed by setting `IsDestroyableBlockage` to true. What this does is make the building be considered impassable and impenetrable pathfinding obstacle to every unit that is not flying or have appropriate `MovementZone` (ones that allow destroyable obstacles to be overcome, e.g `(Infantry|Amphibious)Destroyer`) akin to wall overlays and TerrainTypes.
  - Keep in mind that if an unit has appropriate `MovementZone` but no means to actually destroy an obstacle (such as a weapon that can fire and deal damage at them), they will get stuck trying to go through them instead of pathing around.

In `rulesmd.ini`:
```ini
[SOMEBUILDING]               ; BuildingType
IsDestroyableObstacle=false  ; boolean
```

In addition there is some cleanup to how technos select a weapon to use against walls.

## Summary by CodeRabbit

- **New Features**
	- Enhanced unit pathfinding to better navigate around obstacles, including the ability to consider buildings as destroyable obstacles.
	- Units with specific capabilities can now turn towards their target when firing.
	- Introduced weapon-specific interactions with obstacles, allowing some units to clear blockages using their armaments.
	- Added a power plant enhancer feature to increase power generation under certain conditions.
- **Improvements**
	- Improved the logic for units' interaction with walls and terrains, making navigation and combat more realistic.
	- Added new configuration options to enable improved pathfinding and obstacle handling.
- **Bug Fixes**
	- Refined the behavior of units and buildings in relation to obstacles and pathfinding, fixing issues with navigation and targeting.
- **Documentation**
	- Updated documentation to reflect new features and enhancements in pathfinding and obstacle handling.


mbnq:
Ok, it works without major problems with [my mod](https://github.com/mbnq/RA2YRBF) pretty well.

The only flaw I found is that units attacks terrain objects with IsPassable=yes, while not doing this with build 38.

NVIDIA_Share_iiiAOsq9wa
NVIDIA_Share_ZZlq6rn5V7

Starkku:
- Fixed overridden instructions causing infantry to treat TerrainTypes as impassable
- Fixed conflict with passable TerrainType hook potentially causing issues
- Cleaned up wall weapon selection / check code

src/Ext/Techno/Body.cpp
```diff
@@ -362,6 +364,36 @@ bool TechnoExt::IsTypeImmune(TechnoClass* pThis, TechnoClass* pSource)
...
+	auto const pTechnoType = pThis->GetTechnoType();
+	int weaponIndex = -1;
+	auto pWeapon = TechnoExt::GetCurrentWeapon(pThis, weaponIndex);
+
+	if ((pTechnoType->TurretCount > 0 && !pTechnoType->IsGattling) || !pWallOverlayType)
+		return weaponIndex;
+	else if (weaponIndex == -1)
+		return 0;
+
+	if (!pWeapon || (!pWeapon->Warhead->Wall && (!pWeapon->Warhead->Wood || pWallOverlayType->Armor != Armor::Wood)
```
ZivDero, line 379: Would be nice to pull out these complex conditions into named bools before the if

ZivDero (approved):

Starkku:
The pathfinding 'improvements' failed to actually achieve what they were supposed to and in fact were counterproductive since in many cases they resulted in pathfinding failures when there were too many impassable objects around. It is probably better to leave units that are incapable of firing against buildings to try and go to closest possible cell than fail to move entirely.

That said, this PR now consists (besides the wall weapon selection cleanup I decided to keep in) of a feature that toggles buildings to serve as destroyable obstacles akin to walls or TerrainTypes that are only considered 'passable' for units with appropriate `MovementZone`. Because this is done at zone-level, pathfinding is able to handle it much, much better. Do keep in mind however that if the unit has no weapon capable of actually destroying such obstacle but does have a `MovementZone` that allows pathing through it, it will get stuck trying to go through them.

Issue #1216, opened 2024-03-04
Irregular jumpjet facing and firing behavior when attacking with an OmniFire=no weapon


mevitar:
### Description

If i understood correctly, jumpjets with OmniFire=no on their weapons should now turn to face their target properly. However, when turning, not only they glitch out visually, but they will also start firing their weapon when not facing the target yet. And to make it more inconsistent, sometimes the same jumpjet will _**not fire until it turns**_, only to again start firing before it turns after next attack order.

The visual glitch happens only to voxel jumpjets, didn't notice it with shp jumpjets. They still fire at a target before they start facing it, though.

### Phobos Version

https://github.com/Phobos-developers/Phobos/commit/dd201d39c9dd894505451820d02def6ec8669adb

### Steps to reproduce

1. use a jumpjet without a turret and with OmniFire=no on its weapons
2. order the jumpjet to attack behind it

### Expected behaviour

A jumpjet should turn properly without glitching and start firing once facing the target

### Actual behaviour

![jumpjetfirebug](https://github.com/Phobos-developers/Phobos/assets/1313344/6efc9712-769d-47c2-acf7-43ed0b93760a)

### Additional context

OmniFire=yes and OmniFire.TurnToTarget=yes on the weapon makes it so the jumpjet doesn't glitch anymore, but it will also start firing immediately, before facing the target (i think that's intended in this case, though).

### Checklist

- [X] The issue happens on the **latest official** version of Phobos and wasn't fixed yet.
- [X] I agree to elaborate the details if requested and provide thorough testing if the bugfix is implemented.
- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search and read the issue list to find a similar issue and didn't find it.
- [X] I have attached as much information as possible *(screenshots, gifs, videos, debug and exception logs, etc)*.

PR #1217, opened 2024-03-05
Fix issue #1216

chaserli:
@mevitar could you plz make some regression tests on this? Apart from testing if it fixes your issue, but also tests if it doesn't break anything

Fix issue #1216


mevitar:
> @mevitar could you plz make some regression tests on this? Apart from testing if it fixes your issue, but also tests if it doesn't break anything

Hard to tell without knowing exactly what am i supposed to look for.
What is the area where that new hook is placed in? Just jumpjet locomotor related, or something more specific? Would help with testing if i knew.
The issue itself appears to be fixed, though.

Issue #1107, opened 2023-07-18
When I press "esc" key and get a game crush on game.


PacrateOfficial:
### Description

When I press "esc" key and get a game crush on game.
This is a "Exception code: C0000005 at 61288900".

### Phobos Version

B36

### Conditions to reproduce

Mental Omega 3, I want to change the INI file, but I don't know which of its codes have this phenomenon.
This is a guess only.

### INI code

```ini
No response
```

### Steps to reproduce

1.Launch a campaign or skirmish.
2.Press "esc" key or Click menu.
3.Immediately get a game crush.
4.I don't know why this phenomenon appears.

### Expected behaviour

It should not appear the phenomenon.

### Actual behaviour

When I press "esc" key or Click menu on Screen of Northeast and get a game crush on game.

### Additional context

[debug.log](https://github.com/Phobos-developers/Phobos/files/12081156/debug.log)
[except.txt](https://github.com/Phobos-developers/Phobos/files/12081157/except.txt)
[syringe.log](https://github.com/Phobos-developers/Phobos/files/12081158/syringe.log)

https://github.com/Phobos-developers/Phobos/assets/139878231/8a894c10-3e7f-4fcc-8d32-80e4ffc8e87c

### Checklist

- [X] The issue happens on the **latest official** version of Phobos and wasn't fixed yet.
- [X] I agree to elaborate the details if requested and provide thorough testing if the bugfix is implemented.
- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search and read the issue list to find a similar issue and didn't find it.
- [X] I have attached as much information as possible *(screenshots, gifs, videos, debug and exception logs, etc)*.

Metadorius:
Please, upload the `extcrashdump.dmp` file (you may archive it and split into multiple files around 20mb size to upload here) from the snapshot folder that got created when you pressed "yes" button on the crash message. It will help us know what caused the crash.

PacrateOfficial:
[extcrashdump.part01.zip](https://github.com/Phobos-developers/Phobos/files/12081458/extcrashdump.part01.zip)
[extcrashdump.part02.zip](https://github.com/Phobos-developers/Phobos/files/12081460/extcrashdump.part02.zip)
[extcrashdump.part03.zip](https://github.com/Phobos-developers/Phobos/files/12081462/extcrashdump.part03.zip)
[...]

PR #1220, opened 2024-03-10
Fixes issue #1107

HyperionFrog:
Minor fix. The issue likely results from a null pointer exception.

## Summary by CodeRabbit

- **Bug Fixes**
	- Fixed issues with saving game mode options.
	- Removed unintended light tint from buildings.
	- Prevented crashes when returning to the menu after loading a game.
	- Resolved problems with drawing map previews.
	- Corrected map lighting corruption.
	- Addressed control transfer issues with mind-controlled vehicles.


chaserli:
Could you plz check if swizzling that ptr the dumb way also works

PR #1223, opened 2024-03-11
Fix a crash with SkirmishUnlimitedColors and multiplayer

### LINKED ISSUE #1222: SkirmishUnlimitedColors crashes when attempting to play as observer
### Description

Setting `SkirmishUnlimitedColors=true` and attempting to play as observer will crash the game.

The feature changes the function at `0x69A310` to not convert player color index to scheme index but fails to handle the special case of player color index -2 which means it returns scheme index 5 from the lookup table, corresponding to the full-shaded variant of colorscheme named `LightGrey` with vanilla rules. This is used for observer's loading screen text, which is why it crashes here.

Ares adds a bespoke observer color scheme setting (defaulting to `LightGrey`) that is returned in this case, but since this feature skips Ares color stuff entirely it probably shouldn't concern itself with that. At the very least it should return the color scheme index of `LightGrey` if called with -2 as param, as it is fetched with name not hardcoded index position elsewhere concerning observers. Alternatively patch `0x642BDA` to do it directly instead.

### Phobos Version

b38

### Conditions to reproduce

Playing on CnCNet client, as it is more or less required to fully support this feature in the first place.

### INI code

```ini
[General]
SkirmishUnlimitedColors=true
```

### Steps to reproduce

1. Set `[General]` -> `SkirmishUnlimitedColors` to true in `rulesmd.ini`.
2. Attempt to start a game as observer.

### Expected behaviour

Doesn't crash and shows loading screen text in an appropriate color scheme.

### Actual behaviour

Game crashes cause it tries to render text using invalid color scheme.

### Additional context

_No response_

### Checklist

MortonPL:
Closes #1222.

### DIFF
--- a/Phobos.vcxproj
+++ b/Phobos.vcxproj
@@ -153,6 +153,7 @@
     <ClCompile Include="src\Misc\Hooks.LaserDraw.cpp" />
     <ClCompile Include="src\Misc\Hooks.Timers.cpp" />
     <ClCompile Include="src\Misc\Hooks.INIInheritance.cpp" />
+    <ClCompile Include="src\Misc\Hooks.SkirmishColors.cpp" />
     <ClCompile Include="src\New\Type\Affiliated\TypeConvertGroup.cpp" />
     <ClCompile Include="src\Ext\BuildingType\Hooks.Upgrade.cpp" />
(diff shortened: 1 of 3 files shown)


Starkku (approved):
Looks good and fixes the crashing.

PR #1229, opened 2024-03-24
Universal deploy from any techno into any techno in Warheads and Superweapons

FS-21:
When the Ares Convert logic is used in Warheads or Superweapons we can use a new tag `Convert.UseUniversalDeploy=true` for allowing the conversion between different technos or structure into structure.

Is a smaller version of the PR 1211 that I verified that doesn't desync online.


FS-21:
Because there are other attempts of techno2techno I'll close this and use it internally until any of those attempts is merged.

PR #1241, opened 2024-04-06
New & restored TerrainType features

Starkku:
## Animated TerrainTypes

- Length of the animation can now be customized by setting `AnimationLength` as well, defaulting to half (or quarter if damaged frames are enabled) the number of frames in TerrainType's image.

In `rulesmd.ini`:
```ini
[SOMETERRAINTYPE]  ; TerrainType
AnimationLength=   ; integer, number of frames
```

## Damaged frames and crumbling animation

- By default game shows damage frame only for TerrainTypes alive at only 1 point of health left. Because none of the original game TerrainType assets were made with this in mind, the logic is now locked behind a new key `HasDamagedFrames`.
  - Instead of showing at 1 point of HP left, TerrainTypes switch to damaged frames once their health reaches `[AudioVisual]` -> `ConditionYellow.Terrain` percentage of their maximum health. Defaults to `ConditionYellow` if not set.
- In addition, TerrainTypes can now show 'crumbling' animation after their health has reached zero and before they are deleted from the map by setting `HasCrumblingFrames` to true.
  - Crumbling frames start from first frame after both regular & damaged frames and ends at halfway point of the frames in TerrainType's image.
    - Note that the number of regular & damage frames considered for this depends on value of `HasDamagedFrames` and for `IsAnimated` TerrainTypes, `AnimationLength`. Exercise caution and ensure there are correct amount of frames to display.
  - Sound event from `CrumblingSound` (if set) is played when crumbling animation starts playing.
  - Destroy animation & sound only play after crumbling animation has finished.

In `rulesmd.ini`:
```ini
[AudioVisual]
ConditionYellow.Terrain=  ; floating-point value

[SOMETERRAINTYPE]         ; TerrainType
HasDamagedFrames=false    ; boolean
HasCrumblingFrames=false
CrumblingSound=           ; Sound
```

## Custom palette

- You can now specify custom palette for TerrainTypes in similar manner as TechnoTypes can.
  - Note that this palette behaves like an object palette and does not use tint etc. that have been applied to the tile the TerrainType resides on like a TerrainType using tile palette would.

In `artmd.ini`:
```ini
[SOMETERRAINTYPE]  ; TerrainType
Palette=           ; filename - excluding .pal extension and three-character theater-specific suffix
```

### DIFF
--- a/src/Ext/TerrainType/Body.cpp
+++ b/src/Ext/TerrainType/Body.cpp
@@ -1,5 +1,6 @@
 #include "Body.h"
 
+#include <AnimClass.h>
 #include <TacticalClass.h>
 #include <TerrainClass.h>
 #include <TerrainTypeClass.h>
@@ -16,6 +17,14 @@ int TerrainTypeExt::ExtData::GetTiberiumGrowthStage()
 int TerrainTypeExt::ExtData::GetCellsPerAnim()
 {
 	return GeneralUtils::GetRangedRandomOrSingleValue(this->SpawnsTiberium_CellsPerAnim.Get());
+}
+
+void TerrainTypeExt::ExtData::PlayDestroyEffects(CoordStruct coords)
+{
+	VocClass::PlayIndexAtPos(this->DestroySound.Get(-1), coords);
+
+	if (auto const pAnimType = this->DestroyAnim)
+		GameCreate<AnimClass>(pAnimType, coords);
 }
 
 void TerrainTypeExt::Remove(TerrainClass* pTerrain)
@@ -46,6 +55,10 @@ void TerrainTypeExt::ExtData::Serialize(T& Stm)
 		.Process(this->MinimapColor)
 		.Process(this->IsPassable)
 		.Process(this->CanBeBuiltOn)
+		.Process(this->HasDamagedFrames)
+		.Process(this->HasCrumblingFrames)
+		.Process(this->CrumblingSound)
+		.Process(this->AnimationLength)
 		;
 }
 
@@ -70,6 +83,11 @@ void TerrainTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 
 	this->IsPassable.Read(exINI, pSection, "IsPassable");
 	this->CanBeBuiltOn.Read(exINI, pSection, "CanBeBuiltOn");
+
+	this->HasDamagedFrames.Read(exINI, pSection, "HasDamagedFrames");
+	this->HasCrumblingFrames.Read(exINI, pSection, "HasCrumblingFrames");
+	this->CrumblingSound.Read(exINI, pSection, "CrumblingSound");
+	this->AnimationLength.Read(exINI, pSection, "AnimationLength");
 
 	//Strength is already part of ObjecTypeClass::ReadIni Duh!
 	//this->TerrainStrength.Read(exINI, pSection, "Strength");
--- a/src/Ext/TerrainType/Body.cpp
+++ b/src/Ext/TerrainType/Body.cpp
@@ -19,7 +19,7 @@ int TerrainTypeExt::ExtData::GetCellsPerAnim()
 	return GeneralUtils::GetRangedRandomOrSingleValue(this->SpawnsTiberium_CellsPerAnim.Get());
 }
 
-void TerrainTypeExt::ExtData::PlayDestroyEffects(CoordStruct coords)
+void TerrainTypeExt::ExtData::PlayDestroyEffects(const CoordStruct& coords)
 {
 	VocClass::PlayIndexAtPos(this->DestroySound, coords);
 
@@ -98,6 +98,9 @@ void TerrainTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 
 	this->PaletteFile.Read(pArtINI, pArtSection, "Palette");
 	this->Palette = GeneralUtils::BuildPalette(this->PaletteFile);
+
+	if (GeneralUtils::IsValidString(this->PaletteFile) && !this->Palette)
+		Debug::Log("[Developer warning] [%s] has Palette=%s set but no palette file was loaded (missing file or wrong filename). Missing palettes cause issues with lighting recalculations.\n", pArtSection, this->PaletteFile);
 }
 
 void TerrainTypeExt::ExtData::LoadFromStream(PhobosStreamReader& Stm)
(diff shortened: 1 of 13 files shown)


src/Ext/TerrainType/Hooks.cpp
```diff
@@ -122,22 +210,74 @@ DEFINE_HOOK(0x48381D, CellClass_SpreadTiberium_CellSpread, 0x6)
...
+{
+	GET(TerrainClass*, pThis, ESI);
+
+	auto const pTypeExt = TerrainTypeExt::ExtMap.Find(pThis->Type);
+	double condYellow = RulesExt::Global()->ConditionYellow_Terrain.Get(RulesClass::Instance->ConditionYellow);
+
+	if (!pThis->Type->IsAnimated && pTypeExt->HasDamagedFrames && TerrainTypeTemp::PriorHealthRatio > condYellow && pThis->GetHealthPercentage() <= condYellow)
+	{
+		pThis->IsCrumbling = true; // Dirty hack to get game to redraw the art reliably.
+		MapClass::Instance->Logics.get().AddObject(pThis, false);
```
chaserli, line 252:
```suggestion
		MapClass::Instance->Logics->AddObject(pThis, false);
```

src/Ext/TerrainType/Body.cpp
```diff
@@ -18,6 +19,14 @@ int TerrainTypeExt::ExtData::GetCellsPerAnim()
 	return GeneralUtils::GetRangedRandomOrSingleValue(this->SpawnsTiberium_CellsPerAnim.Get());
 }
 
+void TerrainTypeExt::ExtData::PlayDestroyEffects(CoordStruct coords)
+{
+	VocClass::PlayIndexAtPos(this->DestroySound.Get(-1), coords);
+
+	if (auto const pAnimType = this->DestroyAnim.Get(nullptr))
+		GameCreate<AnimClass>(pAnimType, coords);
```
chaserli, line 27: I see you are a fan of `Nullable` everything that defaults to null, which is unnecessary. Parsing failure for Valueable leads to nullptr for pointer types and -1 to idx types by default.

src/Ext/TerrainType/Hooks.cpp
```diff
@@ -51,6 +55,90 @@ DEFINE_HOOK(0x71C84D, TerrainClass_AI_Animated, 0x6)
...
+	enum { ReturnFromFunction = 0x71C839, SkipCheck = 0x71C7C2 };
+
+	GET(TerrainClass*, pThis, ESI);
+
+	auto const pTypeExt = TerrainTypeExt::ExtMap.Find(pThis->Type);
+
+	if (pTypeExt->HasDamagedFrames && pThis->Health > 0)
+	{
+		if (!pThis->Type->IsAnimated && !pThis->Type->IsFlammable)
+			MapClass::Instance->Logics.get().Remove(pThis);
```
chaserli, line 69:
```suggestion
			MapClass::Instance->Logics->Remove(pThis);
```

src/Ext/TerrainType/Hooks.cpp
```diff
@@ -122,22 +210,74 @@ DEFINE_HOOK(0x48381D, CellClass_SpreadTiberium_CellSpread, 0x6)
...
-	if (auto const pTerrainExt = TerrainTypeExt::ExtMap.Find(pThis->Type))
+	auto const pTypeExt = TerrainTypeExt::ExtMap.Find(pThis->Type);
+
+	// Skip over the removal of the tree as well as destroy sound/anim (for now) if the tree has crumble animation.
+	if (pThis->IsCrumbling && pTypeExt->HasCrumblingFrames)
 	{
-		auto const nCoords = pThis->GetCoords();
-		VocClass::PlayIndexAtPos(pTerrainExt->DestroySound.Get(-1), nCoords);
+		// Needs to be added to the logic layer for the anim to work.
+		MapClass::Instance->Logics.get().AddObject(pThis, false);
```
chaserli, line 271:
```suggestion
		MapClass::Instance->Logics->AddObject(pThis, false);
```

chaserli (changes requested):
haven't looke into the impl yet, only a few codestyle issues

src/Ext/TerrainType/Body.h
```diff
@@ -50,6 +63,7 @@ class TerrainTypeExt
 
 		int GetTiberiumGrowthStage();
 		int GetCellsPerAnim();
+		void PlayDestroyEffects(CoordStruct coords);
```
chaserli, line 66:
```suggestion
		void PlayDestroyEffects(const CoordStruct& coords);
```

src/Ext/TerrainType/Hooks.cpp
```diff
@@ -51,6 +55,90 @@ DEFINE_HOOK(0x71C84D, TerrainClass_AI_Animated, 0x6)
...
+	enum { ReturnFromFunction = 0x71C839, SkipCheck = 0x71C7C2 };
+
+	GET(TerrainClass*, pThis, ESI);
+
+	auto const pTypeExt = TerrainTypeExt::ExtMap.Find(pThis->Type);
+
+	if (pTypeExt->HasDamagedFrames && pThis->Health > 0)
+	{
+		if (!pThis->Type->IsAnimated && !pThis->Type->IsFlammable)
+			MapClass::Instance->Logics->Remove(pThis);
```
chaserli, line 69: `Logics` is static so the `Instance->` is not needed. Best way is to remove that from MapClass and move to LogicClass in yrpp, and then name it `LogicClass::Instance`

src/Ext/TerrainType/Hooks.cpp
```diff
@@ -122,22 +210,74 @@ DEFINE_HOOK(0x48381D, CellClass_SpreadTiberium_CellSpread, 0x6)
...
+
+	GET(TerrainClass*, pThis, ESI);
+
+	auto const pTypeExt = TerrainTypeExt::ExtMap.Find(pThis->Type);
+
+	if (!pThis->IsCrumbling && pTypeExt->HasCrumblingFrames)
+	{
+		// Needs to be added to the logic layer for the anim to work.
+		MapClass::Instance->Logics->AddObject(pThis, false);
+		VocClass::PlayIndexAtPos(pTypeExt->CrumblingSound.Get(-1), pThis->GetCoords());
```
chaserli, line 225: Given that the default value for ValueableIdx is -1, NullableIdx is not necessary since you only used .Get(-1). Same for other occurrences in existing code such as DetachedReport in AnimTypeExt.

chaserli (changes requested):
will look into details later

Starkku:
Fixed the issues pointed out in reviews so far.

PR #1246, opened 2024-04-25
Trajectory projectiles enhancement

Coronia:
Some enhancement for trajectory projectiles, especially for Bombard Trajectory. Many thanks to **NaotoYuuki** who provided trajectory scatter logic and the prototypes of vertical and meteor projectiles, which are used as the base of these additions.

1 - Both Straight Trajectory and Bombard Trajectory now support `Inaccurate=true`, `BallisticScatter.Max` and `BallisticScatter.Min` keys, which allows them to scatter when picking targets.

2 - Bombard Trajectory, instead of simply flying to the top of the target and free fall, is now extended into a series of trajectories which flying to a 'turning point' first, and then turning to the targets or respawning on their top. The behaviors of the projectile and the position of the turning point could be customized by the following tags:
  - `Trajectory.Bombard.FallPercent` controls the distance of the turning point by its percentage of the total distance between attacker and intended target. If set to 0%, then it'll fly up vertically. If set to 100%, then it'll travel to the top of the intended target.
    - For each launch the turning point percentage could add or minus a random value, which is not greater than `Trajectory.Bombard.FallPercentShift`. If set to 0%, random shift will be disabled.
    - You can also makes the turning point scatter in a circle with `Trajectory.Bombard.FallScatterRange` as its radius. If set to 0, random scatter will be disabled.
  - `Trajectory.Bombard.FreeFallOnTarget` controls how it'll hit the intended target. If set to true, the projectile will be respawned above the intended target and free fall. If set to false, the projectile will travel to the intended target from the turning point.
  - `Trajectory.Bombard.NoLaunch` controls whether the attacker will fire the projectile by itself. If set to true, projectile will directly fall from the turning point.
  - `Trajectory.Bombard.FallSpeed` controls the initial speed of the projectile after it turns. If set to 0.0, then it'll use `Trajectory.Speed`. Can't work together with `Trajectory.Bombard.FreeFallOnTarget=true`.
  - `Trajectory.Bombard.TargetSnapDistance` controls the maximum distance in cells from intended target the projectile can be at moment of detonation to make the projectile 'snap' on the intended target. Set to 0 to disable snapping.

In `rulesmd.ini`:
```ini
[SOMEPROJECTILE]                          ; Projectile
Trajectory=Bombard                        ; Trajectory type
Trajectory.Bombard.Height=0.0             ; double
Trajectory.Bombard.FallPercent=1.0        ; double
Trajectory.Bombard.FallPercentShift=0.0   ; double
Trajectory.Bombard.FallScatterRange=0.0   ; floating point value
Trajectory.Bombard.FreeFallOnTarget=true  ; boolean
Trajectory.Bombard.NoLaunch=false         ; boolean
Trajectory.Bombard.FallSpeed=0.0          ; double
Trajectory.Bombard.TargetSnapDistance=0.5 ; floating point value
```
If all values are set to default, it'll perform as the old Bombard Trajectory, so those who have made use of it don't need to modify their ini if they want it to behave the same.

NOTE: Both the old and new Bombard Trajectory will crash when firing a `Voxel=true` projectile, which is unresolved yet. Avoid using it for now, and if someone could make it work I'll be really grateful.

### DIFF
--- a/src/Ext/Bullet/Trajectories/BombardTrajectory.cpp
+++ b/src/Ext/Bullet/Trajectories/BombardTrajectory.cpp
@@ -1,6 +1,6 @@
 #include "BombardTrajectory.h"
-
-#include <Ext/BulletType/Body.h>
+#include <AnimClass.h>
+#include <Ext/Bullet/Body.h>
 
 PhobosTrajectory* BombardTrajectoryType::CreateInstance() const
 {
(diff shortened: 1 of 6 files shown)


src/Ext/Bullet/Trajectories/BombardTrajectory.cpp
```diff
@@ -72,14 +195,59 @@ void BombardTrajectory::OnAIVelocity(BulletClass* pBullet, BulletVelocity* pSpee
...
+				pPosition->X = pBullet->Location.X + pBullet->Velocity.X;
+				pPosition->Y = pBullet->Location.Y + pBullet->Velocity.Y;
+				pPosition->Z = pBullet->Location.Z + pBullet->Velocity.Z;
+			}
+			else
+			{
+				pSpeed->X = 0.0;
+				pSpeed->Y = 0.0;
+				pSpeed->Z = 0.0;
+				if (this->FallPercent != 1.0) // change position and recreate laser trail
```
FS-21, line 213: Not a mistake, but for better readibility put a empty line before this "if"

FS-21 (commented):
The code looks fine

PR #1247, opened 2024-04-25
Part of a small fix on GlobalVariables

a851903106:
Previously while using phobos I noticed that the game was passing GlobalVariables with an error. due to a technical glitch I couldn't get to the root of the problem so I opted for an alternative method of exporting globals.ini and re-reading the parameters.

PS: There may be a better way, but that's not something I can do anymore.

### DIFF
--- a/src/Ext/Scenario/Body.cpp
+++ b/src/Ext/Scenario/Body.cpp
@@ -57,6 +57,17 @@ void ScenarioExt::ExtData::ReadVariables(bool bIsGlobal, CCINIClass* pINI)
 				var.Value = 0;
 		}
 	}
+
+	if (bIsGlobal)
+	{
+		ScenarioExt::Global()->LoadVariablesToFile(true);
+
+		if (!Phobos::Config::SaveVariablesOnScenarioEnd)
+		{
+			// Is it better not to delete the file?
+			DeleteFileA("globals.ini");
+		}
+	}
 }
 
 void ScenarioExt::ExtData::SaveVariablesToFile(bool isGlobal)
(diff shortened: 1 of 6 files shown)


src/Ext/Scenario/Body.cpp
```diff
@@ -72,12 +77,49 @@ void ScenarioExt::ExtData::SaveVariablesToFile(bool isGlobal)
...
+		}
+	}
+
+	pFile->Close();
+
+	if (!Phobos::Config::SaveVariablesOnScenarioEnd)
+	{
+		// Is it better not to delete the file?
+		DeleteFileA(fileName);
+	}
```
FS-21, line 120:
```suggestion
	if (!Phobos::Config::SaveVariablesOnScenarioEnd)
		DeleteFileA(fileName); // Is it better not to delete the file?
```

src/Ext/Scenario/Hooks.Variables.cpp
```diff
@@ -83,7 +83,66 @@ DEFINE_HOOK(0x685EB1, PhobosSaveVariables, 0x5)//Lose
...
 
 	return 0;
 }
 
+// I'm not sure how it works, this seems to solve the problem for now.
+DEFINE_HOOK(0x4C6217, ScenarioClass_LoadVariables, 0x5)
+{
+	ScenarioExt::Global()->LoadVariablesToFile(true);
+
+	return 0x4C622F;
```
FS-21, line 99:
```suggestion
	ScenarioExt::Global()->LoadVariablesToFile(true);
	return 0x4C622F;
```

src/Ext/Scenario/Hooks.Variables.cpp
```diff
@@ -83,7 +83,66 @@ DEFINE_HOOK(0x685EB1, PhobosSaveVariables, 0x5)//Lose
...
+
+					return AltNextScenario;
+				}
+			}
+		}
+
+		if (strcmp(ScenarioClass::Instance->NextScenario, ""))
+		{
+			return NextScenario;
+		}
```
FS-21, line 143:
```suggestion
		if (strcmp(ScenarioClass::Instance->NextScenario, ""))
			return NextScenario;
```

FS-21 (commented):
Well written, only tweaks and if there are no new reviewers I suggest label it with the "Needs testing"

PR #1248, opened 2024-04-25
Fix potential desync caused by cursor on DeploysInto vehicles

Starkku:
Game removes deploying vehicles from map temporarily to check if there's enough space to deploy into a building when displaying allow/disallow deploy cursor. This can cause desyncs if there are certain types of units around the deploying unit.

Only reasonable way to solve this is to perform the cell clear check on every client per frame and use that result in cursor display which is client-specific. This is now implemented in multiplayer games only.

### DIFF
--- a/src/Ext/Techno/Body.cpp
+++ b/src/Ext/Techno/Body.cpp
@@ -384,6 +384,7 @@ void TechnoExt::ExtData::Serialize(T& Stm)
 		.Process(this->DeployFireTimer)
 		.Process(this->ForceFullRearmDelay)
 		.Process(this->WHAnimRemainingCreationInterval)
+		.Process(this->CanCurrentlyDeployIntoBuilding)
 		;
 }
 
--- a/src/Ext/Techno/Hooks.DeploysInto.cpp
+++ b/src/Ext/Techno/Hooks.DeploysInto.cpp
@@ -116,3 +116,42 @@ DEFINE_HOOK(0x7396AD, UnitClass_Deploy_CreateBuilding, 0x6)
 
 	return 0x7396B3;
 }
+
+// Game removes deploying vehicles from map temporarily to check if there's enough
+// space to deploy into a building when displaying allow/disallow deploy cursor.
+// This can cause desyncs if there are certain types of units around the deploying unit.
+// Only reasonable way to solve this is to perform the cell clear check on every client per frame
+// and use that result in cursor display which is client-specific. This is now implemented in multiplayer games only.
+#pragma region DeploysIntoDesyncFix
+
+DEFINE_HOOK(0x73635B, UnitClass_AI_DeploysIntoDesyncFix, 0x6)
+{
+	if (!SessionClass::Instance->IsMultiplayer())
+		return 0;
+
+	GET(UnitClass*, pThis, ESI);
+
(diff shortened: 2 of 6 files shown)


Metadorius (approved):

mbnq:
Tested with [BF 0.8.3507](https://github.com/mbnq/RA2YRBF/releases/download/v0.8.35/RA2YRBF_0_8_3507.zip) [DebugMap](https://github.com/Phobos-developers/Phobos/files/14944228/tutorial_debug.zip)

https://youtu.be/Tk-etJFy-CA  
(processing to 1440p in progress)

Short summary:
With Phobos build 38 it's very easy to reproduce desync spaming with deploy and clicking. With this Fix it is no longer possible to reproduce. Details [on the video](https://youtu.be/Tk-etJFy-CA).

src/Ext/Techno/Body.cpp
```diff
@@ -384,6 +384,7 @@ void TechnoExt::ExtData::Serialize(T& Stm)
 		.Process(this->DeployFireTimer)
 		.Process(this->ForceFullRearmDelay)
 		.Process(this->WHAnimRemainingCreationInterval)
+		.Process(this->CanCurrentlyDeployIntoBuilding)
```
chaserli, line 387: I don't get it, is this desync introduced by phobos or already present in vanilla? and if this dirty fix is reserved for multiplayer only, do you still need to serialize it?

chaserli (changes requested):

src/Ext/Techno/Body.cpp
Starkku: It is technically a vanilla issue, been there since RA1 in fact. I am uncertain of if mod-added content and/or extension changes from Ares or Phobos expose it or make it more common, though.

'Dirty' fix is tiny bit disingenuous here, given that the original approach of removing the unit from the map in code that is only called in local client is hazardous as hell to begin with. I did experiment with other approaches first, skipping occupier pointer checks in `CellClass::ClearToBuild()` for cells containing the unit that called it from cursor code but all of that still desynced. Making it so that the code runs on all clients seems to be the only safe option here.

As for serializing, there is no real need to serialize it I guess. It is recalculated on every frame so even if someone implemented MP saves it wouldn't really cause any issues.

src/Ext/Techno/Hooks.DeploysInto.cpp
```diff
@@ -116,3 +116,42 @@ DEFINE_HOOK(0x7396AD, UnitClass_Deploy_CreateBuilding, 0x6)
...
+
+	if (pThis->Type->DeploysInto)
+		TechnoExt::ExtMap.Find(pThis)->CanCurrentlyDeployIntoBuilding = TechnoExt::CanDeployIntoBuilding(pThis);
+
+	return 0;
+}
+
+DEFINE_HOOK(0x73FEC1, UnitClass_WhatAction_DeploysIntoDesyncFix, 0x6)
+{
+	if (!SessionClass::Instance->IsMultiplayer())
```
chaserli, line 142:
```suggestion
	if (!SessionClass::IsMultiplayer())
```

chaserli (approved):

PR #1252, opened 2024-05-03
Scripts code clean up & tweaks

FS-21:
Minor changes:
- Moved some Techno functions from ScriptExt to TechnoExt
- Tweaked a bit the existing script actions.
- Moved ApplyMindControlRangeLimit call into OnEarlyUpdate().

No Credits/What's new required since these changes aren't features.

### DIFF
--- a/src/Ext/Techno/Body.Update.cpp
+++ b/src/Ext/Techno/Body.Update.cpp
@@ -40,6 +40,7 @@ void TechnoExt::ExtData::OnEarlyUpdate()
 	this->ApplySpawnLimitRange();
 	this->UpdateLaserTrails();
 	this->DepletedAmmoActions();
+	ApplyMindControlRangeLimit(this->OwnerObject());
 }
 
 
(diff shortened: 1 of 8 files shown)


Starkku:
I'll do a more thorough review sometime in near future but is it really necessary to constantly check for null pointers everywhere?

src/Ext/Techno/Body.Update.cpp
```diff
@@ -40,6 +40,7 @@ void TechnoExt::ExtData::OnEarlyUpdate()
 	this->ApplySpawnLimitRange();
 	this->UpdateLaserTrails();
 	this->DepletedAmmoActions();
+	ApplyMindControlRangeLimit(this->OwnerObject());
```
chaserli, line 43: no!

Metadorius: why?

chaserli: unrelated & unnecessary

PR #1255, opened 2024-05-08
[Minor] Attempt to weaken radside desync

chaserli:

### DIFF
--- a/src/Ext/RadSite/Hooks.cpp
+++ b/src/Ext/RadSite/Hooks.cpp
@@ -62,24 +62,23 @@ DEFINE_HOOK(0x5213B4, InfantryClass_AIDeployment_CheckRad, 0x7)
 	auto const pWeapon = pInf->GetDeployWeapon()->WeaponType;
 	int radLevel = 0;
 
-	if (RadSiteExt::ExtMap.size() > 0 && pWeapon)
+	if (RadSiteClass::Array->Count > 0 && pWeapon)
 	{
 		auto const pWeaponExt = WeaponTypeExt::ExtMap.Find(pWeapon);
 		auto const pRadType = pWeaponExt->RadType;
 		auto const warhead = pWeapon->Warhead;
 		auto currentCoord = pInf->GetCell()->MapCoords;
 
-		for (auto const iRad : *RadSiteClass::Array)
+		for (auto const pRadSite : *RadSiteClass::Array)
 		{
-			if (iRad->BaseCell == currentCoord &&
-				iRad->Spread == (int)warhead->CellSpread &&
-				RadSiteExt::ExtMap.Find(iRad)->Type == pRadType
+			if (pRadSite->BaseCell == currentCoord &&
+				pRadSite->Spread == (int)warhead->CellSpread &&
+				RadSiteExt::ExtMap.Find(pRadSite)->Type == pRadType
 				)
 			{
-				radLevel = iRad->GetRadLevel();
+				radLevel = pRadSite->GetRadLevel();
 				break;
 			}
-
 		}
 
 	}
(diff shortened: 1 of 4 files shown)


src/Ext/RadSite/Hooks.cpp
```diff
@@ -69,22 +69,19 @@ DEFINE_HOOK(0x5213B4, InfantryClass_AIDeployment_CheckRad, 0x7)
...
-			{
-				return
-				pair.second->Type == pRadType &&
-					pair.first->BaseCell == currentCoord &&
-					pair.first->Spread == Game::F2I(warhead->CellSpread)
-					;
-			});
-
-		if (it != RadSiteExt::ExtMap.end())
+		for (auto const iRad : *RadSiteClass::Array)
```
Starkku, line 72: I guess for consistency's sake it should be `pRadSite` instead of `iRad`.

Also I think elsewhere iterating these arrays uses `auto const&` for the variable but idk idk if that actually matters.

src/Ext/RadSite/Hooks.cpp
```diff
@@ -138,8 +135,10 @@ DEFINE_HOOK(0x43FB23, BuildingClass_AI_Radiation, 0x5)
 	{
 		CellStruct nCurrentCoord = buildingCoords + *pFoundation;
 
-		for (auto& [pRadSite, pRadExt] : RadSiteExt::ExtMap)
+		for (auto const pRadSite : *RadSiteClass::Array)
```
Starkku, line 138: Same thing about the reference applies here.

src/Ext/RadSite/Hooks.cpp
```diff
@@ -190,8 +189,9 @@ DEFINE_HOOK(0x4DA59F, FootClass_AI_Radiation, 0x5)
 		CellStruct CurrentCoord = pFoot->GetCell()->MapCoords;
 
 		// Loop for each different radiation stored in the RadSites container
-		for (auto& [pRadSite, pRadExt] : RadSiteExt::ExtMap)
+		for (auto const pRadSite : *RadSiteClass::Array)
```
Starkku, line 192: Ditto.

src/Ext/Side/Body.cpp
```diff
@@ -149,8 +149,8 @@ DEFINE_HOOK(0x679A10, SideClass_LoadAllFromINI, 0x5)
 {
 	GET_STACK(CCINIClass*, pINI, 0x4);
 
-	for (auto& [pSide, pSideExt] : SideExt::ExtMap)
-		pSideExt->LoadFromINI(pINI);
+	for (auto const pSide : *SideClass::Array)
```
Starkku, line 152: Ditto.

src/Ext/WarheadType/Detonate.cpp
```diff
@@ -327,11 +329,11 @@ void WarheadTypeExt::ExtData::InterceptBullets(TechnoClass* pOwner, WeaponTypeCl
 	}
 	else
 	{
-		for (auto& [pBullet, pBulletExt] : BulletExt::ExtMap)
+		for (auto const pBullet: *BulletClass::Array)
```
Starkku, line 332: Ditto.

Starkku (changes requested):
Tested the fix, it eradicated the reproducible desync I had with RadSites. Couple of possible minor nitpick changes I guess.

src/Ext/RadSite/Hooks.cpp
chaserli: it doesn't matter in this case, the reference simply avoids copying by using pointer, in this case it's an array of pointers, after compiler optimization the asm is the same

Starkku: In that case that 'feedback' can be ignored.

Starkku (approved):

PR #1261, opened 2024-05-16
Two new trajectories with enhanced straight trajectory

CrimRecya:
New Disperse trajectory and Engrave trajectory with enhanced Straight trajectory.

First, the enhanced Straight trajectory.

- Self-explanatory, is a straight-shot trajectory.
  - `Trajectory.Straight.DetonationDistance` controls the maximum distance in cells from intended target (checked at start of each game frame, before the projectile moves) at which the projectile will be forced to detonate. Set to 0 to disable forced detonation (note that this can cause the projectile to overshoot the target). By setting `Trajectory.Straight.PassThrough=true`, it refers to the distance that projectile should travel from itself when it above 0, and the distance that projectile should move behind the target when it below 0 (use the absolute value).
  - `Trajectory.Straight.PassThrough` enables special case logic where the projectile does not detonate in contact with the target but instead travels up to a distance defined by `Trajectory.Straight.DetonationDistance`. Note that the firing angle of the projectile is adjusted with this in mind, making it fire straight ahead if the target is on same elevation.
  - `Trajectory.Straight.PassDetonate` enables extra detonations when the projectile is traveling.
    - `Trajectory.Straight.PassDetonateWarhead` defined the warhead detonated by `Trajectory.Straight.PassDetonate`, and `Trajectory.Straight.PassDetonateDamage` defined the damage caused by `Trajectory.Straight.PassDetonateWarhead`.
    - `Trajectory.Straight.PassDetonateDelay` controls the delay for detonating the warhead defined by `Trajectory.Straight.Warhead`.
    - `Trajectory.Straight.PassDetonateTimer` controls the initial delay for detonating the warhead defined by `Trajectory.Straight.Warhead`.
    - `Trajectory.Straight.PassDetonateLocal` controls whether `Trajectory.Straight.PassDetonateWarhead` and weapon's `Warhead` are always detonate at ground level. It will also no longer restrict vertical velocity of the projectile when using `Trajectory.Straight.ConfineAtHeight`.
  - `Trajectory.Straight.LeadTimeCalculate` controls whether the projectile need to calculate the lead time of the target when firing. Note that this will not affect the facing of the turret.
  - `Trajectory.Straight.OffsetCoord` controls the offsets of the target. Projectile will aim at this position to attack. It also supports `Inaccurate=yes` and `Trajectory.Straight.LeadTimeCalculate=true` on this basis.
    - `Trajectory.Straight.RotateCoord` controls whether to rotate the projectile's firing direction within the angle bisector of `Trajectory.Straight.OffsetCoord` according to the weapon's `Burst`. Set to 0 to disable this function.
    - `Trajectory.Straight.MirrorCoord` controls whether `Trajectory.Straight.OffsetCoord` need to mirror the lateral value to adapt to the current burst index. At the same time, the rotation direction calculated by `Trajectory.Straight.RotateCoord` will also be reversed, and the rotation angle between each adjacent projectile on each side will not change as a result.
    - `Trajectory.Straight.UseDisperseBurst` controls whether the calculation of `Trajectory.Straight.RotateCoord` is based on its superior's `Trajectory.Disperse.WeaponBurst` of the dispersed trajectory, rather than `Burst` of the weapon. If this value is not appropriate, it will result in unsatisfactory visual displays.
    - `Trajectory.Straight.AxisOfRotation` controls the rotation axis when calculating `Trajectory.Straight.RotateCoord`. The axis will rotates with the unit orientation or the vector that from target position to the source position.
  - `Trajectory.Straight.ProximityImpact` controls the initial proximity fuse times. When there are enough remaining times and the projectile approaches another valid target, it will detonate a warhead defined by `Trajectory.Straight.Warhead` on it. If the times is about to run out, it will also detonate itself at its location. This function can be cancelled by setting to 0. A negative integer means unlimited times. By the way, you can use the weapon's `Warhead` with low versus only to aim at the target, and use the `Trajectory.Straight.ProximityWarhead` to causing actual harm.
    - `Trajectory.Straight.ProximityWarhead` defined the warhead detonated by `Trajectory.Straight.ProximityImpact`, and `Trajectory.Straight.ProximityDamage` defined the damage caused by `Trajectory.Straight.ProximityWarhead`.
    - `Trajectory.Straight.ProximityRadius` controls the range of proximity fuse. It can NOT be set as a negative integer.
    - `Trajectory.Straight.ProximityAllies` controls the damage ratio if the target of proximity fuse is ally. It will not detonate at allies by setting as 0. Note that this is not related to whether the warhead itself affect allies.
    - `Trajectory.Straight.ProximityFlight` controls whether to count units in the air.
  - `Trajectory.Straight.ThroughVehicles` controls whether the projectile will not be obstructed by vehicles or aircrafts on the ground. When it is obstructed, it will be directly detonated at its location. If it still have `Trajectory.Straight.ProximityImpact` times, it will also detonate a `Trajectory.Straight.ProximityImpact` at the location of the obstacle.
  - `Trajectory.Straight.ThroughBuilding` controls whether the projectile will not be obstructed by buildings. When it is obstructed, it will be directly detonated at its location. If it still have `Trajectory.Straight.ProximityImpact` times, it will also detonate a `Trajectory.Straight.ProximityImpact` at the location of the obstacle.
  - `Trajectory.Straight.EdgeAttenuation` controls the edge attenuation ratio of projectile damage, includes `Trajectory.Straight.Damage`. Can NOT be set to a negative integer.
  - `Trajectory.Straight.SubjectToGround` controls whether the projectile should explode when it hits the ground. Note that this will make AI search for suitable attack locations.
  - `Trajectory.Straight.ConfineAtHeight` controls the height above ground that projectile will try to travel as it can. It can not move down from the cliff by setting `SubjectToCliffs=yes`. It can be cancelled by setting as a non positive integer. It will be forcibly cancelled by setting `Trajectory.Speed` above 256.

In `rulesmd.ini`:
```ini
Trajectory=Straight                             ; Trajectory type
Trajectory.Straight.DetonationDistance=0.4      ; floating point value
Trajectory.Straight.TargetSnapDistance=0.5      ; floating point value
Trajectory.Straight.PassThrough=false           ; boolean
Trajectory.Straight.PassDetonate=false          ; boolean
Trajectory.Straight.PassDetonateWarhead=        ; WarheadType
Trajectory.Straight.PassDetonateDamage=0        ; integer
Trajectory.Straight.PassDetonateDelay=1         ; integer
Trajectory.Straight.PassDetonateTimer=0         ; integer
Trajectory.Straight.PassDetonateLocal=false     ; boolean
Trajectory.Straight.LeadTimeCalculate=false     ; boolean
Trajectory.Straight.OffsetCoord=0,0,0           ; integer - Forward,Lateral,Height
Trajectory.Straight.RotateCoord=0               ; floating point value
Trajectory.Straight.MirrorCoord=true            ; boolean
Trajectory.Straight.UseDisperseBurst=false      ; boolean
Trajectory.Straight.AxisOfRotation=0,0,1        ; integer - Forward,Lateral,Height
Trajectory.Straight.ProximityImpact=0           ; integer
Trajectory.Straight.ProximityWarhead=           ; WarheadType
Trajectory.Straight.ProximityDamage=0           ; integer
Trajectory.Straight.ProximityRadius=0.7         ; floating point value
Trajectory.Straight.ProximityAllies=0           ; floating point value
Trajectory.Straight.ProximityFlight=false       ; boolean
Trajectory.Straight.ThroughVehicles=true        ; boolean
Trajectory.Straight.ThroughBuilding=true        ; boolean
Trajectory.Straight.EdgeAttenuation=1.0         ; floating point value
Trajectory.Straight.SubjectToGround=false       ; boolean
Trajectory.Straight.ConfineAtHeight=0           ; integer
```

```{note}
- Make sure you set a low `Trajectory.Straight.ProximityRadius` value unless necessary.
- Note that if `Trajectory.Straight.ProximityFlight` is true, it will consume more CPU performance.
- Note that the listed Warheads in `Trajectory.Straight.PassDetonateWarhead` and `Trajectory.Straight.ProximityWarhead` must be listed in `[Warheads]` for them to work.
```

2024-04-2623-21-28-ezgif com-video-to-gif-converter
`Trajectory.Straight.ConfineAtHeight`

2024-05-2100-24-23-ezgif com-video-to-gif-converter
`Trajectory.Straight.OffsetCoord`

2024-05-2100-17-46-ezgif com-video-to-gif-converter
`Trajectory.Straight.PassDetonate`

2024-05-2100-48-34-ezgif com-video-to-gif-converter
`Trajectory.Straight.ProximityImpact`

2024-05-1114-49-41-ezgif com-video-to-gif-converter
`Trajectory.Straight.LeadTimeCalculate`

Then, the new Disperse trajectory.

- Its trajectory looks a bit like a `ROT`, but its settings are more flexible. It also has a unique trajectory. And as its name suggests, it can even spread out more weapons.
  - `Trajectory.Disperse.UniqueCurve` controls whether to enable simulated Qian Xuesen trajectory. After enabling this, it will NOT respect the following 19 items and `Trajectory.Speed`.
  - `Trajectory.Disperse.PreAimCoord` controls the initial flight direction of the projectile, and after reaching this coordinate, it will begin to turn towards the target direction. When it is set to 0,0,0 , it will directly face the target.
    - `Trajectory.Disperse.RotateCoord` controls whether to rotate the projectile's firing direction within the angle bisector of `Trajectory.Disperse.PreAimCoord` according to the weapon's `Burst`. Set to 0 to disable this function.
    - `Trajectory.Disperse.MirrorCoord` controls whether `Trajectory.Disperse.PreAimCoord` need to mirror the lateral value to adapt to the current burst index. At the same time, the rotation direction calculated by `Trajectory.Disperse.RotateCoord` will also be reversed, and the rotation angle between each adjacent projectile on each side will not change as a result.
    - `Trajectory.Disperse.FacingCoord` controls whether the forward direction in `Trajectory.Disperse.PreAimCoord` is depending on the orientation of the firer. By default, it will depend on the vector between the firer and the target.
    - `Trajectory.Disperse.ReduceCoord` controls whether `Trajectory.Disperse.PreAimCoord` defines the initial movement coordinates when the attack distance is 10 cells, and the actual initial movement coordinates will change with the length of the attack distance. It can be simply understood as an optimization term aimed at ensuring hits at close range.
    - `Trajectory.Disperse.UseDisperseBurst` controls whether the calculation of `Trajectory.Disperse.RotateCoord` is based on its superior's `Trajectory.Disperse.WeaponBurst` of the dispersed trajectory, rather than `Burst` of the weapon. If this value is not appropriate, it will result in unsatisfactory visual displays.
    - `Trajectory.Disperse.AxisOfRotation` controls the rotation axis when calculating `Trajectory.Disperse.RotateCoord`. The axis will rotates with the unit orientation or the vector that from target position to the source position.
  - `Trajectory.Disperse.LaunchSpeed` controls the initial flight speed of the projectile, it will be directly reduced to this value if it exceeds `Trajectory.Speed`.
  - `Trajectory.Disperse.Acceleration` controls the acceleration of the projectile's flight speed, increasing the speed per frame according to this value, the final speed will not exceed `Trajectory.Speed`.
  - `Trajectory.Disperse.ROT` controls the turning speed of the projectile and can determine the turning radius of the projectile. The turning speed will increase with the increase of speed.
  - `Trajectory.Disperse.LockDirection` controls whether after reaching `Trajectory.Disperse.PreAimCoord` and completing the first turn, the projectile will lock the direction of motion without further adjustment.
  - `Trajectory.Disperse.CruiseEnable` controls whether after reaching `Trajectory.Disperse.PreAimCoord` and completing the first turn, the projectile will maintain altitude while moving towards the target.
    - `Trajectory.Disperse.CruiseUnableRange` controls how far away it is from the target to end the cruise phase, no longer maintaining the cruise altitude, and begins to move directly towards the target. If the distance is already below this value by the time `Trajectory.Disperse.PreAimCoord` is reached, the cruise phase will be skipped.
  - `Trajectory.Disperse.LeadTimeCalculate` controls whether the projectile need to calculate the lead time of the target during motion.
  - `Trajectory.Disperse.TargetSnapDistance` controls the maximum distance in cells from intended target the projectile can be at moment of detonation to make the projectile 'snap' on the intended target. Set to 0 to disable snapping.
  - `Trajectory.Disperse.RetargetRadius` controls the radius of the projectile to search for a new target after losing its original target. If it hasn't arrived `Trajectory.Disperse.PreAimCoord` yet, the last coordinate of the original target is taken as the center of the searching circle. Otherwise, the coordinate of the distance in front of the projectile is taken as the center of the circle. Set to 0 indicates that this function is not enabled, and it will still attempt to attack the original target's location. If it is set to a negative value, it will directly self explode in place.
    - `Trajectory.Disperse.RetargetAllies` controls whether the projectile chooses allies as its target when searching for new targets after losing its original target.
  - `Trajectory.Disperse.SuicideShortOfROT` controls whether the projectile will explode when it detected its insufficient turning ability.
  - `Trajectory.Disperse.SuicideAboveRange` controls the projectile to self destruct directly after reaching the flight distance. Set to 0 to disable suicide.
  - `Trajectory.Disperse.SuicideIfNoWeapon` controls whether the projectile will self destruct after the number of times it spreads the weapon has been exhausted. If `Trajectory.Disperse.WeaponCount` set to 0, this will not be enabled.
  - `Trajectory.Disperse.Weapons` defined the dispersal weapons of the projectile.
    - `Trajectory.Disperse.WeaponBurst` defined how many corresponding weapons each time the projectile will fire. When the quantity is lower than `Trajectory.Disperse.Weapons`, it will use the last value.
    - `Trajectory.Disperse.WeaponCount` controls how many times the projectile can spread the weapon. Set to a negative value means unlimited times.
    - `Trajectory.Disperse.WeaponDelay` controls the delay for dispersing the weapons defined by `Trajectory.Disperse.Weapons`.
    - `Trajectory.Disperse.WeaponTimer` controls the initial delay for dispersing the weapons defined by `Trajectory.Disperse.Weapons`.
    - `Trajectory.Disperse.WeaponScope` controls the weapon dispersing timer to start counting only within this distance of reaching the target. Set to 0 to disable this function. Set to a negative value means it will only disperse the weapon at most once before detonation.
    - `Trajectory.Disperse.WeaponSeparate` controls whether the projectile no longer fire all the weapons in `Trajectory.Disperse.Weapons` at once and instead fire a group of weapons in the list order, following `Trajectory.Disperse.WeaponBurst`.
    - `Trajectory.Disperse.WeaponRetarget` controls whether the dispersed weapons will find new targets on their own. Using the range of weapons to search new targets.
    - `Trajectory.Disperse.WeaponLocation` controls whether the dispersed weapons will search for new targets at the center of the spreading position, otherwise they will focus on the original target. If a suitable target cannot be found, it will randomly attack the ground.
    - `Trajectory.Disperse.WeaponTendency` controls whether the dispersed weapons will choose the original target as the first new target in each group of weapons.
    - `Trajectory.Disperse.WeaponToAllies` controls whether the dispersed weapons will choose allies as targets when searching for new targets.
    - `Trajectory.Disperse.WeaponToGround` controls whether the dispersed weapons will only choose the ground as their targets.

In `rulesmd.ini`:
```ini
Trajectory=Disperse                             ; Trajectory type
Trajectory.Disperse.UniqueCurve=false           ; boolean
Trajectory.Disperse.PreAimCoord=0,0,0           ; integer - Forward,Lateral,Height
Trajectory.Disperse.RotateCoord=0               ; floating point value
Trajectory.Disperse.MirrorCoord=true            ; boolean
Trajectory.Disperse.FacingCoord=false           ; boolean
Trajectory.Disperse.ReduceCoord=true            ; boolean
Trajectory.Disperse.UseDisperseBurst=false      ; boolean
Trajectory.Disperse.AxisOfRotation=0,0,1        ; integer - Forward,Lateral,Height
Trajectory.Disperse.LaunchSpeed=0               ; floating point value
Trajectory.Disperse.Acceleration=10.0           ; floating point value
Trajectory.Disperse.ROT=30.0                    ; floating point value
Trajectory.Disperse.LockDirection=false         ; boolean
Trajectory.Disperse.CruiseEnable=false          ; boolean
Trajectory.Disperse.CruiseUnableRange=5         ; floating point value
Trajectory.Disperse.LeadTimeCalculate=true      ; boolean
Trajectory.Disperse.TargetSnapDistance=0.5      ; floating point value
Trajectory.Disperse.RetargetRadius=0            ; floating point value
Trajectory.Disperse.RetargetAllies=false        ; boolean
Trajectory.Disperse.SuicideShortOfROT=true      ; boolean
Trajectory.Disperse.SuicideAboveRange=0         ; floating point value
Trajectory.Disperse.SuicideIfNoWeapon=true      ; boolean
Trajectory.Disperse.Weapons=                    ; list of WeaponTypes
Trajectory.Disperse.WeaponBurst=                ; list of integers
Trajectory.Disperse.WeaponCount=0               ; integer
Trajectory.Disperse.WeaponDelay=1               ; integer
Trajectory.Disperse.WeaponTimer=0               ; integer
Trajectory.Disperse.WeaponScope=0               ; floating point value
Trajectory.Disperse.WeaponSeparate=false        ; boolean
Trajectory.Disperse.WeaponRetarget=false        ; boolean
Trajectory.Disperse.WeaponLocation=false        ; boolean
Trajectory.Disperse.WeaponTendency=false        ; boolean
Trajectory.Disperse.WeaponToAllies=false        ; boolean
Trajectory.Disperse.WeaponToGround=false        ; boolean
```

```{note}
- Make sure you set a low `Trajectory.Disperse.RetargetRadius` value unless necessary. If you set `Trajectory.Disperse.WeaponRetarget=true`, also make sure you set `Trajectory.Disperse.Weapons` a low `Range` value unless necessary.
- This does not properly support `Arcing=true` projectiles.
- Note that the listed Weapons in `Trajectory.Disperse.Weapons` must be listed in `[WeaponTypes]` for them to work.
- `Trajectory.Disperse.Weapons` now not support customized `LaserThickness`, `Bolt.ColorN`, `Bolt.DisableN`, `Bolt.Arcs`, `Beam.Color`, `Beam.Duration`  and `Beam.Amplitude`.
```

2024-05-2100-50-25-ezgif com-video-to-gif-converter
`Trajectory=Disperse`

2024-05-2102-38-09-ezgif com-video-to-gif-converter
`Trajectory.Disperse.UniqueCurve`

2024-05-2100-53-47-ezgif com-video-to-gif-converter
`Trajectory.Disperse.FacingCoord`

2024-05-2100-59-26-ezgif com-video-to-gif-converter
`Trajectory.Disperse.LeadTimeCalculate`

2024-05-2620-23-27-ezgif com-video-to-gif-converter
`Trajectory.Disperse.RotateCoord`

2024-05-2100-41-03-ezgif com-video-to-gif-converter
`Trajectory.Disperse.Weapons` > Straight : `Inaccurate`

Last, the new Engrave trajectory.

- Visually, like the thermal lance. Calling it 'trajectory' may not be appropriate. It does not read the settings on the weapon.
  - `Trajectory.Engrave.SourceCoord` controls the starting point of engraving line segment. Taking the target as the coordinate center. Specifically, it will start from the firing position when set to 0,0 . The height of the point will always at ground level.
  - `Trajectory.Engrave.TargetCoord` controls the end point of engraving line segment. Taking the target as the coordinate center. The height of the point will always at ground level.
    - `Trajectory.Engrave.MirrorCoord` controls whether `Trajectory.Engrave.SourceCoord` and `Trajectory.Engrave.TargetCoord` need to mirror the lateral value to adapt to the current FLH.
    - `Trajectory.Engrave.TheDuration` controls the duration of the entire engrave process. Set to 0 will automatically use `Trajectory.Engrave.SourceCoord` and `Trajectory.Engrave.TargetCoord` to calculate the process duration.
  - `Trajectory.Engrave.IsLaser` controls whether laser drawing is required.
    - `Trajectory.Engrave.IsSupported` controls whether the engrave laser will be brighter and thicker. Need to set `Trajectory.Engrave.IsHouseColor` or `Trajectory.Engrave.IsSingleColor` to true.
    - `Trajectory.Engrave.IsHouseColor` controls whether set the engrave laser to draw using player's team color. These lasers respect `Trajectory.Engrave.LaserThickness` and `Trajectory.Engrave.IsSupported`.
    - `Trajectory.Engrave.IsSingleColor` controls whether set the engrave laser to draw using only `Trajectory.Engrave.LaserInnerColor`. These lasers respect `Trajectory.Engrave.LaserThickness` and `Trajectory.Engrave.IsSupported`.
    - `Trajectory.Engrave.LaserInnerColor` controls the inner color of the engrave laser.
    - `Trajectory.Engrave.LaserOuterColor` controls the outer color of the engrave laser.
    - `Trajectory.Engrave.LaserOuterSpread` controls the spread color of the engrave laser.
    - `Trajectory.Engrave.LaserThickness` controls the thickness of the engrave laser. Need to set `Trajectory.Engrave.IsHouseColor` or `Trajectory.Engrave.IsSingleColor` to true.
    - `Trajectory.Engrave.LaserDuration` controls the duration of the engrave laser.
    - `Trajectory.Engrave.LaserDelay` controls how often to draw the engrave laser.
  - `Trajectory.Engrave.DamageDelay` controls how often to detonate warheads.

In `rulesmd.ini`:
```ini
Trajectory=Engrave                             ; Trajectory type
Trajectory.Engrave.SourceCoord=0,0             ; integer - Forward,Lateral
Trajectory.Engrave.TargetCoord=0,0             ; integer - Forward,Lateral
Trajectory.Engrave.MirrorCoord=true            ; boolean
Trajectory.Engrave.TheDuration=0               ; integer
Trajectory.Engrave.IsLaser=true                ; boolean
Trajectory.Engrave.IsSupported=false           ; boolean
Trajectory.Engrave.IsHouseColor=false          ; boolean
Trajectory.Engrave.IsSingleColor=false         ; boolean
Trajectory.Engrave.LaserInnerColor=0,0,0       ; integer - Red,Green,Blue
Trajectory.Engrave.LaserOuterColor=0,0,0       ; integer - Red,Green,Blue
Trajectory.Engrave.LaserOuterSpread=0,0,0      ; integer - Red,Green,Blue
Trajectory.Engrave.LaserThickness=3            ; integer
Trajectory.Engrave.LaserDuration=1             ; integer
Trajectory.Engrave.LaserDelay=1                ; integer
Trajectory.Engrave.DamageDelay=10              ; integer
```

```{note}
- It's best not to let it be intercepted.
```

2024-05-0401-23-49-ezgif com-video-to-gif-converter
`Trajectory=Engrave`

The comprehensive display.

2024-05-3016-02-33-ezgif com-video-to-gif-converter
`Trajectory.Disperse.Weapons` > Straight : `Trajectory.Straight.UseDisperseBurst`

2024-05-2100-35-04-ezgif com-video-to-gif-converter
`Trajectory.Engrave.IsSingleColor`
`Trajectory.Disperse.Weapons` > Straight : `Inaccurate`

2024-05-2201-39-04-ezgif com-video-to-gif-converter
`Trajectory.Engrave.TheDuration`
`Trajectory.Disperse.Weapons` > Straight : `Inaccurate`

2024-05-2102-40-49-ezgif com-video-to-gif-converter
`Trajectory.Engrave.IsSupported`
`Trajectory.Disperse.CruiseEnable`
`Trajectory.Disperse.Weapons` > Disperse : `Trajectory.Disperse.PreAimCoord`
`Trajectory.Straight.ProximityImpact`


src/Ext/Bullet/Hooks.cpp
```diff
@@ -64,6 +64,11 @@ DEFINE_HOOK(0x4666F7, BulletClass_AI, 0x6)
 		}
 	}
 
+	//Let these two types of trajectory draw their own laser to avoid
+	//predicting incorrect positions or pass through targets.
+	if (pBulletExt->Trajectory)
+		return 0;
```
Metadorius, line 70: Isn't this going to disable the laser trails on *all* the trajectories, not just your Engrave one?

Metadorius: Also is there maybe a way to reuse the laser trail type somehow instead of re-defining new parameters on trajectory itself? I haven't looked deeply into the code yet though.

CrimRecya: Not so. Let it skip drawing here and instead let them be drawn after the Trajectory‘s OnAI() because if I don't do this, the laser trails will be drawn before the calculation of changing the velocity direction in each frame. This will cause the laser trails to be drawn in the wrong position too early, resulting in a visual appearance resembling a "bouncing". I haven't come up with a better solution yet.

src/Ext/Bullet/Hooks.cpp
```diff
@@ -363,7 +373,10 @@ DEFINE_HOOK(0x468E9F, BulletClass_Explode_TargetSnapChecks2, 0x6)
 	// Fixes issues with walls etc.
 	if (auto const pExt = BulletExt::ExtMap.Find(pThis))
 	{
-		if (pExt->Trajectory && pExt->Trajectory->Flag == TrajectoryFlag::Straight && !pExt->SnappedToTarget)
+		if (pExt->Trajectory && !pExt->SnappedToTarget
+			&& (pExt->Trajectory->Flag == TrajectoryFlag::Straight
+			|| pExt->Trajectory->Flag == TrajectoryFlag::Disperse
+			|| pExt->Trajectory->Flag == TrajectoryFlag::Engrave))
 			return SkipSetCoordinate;
```
FS-21, line 380: With multiline IFs put brackets please.

src/Ext/Bullet/Hooks.cpp
```diff
@@ -378,7 +391,9 @@ DEFINE_HOOK(0x468D3F, BulletClass_ShouldExplode_AirTarget, 0x6)
 
 	if (auto const pExt = BulletExt::ExtMap.Find(pThis))
 	{
-		if (pExt->Trajectory && pExt->Trajectory->Flag == TrajectoryFlag::Straight)
+		if (pExt->Trajectory && (pExt->Trajectory->Flag == TrajectoryFlag::Straight
+			|| pExt->Trajectory->Flag == TrajectoryFlag::Disperse
+			|| pExt->Trajectory->Flag == TrajectoryFlag::Engrave))
 			return SkipCheck;
```
FS-21, line 397: With multiline IFs put brackets please.

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
```diff
@@ -0,0 +1,886 @@
...
+bool DisperseTrajectory::Load(PhobosStreamReader& Stm, bool RegisterForChange)
+{
+	this->PhobosTrajectory::Load(Stm, false);
+
+	Stm
+		.Process(this->UniqueCurve)
+		.Process(this->PreAimCoord)
+		.Process(this->LaunchSpeed)
+		.Process(this->Acceleration)
+		.Process(this->RateofTurning)
```
FS-21, line 118: Shouldn't be called ROT ?

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
```diff
@@ -0,0 +1,886 @@
...
+bool DisperseTrajectory::Save(PhobosStreamWriter& Stm) const
+{
+	this->PhobosTrajectory::Save(Stm);
+
+	Stm
+		.Process(this->UniqueCurve)
+		.Process(this->PreAimCoord)
+		.Process(this->LaunchSpeed)
+		.Process(this->Acceleration)
+		.Process(this->RateofTurning)
```
FS-21, line 157: As mentioned above I recommend ROT instead of RateofTurning

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
```diff
@@ -0,0 +1,886 @@
...
+		{
+			if (pBullet->Owner)
+			{
+				if (pBullet->Owner->HasTurret())
+					RotateAngle = -(pBullet->Owner->TurretFacing().GetRadian<32>());
+				else
+					RotateAngle = -(pBullet->Owner->PrimaryFacing.Current().GetRadian<32>());
+			}
+			else
+				RotateAngle = Math::atan2(pBullet->TargetCoords.Y - TheSource.Y , pBullet->TargetCoords.X - TheSource.X);
```
FS-21, line 287: Brackets in "else" because you started with brackets in "if"

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
```diff
@@ -0,0 +1,886 @@
...
+	if (this->WeaponScope <=0 || pBullet->TargetCoords.DistanceFrom(pBullet->Location) <= this->WeaponScope)
+	{
+		if (this->WeaponCount > 0)
+		{
+			if (pOwner)
+			{
+				if (PrepareDisperseWeapon(pBullet, pOwner))
+					return true;
+			}
+			else
```
FS-21, line 316: Put the else in brackets.

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
```diff
@@ -0,0 +1,886 @@
...
+	}
+
+	if (this->SuicideAboveRange > 0)
+	{
+		double BulletSpeed = this->LaunchSpeed;
+		if (this->UniqueCurve)
+		{
+			BulletSpeed = pBullet->Velocity.Magnitude();
+		}
+		this->SuicideAboveRange -= BulletSpeed;
```
FS-21, line 436: Put a blank line for separating the brackets and the new code.

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
```diff
@@ -0,0 +1,886 @@
...
+bool DisperseTrajectory::BulletRetargetTechno(BulletClass* pBullet, HouseClass* pOwner)
+{
+	bool Check = false;
+	if (!pBullet->Target)
+		Check = true;
+	else if (TechnoClass* pTarget = abstract_cast<TechnoClass*>(pBullet->Target))
+		if (pTarget->IsDead() || pTarget->InLimbo)
+			Check = true;
+
+	if (Check)
```
FS-21, line 456: You can reduce in 1 level if you write:

if (!Check)
    return false;

Everything else what was inside the "if (Check)"

FS-21, line 456: Sicne is a large block inside the "if (Check)" You could reduce 1 level if:

if (!Check)
    return false;
    
And there here everything else, including the return false of the end;

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
```diff
@@ -0,0 +1,886 @@
...
+			if (pBullet->Velocity.Z > -160.0)
+				pBullet->Velocity.Z -= 4.0;
+
+			if (FutureLocation <= TargetLocation.Z)
+				this->InStraight = true;
+			else if (FutureLocation <= pBullet->SourceCoords.Z)
+				this->InStraight = true;
+		}
+	}
+
```
FS-21, line 548: Remove this gap

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
```diff
@@ -0,0 +1,886 @@
...
+{
+	CoordStruct TargetVelocity
+	{
+		TargetLocation.X - pBullet->Location.X,
+		TargetLocation.Y - pBullet->Location.Y,
+		TargetLocation.Z - pBullet->Location.Z
+	};
+
+	BulletVelocity MoveToVelocity = pBullet->Velocity;
+
```
FS-21, line 599: Maybe is better if these 2 lines are together without a gap

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
```diff
@@ -0,0 +1,886 @@
...
+					}
+
+					if (this->WeaponTendency && i == 0 && j == 0)
+						BulletTarget = pBullet->Target ? pBullet->Target : MapClass::Instance->TryGetCellAt(pBullet->TargetCoords);
+
+					CreateDisperseBullets(pBullet, pWeapon, BulletTarget, pOwner);
+				}
+			}
+		}
+
```
FS-21, line 738: Remove the gap

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
```diff
@@ -0,0 +1,886 @@
...
+			}
+
+			else
+			{
+				pLaser = GameCreate<LaserDrawClass>(pBullet->Location, BulletTarget->GetCoords(),
+					pWeapon->LaserInnerColor, pWeapon->LaserOuterColor, pWeapon->LaserOuterSpread, pWeapon->LaserDuration);
+				pLaser->IsHouseColor = false;
+			}
+
+			pLaser->Thickness = 3;//LaserThickness
```
FS-21, line 854: If is a leftover from old code remove this, if not put a little gap please.

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
```diff
@@ -0,0 +1,886 @@
...
+			if (pWeapon->IsAlternateColor)
+				pEBolt->AlternateColor = true;
+
+			else
+				pEBolt->AlternateColor = false;
+
+			if (auto pWeaponTypeExt = WeaponTypeExt::ExtMap.Find(pWeapon))
+				if (pWeaponTypeExt->Bolt_Disable1)
+
+		//	Bolt.Color1,Bolt.Color2,Bolt.Color3
```
FS-21, line 868: Remember remove the old code if you finished this block of code

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
```diff
@@ -0,0 +1,886 @@
...
+			pRadBeamType = RadBeamType::Temporal;
+		else
+			pRadBeamType = RadBeamType::RadBeam;
+
+		if (RadBeam* pRadBeam = RadBeam::Allocate(pRadBeamType))
+		{
+			pRadBeam->SetCoordsSource(pBullet->Location);
+			pRadBeam->SetCoordsTarget(BulletTarget->GetCoords());
+			pRadBeam->Color = pWeapon->Warhead->Temporal ? ColorStruct { 255, 100, 0 } : ColorStruct { 128, 200, 255 };//Beam.Color
+		//	pRadBeam->Period = ;//Beam.Duration
```
FS-21, line 888: Remember delete old commented code.

src/Ext/Bullet/Trajectories/EngraveTrajectory.cpp
```diff
@@ -0,0 +1,419 @@
...
+bool EngraveTrajectory::OnAI(BulletClass* pBullet)
+{
+	if (!pBullet->Owner)
+		return true;
+
+	if (this->TechnoInLimbo != pBullet->Owner->InLimbo)
+		return true;
+
+	this->CheckTimesLimit -= 1;
+	if (this->CheckTimesLimit < 0)
```
FS-21, line 275: Since "else" has brackets this "if" too

src/Ext/Bullet/Trajectories/EngraveTrajectory.cpp
```diff
@@ -0,0 +1,419 @@
...
+		return true;
+	else//SetLocation() seems to work wrong if I put this part into OnAIVelocity().
```
FS-21: Just for easy lecture put a little separation in the coment please

src/Ext/Bullet/Trajectories/EngraveTrajectory.cpp
```diff
@@ -0,0 +1,419 @@
...
+
+		if (abs(CheckDifference) >= 384)
+		{
+			if (CheckDifference > 0)
+			{
+				BulletCoords.Z += CheckDifference;
+				pBullet->SetLocation(BulletCoords);
+				this->SetItsLocation = false;
+			}
+			else
```
FS-21, line 306: if "if" has brackets then "else" must have brackets too

src/Ext/Bullet/Trajectories/EngraveTrajectory.cpp
```diff
@@ -0,0 +1,419 @@
...
+			pLaser->IsHouseColor = true;
+		}
+
+		else if (this->IsSingleColor)
+		{
+			pLaser = GameCreate<LaserDrawClass>(FireCoord, pBullet->Location, this->LaserInnerColor,
+				ColorStruct { 0, 0, 0 }, ColorStruct { 0, 0, 0 }, this->LaserDuration);
+			pLaser->IsHouseColor = true;
+		}
+
```
FS-21: remove this little gap

src/Ext/Bullet/Trajectories/EngraveTrajectory.cpp
```diff
@@ -0,0 +1,419 @@
...
+	this->LaserTimer += 1;
+	this->LaserTimer %= this->LaserDelay;
+	this->DamageTimer += 1;
+	this->DamageTimer %= this->DamageDelay;
+
+	return false;
+}
+
+void EngraveTrajectory::OnAIPreDetonate(BulletClass* pBullet){}//do nothing
```
FS-21, line 393: Just for easy lecture put a little separation in the coment after the "else" please (Nothing personal :-D )

CrimRecya: Thank you for your comment. I have made all these changes OvO.

src/Ext/Bullet/Trajectories/StraightTrajectory.cpp
```diff
@@ -70,40 +172,99 @@ bool StraightTrajectory::Save(PhobosStreamWriter& Stm) const
...
+			BulletDetonateLastCheck(pBullet, StraightSpeed);
+		}
+		else
+		{
+			if (this->ConfineAtHeight > 0)
+			{
+				if (PassAndConfineAtHeight(pBullet, StraightSpeed))
+					return true;
+			}
+		}
```
FS-21, line 250:
```suggestion
		else if (this->ConfineAtHeight > 0)
		{
			if (PassAndConfineAtHeight(pBullet, StraightSpeed))
				return true;
		}
```

src/Ext/Bullet/Trajectories/StraightTrajectory.cpp
```diff
@@ -124,100 +285,867 @@ void StraightTrajectory::OnAIVelocity(BulletClass* pBullet, BulletVelocity* pSpe
...
+		return true;
+	}
+
+	this->CheckTimesLimit -= 1;
+	if (this->CheckTimesLimit < 0)
+		return true;
+
+	if (!this->PassThrough && this->DetonationDistance > 0
+		&& pBullet->TargetCoords.DistanceFrom(pBullet->Location) < this->DetonationDistance)
+		return true;
```
FS-21, line 543:
```suggestion
	if (!this->PassThrough && this->DetonationDistance > 0
		&& pBullet->TargetCoords.DistanceFrom(pBullet->Location) < this->DetonationDistance)
	{
		return true;
	}
```

src/Ext/Bullet/Trajectories/StraightTrajectory.cpp
```diff
@@ -124,100 +285,867 @@ void StraightTrajectory::OnAIVelocity(BulletClass* pBullet, BulletVelocity* pSpe
...
+
+	if (CellClass* pCell = MapClass::Instance->TryGetCellAt(pBullet->Location))
+		return false;
+	else
+		return true;
+}
+
+void StraightTrajectory::BulletDetonateLastCheck(BulletClass* pBullet, double StraightSpeed)
+{
+	if (StraightSpeed < 512.0)
```
FS-21, line 556: Reduce 1 level if

if (StraightSpeed >= 512.0)
	return;

Here all the code that was inside the "if (StraightSpeed < 512.0)"

src/Ext/Bullet/Trajectories/StraightTrajectory.cpp
```diff
@@ -124,100 +285,867 @@ void StraightTrajectory::OnAIVelocity(BulletClass* pBullet, BulletVelocity* pSpe
...
+		double LocationDistance = pBullet->Location.DistanceFrom(pBullet->TargetCoords);
 
-		if (pCell)
-			coords = pCell->GetCoordsWithBridge();
+		if (!this->PassThrough)
+		{
+			if (LocationDistance < StraightSpeed)
+			{
+				VelocityCheck = 2;
+			}
```
FS-21, line 574:
```suggestion
			if (LocationDistance < StraightSpeed)
				VelocityCheck = 2;
```

src/Ext/Bullet/Trajectories/StraightTrajectory.cpp
```diff
@@ -124,100 +285,867 @@ void StraightTrajectory::OnAIVelocity(BulletClass* pBullet, BulletVelocity* pSpe
...
+		{
+			if (ThisSize >= Capacity)
+				break;
+
+			while (pObject)
+			{
+				auto const pTechno = abstract_cast<TechnoClass*>(pObject);
+				pObject = pObject->NextObject;
+
+				if (pTechno)
```
FS-21, line 639: You can reduce 1 level if you write:

if (!pTechno)
	continue;

And after that Here the rest of the "if (pTechno)" code.

src/Ext/Bullet/Trajectories/StraightTrajectory.cpp
```diff
@@ -124,100 +285,867 @@ void StraightTrajectory::OnAIVelocity(BulletClass* pBullet, BulletVelocity* pSpe
...
+				RecCells.push_back(mCurCell);
+				mCurCell.X -= mTheUnit.X;
+				mCurCell.Y += mTheUnit.Y;
+				RecCells.push_back(mCurCell);
+				mCurCell.X += mTheUnit.X;
+				RecCells.push_back(mCurCell);
+			}
+		}
+	}
+
```
FS-21, line 797: Remove the gap bewteen if & else

src/Ext/Bullet/Hooks.cpp
CrimRecya: I tried to rewrite the comments for these two lines. This is a modification made for all trajectories. The purpose is to draw the laser trails after the velocity calculation for this frame is completed, so as not to draw to the position before the velocity calculation due to drawing one frame earlier. Apologize for my poor English, I think this translation software has helped me express myself clearly.

Metadorius: It is fine, your English is pretty good from what I've seen. We will do a detailed review later. Thanks for your contribution!

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
```diff
@@ -0,0 +1,890 @@
...
+			StandardVelocityChange(pBullet);
+
+		VelocityUp = true;
+	}
+
+	if (VelocityUp)
+	{
+		if (CalculateBulletVelocity(pBullet, this->LaunchSpeed))
+			return true;
+	}
```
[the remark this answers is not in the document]
CrimRecya: Thank you very much for your correction.

FS-21:
By the way, what is exactly a "engrave" trajectory? because I don't get the concept in the documentation. Maybe if I was an English native I could understand.

If possible I would put the right gif you created in the right place documentation for easy understanding :-/
I only recognice the disperse trajectory in the gifs because it works like a shotgun.

For example what trajectory was used in the last gif that remembers me the Protoss Colossus dual weapon
image

FS-21 (commented):
I think the latest changes are well written so I have no complains with the code.

Another story is all tags were tested in online for hunting a possible desync.

CrimRecya:
> By the way, what is exactly a "engrave" trajectory? because I don't get the concept in the documentation. Maybe if I was an English native I could understand.
> 
> If possible I would put the right gif you created in the right place documentation for easy understanding :-/ I only recognice the disperse trajectory in the gifs because it works like a shotgun.
> 
> For example what trajectory was used in the last gif that remembers me the Protoss Colossus dual weapon image

Apologize for my poor English. The Last gif is the "engrave" trajectory. And thank you for your suggestion. I will create a few more GIF images these days and place them appropriately.

CrimRecya:
Updated introduction. The resources from mod MO3 are for display purposes only.

chaserli (changes requested):
Impressive work. The quantity is so huge that I'm afraid it's almost impossible to review and merge all of these on 1 single PR. imo I suggest detach these into several commits:
* The "engrave" one
* The "disperse" one
* As for the straight trajectory one, see if it's possible to incorporate the "inaccurate" part of #1246. The biggest concern I noticed is that vector of  "Causality" thing you made: Are you sure the technos always live longer than your bullet? If not then pointer invalidation will be quite a big trouble for perf. Try to see if there's a simpler way

CrimRecya:
> Impressive work. The quantity is so huge that I'm afraid it's almost impossible to review and merge all of these on 1 single PR. imo I suggest detach these into several commits:
> 
> * The "engrave" one
> * The "disperse" one
> * As for the straight trajectory one, see if it's possible to incorporate the "inaccurate" part of [Trajectory projectiles enhancement #1246](https://github.com/Phobos-developers/Phobos/pull/1246). The biggest concern I noticed is that vector of  "Causality" thing you made: Are you sure the technos always live longer than your bullet? If not then pointer invalidation will be quite a big trouble for perf. Try to see if there's a simpler way

Okay. But I have a question about splitting here that whether [the state of mutual reference between my Trajectories, such as Disperse using Straight's ‘UseDisperseBurst’ tag, or Disperse, Straight, and Engrave using CheckSubjectToGround() in Hooks.Obstacles’s at the same time] may cause splitting problems.
And for the ‘Inaccurate’ part of Straight, I already have the corresponding part here, and the issue of null pointers in the Casualty part is actually no longer returning Vector during Merge step, which has passed hundreds of tests in game and no need to worry about this anymore :).

chaserli:
> Okay. But I have a question about splitting here that whether [the state of mutual reference between my Trajectories, such as Disperse using Straight's ‘UseDisperseBurst’ tag, or Disperse, Straight, and Engrave using CheckSubjectToGround() in Hooks.Obstacles’s at the same time] may cause splitting problems. And for the ‘Inaccurate’ part of Straight, I already have the corresponding part here, and the issue of null pointers in the Casualty part is actually no longer returning Vector during Merge step, which has passed hundreds of tests in game and no need to worry about this anymore :).

Got a point. Excuse me for the oversight, I'll have a look later on this one.

PR #1263, opened 2024-05-18
Minor crit enhancement

Coronia:
1 - Crit chance and extra damage can now be determined differently when target health is below certain percentage

- `Crit.Chance` determines chance for a critical hit to occur. By default this is checked once when the Warhead is detonated and every target that is susceptible to critical hits will be affected. If `Crit.ApplyChancePerTarget` is set, then whether or not the chance roll is successful is determined individually for each target. Can be set to a single value for all conditions, or a list of values which would work when target health is below certain percentage. Value from position matching the position from `Crit.AffectBelowPercent` is used if found, or 0.0 if not found.
- `Crit.ExtraDamage` determines the damage dealt by the critical hit. If `Crit.Warhead` is set, the damage is used to detonate the specified Warhead on each affected target, otherwise the damage is directly dealt based on current Warhead's `Verses` settings. Can be set to a single value for all conditions, or a list of values which would work when target health is below certain percentage. Value from position matching the position from `Crit.AffectBelowPercent` is used if found, or 0 if not found.
- `Crit.AffectBelowPercent`, if set to a single value, determines minimum percentage of their maximum `Strength` that targets must have left to be affected by a critical hit. If set to a list of values, it'll further determine `Crit.Chance` and `Crit.ExtraDamage` when target health is below the certain percentage listed here.
  - Health percentage values in the list should be in descending order, or the health check could go wrong.

2 - Add AE multiplier for crit extra damage

- `Crit.ExtraDamage.Multiplier` and `Crit.ExtraDamage.Bonus` can be used to multiply the critical hit extra damage or grant a fixed bonus to it for the object the effect is attached to, respectively. They also respect `Crit.AllowWarheads` and `Crit.DisallowWarheads`.

In `rulesmd.ini`:
```ini
[SOMEWARHEAD]                       ; Warhead
Crit.Chance=                        ; list of floating-point values (percentage or absolute) (0.0-1.0)
Crit.ExtraDamage=                   ; list of integers
Crit.AffectBelowPercent=            ; list of floating-point values (percentage or absolute) (0.0-1.0)

[SOMEATTACHEFFECT]                           ; AttachEffectType
Crit.ExtraDamage.Multiplier=1.0              ; floating point value
Crit.ExtraDamage.Bonus=0                     ; integer
```

NOTE:
- If using crit logic in the original way (single value for `Crit.Chance`, `Crit.ExtraDamage` and `Crit.AffectBelowPercent`), it'll work exactly the same.

Example 1:
- when target health is between 100%-75% it'll have a 10% chance to deal crit with 10 extra damage.
- when target health is between 75%-50% it'll have a 20% chance to deal crit with 10 extra damage.
- when target health is between 50%-25% it'll have a 20% chance to deal crit with 20 extra damage.
- when target health is between 25%-0% it'll have a 50% chance to deal crit with 50 extra damage.
```ini
[SOMEWARHEAD]
Crit.Chance=10%,20%,20%,50%
Crit.ExtraDamage=10,10,20,50
Crit.AffectBelowPercent=100%,75%,50%,25%
```

Example 2:
use this new logic to simply increase damage based on the target's remaining health.
since `Crit.Chance` is set to a single value, it means that it'll be 100% for any health percentage.
```ini
[SOMEWARHEAD]
Crit.Chance=1.0
Crit.ExtraDamage=10,20,30,40,50,60,70,80,90,100
Crit.AffectBelowPercent=100%,90%,80%,70%,60%,50%,40%,30%,20%,10%
```

## Summary by CodeRabbit

- **Updates**
	- Updated a submodule or dependency that may include new features and bug fixes, potentially enhancing project functionality.


FS-21 (commented):
The code looks fine.

docs/Whats-New.md
```diff
@@ -403,6 +403,9 @@ New:
 - Custom tint on TechnoTypes (by Starkku)
 - Revenge weapon (by Starkku)
 - AttachEffect types with new features like custom tint and weapon range modifier (by Starkku)
+- Warheads spawning powerup crates (by Starkku)
```
Starkku, line 406: This line is a duplicate, I guess it is an error from solving merge conflicts when merging changes from develop.

src/Ext/WarheadType/Detonate.cpp
```diff
@@ -316,7 +334,37 @@ void WarheadTypeExt::ExtData::ApplyCrit(HouseClass* pHouse, TechnoClass* pTarget
...
+
+		if (pType->Crit_AllowWarheads.size() > 0 && !pType->Crit_AllowWarheads.Contains(this->OwnerObject()))
+			continue;
+
+		if (pType->Crit_DisallowWarheads.size() > 0 && pType->Crit_DisallowWarheads.Contains(this->OwnerObject()))
+			continue;
+
+		damage = static_cast<int>(damage * pType->Crit_ExtraDamage_Multiplier);
+		extraDamageBonus += pType->Crit_ExtraDamage_Bonus;
+	}
```
Starkku, line 365: Refactor the code so it only loops through this list once for both chance and the damage multipliers. As it stands it is very inefficient.

Starkku (changes requested):
As far as I can tell code looks mostly fine even though it massively complicates the functionality here. Couple of things, one minor and another bigger thing to look at though.

src/Ext/WarheadType/Detonate.cpp
Coronia: already applied these changes

Coronia:
To make it clearer that this change allow a list of number as input, I've added the following 3 tags. If both these tags and the original ones are set, they'll override the old ones.

In `rulesmd.ini`:
```ini
[SOMEWARHEAD]                       ; Warhead
Crit.Chances=                       ; list of floating-point values (percentage or absolute) (0.0-1.0), overrides Crit.Chance
Crit.ExtraDamages=                  ; list of integers, overrides Crit.ExtraDamage
Crit.AffectBelowPercents=           ; list of floating-point values (percentage or absolute) (0.0-1.0), overrides Crit.AffectBelowPercent
```

Coronia:
this implementation is way too complex but has little actual usage. It'll be more beneficial to do it in the warhead level, i.e. detonate different warheads based on the target's strength. As for the fire multiplier for extra damage, maybe adding a `Crit.ExtraDamage.ApplyFirepowerMult` tag would be sufficient instead of altering the AE system

PR #1264, opened 2024-05-19
New AutoDeath condition & convert when AutoDeath

Coronia:
1 - added the following AutoDeath conditions:
- `OwnedByPlayer` / `OwnedByAI`: The object will die if its owner house is / isn't a human player.
- `MoneyExceed` / `MoneyBelow`: The object will die if its owner house has credits greater / smaller than (or equal to) set value.
- `LowPower` / `FullPower`: The object will die if its owner house is in low / full power.
- `PassengersExceed` / `PassengersBelow`: The object will die if its current passenger amount is greater / smaller than (or equal to) set value.

2 - instead of firing AutoDeath when any condition is met, you can now set `OnAnyCondition` to false to make it trigger only when all conditions are met

3 - AutoDeath now support `convert` behavior. If the object is a **unit**, it will be changed into the TechnoType in `Convert.AutoDeath` instead of destroyed.

In `rulesmd.ini`:
```ini
[SOMETECHNO]                                   ; TechnoType
AutoDeath.Behavior=                            ; enumeration (kill | vanish | sell | convert), default not set
AutoDeath.OwnedByPlayer=false                  ; boolean
AutoDeath.OwnedByAI=false                      ; boolean
AutoDeath.MoneyExceed=-1                       ; integer
AutoDeath.MoneyBelow=-1                        ; integer
AutoDeath.LowPower=false                       ; boolean
AutoDeath.FullPower=false                      ; boolean
AutoDeath.PassengersExceed=-1                  ; integer
AutoDeath.PassengersBelow=-1                   ; integer
AutoDeath.OnAnyCondition=true                  ; boolean

Convert.AutoDeath=                             ; TechnoType
```

### DIFF
--- a/src/Ext/Techno/Body.Update.cpp
+++ b/src/Ext/Techno/Body.Update.cpp
@@ -135,12 +135,124 @@ bool TechnoExt::ExtData::CheckDeathConditions(bool isInLimbo)
 	// Self-destruction must be enabled
 	const auto howToDie = pTypeExt->AutoDeath_Behavior.Get();
 	const auto pVanishAnim = pTypeExt->AutoDeath_VanishAnimation.Get();
+	const auto pConvert = pTypeExt->AutoDeath_Convert.Get();
+	const bool anyMatched = pTypeExt->AutoDeath_ContentIfAnyMatch;
+
+	// Death by money
+
+	if (pTypeExt->AutoDeath_MoneyExceed >= 0)
+	{
+		if (pThis->Owner && pThis->Owner->Available_Money() >= pTypeExt->AutoDeath_MoneyExceed)
+		{
+			if (anyMatched)
+			{
+				TechnoExt::KillSelf(pThis, howToDie, pVanishAnim, isInLimbo, pConvert);
+				return true;
+			}
+		}
+		else if (!anyMatched)
+		{
+			return false;
+		}
+	}
+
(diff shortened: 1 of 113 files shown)


mbnq:
Would be nice if there was also:
`AutoDeath.IsAI=`
That would work when owner of techno isn't human player.

Coronia:
> ```ini
> AutoDeath.OwnedByPlayer
> ```

src/Ext/Techno/Body.Update.cpp
MortonPL, line 1: This whole bit should be rearranged IMO. There's many repeated condition checks and calls.

MortonPL (changes requested):

chaserli (changes requested):
Your intention seems to make it an autonomous trigger event/action system, at this point we might need to redesign the whole thing instead of stacking things up.

One way is to regroup it into the AttachEffect system, so that you can apply an AE under a given condition. That might be a huge change to the code structure.

Another easier way is to develop such condition-consequence system, which sounds already much simpler. 

To start with, you can wrap all these, including the old ones, into a standalone entity.

Metadorius:
> One way is to regroup it into the AttachEffect system, so that you can apply an AE under a given condition. That might be a huge change to the code structure.

Why not just fire a weapon on itself instead?

PR #1265, opened 2024-05-19
[Minor] Change a human player condition

Coronia:
`HouseClass::IsHumanPlayer` can't cover the case when player is controlling multiple houses in campaign. Should use `HouseClass::IsControlledByHuman()` in these cases.


chaserli (commented):
* Did the previous fix work?
* Under which circumstances will this change help? confirmed?

PR #1267, opened 2024-05-20
Improve Phobos warhead effect CellSpread behaviour

Starkku:
Fixes effect + `CellSpread`to interaction with buildings correctly e.g hitting any cell on foundation counts as a hit instead of having to hit the top-leftmost cell (cell 0).

Also reimplemented air unit checks using `AircraftTrackerClass`.


DmitryVolkov666:
seems to be working as expected

Metadorius (approved):

PR #1274, opened 2024-06-01
Couple of minor digital display improvements

Starkku:
- `Offset` and `Offset.ShieldDelta` Y-axis offset behaved inconsistently compared to everything else in the game, with negative values shifting down and positive up. It now works the other way around like `PixelSelectionBracketDelta`, `YDrawOffset` etc. do.
- Added `ValueScaleDivisor` (taking suggestions for name) that can be used to adjust the scale of displayed values. This is given as an integer to prevent rounding issues with floating point values and is used to divide the values.


Metadorius:
What are your thoughts about a migration script entry?

Metadorius:
Also perhaps `ValueScaleDivisor` should be done for floating damage numbers too?

Starkku:
> What are your thoughts about a migration script entry?

For the offset? Could probably be added.

> Also perhaps ValueScaleDivisor should be done for floating damage numbers too?

Possibly, but I figured since those are treated as a debug tool currently leaving the actual damage numbers intact makes more sense there.

MortonPL:
> `Offset` and `Offset.ShieldDelta` Y-axis offset behaved inconsistently compared to everything else in the game, with negative values shifting down and positive up. It now works the other way around like `PixelSelectionBracketDelta`, `YDrawOffset` etc. do.

This must be documented.

Metadorius (approved):

Metadorius (approved):

PR #1275, opened 2024-06-01
Strafing.Shots > 5

chaserli:
Now works with any projectile (no longer requires ROT <=1)

### DIFF
--- a/src/Ext/Aircraft/Hooks.cpp
+++ b/src/Ext/Aircraft/Hooks.cpp
@@ -3,6 +3,7 @@
 
 
 #include <Ext/Aircraft/Body.h>
+#include <Ext/Techno/Body.h>
 #include <Ext/Anim/Body.h>
 #include <Ext/WeaponType/Body.h>
 #include <Utilities/Macro.h>
@@ -11,9 +12,12 @@ DEFINE_HOOK(0x417FF1, AircraftClass_Mission_Attack_StrafeShots, 0x6)
 {
 	GET(AircraftClass*, pThis, ESI);
 
+	auto pExt = TechnoExt::ExtMap.Find(pThis);
+
 	if (pThis->MissionStatus < (int)AirAttackStatus::FireAtTarget2_Strafe
 		|| pThis->MissionStatus >(int)AirAttackStatus::FireAtTarget5_Strafe)
 	{
+		pExt->Strafe_BombsDroppedThisRound = 0;
 		return 0;
 	}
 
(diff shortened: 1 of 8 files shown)


src/Ext/Techno/Body.h
```diff
@@ -34,6 +34,7 @@ class TechnoExt
 		CDTimerClass AutoDeathTimer;
 		AnimTypeClass* MindControlRingAnimType;
 		int DamageNumberOffset;
+		int Strafe_HowManyShitIPooped;
```
Starkku, line 37: Change the variable name. I know Ares has le funny names here and there but we don't need to copy that.

src/Ext/Aircraft/Hooks.cpp
```diff
@@ -32,6 +53,40 @@ DEFINE_HOOK(0x417FF1, AircraftClass_Mission_Attack_StrafeShots, 0x6)
 	return 0;
 }
 
+__forceinline BOOL Weapon_Is_Strafe(WeaponTypeClass const* wpnType)
```
Starkku, line 56: `BOOL`, `TRUE` and `FALSE` don't need to be capitalized all over the place. Also probably use `pWeapon` for the `WeaponTypeClass` pointer name, as is usually the case across the project where only a single weapon pointer is present.

src/Ext/Aircraft/Hooks.cpp
```diff
@@ -19,9 +20,29 @@ DEFINE_HOOK(0x417FF1, AircraftClass_Mission_Attack_StrafeShots, 0x6)
 
 	int weaponIndex = pThis->SelectWeapon(pThis->Target);
 	auto pWeaponExt = WeaponTypeExt::ExtMap.Find(pThis->GetWeapon(weaponIndex)->WeaponType);
+
+	auto const& sst = pWeaponExt->Strafing_Shots;
```
Starkku, line 24: Could probably be renamed to something more immediately apparent (`strafingShots`, `maxShots`, `shots`). The `Get()` calls after `isset()` check can also probably be safely ditched. That or assign to a variable that is used thereafter instead.

Starkku (changes requested):
`Strafing.Shots should not control whether or not the attack is strafing, it should logically be separated to a different key, probably called just S`trafing=bool`. `Strafing.Shots` just controls how many times it fires, you might want to make a strafing weapon that isn't `ROT` < 2 and not change that and Phobos should intuitively support that.

Some code style nitpicks as well.

Starkku (approved):

Starkku:
Tested and it works fine. Took a liberty of fixing some minor code style issues and updating YRpp with better definitions for the unknown `AircraftClass` fields as well as adjusting couple of things in the documentation (this isn't really a Phobos bug fix). Merging since it is more or less complete.

PR #1280, opened 2024-06-14
[Minor] allow scatter as an option for animtype CreateUnit.Mission

chaserli:


Starkku:
The whole business with the `scatter` is not really needed here, you can remove the `InAir()` check from [this row](https://github.com/Phobos-developers/Phobos/blob/9e1160fa7f2b195bc0d9781a3a46e5867c735951/src/Ext/Anim/Hooks.AnimCreateUnit.cpp#L147) and add it to [these](https://github.com/Phobos-developers/Phobos/blob/9e1160fa7f2b195bc0d9781a3a46e5867c735951/src/Ext/Anim/Hooks.AnimCreateUnit.cpp#L162) [rows](https://github.com/Phobos-developers/Phobos/blob/9e1160fa7f2b195bc0d9781a3a46e5867c735951/src/Ext/Anim/Hooks.AnimCreateUnit.cpp#L168) (change to `else if`) and any jumpjet spawned from anim that does not have `CreateUnit.AlwaysSpawnOnGround` set to true (a question can be raised if there should be an additional flag here instead but I imagine this will cover vast majority of use cases - if people want to force ground spawn they can set this flag) will ascend automatically upon spawning without spinning around and retaining their initial facing as well as whatever mission they're set to have.

image

PR #1288, opened 2024-06-23
[Customized] Technos keep range after firing

CrimRecya:
- Technos will maintain a suitable distance after firing if `KeepRange` is not set to 0.
  - `KeepRange` controls how long the distance to maintain when the techno's ROF timer is ticking. What is actually read is its absolute value. If it is a positive value, it will be stayed outside this distance, just like it has a special `MinimumRange` after firing. If it is a negative value, it will be kept as close as possible to this distance, just like it has a special `Range` after firing. In addition, if the effective range section is too small, it will be considered unable to fire. It is best to have an effective range of 1.0, and 2.0 is best for Infantry.
    - `KeepRange.AllowAI` controls whether this function is effective for computer.
    - `KeepRange.AllowPlayer` controls whether this function is effective for human.

In `rulesmd.ini`:
```ini
KeepRange=0                  ; floating point value
KeepRange.AllowAI=false      ; boolean
KeepRange.AllowPlayer=false  ; boolean
```


CrimRecya:
> If positive KeepRange= is higher than weapon's Range=, or if negative KeepRange= is lower than weapon's MinimumRange=, then the unit will go into melee range between shots. This behavior is probably fine for negative KeepRange=. However, for positive values, i was expecting something like this: a unit with Range=2, KeepRange=8 and long ROF would get close to fire, and then run away to 8 range, and then come back to Range=2 once its ROF timer ends. Unfortunately, that's not the case.
> 
> Otherwise appears to be working fine.

Thank you for testing. The original design intention of this feature is indeed to allow technos to make corresponding actions within its range. Therefore, I use this flag to replace the weapon's `Range`/`MinimumRange` when the techno is rearming.
I have tested replacing both at the same time to achieve movement beyond the actual range, but this *may* still cause the techno to shift from attacking targets to moving towards them. This may require more effort.

TaranDahl (approved):
Tested locally. Generally works as expected.
The technos may acts stupid if a short range is given to keep. But I think it is ww's stupid pathfinding that is to blame.
Should also work properly online from the code perspective. Didn't tested, though.

PR #1291, opened 2024-06-25
[Customized] Projectile subject to ground check before firing

CrimRecya:

If set `SubjectToGround` to true ,it will predict will predict the height of the connecting straight line from the bullet's source coordinates to target coordinates. If the predicted height is lower than the ground height of the current predicted position, the firer will also consider targets blocked by such obstacles as out of range and will attempt to reposition themselves.

In `rulesmd.ini`:
```ini
[SOMEPROJECTILE]              ; Projectile
SubjectToGround=false         ; boolean
```


Coronia (changes requested):

docs/New-or-Enhanced-Logics.md
```diff
@@ -683,6 +683,7 @@ Shrapnel.UseWeaponTargeting=false  ; boolean
 
 - It is now possible to make projectiles consider either land or water as obstacles that block their path by setting `SubjectToLand/Water` to true, respectively. Weapons firing such projectiles will consider targets blocked by such obstacles as out of range and will attempt to reposition themselves so they can fire without being blocked by the said obstacles before firing and if `SubjectToLand/Water.Detonate` is set to true, the projectiles will detonate if they somehow manage to collide with the said obstacles.
   - `Level=true` projectiles detonate on tiles belonging to non-water tilesets by default, but will not consider such tiles as true obstacles. This behaviour can be overridden by setting these keys.
+- As for `SubjectToGround`, if set it to true ,it will predict will predict the height of the connecting straight line from the bullet's source coordinates to target coordinates. If the predicted height is lower than the ground height of the current predicted position, the firer will also consider targets blocked by such obstacles as out of range and will attempt to reposition themselves.
```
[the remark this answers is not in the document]
CrimRecya: Get this.

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -575,29 +575,56 @@ namespace BurstFLHTemp
...
+
+	auto const pTypeExt = pExt->TypeExtData;
+
 	if (weaponIndex < 0)
+	{
+		FootClass* currentPassenger = pThis->Passengers.FirstPassenger;
+		const int passengerIndex = -weaponIndex - 1;
+
+		for (int i = 0; i < passengerIndex && currentPassenger; i++)
+			currentPassenger = abstract_cast<FootClass*>(currentPassenger->NextObject);
```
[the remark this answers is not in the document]
CrimRecya: I think it's not necessary because it's only one line here and it's clear enough

src/Ext/Bullet/Hooks.Obstacles.cpp
```diff
@@ -82,6 +85,27 @@ class BulletObstacleHelper
...
 	}
+
+	static CoordStruct AddFLHToSourceCoords(const CoordStruct sourceCoords, TechnoClass* const pTechno, AbstractClass* const pTarget, bool& subjectToGround)
+	{
+		if (!subjectToGround)
+			return sourceCoords;
+
+		if (pTechno->IsInAir() || pTarget->IsInAir())
+		{
+			subjectToGround = false;
```
Coronia, line 96: is there a need to prohibit air unit and target from using this?

src/Ext/Bullet/Hooks.Obstacles.cpp
```diff
@@ -166,7 +189,11 @@ DEFINE_HOOK(0x6F7647, TechnoClass_InRange_Obstacles, 0x5)
 	auto pTechno = InRangeTemp::Techno;
 
 	if (!pObstacleCell)
-		pObstacleCell = BulletObstacleHelper::FindFirstImpenetrableObstacle(*pSourceCoords, targetCoords, pTechno, pTarget, pTechno->Owner, pWeapon, true);
+	{
+		auto subjectToGround = BulletTypeExt::ExtMap.Find(pWeapon->Projectile)->SubjectToGround.Get(); // Make AI search for suitable attack locations.
```
Coronia, line 193: not need to use Get() here since Valueable always has a valid value

CrimRecya: Without this, compilation would not be possible because `Valueable<bool>` cannot be passed as `bool&`

src/Ext/Bullet/Hooks.Obstacles.cpp
CrimRecya: Updated.
When the air force is in use, they cannot find a suitable location normally. When the target is the air force, I think there is no need to make additional calculations.

PR #1294, opened 2024-06-25
[General Purpose] Enhanced Straight trajectory

CrimRecya:

- Self-explanatory, is a straight-shot trajectory.
  - `Trajectory.Straight.DetonationDistance` controls the maximum distance in cells from intended target (checked at start of each game frame, before the projectile moves) at which the projectile will be forced to detonate. Set to 0 to disable forced detonation (note that this can cause the projectile to overshoot the target). If `Trajectory.Straight.ApplyRangeModifiers` is set to true, any applicable weapon range modifiers from the firer are applied here as well. By setting `Trajectory.Straight.PassThrough=true`, it refers to the distance that projectile should travel from itself when it above 0, and the distance that projectile should move behind the target when it below 0 (use the absolute value).
  - `Trajectory.Straight.PassThrough` enables special case logic where the projectile does not detonate in contact with the target but instead travels up to a distance defined by `Trajectory.Straight.DetonationDistance`. Note that the firing angle of the projectile is adjusted with this in mind, making it fire straight ahead if the target is on same elevation.
  - `Trajectory.Straight.PassDetonate` enables extra detonations when the projectile is traveling.
    - `Trajectory.Straight.PassDetonateWarhead` defined the warhead detonated by `Trajectory.Straight.PassDetonate`, and `Trajectory.Straight.PassDetonateDamage` defined the damage caused by `Trajectory.Straight.PassDetonateWarhead`.
    - `Trajectory.Straight.PassDetonateDelay` controls the delay for detonating the warhead defined by `Trajectory.Straight.Warhead`.
    - `Trajectory.Straight.PassDetonateInitialDelay` controls the initial delay for detonating the warhead defined by `Trajectory.Straight.Warhead`.
    - `Trajectory.Straight.PassDetonateLocal` controls whether `Trajectory.Straight.PassDetonateWarhead` and weapon's `Warhead` are always detonate at ground level. It will also no longer restrict vertical velocity of the projectile when using `Trajectory.Straight.ConfineAtHeight`.
  - `Trajectory.Straight.LeadTimeCalculate` controls whether the projectile need to calculate the lead time of the target when firing. Note that this will not affect the facing of the turret.
  - `Trajectory.Straight.OffsetCoord` controls the offsets of the target. Projectile will aim at this position to attack. It also supports `Inaccurate=yes` and `Trajectory.Straight.LeadTimeCalculate=true` on this basis.
    - `Trajectory.Straight.RotateCoord` controls whether to rotate the projectile's firing direction within the angle bisector of `Trajectory.Straight.OffsetCoord` according to the weapon's `Burst`. Set to 0 to disable this function.
    - `Trajectory.Straight.MirrorCoord` controls whether `Trajectory.Straight.OffsetCoord` need to mirror the lateral value to adapt to the current burst index. At the same time, the rotation direction calculated by `Trajectory.Straight.RotateCoord` will also be reversed, and the rotation angle between each adjacent projectile on each side will not change as a result.
    - `Trajectory.Straight.UseDisperseBurst` controls whether the calculation of `Trajectory.Straight.RotateCoord` is based on its superior's `Trajectory.Disperse.WeaponBurst` of the dispersed trajectory, rather than `Burst` of the weapon. If this value is not appropriate, it will result in unsatisfactory visual displays.
    - `Trajectory.Straight.AxisOfRotation` controls the rotation axis when calculating `Trajectory.Straight.RotateCoord`. The axis will rotates with the unit orientation or the vector that from target position to the source position.
  - `Trajectory.Straight.ProximityImpact` controls the initial proximity fuse times. When there are enough remaining times and the projectile approaches another valid target, it will detonate a warhead defined by `Trajectory.Straight.Warhead` on it. If the times is about to run out, it will also detonate itself at its location. This function can be cancelled by setting to 0. A negative integer means unlimited times. By the way, you can use the weapon's `Warhead` with low versus only to aim at the target, and use the `Trajectory.Straight.ProximityWarhead` to causing actual harm.
    - `Trajectory.Straight.ProximityWarhead` defined the warhead detonated by `Trajectory.Straight.ProximityImpact`, and `Trajectory.Straight.ProximityDamage` defined the damage caused by `Trajectory.Straight.ProximityWarhead`.
    - `Trajectory.Straight.ProximityRadius` controls the range of proximity fuse. It can NOT be set as a negative integer.
    - `Trajectory.Straight.ProximityDirect` controls whether let the target receive damage instead of detonating the warhead.
    - `Trajectory.Straight.ProximityAllies` controls whether allies will also trigger the proximity fuse.
    - `Trajectory.Straight.ProximityFlight` controls whether to count units in the air.
  - `Trajectory.Straight.ThroughVehicles` controls whether the projectile will not be obstructed by vehicles or aircrafts on the ground. When it is obstructed, it will be directly detonated at its location. If it still have `Trajectory.Straight.ProximityImpact` times, it will also detonate a `Trajectory.Straight.ProximityImpact` at the location of the obstacle.
  - `Trajectory.Straight.ThroughBuilding` controls whether the projectile will not be obstructed by buildings. When it is obstructed, it will be directly detonated at its location. If it still have `Trajectory.Straight.ProximityImpact` times, it will also detonate a `Trajectory.Straight.ProximityImpact` at the location of the obstacle.
  - `Trajectory.Straight.EdgeAttenuation` controls the edge attenuation ratio of projectile damage, includes `Trajectory.Straight.Damage`. Can NOT be set to a negative integer.
  - `Trajectory.Straight.SubjectToGround` controls whether the projectile should explode when it hits the ground. Note that this will not make AI search for suitable attack locations.
  - `Trajectory.Straight.ConfineAtHeight` controls the height above ground that projectile will try to travel as it can. It can not move down from the cliff by setting `SubjectToCliffs=yes`. It can be cancelled by setting as a non positive integer. It will be forcibly cancelled by setting `Trajectory.Speed` above 256.

In `rulesmd.ini`:
```ini
Trajectory=Straight                             ; Trajectory type
Trajectory.Straight.ApplyRangeModifiers=false   ; boolean
Trajectory.Straight.DetonationDistance=0.4      ; floating point value
Trajectory.Straight.TargetSnapDistance=0.5      ; floating point value
Trajectory.Straight.PassThrough=false           ; boolean
Trajectory.Straight.PassDetonate=false          ; boolean
Trajectory.Straight.PassDetonateWarhead=        ; WarheadType
Trajectory.Straight.PassDetonateDamage=0        ; integer
Trajectory.Straight.PassDetonateDelay=1         ; integer
Trajectory.Straight.PassDetonateInitialDelay=0  ; integer
Trajectory.Straight.PassDetonateLocal=false     ; boolean
Trajectory.Straight.LeadTimeCalculate=false     ; boolean
Trajectory.Straight.OffsetCoord=0,0,0           ; integer - Forward,Lateral,Height
Trajectory.Straight.RotateCoord=0               ; floating point value
Trajectory.Straight.MirrorCoord=true            ; boolean
Trajectory.Straight.UseDisperseBurst=false      ; boolean
Trajectory.Straight.AxisOfRotation=0,0,1        ; integer - Forward,Lateral,Height
Trajectory.Straight.ProximityImpact=0           ; integer
Trajectory.Straight.ProximityWarhead=           ; WarheadType
Trajectory.Straight.ProximityDamage=0           ; integer
Trajectory.Straight.ProximityRadius=0.7         ; floating point value
Trajectory.Straight.ProximityDirect=false       ; boolean
Trajectory.Straight.ProximityAllies=false       ; boolean
Trajectory.Straight.ProximityFlight=false       ; boolean
Trajectory.Straight.ThroughVehicles=true        ; boolean
Trajectory.Straight.ThroughBuilding=true        ; boolean
Trajectory.Straight.EdgeAttenuation=1.0         ; floating point value
Trajectory.Straight.SubjectToGround=false       ; boolean
Trajectory.Straight.ConfineAtHeight=0           ; integer
```

```{note}
- Make sure you set a low `Trajectory.Straight.ProximityRadius` value unless necessary.
- Note that if `Trajectory.Straight.ProximityFlight` is true, it will consume more CPU performance.
- Note that the listed Warheads in `Trajectory.Straight.PassDetonateWarhead` and `Trajectory.Straight.ProximityWarhead` must be listed in `[Warheads]` for them to work.
```


src/Ext/Bullet/Hooks.cpp
```diff
@@ -64,6 +64,12 @@ DEFINE_HOOK(0x4666F7, BulletClass_AI, 0x6)
 		}
 	}
 
+	//Because the laser trails will be drawn before the calculation of changing the velocity direction in each frame.
+	//This will cause the laser trails to be drawn in the wrong position too early, resulting in a visual appearance resembling a "bouncing".
```
chaserli, line 68: Can you provide some ini code for me to quickly reproduce this issue? on 336

src/Ext/BulletType/Body.h
```diff
@@ -46,6 +46,10 @@ class BulletTypeExt
 		Valueable<bool> Arcing_AllowElevationInaccuracy;
 		Valueable<WeaponTypeClass*> ReturnWeapon;
 
+		//If simply put these part into Trajectory theirselves, firing these bullets will crash the game after load the saved games.
```
chaserli, line 49: Did you figure out why? even ValueableIdx doesn't work?

src/Ext/Bullet/Trajectories/StraightTrajectory.h
```diff
@@ -50,15 +146,58 @@ class StraightTrajectory final : public PhobosTrajectory
 	virtual TrajectoryCheckReturnType OnAITargetCoordCheck(BulletClass* pBullet) override;
 	virtual TrajectoryCheckReturnType OnAITechnoCheck(BulletClass* pBullet, TechnoClass* pTechno) override;
 
+	struct CasualtyData
+	{
+		TechnoClass* pCasualty;
```
chaserli, line 151: I still don't quite buy it. There might be a chance that the techno is killed/limboed/removed while the bullet is on the way, right?

src/Ext/Bullet/Hooks.cpp
CrimRecya: You can use the original bombard trajectory and give it a high speed (>300) to see this phenomenon. This problem will occur if the bullet (use trajectory) is turned without this modification. And this phenomenon will be very obvious if the bullet is very fast.

src/Ext/BulletType/Body.h
CrimRecya: Unfortunately, I have not found the root cause of this problem. I have also tried using Nullable before, but this problem still exists. In the end I tried to put this here like other bullets, then the problem is solved miraculously. I hope someone can find a better solution.

src/Ext/Bullet/Trajectories/StraightTrajectory.h
CrimRecya: Yes, this struct is only used to determine if the techno is duplicated. And you can see lines 834 to 913, it will not return to the vector in the end if the pointer is no longer valid.

src/Ext/Bullet/Hooks.cpp
chaserli: sounds fair, let me commit this thing first so that you won't need to include it in all your traj prs

src/Ext/BulletType/Body.h
CrimRecya: When I was testing, it was certain that all `Warhead` used here was included in `[Warheads]`. The magical thing is that if I don't do this, when playing a complete game, it won't have any problems yet. But if reading the saved game to continue, although these warheads will still be considered as not nullptr, they actually irrational and crash the game in bullet‘s CreateBullet(). Thank you for your suggestion, I'll try it on the weekend. :)

CrimRecya: > That sound like bug, if it was `nullptr` then de-serialize as random pointer and crash the game , something must be wrong on the code for sure.

I guess this may be due to some issues occurring from saving the game to reading the saved game. But actually, I don't understand this aspect yet, I have written the relevant parts according to the existing ones. So I don't understand why it's not work to put it inside in its own part, but it's okay to put it outside in the extdata. If it is really a problem with saving, I think I need some help.

CrimRecya: This problem has been solved! I have put them all back to themselves :) .

CrimRecya:
Changes:
    - `Trajectory.Straight.ProximityDirect` controls whether let the target receive damage instead of detonating the warhead.
    - `Trajectory.Straight.ProximityAllies` controls whether allies will also trigger the proximity fuse.

In `rulesmd.ini`:
```ini
Trajectory.Straight.ProximityDirect=false       ; boolean
Trajectory.Straight.ProximityAllies=false       ; boolean
```

then separate the allies damage multiplier to a more comprehensive function in #1307 .

Starkku:
It is a small code style thing but it is consistent across all of the code you write, so figured I'll mention it when it is on top of my mind.

Phobos code should use camelCase (first letter of the first word is uncapitalized) for local variables names, not PascalCase (caps).

CrimRecya:
> It is a small code style thing but it is consistent across all of the code you write, so figured I'll mention it when it is on top of my mind.
> 
> Phobos code should use camelCase (first letter of the first word is uncapitalized) for local variables names, not PascalCase (caps). Pointers specifically are prefixed with `p` but no other prefixes (such a `b` for bools or `n` for integers) are in actual use, despite some of it appearing in the codebase.

Get it. I will check them all out these days :) .

Starkku:
Currently (in this branch) any `Trajectory=Straight` weapon fired by AircraftType requires `Trajectory.Straight.LeadTimeCalculate` to be set to true to function correctly, and this will not work on BuildingTypes due to a check. Simple solution might be to force this behaviour on for AircraftType-fired straight trajectory bullets, proper solution would likely require more research into why it doesn't work properly in the first place.

CrimRecya:
> Currently (in this branch) any `Trajectory=Straight` weapon fired by AircraftType requires `Trajectory.Straight.LeadTimeCalculate` to be set to true to function correctly, and this will not work on BuildingTypes due to a check. Simple solution might be to force this behaviour on for AircraftType-fired straight trajectory bullets, proper solution would likely require more research into why it doesn't work properly in the first place.

Thx for testing, I will ty to fix it later. :o

CrimRecya:
> Currently (in this branch) any `Trajectory=Straight` weapon fired by AircraftType requires `Trajectory.Straight.LeadTimeCalculate` to be set to true to function correctly, and this will not work on BuildingTypes due to a check. Simple solution might be to force this behaviour on for AircraftType-fired straight trajectory bullets, proper solution would likely require more research into why it doesn't work properly in the first place.

It should be able to work normally now. :)

Starkku:
I just reviewed the PR and the code & logic are mostly sound, to the extent that I can follow it (more on that later), there are some inconsistencies with existing codebase regarding use of `const` and explicit type declarations vs. use of `auto` but it is all technically correct and contribution guide does not comment on this so I'd need to consult with other devs on that so no need to worry about it now.

I do have concerns about the verbosity of some of the code though. Particularly the obstacle (walls etc) and the proximity rectangle checks. There is existing projectile obstacle check functions in vanilla game, as well as some Phobos-specific ones in `Ext\Bullet\Hooks.Obstacles.cpp` (which could be moved elsewhere if there's need to access them outside that file) but I cannot ascertain if they would have fit here so was that considered during development of this feature? As for the proximity check, I have trouble following the exact logic here but it seems like a lot of code for simply calculating cell offsets that game usually does with a table of `CellStruct` offsets (which could be dynamically constructed based on a given range here I suppose - if it is always same per type it could also only be done once) but maybe I am missing something critical here.

CrimRecya:
> I just reviewed the PR and the code & logic are mostly sound, to the extent that I can follow it (more on that later), there are some inconsistencies with existing codebase regarding use of `const` and explicit type declarations vs. use of `auto` but it is all technically correct and contribution guide does not comment on this so I'd need to consult with other devs on that so no need to worry about it now.
> 
> I do have concerns about the verbosity of some of the code though. Particularly the obstacle (walls etc) and the proximity rectangle checks. There is existing projectile obstacle check functions in vanilla game, as well as some Phobos-specific ones in `Ext\Bullet\Hooks.Obstacles.cpp` (which could be moved elsewhere if there's need to access them outside that file) but I cannot ascertain if they would have fit here so was that considered during development of this feature? As for the proximity check, I have trouble following the exact logic here but it seems like a lot of code for simply calculating cell offsets that game usually does with a table of `CellStruct` offsets (which could be dynamically constructed based on a given range here I suppose - if it is always same per type it could also only be done once) but maybe I am missing something critical here.

It's like this: When the velocity of the projectile is greater than 1 grid per frame, the additional wall check is used mainly to deal with the situation where a straight projectile can pass through the target, but there are walls behind the target. For the position behind the target，although it will not be considered an obstacle during launch, vanilla check cannot perfectly block the projectile in the wall position because sometimes walls are just right between two updates' positions, and then the projectile will pass through them.
As for the rectangular grid check here, I didn't find a tilted rectangular range search function like circular range search, so I wrote this here to find the grids within a certain range on both sides of the projectile's forward direction through the rectangles' four vertices, so that I can find the target who may should be damaged.

src/Ext/Bullet/Trajectories/StraightTrajectory.h
```diff
@@ -49,17 +114,45 @@ class StraightTrajectory final : public PhobosTrajectory
...
-	StraightTrajectoryType const* Type;
+	int PassDetonateDamage;
+	CDTimerClass PassDetonateTimer;
+	CoordStruct OffsetCoord;
+	bool UseDisperseBurst;
+	int ProximityImpact;
+	int ProximityDamage;
+	int RemainingDistance;
+	TechnoClass* ExtraCheck;
+	std::map<TechnoClass*, int> TheCasualty;
```
chaserli, line 127: Is it possible that you think of a different implementation that avoids storing these 2 members? I'm still not convinced it's safe for 2 reasons: pointer invalidation and swizzle. 

std::map is a bst, there's no way Ares' swizzler would work here. I assume you probably saw the wrong examples in HouseExt but unfortunately they should have been corrected long ago.

chaserli (changes requested):

src/Ext/Bullet/Trajectories/StraightTrajectory.h
CrimRecya: Let me explain:
In a certain frame, when the projectile finds a unit that will block it, it is recorded in `ExtraCheck`, and then set the target of the projectile as it, and recalculate the speed. At the next frame, `ExtraCheck` will no longer be taken out for use. Instead, when it is checked to be non empty, it is directly judged at the beginning that it should be detonated
And I use `TheCasualty` to record whether the unit pointed to by the pointer has been hurt in a short time. It is only used to compare the value of the pointer itself. All the pointers that need to be used must still be just obtained from the range. These units will eventually be sorted by distance and `UniqueID`.
I can add some comments for codes. If necessary, I think I can also replace 'thecaculty' with a record `<UniqueID, frames>`, because I just need a "name" for comparison.

chaserli: <int,int> is fine. Just don't use pointers cuz I don't think you swizzle

CrimRecya: Copy that. I will modify it later.

docs/New-or-Enhanced-Logics.md
```diff
@@ -740,18 +740,74 @@ Trajectory.Speed=100.0  ; floating point value
...
+  - `Trajectory.Straight.PassThrough` enables special case logic where the projectile does not detonate in contact with the target but instead travels up to a distance defined by `Trajectory.Straight.DetonationDistance`. Note that the firing angle of the projectile is adjusted with this in mind, making it fire straight ahead if the target is on same elevation.
+  - `Trajectory.Straight.PassDetonate` enables extra detonations when the projectile is traveling.
+    - `Trajectory.Straight.PassDetonateWarhead` defines the warhead detonated by `Trajectory.Straight.PassDetonate`, and `Trajectory.Straight.PassDetonateDamage` defines the damage caused by `Trajectory.Straight.PassDetonateWarhead`.
+    - `Trajectory.Straight.PassDetonateDelay` controls the delay for detonating the warhead defined by `Trajectory.Straight.Warhead`.
+    - `Trajectory.Straight.PassDetonateTimer` controls the initial delay for detonating the warhead defined by `Trajectory.Straight.Warhead`.
```
chaserli, line 749: according to your description why not just PassDetonateInitalDelay?

src/Ext/Bullet/Trajectories/PhobosTrajectory.cpp
```diff
@@ -62,6 +62,10 @@ void TrajectoryTypePointer::LoadFromINI(CCINIClass* pINI, const char* pSection)
 	if (_ptr)
 	{
 		_ptr->Trajectory_Speed.Read(exINI, pSection, "Trajectory.Speed");
+
+		if (abs(_ptr->Trajectory_Speed) < 1e-10)
+			_ptr->Trajectory_Speed = 0.001;
```
chaserli, line 67:
```suggestion
		_ptr->Trajectory_Speed = Math::max(0.001,_ptr->Trajectory_Speed);
```
btw can Trajectory.Speed be negative?

src/Ext/Bullet/Trajectories/PhobosTrajectory.h
```diff
@@ -9,7 +9,7 @@ enum class TrajectoryFlag : int
 {
 	Invalid = -1,
 	Straight = 0,
-	Bombard = 1,
+	Bombard = 1
```
chaserli, line 12: discard it

src/Ext/Bullet/Trajectories/StraightTrajectory.cpp
```diff
@@ -121,100 +249,1037 @@ void StraightTrajectory::OnAIVelocity(BulletClass* pBullet, BulletVelocity* pSpe
...
+			const auto straightSpeedSquared = pType->Trajectory_Speed * pType->Trajectory_Speed;
+			const auto baseFactor = straightSpeedSquared - targetSpeedSquared;
+			const auto squareFactor = baseFactor * verticalDistanceSquared + straightSpeedSquared * horizonDistanceSquared;
+
+			if (squareFactor > 1e-10)
+			{
+				const auto minusFactor = -(horizonDistance * targetSpeed);
+				int travelTime = 0;
+
+				if (abs(baseFactor) < 1e-10)
```
chaserli, line 299:
```suggestion
				if (std::abs(baseFactor) < 1e-10)
```
I know msvc won't confuse but add std:: just in case.

chaserli (approved):

src/Ext/Bullet/Trajectories/PhobosTrajectory.cpp
CrimRecya: I am not sure if anyone is using this "feature", so I just make sure it is not 0. Let me get rid of it.

PR #1295, opened 2024-06-25
[General Purpose] New Disperse trajectory

CrimRecya:

- Its trajectory looks a bit like a `ROT`, but its settings are more flexible. It also has a unique trajectory. And as its name suggests, it can even spread out more weapons.
  - `Trajectory.Disperse.UniqueCurve` controls whether to enable simulated Qian Xuesen trajectory. After enabling this, it will NOT respect the following 21 items and `Trajectory.Speed`.
  - `Trajectory.Disperse.PreAimCoord` controls the initial flight direction of the projectile, and after reaching this coordinate, it will begin to turn towards the target direction. When it is set to 0,0,0 , it will directly face the target.
    - `Trajectory.Disperse.RotateCoord` controls whether to rotate the projectile's firing direction within the angle bisector of `Trajectory.Disperse.PreAimCoord` according to the weapon's `Burst`. Set to 0 to disable this function.
    - `Trajectory.Disperse.MirrorCoord` controls whether `Trajectory.Disperse.PreAimCoord` need to mirror the lateral value to adapt to the current burst index. At the same time, the rotation direction calculated by `Trajectory.Disperse.RotateCoord` will also be reversed, and the rotation angle between each adjacent projectile on each side will not change as a result.
    - `Trajectory.Disperse.FacingCoord` controls whether the forward direction in `Trajectory.Disperse.PreAimCoord` is depending on the orientation of the firer. By default, it will depend on the vector between the firer and the target.
    - `Trajectory.Disperse.ReduceCoord` controls whether `Trajectory.Disperse.PreAimCoord` defines the initial movement coordinates when the attack distance is 10 cells, and the actual initial movement coordinates will change with the length of the attack distance. It can be simply understood as an optimization term aimed at ensuring hits at close range.
    - `Trajectory.Disperse.UseDisperseBurst` controls whether the calculation of `Trajectory.Disperse.RotateCoord` is based on its superior's `Trajectory.Disperse.WeaponBurst` of the dispersed trajectory, rather than `Burst` of the weapon. If this value is not appropriate, it will result in unsatisfactory visual displays.
    - `Trajectory.Disperse.AxisOfRotation` controls the rotation axis when calculating `Trajectory.Disperse.RotateCoord`. The axis will rotates with the unit orientation or the vector that from target position to the source position.
  - `Trajectory.Disperse.LaunchSpeed` controls the initial flight speed of the projectile.
  - `Trajectory.Disperse.Acceleration` controls the acceleration of the projectile's flight speed, increasing or decreasing the speed per frame according to this value, the final speed will be `Trajectory.Speed`.
  - `Trajectory.Disperse.ROT` controls the turning speed of the projectile which refers to the maximum angle that the projectile can rotate per frame in terms of direction.
  - `Trajectory.Disperse.LockDirection` controls whether after reaching `Trajectory.Disperse.PreAimCoord` and completing the first turn, the projectile will lock the direction of motion without further adjustment.
  - `Trajectory.Disperse.CruiseEnable` controls whether after reaching `Trajectory.Disperse.PreAimCoord` and completing the first turn, the projectile will maintain altitude while moving towards the target.
    - `Trajectory.Disperse.CruiseUnableRange` controls how far away it is from the target to end the cruise phase, no longer maintaining the cruise altitude, and begins to move directly towards the target. If the distance is already below this value by the time `Trajectory.Disperse.PreAimCoord` is reached, the cruise phase will be skipped.
    - `Trajectory.Disperse.CruiseAltitude` controls the minimum altitude of the projectile in the cruise phase.
    - `Trajectory.Disperse.CruiseAlongLevel` controls whether to calculate `Trajectory.Disperse.CruiseAltitude` by the height of the projectile and the ground height of the current position, otherwise it will be calculated by the height of the projectile and the height of the launch position.
  - `Trajectory.Disperse.LeadTimeCalculate` controls whether the projectile need to calculate the lead time of the target during motion.
  - `Trajectory.Disperse.RecordSourceCoord` controls whether the projectile needs to record the launch position, which will be used for the weapons in `Trajectory.Disperse.Weapons`. It can be nested and inherited when both sides set this to true.
  - `Trajectory.Disperse.RetargetAllies` controls whether the projectile chooses allies as its target when searching for new targets after losing its original target.
  - `Trajectory.Disperse.RetargetRadius` controls the radius of the projectile to search for a new target after losing its original target. If it hasn't arrived `Trajectory.Disperse.PreAimCoord` yet, the last coordinate of the original target is taken as the center of the searching circle. Otherwise, the coordinate of the distance in front of the projectile is taken as the center of the circle. Set to 0 indicates that this function is not enabled, and it will still attempt to attack the original target's location. If it is set to a negative value, it will directly self explode in place.
  - `Trajectory.Disperse.TargetSnapDistance` controls the maximum distance in cells from intended target the projectile can be at moment of detonation to make the projectile 'snap' on the intended target. Set to 0 to disable snapping.
  - `Trajectory.Disperse.SuicideAboveRange` controls the projectile to self destruct directly after reaching the flight distance. Set to 0 to disable suicide. When set to a negative number, its absolute value represents a multiple of the initial distance.
  - `Trajectory.Disperse.SuicideShortOfROT` controls whether the projectile will explode when it detected its insufficient turning ability.
  - `Trajectory.Disperse.SuicideIfNoWeapon` controls whether the projectile will self destruct after the number of times it spreads the weapon has been exhausted. If `Trajectory.Disperse.WeaponCount` set to 0, this will not be enabled.
  - `Trajectory.Disperse.Weapons` defines the dispersal weapons of the projectile.
    - `Trajectory.Disperse.WeaponBurst` defines how many corresponding weapons each time the projectile will fire. When the quantity is lower than `Trajectory.Disperse.Weapons`, it will use the last value.
    - `Trajectory.Disperse.WeaponCount` controls how many times the projectile can spread the weapon. Set to a negative value means unlimited times.
    - `Trajectory.Disperse.WeaponDelay` controls the delay for dispersing the weapons defined by `Trajectory.Disperse.Weapons`.
    - `Trajectory.Disperse.WeaponInitialDelay` controls the initial delay for dispersing the weapons defined by `Trajectory.Disperse.Weapons`.
    - `Trajectory.Disperse.WeaponEffectiveRange` controls the weapon dispersing timer to start counting only within this distance of reaching the target. Set to 0 to disable this function. Set to a negative value means it will only disperse the weapon at most once before detonation.
    - `Trajectory.Disperse.WeaponSeparate` controls whether the projectile no longer fire all the weapons in `Trajectory.Disperse.Weapons` at once and instead fire a group of weapons in the list order, following `Trajectory.Disperse.WeaponBurst`.
    - `Trajectory.Disperse.WeaponRetarget` controls whether the dispersed weapons will find new targets on their own. Using the `Range`, `CanTarget`, required `AttachedEffects` of weapons to search new targets.
    - `Trajectory.Disperse.WeaponLocation` controls whether the dispersed weapons will search for new targets at the center of the spreading position, otherwise they will focus on the original target.
    - `Trajectory.Disperse.WeaponTendency` controls whether the dispersed weapons will choose the original target as the first new target in each group of weapons.
    - `Trajectory.Disperse.WeaponHolistic` controls whether the dispersed weapons will choose targets that are in different states from the original target (in air and on ground).
    - `Trajectory.Disperse.WeaponMarginal` controls whether the dispersed weapons will choose unimportant items such as trees and streetlights as secondary targets.
    - `Trajectory.Disperse.WeaponToAllies` controls whether the dispersed weapons will choose allies as targets when searching for new targets.
    - `Trajectory.Disperse.WeaponDoRepeat` controls whether the dispersed weapons will select duplicate targets when the number of targets is insufficient. If it is set to true and the weapon can select both the technos and the ground as targets, the technos will be prioritized, then if all non-repeating technos have been selected and the weapon can still be launched at this time (in each round of salvo), it will start selecting duplicate technos. If it is set to false and the weapon can select both the technos and the ground as targets, the technos will be prioritized, followed by the ground cells, then if all non-repeating technos have been selected and the weapon can still be launched at this time (in each round of salvo), it will stop firing remaining bursts.

In `rulesmd.ini`:
```ini
Trajectory=Disperse                             ; Trajectory type
Trajectory.Disperse.UniqueCurve=false           ; boolean
Trajectory.Disperse.PreAimCoord=0,0,0           ; integer - Forward,Lateral,Height
Trajectory.Disperse.RotateCoord=0               ; floating point value
Trajectory.Disperse.MirrorCoord=true            ; boolean
Trajectory.Disperse.FacingCoord=false           ; boolean
Trajectory.Disperse.ReduceCoord=true            ; boolean
Trajectory.Disperse.UseDisperseBurst=false      ; boolean
Trajectory.Disperse.AxisOfRotation=0,0,1        ; integer - Forward,Lateral,Height
Trajectory.Disperse.LaunchSpeed=0               ; floating point value
Trajectory.Disperse.Acceleration=10.0           ; floating point value
Trajectory.Disperse.ROT=10.0                    ; floating point value
Trajectory.Disperse.LockDirection=false         ; boolean
Trajectory.Disperse.CruiseEnable=false          ; boolean
Trajectory.Disperse.CruiseUnableRange=5         ; floating point value
Trajectory.Disperse.CruiseAltitude=800          ; integer
Trajectory.Disperse.CruiseAlongLevel=false      ; boolean
Trajectory.Disperse.LeadTimeCalculate=true      ; boolean
Trajectory.Disperse.RecordSourceCoord=false     ; boolean
Trajectory.Disperse.RetargetAllies=false        ; boolean
Trajectory.Disperse.RetargetRadius=0            ; floating point value
Trajectory.Disperse.TargetSnapDistance=0.5      ; floating point value
Trajectory.Disperse.SuicideAboveRange=-3.0      ; floating point value
Trajectory.Disperse.SuicideShortOfROT=false     ; boolean
Trajectory.Disperse.SuicideIfNoWeapon=true      ; boolean
Trajectory.Disperse.Weapons=                    ; list of WeaponTypes
Trajectory.Disperse.WeaponBurst=                ; list of integers
Trajectory.Disperse.WeaponCount=0               ; integer
Trajectory.Disperse.WeaponDelay=1               ; integer
Trajectory.Disperse.WeaponInitialDelay=0        ; integer
Trajectory.Disperse.WeaponEffectiveRange=0      ; floating point value
Trajectory.Disperse.WeaponSeparate=false        ; boolean
Trajectory.Disperse.WeaponRetarget=false        ; boolean
Trajectory.Disperse.WeaponLocation=false        ; boolean
Trajectory.Disperse.WeaponTendency=false        ; boolean
Trajectory.Disperse.WeaponHolistic=false        ; boolean
Trajectory.Disperse.WeaponMarginal=false        ; boolean
Trajectory.Disperse.WeaponToAllies=false        ; boolean
Trajectory.Disperse.WeaponDoRepeat=false        ; boolean
```

```{note}
- Make sure you set a low `Trajectory.Disperse.RetargetRadius` value unless necessary. If you set `Trajectory.Disperse.WeaponRetarget=true`, also make sure you set `Trajectory.Disperse.Weapons` a low `Range` value unless necessary.
- This does not properly support `Arcing=true` projectiles.
- Note that the listed Weapons in `Trajectory.Disperse.Weapons` must be listed in `[WeaponTypes]` for them to work.
- `Trajectory.Disperse.Weapons` now not support customized `Bolt.ColorN`.
```


src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
```diff
@@ -0,0 +1,1379 @@
...
+
+	return true;
+}
+
+void DisperseTrajectory::OnUnlimbo(BulletClass* pBullet, CoordStruct* pCoord, BulletVelocity* pVelocity)
+{
+	auto const pType = this->GetTrajectoryType<DisperseTrajectoryType>(pBullet);
+
+	this->UniqueCurve = pType->UniqueCurve;
+	this->PreAimCoord = pType->PreAimCoord;
```
chaserli, line 259: I wonder if these can be done in its constructor, because OnUnlimbo is called right after that.

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
```diff
@@ -0,0 +1,1379 @@
...
+					const CoordStruct theTargetCoord = pTarget->GetCoords();
+					animIndex = static_cast<int>((Math::atan2(theTargetCoord.Y - theSourceCoord.Y , theTargetCoord.X - theSourceCoord.X) + Math::TwoPi + Math::Pi) * animCounts / Math::TwoPi - (animCounts / 8) + 0.5) % animCounts;
+				}
+			}
+			else
+			{
+				animIndex = ScenarioClass::Instance->Random.RandomRanged(0 , animCounts - 1);
+			}
+
+			if (AnimClass* const pAnim = GameCreate<AnimClass>(pWeapon->Anim[animIndex], pBullet->Location))
```
chaserli, line 1298: sanity check only on the type. In GameCreate, allocation failure leads directly to game crash.

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
```diff
@@ -0,0 +1,1379 @@
...
+		}
+	}
+	else
+	{
+		return;
+	}
+
+	if (pWeapon->Report.Count > 0)
+	{
+		const int reportIndex = pWeapon->Report.GetItem(ScenarioClass::Instance->Random.RandomRanged(0 , pWeapon->Report.Count - 1));
```
chaserli, line 1312: Sound effects use the other rng

chaserli (changes requested):
This one is fully decoupled and can/should go first among all these 3 new trajectories. I might have a few concerns about potential overhead in BulletRetargetTechno but it should be fine in game.

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
CrimRecya: I can only say that I am not sure about this. But thousands of games have proven that this is not necessary, and this is also what the sample trajectory template tells me.

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
CrimRecya: I'm sorry I didn't understand what you meant. Can you clearly tell me what needs to be modified?

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
CrimRecya: What does this need to be replaced with?

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
CrimRecya: I have tried to make modifications to it. If I misunderstood your meaning, please point it out.

chaserli: I mean you don't need to check if result of GameCreate is null

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
chaserli: Check how the game handles weapon reports, it's not this rng. Using this might lead to desync, dunno

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
CrimRecya: Get it. Changed.

src/Ext/Bullet/Trajectories/DisperseTrajectory.cpp
CrimRecya: I tried using `template <typename T> void Serialize(T& Stm);` like other ext classes, Although there were no issues with using it in `Load`, there were red wavy lines that I couldn't solve when using it in `Save`. :(

CrimRecya: May I ask how I should proceed?

src/Ext/Bullet/Trajectories/DisperseTrajectory.h
```diff
@@ -0,0 +1,248 @@
...
+	Valueable<bool> WeaponLocation;
+	Valueable<bool> WeaponTendency;
+	Valueable<bool> WeaponToAllies;
+	Valueable<bool> WeaponToGround;
+};
+
+class DisperseTrajectory final : public PhobosTrajectory
+{
+public:
+	DisperseTrajectory() : PhobosTrajectory(TrajectoryFlag::Disperse)
```
chaserli, line 88: This one is used for s/l

src/Ext/Bullet/Trajectories/DisperseTrajectory.h
```diff
@@ -0,0 +1,248 @@
...
+		, CurrentBurst { 0 }
+		, ThisWeaponIndex { 0 }
+		, LastTargetCoord {}
+		, PreAimDistance { 0 }
+		, LastReviseMult { 0 }
+		, FirepowerMult { 1.0 }
+	{}
+
+	DisperseTrajectory(PhobosTrajectoryType const* pType) : PhobosTrajectory(TrajectoryFlag::Disperse)
+		, UniqueCurve { false }
```
chaserli, line 135: and this one is used for creation in CreateInsanstance which is right before Unlimbo, so you can initialize the values here with pType

chaserli (changes requested):
If I understand this correctly

src/Ext/Bullet/Trajectories/DisperseTrajectory.h
CrimRecya: Thank you so much. I'll have a try.

chaserli: it's a constructor, you know for sure your pType is DisperseTrajectoryType const*. Just do it in those {}

btw I've updated the boilerplate

CrimRecya: Get it.

chaserli: not like that. The fact is no one touched these boilerplates until you do. You could have just think for yourself and don't have to follow the wrong example. The boilerplates were not entirely handy. It leads to your confusing impl here, for example:

* You copied lots of Type properties into trajectory, are they ever changed? if not then why not just copy the type instance or pointer?
* You passed PhobosTrajectoryType* as arg for ctor rather than DisperseTrajectoryType*, for no reason
* You initialized all those properties and then reset again in ctor, for no reason

If you find something global was written in a strange way then you can just change it, provided that you understand what it does

Metadorius: I feel like you're being a bit too harsh, how can a newcomer figure how it is supposed to work if all they have is a not very correct example?

chaserli: > I feel like you're being a bit too harsh, how can a newcomer figure how it is supposed to work if all they have is a not very correct example?

I just updated the boilerplate again, but as expected that's far from enough and should require a lot of work on it. Problem is neither of us maintainers has time and energy for it.
So I think if @CrimRecya keep working on these he could try fixing them first before proceed.
For example, I demoed how to fix type ptr swizzle in a236394. There's a lot more to do. At the end of the day raw traj(type) pointers should be no longer in use.

TaranDahl:
I use this as a substitute for the ROT projectile. It performs well in local tests and solves many problems caused by the design defects of the ROT projectile.

CrimRecya:
Has been completely rewrited.

PR #1299, opened 2024-07-07
Allow vehicles to retain target on move command

Starkku:
### Retain target on movement command

- It is now possible for vehicles to retain their target when issued movement command by setting `KeepTargetOnMove` to true.
  - Note that no check is done whether or not the vehicle or the weapon can actually fire while moving, this is on modder's discretion.
  - The target is automatically reset if the vehicle moves beyond the weapon's range from the target.
- `KeepTargetOnMove.ExtraDistance` can be used to modify the distance considered 'out of range' from target (it is added to weapon range), negative values work to reduce the distance.

In `rulesmd.ini`:
```ini
[SOMEVEHICLE]                     ; VehicleType
KeepTargetOnMove=false            ; boolean
KeepTargetOnMove.ExtraDistance=0  ; floating point value, distance in cells
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -253,6 +253,7 @@ This page lists all the individual contributions to the project by their author.
   - EM Pulse cannon logic improvements
   - `<Player @ X>` as owner for pre-placed objects
   - Custom exit cell for infantry factory
+  - Vehicles keeping target on move command
 - **Morton (MortonPL)**:
   - `XDrawOffset` for animations
   - Shield passthrough & absorption
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -1226,6 +1226,20 @@ IronCurtain.KeptOnDeploy=       ; boolean, default to [CombatDamage]->IronCurtai
 ForceShield.KeptOnDeploy=       ; boolean, default to [CombatDamage]->ForceShield.KeptOnDeploy
 ```
 
+### Retain target on movement command
+
+- It is now possible for vehicles to retain their target when issued movement command by setting `KeepTargetOnMove` to true.
+  - Note that no check is done whether or not the vehicle or the weapon can actually fire while moving, this is on modder's discretion.
+  - The target is automatically reset if the vehicle moves beyond the weapon's range from the target.
+- `KeepTargetOnMove.ExtraDistance` can be used to modify the distance considered 'out of range' from target (it is added to weapon range), negative values work to reduce the distance.
+
+In `rulesmd.ini`:
+```ini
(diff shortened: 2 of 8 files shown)


TaranDahl:
Very useful improvement.
Adding a global version tag in [General] would be better I think, as the flags may be used by many UnitTypes.

mevitar:
I had a weird instance of a subterranean harvester getting stuck underground and able to move around, but never back up to the ground.
And after a while, the game crashed. Here's the link:
https://drive.google.com/file/d/1Zbul8kBow3Q-zuzHt__L86NkuqHpqYLJ

mevitar:
https://drive.google.com/file/d/111p6BCPOZ2Js3qFYof9DcgTjbJOj1iYm
Another crash in this newest build available. No idea what's the reason.
The except.txt looks very similar (3 topmost elements on the stack are the same).

mevitar:
Besides these crashes that might or might not be related, both tags seem to work fine.

CrimRecya (approved):
LGTM. And I didn't encounter any problems during the testing either.

Metadorius:
> Adding a global version tag in [General] would be better I think, as the flags may be used by many UnitTypes.

@Starkku what do you think?

Speederovsky:
This was added with designing specific units in mind, not for general application - think units like Cyclone in SC2. Giving this to _all globally_ would very likely not improve general unit behavior at all and annoy many players when a lot of imprecise "target out of range" situations happens and combat gets interrupted. But you can try, though definitely not _instead_ of the unit-specific tag.

Metadorius:
Yeah in no way that is going to be "instead".

ZivDero:
> Yeah in no way that is going to be "instead".

This sounds like a worse version of queue move.

Metadorius:
> > Yeah in no way that is going to be "instead".
> 
> This sounds like a worse version of queue move.

image
*Waiting for the day ZivDero restores proper Q-Move*

ZivDero:
> > > Yeah in no way that is going to be "instead".
> > 
> > 
> > This sounds like a worse version of queue move.
> 
> image _Waiting for the day ZivDero restores proper Q-Move_

As far as I can see, it's a matter of 2 hooks, each maybe a couple lines long. I can instruct you on it :P

Starkku:
This was pretty much explicitly trying to imitate Terran Cyclone's behaviour from SC2 rather than aiming for any sort of general unit behaviour improvement just like @Speederovsky stated so I am somewhat hesitant to add the global control and consider it to be a poor idea but if there are people who insist on it I guess I could.

Metadorius:
No global control it is then.

PR #1302, opened 2024-07-07
Airburst/Splits reimplementation

Starkku:
TL;DR - Core differences here to Ares' implementation:
- Projectiles created from `AirburstWeapon` remember their WeaponType and thus can apply radiation.
- Previously hardcoded values are now customizable via `RetargetSelf.Probability`, `Splits.TargetingDistance` and `Splits.TargetCellRange`.
- More elaborate target filtering can be enabled for `Splits` via `Splits.UseWeaponTargeting`.
- `Airburst` can now respect `Cluster` by setting `Airburst.UseCluster`. `Airburst.RandomClusters` controls if they're spread randomly or evenly across the affected area.
- Firepower multipliers from the original firer of the weapon can be applied on `AirburstWeapon` by setting `AirburstWeapon.ApplyFirepowerMult`.

### Airburst & Splits

- `AirburstWeapon` logic has been reimplemented and thus there are several additions & changes to it.
- `Splits` can be set to true to use projectile splitting logic from Firestorm, with the number of split projectiles defined by `Cluster`.
  - `RetargetAccuracy` defines the probability that the splitted projectiles head to the same target as the original projectile.
  - `RetargetSelf` determines if it is possible for the splitted projectiles to aim at the firer of the original projectile.
    - `RetargetSelf.Probability` is the probability that if the original firer is chosen as a target, it is kept as the target instead of rerolled to another.
  - `Splits.TargetingDistance` is the distance in cells that any potential target has to be within from the original target coordinates to be eligible for targeting by the splitted projectiles.
  - `Splits.TargetCellRange` is the distance in whole cells from the original target cell from which the splitted projectiles can pick new target cells if not enough TechnoType targets were found nearby.
  - `Splits.UseWeaponTargeting`, if set to true, enables weapon targeting filter for when checking targets for splitted projectiles. Target's `LegalTarget` setting, Warhead `Verses` against `Armor` as well as `AirburstWeapon` weapon targeting filters & AttachEffect filters will be checked.
    - Do note that this overrides checking Warhead for `AffectsAllies/Owner/Enemies` for targeting. You can use `CanTargetHouses` on `AirburstWeapon` to achieve similar behaviour, however.
- Behaviour for if `Airburst` is set to true can also be customized.
  - `AirburstSpread` is the distance in cells that the effect covers, with each cell in range being targeted by `AirburstWeapon` by default.
  - `Airburst.UseCluster`, if set to true, makes it so that only number of cells in the affected area dictated by `Cluster` will be affected, instead of all of them.
    - If `Airburst.RandomClusters` is set to true, the cells affected will be picked by random. Otherwise they will be evenly spaced (counting from center to edges of affected area).
- `AroundTarget` controls whether or not targets for projectiles created by `Airburst` or `Splits` are checked for in area around the original projectile's intended target, or where the original projectile detonated. Defaults to value of `Splits`.
- `AirburstWeapon.ApplyFirepowerMult` determines whether or not firepower modifiers from the firer of the original projectile are applied on the projectiles created from `AirburstWeapon`.

In `rulesmd.ini`:
```ini
[SOMEPROJECTILE]                         ; Projectile
Splits=                                  ; boolean
RetargetAccuracy=0.0                     ; floating point value, percents or absolute (0.0-1.0)
RetargetSelf=true                        ; boolean
RetargetSelf.Probability=0.5             ; floating point value, percents or absolute (0.0-1.0)
Splits.TargetingDistance=5.0             ; floating point value, distance in cells
Splits.TargetCellRange=3                 ; integer, cell offset
Splits.UseWeaponTargeting=false          ; boolean
AirburstSpread=1.5                       ; floating point value, distance in cells
Airburst.UseCluster=false                ; boolean
Airburst.RandomClusters=false            ; boolean
AroundTarget=                            ; boolean
AirburstWeapon.ApplyFirepowerMult=false  ; boolean
```

**Note:**
`Splits`, `AirburstSpread`, `RetargetAccuracy`, `RetargetSelf` and `AroundTarget`, beyond the other additions, should function similarly to the equivalent features introduced by Ares and take precedence over them if Phobos is used together with Ares.

### DIFF
--- a/src/Ext/Bullet/Hooks.DetonateLogics.cpp
+++ b/src/Ext/Bullet/Hooks.DetonateLogics.cpp
@@ -364,3 +364,231 @@ DEFINE_HOOK(0x469AA4, BulletClass_Logics_Extras, 0x5)
 
 	return 0;
 }
+
+#pragma region Airburst
+
+static bool IsAllowedSplitsTarget(TechnoClass* pSource, HouseClass* pOwner, WeaponTypeClass* pWeapon, TechnoClass* pTarget, bool useWeaponTargeting)
+{
+	auto const pWH = pWeapon->Warhead;
+
+	if (useWeaponTargeting)
+	{
+		auto const pType = pTarget->GetTechnoType();
+
+		if (!pType->LegalTarget || GeneralUtils::GetWarheadVersusArmor(pWH, pType->Armor) == 0.0)
+			return false;
+
+		auto const pWeaponExt = WeaponTypeExt::ExtMap.Find(pWeapon);
+
+		if (!EnumFunctions::CanTargetHouse(pWeaponExt->CanTargetHouses, pOwner, pTarget->Owner)
+			|| !EnumFunctions::IsCellEligible(pTarget->GetCell(), pWeaponExt->CanTarget, true, true)
+			|| !EnumFunctions::IsTechnoEligible(pTarget, pWeaponExt->CanTarget))
+		{
+			return false;
+		}
+
+		if (!pWeaponExt->HasRequiredAttachedEffects(pTarget, pSource))
+			return false;
+	}
+	else
+	{
+		if (!WarheadTypeExt::ExtMap.Find(pWH)->CanTargetHouse(pOwner, pTarget))
+			return false;
+	}
+
+	return true;
+}
+
+// Disable Ares' Airburst implementation.
+DEFINE_PATCH(0x469EBA, 0x90, 0x90, 0x90, 0x90, 0x90, 0x90);
+
+DEFINE_HOOK(0x468EB3, BulletClass_Explodes_AirburstCheck1, 0x6)
+{
+	enum { Continue = 0x468EC7, Skip = 0x468FF4 };
+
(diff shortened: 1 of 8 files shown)


src/Ext/Bullet/Hooks.DetonateLogics.cpp
```diff
@@ -365,3 +365,230 @@ DEFINE_HOOK(0x46A290, BulletClass_Logics_Extras, 0x5)
...
+				targets.RemoveItem(index);
+			}
+
+			if (pTarget)
+			{
+
+				if (auto const pBullet = pTypeSplits->CreateBullet(pTarget, pThis->Owner, damage, pWeapon->Warhead, pWeapon->Speed, pWeapon->Bright))
+				{
+					pBullet->WeaponType = pWeapon;
+					pBullet->Range = projectileRange;
```
CrimRecya, line 569: In order to inherit the bullet's `house` after its firer death normally, what about add
```
BulletExt::ExtMap.Find(pBullet)->FirerHouse = pOwner;
```
here.

PR #1304, opened 2024-07-11
Animatable template & animation transparency customization settings

Starkku:
Split from #1300 

-------------

### Customizable animation transparency settings

- `Translucency.Cloaked` can be used to override `Translucency` on animations attached to currently cloaked TechnoTypes.
- Both `Translucency` and `Translucency.Cloaked` can use the new keyframe system to animate along with the animation. Read more about the keyframe system below.
- If interpolation is enabled, the keyframe values are clamped to valid transparency values (0,25,50 and 75), e.g a value of 1.5 would become 0 and 56.525 would become 50 and so on.

In `artmd.ini`:
```ini
[SOMEANIM]             ; AnimationType
Translucency=0         ; integer - only accepted values are 75, 50, 25 and 0.
Translucency.Cloaked=  ; integer - only accepted values are 75, 50, 25 and 0.
```

----------------

### Keyframe animations (Animatable template)

- Some features use keyframe-based animation system to define animations in INI. Defined in INI it looks something like following.

```ini
[SOMESECTION]
BASEKEY.KeyframeN.Value=        ; Key-dependant value type
BASEKEY.KeyframeN.Percentage=   ; floating point value, percents or absolute
BASEKEY.KeyframeN.Absolute=     ; integer, zero-based frame index
BASEKEY.Interpolation=none      ; Interpolation mode (none|linear)
BASEKEY.ResetData=false         ; boolean
```

- `BASEKEY` is whatever base key name the feature in question may use. `N` is zero-based keyframe index. If no keyframes are defined, a single value from `BASEKEY` is attempted to be parsed instead.
  - `Value` is a key/feature-dependant value type associated with that keyframe.
  - `Percentage` is the percentage through the animation's frames where the keyframe becomes active. It is also possible to instead use zero-based frame index via `Absolute` which takes precedence over percentage, albeit it is internally converted to a percentage value.
  - `Interpolation` controls interpolation of values between keyframes. The behaviour here may depend on the value type in use, as not all value types may be interpolatable well or at all.
  - `ResetData` if set to true makes it so that all existing keyframe data is reset before parsing. Can be used to reset keyframes when redefining them in map files etc.

```{note}
Keyframes are expected to be defined with no duplicates for Percentage or Absolute. Failure to do so will crash the game and output developer warnings about offending keys to the log.
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -266,6 +266,7 @@ This page lists all the individual contributions to the project by their author.
   - Customizing effect of level lighting on air units
   - Reimplemented `Airburst` & `Splits` logic with more customization options
   - Buildings considered as destroyable pathfinding obstacles
+  - Animation transparency customization settings
   - Animation visibility customization settings
   - Light effect customizations
   - Building unit repair customizations
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -53,6 +53,7 @@ This page lists all the individual contributions to the project by their author.
   - Ported XNA CnCNet Client MP save handling
   - Retint fix toggle
   - Voxel drawing invisible sections skip
+  - Animatable template
 - **Uranusian (Thrifinesma)**:
   - Mind Control enhancement
   - Custom warhead splash list
@@ -297,6 +298,7 @@ This page lists all the individual contributions to the project by their author.
   - Armed building guard mission retry delay customization
   - Building turret idle/firing/low power animations
   - Animation theater/tile palette toggle
+  - Animatable template
 - **Morton (MortonPL)**:
   - `XDrawOffset` for animations
   - Shield passthrough & absorption
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -577,24 +577,19 @@ AttachedSystem=  ; ParticleSystemType
 ### Customizable animation transparency settings
 
 - `Translucency.Cloaked` can be used to override `Translucency` on animations attached to currently cloaked TechnoTypes.
-- Properties of the three transparency stages used by `Translucent=true`can now be customized per animation type.
-  - `Translucent.StageX.Percent` controls at what percentage through the animation's frames the applicable translucency stage becomes active.
-    - If `Translucent.StageX.Frame` is set, this explicit frame value is used instead.
-  - `Translucent.StageX.Translucency` controls the transparency level for the stage.
+- `Translucent=true` animated transparency is now fully controllable via new keyframe settings.
+  - Keyframes are defined via `Translucent.KeyframeN.*` settings where N is zero-based keyframe index.
+  - `Translucent.KeyframeN.Value` is keyframe's transparency value.
+  - `Translucent.KeyframeN.Percentage` is the percentage through the animation's frames where the keyframes becomes active. It is also possible to instead use zero-based frame index via `Translucent.KeyframeN.Absolute` which takes precedence over percentage.
+  - Keyframes should be listed in ascending order with no duplicate percentages / frames across different keyframes. Failure to do so can produce unpredictable results.
 
 In `artmd.ini`:
 ```ini
 [SOMEANIM]                          ; AnimationType
 Translucency.Cloaked=               ; integer - only accepted values are 75, 50, 25 and 0.
-Translucent.Stage1.Percent=0.2      ; floating point value, percents or absolute
-Translucent.Stage1.Frame=           ; integer, 0-based frame index
-Translucent.Stage1.Translucency=25  ; integer - only accepted values are 75, 50, 25 and 0.
-Translucent.Stage2.Percent=0.4      ; floating point value, percents or absolute
-Translucent.Stage2.Frame=           ; integer, 0-based frame index
-Translucent.Stage2.Translucency=50  ; integer - only accepted values are 75, 50, 25 and 0.
-Translucent.Stage3.Percent=0.6      ; floating point value, percents or absolute
-Translucent.Stage3.Frame=           ; integer, 0-based frame index
-Translucent.Stage3.Translucency=75  ; integer - only accepted values are 75, 50, 25 and 0.
+Translucent.KeyframeN.Value=        ; integer - only accepted values are 75, 50, 25 and 0.
+Translucent.KeyframeN.Percentage=   ; floating point value, percents or absolute
+Translucent.KeyframeN.Absolute=     ; integer, zero-based frame index
 ```
 
 ### Customizable animation visibility settings
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -581,7 +581,6 @@ AttachedSystem=  ; ParticleSystemType
   - Keyframes are defined via `Translucent.KeyframeN.*` settings where N is zero-based keyframe index.
   - `Translucent.KeyframeN.Value` is keyframe's transparency value.
   - `Translucent.KeyframeN.Percentage` is the percentage through the animation's frames where the keyframes becomes active. It is also possible to instead use zero-based frame index via `Translucent.KeyframeN.Absolute` which takes precedence over percentage.
-  - Keyframes should be listed in ascending order with no duplicate percentages / frames across different keyframes. Failure to do so can produce unpredictable results.
   - `Translucent.Interpolation` controls interpolation of values between keyframes. These values are still clamped to the allowed transparency values of 0,25,50 and 75.
 
 In `artmd.ini`:
@@ -592,6 +591,10 @@ Translucent.KeyframeN.Value=        ; integer - only accepted values are 75, 50,
 Translucent.KeyframeN.Percentage=   ; floating point value, percents or absolute
 Translucent.KeyframeN.Absolute=     ; integer, zero-based frame index
 Translucent.Interpolation=none      ; Interpolation mode (none|linear)
+```
+
+```{note}
+Keyframes are expected to be defined in ascending order with no duplicates. Failure to do so will crash the game and output developer warnings about offending keys to the log.
 ```
 
 ### Customizable animation visibility settings
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -574,6 +574,29 @@ In `artmd.ini`:
 AttachedSystem=  ; ParticleSystemType
 ```
 
+### Customizable animation transparency settings
+
+- `Translucency.Cloaked` can be used to override `Translucency` on animations attached to currently cloaked TechnoTypes.
+- Properties of the three transparency stages used by `Translucent=true`can now be customized per animation type.
+  - `Translucent.StageX.Percent` controls at what percentage through the animation's frames the applicable translucency stage becomes active.
+    - If `Translucent.StageX.Frame` is set, this explicit frame value is used instead.
+  - `Translucent.StageX.Translucency` controls the transparency level for the stage.
+
+In `artmd.ini`:
+```ini
+[SOMEANIM]                          ; AnimationType
+Translucency.Cloaked=               ; integer - only accepted values are 75, 50, 25 and 0.
+Translucent.Stage1.Percent=0.2      ; floating point value, percents or absolute
+Translucent.Stage1.Frame=           ; integer, 0-based frame index
+Translucent.Stage1.Translucency=25  ; integer - only accepted values are 75, 50, 25 and 0.
+Translucent.Stage2.Percent=0.4      ; floating point value, percents or absolute
+Translucent.Stage2.Frame=           ; integer, 0-based frame index
+Translucent.Stage2.Translucency=50  ; integer - only accepted values are 75, 50, 25 and 0.
+Translucent.Stage3.Percent=0.6      ; floating point value, percents or absolute
+Translucent.Stage3.Frame=           ; integer, 0-based frame index
+Translucent.Stage3.Translucency=75  ; integer - only accepted values are 75, 50, 25 and 0.
+```
+
 ### Customizable animation visibility settings
 
 - It is now possible to customize which players can see an animation using `VisibleTo`.
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -582,6 +582,7 @@ AttachedSystem=  ; ParticleSystemType
   - `Translucent.KeyframeN.Value` is keyframe's transparency value.
   - `Translucent.KeyframeN.Percentage` is the percentage through the animation's frames where the keyframes becomes active. It is also possible to instead use zero-based frame index via `Translucent.KeyframeN.Absolute` which takes precedence over percentage.
   - Keyframes should be listed in ascending order with no duplicate percentages / frames across different keyframes. Failure to do so can produce unpredictable results.
+  - `Translucent.Interpolation` controls interpolation of values between keyframes. These values are still clamped to the allowed transparency values of 0,25,50 and 75.
 
 In `artmd.ini`:
 ```ini
@@ -590,6 +591,7 @@ Translucency.Cloaked=               ; integer - only accepted values are 75, 50,
 Translucent.KeyframeN.Value=        ; integer - only accepted values are 75, 50, 25 and 0.
 Translucent.KeyframeN.Percentage=   ; floating point value, percents or absolute
 Translucent.KeyframeN.Absolute=     ; integer, zero-based frame index
+Translucent.Interpolation=none      ; Interpolation mode (none|linear)
 ```
 
 ### Customizable animation visibility settings
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -577,11 +577,8 @@ AttachedSystem=  ; ParticleSystemType
 ### Customizable animation transparency settings
 
 - `Translucency.Cloaked` can be used to override `Translucency` on animations attached to currently cloaked TechnoTypes.
-- `Translucent=true` animated transparency is now fully controllable via new keyframe settings.
-  - Keyframes are defined via `Translucent.KeyframeN.*` settings where N is zero-based keyframe index.
-  - `Translucent.KeyframeN.Value` is keyframe's transparency value.
-  - `Translucent.KeyframeN.Percentage` is the percentage through the animation's frames where the keyframes becomes active. It is also possible to instead use zero-based frame index via `Translucent.KeyframeN.Absolute` which takes precedence over percentage.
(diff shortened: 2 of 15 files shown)


Starkku:
Now uses @Metadorius's `Animatable` template that allows defining arbitrary keyframe animations in INI.

Might require testing to see if there's a noticeable performance overhead from the template.

src/Utilities/TemplateDef.h
```diff
@@ -1400,8 +1424,70 @@ if(_strcmpi(parser.value(), #name) == 0){ value = __uuidof(name ## LocomotionCla
...
+			result = ColorStruct { r, g, b };
+			break;
+		}
+		default:
+			break;
+		}
+
+		return result;
+	}
+}
```
Metadorius, line 1490: I think you can collapse all that by defining a primitive interpolation or two, say, on double, and routing all other specializations to that. Although I would probably make interpolation mode a template param (not sure on that?) and write different functions per each interpolation, and have a single function that dispatches stuff. Could also make it a separate file for clarity, this file is getting quite big.

src/Ext/Anim/Hooks.cpp
```diff
@@ -329,6 +329,84 @@ DEFINE_HOOK(0x4232BF, AnimClass_DrawIt_MakeInfantry, 0x6)
...
+		if (translucencyLevel >= 15)
+			return ReturnFromFunction;
+
+		auto const pTypeExt = AnimTypeExt::ExtMap.Find(pType);
+		int currentFrame = pThis->Animation.Value;
+		int frames = pType->End;
+
+		if (pTypeExt->Translucent_Keyframes.KeyframeData.size() > 0)
+		{
+			flags |= pTypeExt->Translucent_Keyframes.Get(static_cast<double>(currentFrame) / frames);
```
Metadorius, line 393: Isn't there a timer somewhere that will happily already give you a percentage?

src/Ext/Anim/Hooks.cpp
```diff
@@ -329,6 +329,84 @@ DEFINE_HOOK(0x4232BF, AnimClass_DrawIt_MakeInfantry, 0x6)
...
+		}
+		else
+		{
+			if (currentFrame > frames * 0.6)
+				flags |= BlitterFlags::TransLucent75;
+			else if (currentFrame > frames * 0.4)
+				flags |= BlitterFlags::TransLucent50;
+			else if (currentFrame > frames * 0.2)
+				flags |= BlitterFlags::TransLucent25;
```
Metadorius, line 402: Not sure if we should hardcode that?

Starkku: This is the default `Translucent` behaviour used if the keyframes are not in place.

src/Ext/Anim/Hooks.cpp
Starkku: No. `StageClass` deals with integer increments.

src/Ext/Anim/Hooks.cpp
```diff
@@ -329,6 +329,84 @@ DEFINE_HOOK(0x4232BF, AnimClass_DrawIt_MakeInfantry, 0x6)
...
+		}
+
+		if (translucency <= 0)
+		{
+			if (translucencyLevel)
+			{
+				if (translucencyLevel > 15)
+					return ReturnFromFunction;
+				else if (translucencyLevel > 10)
+					flags |= BlitterFlags::TransLucent50;
```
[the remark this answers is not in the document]
Starkku: No, not 75.

docs/New-or-Enhanced-Logics.md
```diff
@@ -470,6 +470,24 @@ In `artmd.ini`:
...
 ```
 
+### Customizable animation transparency settings
+
+- `Translucency.Cloaked` can be used to override `Translucency` on animations attached to currently cloaked TechnoTypes.
+- `Translucent=true` animated transparency is now fully controllable via new keyframe settings.
+  - Keyframes are defined via `Translucent.KeyframeN.*` settings where N is zero-based keyframe index.
+  - `Translucent.KeyframeN.Value` is keyframe's transparency value.
+  - `Translucent.KeyframeN.Percentage` is the percentage through the animation's frames where the keyframes becomes active. It is also possible to instead use zero-based frame index via `Translucent.KeyframeN.Absolute` which takes precedence over percentage.
+  - Keyframes should be listed in ascending order with no duplicate percentages / frames across different keyframes. Failure to do so can produce unpredictable results.
```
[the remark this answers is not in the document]
Starkku: Sure if you can think of a way to check for this without adding another massive layer of complexity.

src/Ext/Anim/Hooks.cpp
```diff
@@ -329,6 +329,84 @@ DEFINE_HOOK(0x4232BF, AnimClass_DrawIt_MakeInfantry, 0x6)
...
+			else if (translucency == 50)
+				flags |= BlitterFlags::TransLucent50;
+			else if (translucency == 25)
+				flags |= BlitterFlags::TransLucent25;
+		}
+	}
+	else
+	{
+		if (translucencyLevel >= 15)
+			return ReturnFromFunction;
```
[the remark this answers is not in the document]
Starkku: Reverse-engineered game code. `translucencyLevel` is some sort of stage/timer thing set on building anims when they need to be drawn at certain transparency when they are normally not, contrary to the YRpp description seems to go from 0 to 15 and is directly copied from the parent building's field `0x6ED`. Rest is just mapping the `Translucency` integer to correct blitter flags.

Perhaps, the translucency stages are basically self-explanatory though and this is probably the only place where whatever the shit `translucencyLevel` is supposed to mean is used.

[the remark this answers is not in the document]
Metadorius: could leave a comment where stuff is reversed and where custom additions start and we're gucci I htink?

src/Ext/Anim/Hooks.cpp
Metadorius: oh, a comment explaining that would be good in this case

src/Utilities/TemplateDef.h
Starkku: Honestly I question the need for interpolation modes beyond none and linear in the first place anyway.

Metadorius: It depends on the use case. Although I can agree that the specific use cases I am thinking of are rather niche.

What I had in mind was a buildup anim/deploy techno attachment FLH translation, for instance, or maybe just movable FLH/TurretOffset for animated units.

Metadorius: It is probably still good to lay out the ground for easier extensibility while we're at it.

Starkku: > I think you can collapse all that by defining a primitive interpolation or two, say, on double, and routing all other specializations to that

This is heavily case-specific. It works for numeric values like integers (can just truncate the double result) but pretty much anything else does not work like that. ColorStruct, one of the examples I added in wouldn't work like this for an example - you can't just create the color data out of thin air from a floating point value.

docs/New-or-Enhanced-Logics.md
```diff
@@ -476,6 +476,22 @@ In `artmd.ini`:
...
+  - If interpolation is enabled, the keyframe values are clamped to valid transparency values (0,25,50 and 75), e.g a value of 1.5 would become 0 and 56.525 would become 50 and so on.
+
+In `artmd.ini`:
+```ini
+[SOMEANIM]                          ; AnimationType
+Translucency.Cloaked=               ; integer - only accepted values are 75, 50, 25 and 0.
+Translucent.KeyframeN.Value=        ; integer - only accepted values are 75, 50, 25 and 0.
+Translucent.KeyframeN.Percentage=   ; floating point value, percents or absolute
+Translucent.KeyframeN.Absolute=     ; integer, zero-based frame index
+Translucent.Interpolation=none      ; Interpolation mode (none|linear)
```
Metadorius, line 492:
```suggestion
Translucent.KeyframeN.Value=        ; integer - only accepted values are 75, 50, 25 and 0.
Translucent                         ; animatable translucency value - only accepted values are 75, 50, 25 and 0.
```

This way we only need to update one place when we update the template. Also need to add a link to animatable template description.

Accepting ideas on better syntax in the code block.

src/Utilities/TemplateDef.h
```diff
@@ -1734,3 +1757,210 @@ bool Damageable<T>::Save(PhobosStreamWriter& Stm) const
...
+requires MultiflagReadable<T, TExtraArgs...>
+void __declspec(noinline) MultiflagNullableVector<T, TExtraArgs...>::Read(INI_EX& parser, const char* const pSection, const char* const pBaseFlag, TExtraArgs&... extraArgs)
+{
+	char flagName[0x40];
+	for (size_t i = 0; ; ++i)
+	{
+		T dataEntry {};
+
+		// we expect %d for array number then %s for the subflag name, so we replace %s with itself (but escaped)
+		_snprintf_s(flagName, sizeof(flagName), pBaseFlag, i, "%s");
```
Metadorius, line 1802: I am pretty sure I used `%%` somewhere around here? Just in case, it was not a typo, it's just a way to escape the `%` character because we do formatting twice and need it in the second formatting pass.

CREDITS.md
```diff
@@ -245,6 +245,7 @@ This page lists all the individual contributions to the project by their author.
   - Customizing effect of level lighting on air units
   - Reimplemented `Airburst` & `Splits` logic with more customization options
   - Buildings considered as destroyable pathfinding obstacles
+  - Animation transparency customization settings
```
Metadorius, line 248: Could you add us as authors of animatable template too?

src/Utilities/TemplateDef.h
Starkku: Yeah but it didn't actually work

Metadorius: :isothonk:

That's weird because the docs say it's the correct way to escape... oh well

Starkku: You don't need to escape it in the args, only in the format specifier if you actually want it to include % as a character. Having double % in the args was causing it to print it twice in the buffer so parsing didn't work.

Starkku:
Updated credits & docs, ready for another review pass and/or further testing if needed, if not then merge.

src/Ext/Anim/Hooks.cpp
```diff
@@ -469,6 +469,90 @@ DEFINE_HOOK(0x423061, AnimClass_DrawIt_Visibility, 0x6)
...
+
+		auto const pTypeExt = AnimTypeExt::ExtMap.Find(pType);
+		int currentFrame = pThis->Animation.Value;
+		int frames = pType->End;
+
+		// New addition: Keyframeable Translucent stages.
+		if (pTypeExt->Translucent_Keyframes.KeyframeData.size() > 0)
```
Metadorius: I do think we should make a shortcut function to check whether the template values are set

src/Utilities/TemplateDef.h
```diff
@@ -1807,3 +1830,210 @@ bool Damageable<T>::Save(PhobosStreamWriter& Stm) const
...
+{
+	return Savegame::WritePhobosStream(Stm, this->Percentage)
+		&& Savegame::WritePhobosStream(Stm, this->Value);
+}
+
+template <typename TValue>
+TValue Animatable<TValue>::Get(double const percentage) const noexcept
+{
+	// This currently assumes the keyframes are ordered and there are no duplicates for same frame/percentage.
+	// Thing is still far from lightweight as searching for the correct items requires going through the vector.
```
Metadorius: I do think making a volatile cache LUT on the fly based on results of calculation would be a smart idea, just like voxel caching is done.

Otherwise could add bisection to optimize.

src/Utilities/TemplateDef.h
```diff
@@ -1807,3 +1830,210 @@ bool Damageable<T>::Save(PhobosStreamWriter& Stm) const
 		&& Savegame::WritePhobosStream(Stm, this->ConditionYellow)
 		&& Savegame::WritePhobosStream(Stm, this->ConditionRed);
 }
+
+// MultiflagValueableVector
+
+template<typename T, typename... TExtraArgs>
+requires MultiflagReadable<T, TExtraArgs...>
+void __declspec(noinline) MultiflagValueableVector<T, TExtraArgs...>::Read(INI_EX& parser, const char* const pSection, const char* const pBaseFlag, TExtraArgs&... extraArgs)
```
Metadorius: I wonder whether it is a good idea to add a `.ResetData=true/false` subflag (not serialized, just controls the behavior) here for overriding in map files? If not set - list entries replace the old ones one by one, if set - keyframes are re-set and it's read anew.

src/Utilities/TemplateDef.h
```diff
@@ -1807,3 +1830,210 @@ bool Damageable<T>::Save(PhobosStreamWriter& Stm) const
...
+}
+
+template <typename TValue>
+void __declspec(noinline) Animatable<TValue>::Read(INI_EX& parser, const char* const pSection, const char* const pBaseFlag, absolute_length_t absoluteLength)
+{
+	char flagName[0x40];
+
+	// we expect "BaseFlagName.%s" here
+	_snprintf_s(flagName, sizeof(flagName), pBaseFlag, "Keyframe%d.%s");
+	this->KeyframeData.Read(parser, pSection, flagName, absoluteLength);
```
Metadorius, line 2217: I propose to add a shortcut of `BaseFlag=value` reading if the `KeyframeData` is empty. It will populate it with 1 keyframe with no interpolation, effectively allowing to use the template the same as a simple `Valueable<T>` if user wishes, effectively making it a drop-in extension for the latter and allowing to animate anything.

This will also work nice with `.ResetData` subflag if the user wishes to redefine to a simple value.

src/Utilities/TemplateDef.h
```diff
@@ -1807,3 +1830,210 @@ bool Damageable<T>::Save(PhobosStreamWriter& Stm) const
...
+	char flagName[0x40];
+
+	Nullable<double> percentageTemp {};
+	Nullable<absolute_length_t> absoluteTemp {};
+
+	_snprintf_s(flagName, sizeof(flagName), pBaseFlag, "Percentage");
+	percentageTemp.Read(parser, pSection, flagName);
+
+	_snprintf_s(flagName, sizeof(flagName), pBaseFlag, "Absolute");
+	absoluteTemp.Read(parser, pSection, flagName);
```
Metadorius: We probably should check whether absoluteLength is not 0 and if it is 0 then not read absolute at all

Metadorius (changes requested):
@Starkku @ZivDero @CrimRecya What do you think of robustness of the design for the use of multiflag vectors with LaserTrailDataEntry, for example? And other multiflag vectors if you can think of any.

LGTM otherwise

src/Utilities/TemplateDef.h
Metadorius: Also remind me, what does the template return if there's no values? It probably should return the default for T?

src/Utilities/Interpolation.h
```diff
@@ -0,0 +1,64 @@
...
+
+#include <YRPPCore.h>
+#include "Enum.h"
+
+namespace detail
+{
+	template <typename T>
+	inline T interpolate(T& first, T& second, double percentage, InterpolationMode mode)
+	{
+		return first;
```
ZivDero, line 12: This needs to throw some kind of warning at the user so that they know they're missing an implementation.

Starkku:
Updated with feedback addressed.

The translucency customization code has been refactored a bit, instead of keyframes affecting `Translucent` it is now tied to `Translucency` to be more consistent with how the template works with animatable values. `Translucency.Cloaked` can also now be keyframed.

Metadorius (approved):
LGTM

src/Utilities/TemplateDef.h
```diff
@@ -1870,3 +1893,262 @@ bool Damageable<T>::Save(PhobosStreamWriter& Stm) const
...
+// MultiflagValueableVector
+
+template<typename T, typename... TExtraArgs>
+requires MultiflagReadable<T, TExtraArgs...>
+void __declspec(noinline) MultiflagValueableVector<T, TExtraArgs...>::Read(INI_EX& parser, const char* const pSection, const char* const pBaseFlag, TExtraArgs&... extraArgs)
+{
+	char flagName[0x40];
+	for (size_t i = 0; ; ++i)
+	{
+		T dataEntry {};
```
Metadorius: rules:
```ini
Some0.Bull=1
Some0.Shit=2
Some1.Bull=3
Some1.Shit=4
```

map:
```ini

Some1.Bull=0
```

expected result:
```ini
Some0.Bull=1
Some0.Shit=2
Some1.Bull=0
Some1.Shit=4
```

I think it currently will result in
```ini
Some0.Bull=1
Some0.Shit=2
Some1.Bull=0
Some1.Shit=<default>
```

which is not expected

I propose to change the vector to not create a new `T dataEntry {}` always but use the existing one if it exists, and to add `.ResetValue` but per each entry; the only thing it would do is force `T dataEntry {}` always.

Also should move the whole vector ResetValues here from Animatable

Metadorius (dismissed):

Starkku:
Implemented following changes:

- Rename `ResetData`  to `ResetValues` and move from `Animatable` to `MultiflagValueable/NullableVector`
- Make `MultiflagValueable/NullableVector` retain existing values on further parsing passes if unchanged unless explicitly reset.
- Clean up repetitive code in  `MultiflagValueable/NullableVector` with helper functions.
- Disallow < 0 values for `KeyFrameData.Percentage `, throw fatal error on parsing if present.
- Allow parsing `KeyFrameData` on further parsing passes if `Percentage` was parsed on previous pass but not on current.
- Fix base flag name getting cut on the Animatable parsing error fatal error message box.
- Place sorted keyframes in separate vector from parsed values, recalculate on every parsing pass to ensure data correctness.

Note about the sorting change: I did explore other solutions but they basically boiled down to the following and every single one of them had in my opinion more issues than the chosen approach.
1.  Sort once on first call to `Get()`: Clean but adds runtime during game itself and not just to loading screen.
2. Sort once after all parsing passes are done: Identifying the final parsing pass reliably is not easily doable, it can be up to 4 times (rules, non-client gamemode INI, map file, `tmcj4f.ini` if the cheat option is on) but at the minimum it is 1.
3. Sort once manually on some kind of 'finalize' function: Manually keeping track of all `Animatable` instances, adding some mechanism after all parsing has been done etc. would be way too much work and potentially way more annoying for devs to use as well.
4. Remove binary search: A possibility but ironically enough linear search solutions that are not terrible for performance are harder to understand and read.

And in case it wasn't clear why this had to be changed to begin with, sorting the parsed data always after each parsing pass was causing problems with the data if re-parsed (e.g keyframe indices no longer match originals, possibly)

Starkku:
Merging, latest changes were approved elsewhere.

PR #1305, opened 2024-07-12
Fix PowersUpToLevel > 0 upgrade animation inconsistencies

Starkku:
Previously, if you had building with `Upgrades` value that is lower than amount of available possible upgrades and the upgrades had `PowersUpToLevel` > 0, inconsistencies (or limitations, given that it was not originally made for scenario like this) in upgrade code would allow it to display the correct base animation image (because when placing an upgrade, game copies the upgrade building's image filename to the target building's animation struct corresponding to that upgrade) but damage animation would likely either be absent or wrong entirely.

This PR corrects these inconsistencies and allows buildings like this to use `PowerUpNAnim` and related settings corresponding to the current powerup level regardless of building type's `Upgrades` value. `PowerUpNAnim` specifically is also used over the upgrade's image filename if it is set. `PowersUpToLevel=-1` etc. still increment's the building's current upgrade level by one and uses that as the powerup level as before. None of this also does really concern upgrades to `WallTower` building, which behave in somewhat different manner to begin with.

This does not extend the limit of upgrades on building to above 3, nor does it change anything to allow fully functional distinct building upgrade anims to be displayed for more than 3 different upgrade types.


Coronia (approved):
codes look fine and work well in our actual test

PR #1307, opened 2024-07-14
[Customized] Damage multiplier for different houses

CrimRecya:
- Warheads are now able to define the extra damage multiplier for owner house, ally houses and enemy houses. The values in warheads are default to respective `[CombatDamage]` -> `DamageXXMultiplier`, which all default to 1.0 .Note that this will not affect damage with ignore defenses.

In `rulesmd.ini`:
```ini
[CombatDamage]
DamageOwnerMultiplier=1.0     ; floating point value
DamageAlliesMultiplier=1.0    ; floating point value
DamageEnemiesMultiplier=1.0   ; floating point value

[SOMEWARHEAD]                 ; Warhead
DamageOwnerMultiplier=        ; floating point value
DamageAlliesMultiplier=       ; floating point value
DamageEnemiesMultiplier=      ; floating point value
```


Sovietianqi (approved):

src/Ext/Techno/Hooks.Shield.cpp
```diff
@@ -17,6 +17,36 @@ DEFINE_HOOK(0x701900, TechnoClass_ReceiveDamage_Shield, 0x6)
...
+		else if (pHouse->IsAlliedWith(args->SourceHouse))
+		{
+			if (pWHExt->DamageAlliesMultiplier != 1.0)
+				calculateDamage = static_cast<int>(*args->Damage * pWHExt->DamageAlliesMultiplier.Get(RulesExt::Global()->DamageAlliesMultiplier));
+		}
+		else
+		{
+			if (pWHExt->DamageEnemiesMultiplier != 1.0)
+				calculateDamage = static_cast<int>(*args->Damage * pWHExt->DamageEnemiesMultiplier.Get(RulesExt::Global()->DamageEnemiesMultiplier));
+		}
```
Coronia, line 44: the current implementation will make those global multipliers being omitted if warhead multipliers set to 1.0, which seems to be wrong. Also you can't just use Nullable variable before asserting if it's set or not. In fact you could just remove the if statement here and let Get function do the job, since it'll pick the default 1.0 if it's not set

		if (pHouse == args->SourceHouse)
			calculateDamage = static_cast<int>(*args->Damage * pWHExt->DamageOwnerMultiplier.Get(RulesExt::Global()->DamageOwnerMultiplier));
		else if (pHouse->IsAlliedWith(args->SourceHouse))
			calculateDamage = static_cast<int>(*args->Damage * pWHExt->DamageAlliesMultiplier.Get(RulesExt::Global()->DamageAlliesMultiplier));
		else
			calculateDamage = static_cast<int>(*args->Damage * pWHExt->DamageEnemiesMultiplier.Get(RulesExt::Global()->DamageEnemiesMultiplier));

Coronia (changes requested):
since Phobos already has an extra warhead logic, suggest to rename this one a bit. 'damage multiplier for different houses' perhaps

src/Ext/Techno/Hooks.Shield.cpp
CrimRecya: My mistake, get it.

Coronia (changes requested):
think it'd better not to be blocked by IgnoreDefenses in order to be consistent with other firepower multipliers. The shield part can be kept as it was

docs/New-or-Enhanced-Logics.md
```diff
@@ -1518,6 +1518,23 @@ In `rulesmd.ini`:
 NotHuman.DeathSequence=  ; integer (1 to 5)
 ```
 
+### Damage multiplier for different houses
+
+- Warheads are now able to define the extra damage multiplier for owner house, ally houses and enemy houses. The values in warheads are default to respective `[CombatDamage]` -> `DamageXXMultiplier`, which all default to 1.0 .Note that this will not affect damage with ignore defenses.
```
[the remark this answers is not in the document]
CrimRecya: Okay, I see.

Coronia (approved):

TaranDahl:
I hope there can be an independent flag control to make the friendly damage modifier ineffective for the main target. If possible. There is such a need: I hope that a certain unit will not take out nearby friendly forces together when attacking the enemy, but in some cases, forced attacks on friendly forces will be required (for example, when an AI's building blocks the road).

PR #1308, opened 2024-07-14
[Highly Customized] Units Rotate Turret When in Idle Action

CrimRecya:
- Now unit with turret without `TurretSpins=true` can looks more vivid when it is in idle.
  - `UnitIdleRotateTurret` controls whether units can rotate their turrets when in idle. Defaults to `[AudioVisual]` -> `UnitIdleRotateTurret`.
  - `UnitIdlePointToMouse` controls whether units will turn their turrets to your mouse when in idle. Defaults to `[AudioVisual]` -> `UnitIdlePointToMouse`.
  - `UnitIdleActionRestartMin` and `UnitIdleActionRestartMax` control the delay from idle to action occurrence together.
  - `UnitIdleActionIntervalMin` and `UnitIdleActionIntervalMax` control the delay between every idle actions together.

In `rulesmd.ini`:
```ini
[AudioVisual]
UnitIdleRotateTurret=false      ; boolean
UnitIdlePointToMouse=false      ; boolean
UnitIdleActionRestartMin=150    ; integer, number of frames
UnitIdleActionRestartMax=300    ; integer, number of frames
UnitIdleActionIntervalMin=150   ; integer, number of frames
UnitIdleActionIntervalMax=450   ; integer, number of frames

UnitIdleRotateTurret=           ; boolean
UnitIdlePointToMouse=           ; boolean
```

rotate


CrimRecya:
Only test in single player yet, may need more test.

Metadorius:
99% sure point to mouse will desync :(

CrimRecya:
> 99% sure point to mouse will desync :(

Oh, that's too bad.

Metadorius:
I think you misunderstood. Random facing will work fine as long as random calls are synchronized. Mouse on the other hand is not synchronized, so if you depend on mouse pointer it will desync because other players don't know where your mouse pointer is.

CrimRecya:
> I think you misunderstood. Random facing will work fine as long as random calls are synchronized. Mouse on the other hand is not synchronized, so if you depend on mouse pointer it will desync because other players don't know where your mouse pointer is.

Get it. now it has changed.

mevitar:
This doesn't work properly for jumpjets with turrets. They will move their turret only if they are pre-placed, and if ordered to scatter from ground. But if they are ordered to move, they won't rotate on idle anymore (i suppose they don't actually count as idle by the game the same way that ground units do).

CrimRecya:
> This doesn't work properly for jumpjets with turrets. They will move their turret only if they are pre-placed, and if ordered to scatter from ground. But if they are ordered to move, they won't rotate on idle anymore (i suppose they don't actually count as idle by the game the same way that ground units do).

Thank you for testing. However, I plan to repull this request at the end of the year because I have also created some other features (i.e. restrictions on turret rotation angle and vehicle body's orientation) at the locations of these hooks, and I have not yet submitted them as they are still under testing. At the same time, I will also find the way to solve the turret rotation limitation of the jumpjets.
Currently, only after executing the attack command and eliminating the target, can they enter this state normally. This is because their mission will be stuck in `move` and always remember their `destination`. If they are commanded to `stop`, they will lose their `destination` but still stuck in `move`, which will put them in a state of standing idly by. Afterwards, I will also submit these fixes together.

CrimRecya:
I will pull a new request soon. Close this first.

PR #1312, opened 2024-07-15
Avoid InvisibleInGame targets in new scripts

FS-21:
Buildings with InvisibleInGame=yes should be ignored in the target selection of the new attack/move actions.

I just found I forgot to send this time ago :-/
No credits or mention needed since it only fix a small bad behaviour.


FS-21:
> Since the scripts were introduced before last stable released, this probably requires a changelog entry under Phobos fixes.

Makes sense :-D

Starkku:
I meant the actual changelog (`What's-New.md`) not the fixes and improvements page misc. section which contains **vanilla game** fixes.

FS-21:
I think I got the location

> I meant the actual changelog (`Whats-New.md`) not the fixes and improvements page misc. section which contains **vanilla game** fixes.

Ok! I think I found the place

PR #1315, opened 2024-07-16
Fix #1106 : Deployer = yes to no infantry conversion sequence update

### LINKED ISSUE #1106: Convert superweapon issue
### Description

1 when I convert a deployed GI into a troop, it cannot move.
2 AI would not continue the command of the converted units, they would stop here.
3 AI would not recognize new units as the old one, so AI would build the old units (as they ingore the build prerequisite).
4 If I put autofire on my convert superweapon, according to 2, AI would continuely build the old units, and as they are all converted, AI would not stop doing this.

### Phobos Version

B36

### Conditions to reproduce

MO3, normal syringe, a little bit change in rules.

### INI code

```ini
[ArsenalMedicSpawn]
......
Convert0.From=E1                  ; list of TechnoTypes
Convert0.To=ARND                    ; TechnoType
Convert0.AffectedHouses=owner     ; list of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
LimboDelivery.Types=UPGRADEBD

[E1]
Prerequisite.Negative=UPGRADEBD
```

### Steps to reproduce

1. building barracks to train GI
2. building arsenal
3. seeing what happens

### Expected behaviour

AI would building new GI rather than the old ones, the new GI would act as normal.

### Actual behaviour

New GI stopped, and AIs are still building normal GI.

### Additional context

XBZS F~2SOZN%3)72V5N8(B
`E}4$7Z)U_$SF}T 5_BTSS0

### Checklist

chaserli:
It's a bad design and thus does not worth a fix
in case someone really wants to, here's one way to it


Coronia (approved):
other than break not being line-feeded properly the codes look fine

PR #1318, opened 2024-07-18
[Highly Customized] Jumpjet Climbing Logic Enhancement

CrimRecya:
In vanilla game, the jumpjet unit would stop horizontal movement as it rose to a relatively high altitude. Here are two options to change this.
- You can now let the jumpjets increase their height earlier by set `JumpjetClimbPredictHeight` to true or simply let them skip the stop check by set `JumpjetClimbWithoutCutOut` to true.

In `rulesmd.ini`:
```ini
[General]
JumpjetClimbPredictHeight=false  ; boolean
JumpjetClimbWithoutCutOut=false  ; boolean
```


src/Ext/Unit/Hooks.Jumpjet.cpp
```diff
@@ -260,3 +261,111 @@ void __stdcall JumpjetLocomotionClass_Unlimbo(ILocomotion* pThis)
...
+	CellClass* const pCell = MapClass::Instance->GetTargetCell(location);
+
+	if (!pCell)
+		return -1;
+
+	int height = pCell->GetFloorHeight(Point2D{ location.X & 0xFF, location.Y & 0xFF });
+	ObjectClass* pObject = pCell->FirstObject;
+
+	while (pObject && pObject->WhatAmI() != AbstractType::Building)
+		pObject = pObject->NextObject;
```
Coronia, line 284: while (pObject && pObject->WhatAmI() != AbstractType::Building)
	{
		pObject = pObject->NextObject;
	}

Coronia (changes requested):
some comments on how certain 'magic number' coming out would be nice, for example:

	if (pCell->FindTechnoNearestTo(Point2D::Empty, false))
		height += 85;
why 85 here?

			const int checkStep = (largeStep >> 7) + ((largeStep & 0x7F) ? 2 : 1);
and what's this for?

src/Ext/Unit/Hooks.Jumpjet.cpp
CrimRecya: I add the notes, and change the last one.

src/Ext/Unit/Hooks.Jumpjet.cpp
```diff
@@ -276,3 +277,109 @@ DEFINE_HOOK(0x54DAC4, JumpjetLocomotionClass_EndPiggyback_Blyat, 0x6)
...
+	bool Skip = false;
+	int GetJumpjetHeightWithOccupyTechno(Point2D location); // Replace sub_485080
+	int JumpjetLocomotionPredictHeight(JumpjetLocomotionClass* pThis); // Replace sub_54D820
+}
+
+int JumpjetRushHelpers::GetJumpjetHeightWithOccupyTechno(Point2D location)
+{
+	CellClass* const pCell = MapClass::Instance->GetTargetCell(location);
+
+	if (!pCell)
```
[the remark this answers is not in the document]
CrimRecya: Oh, you're right, this was my negligence. I will modify it to TryGetCellAt later, thx for pointing it out.

TaranDahl:
Tested locally. Doing good.

TaranDahl (approved):

Coronia:
Doc needs more details about the behaviors of these 2 tags. For example: when'll the aircraft height change if `JumpjetClimbWithoutCutOut` set to true, what'll happen if an aircraft stop during the process of earlier height changing, and what'll happen if these 2 tags being used at the same time

mevitar:
Didn't find any issues. Didn't try in multiplayer.

Coronia (approved):

PR #1319, opened 2024-07-19
[Minor][Needs testing] Attempt to fix the ReloadInTransport crash

Coronia:
Techno with ReloadInTransport=yes suffered from a crash when the transporter's owner is changed (i.e. being mind controlled). This is because it'll move the techno from the original owner's OwnedTransportReloader list to the new owner's, which is actually wrong when the transporter doesn't have Passengers.SyncOwner=yes. Attempted to fix this issue and unified the conditions of ReloadInTransport

### DIFF
--- a/src/Ext/Techno/Hooks.Transport.cpp
+++ b/src/Ext/Techno/Hooks.Transport.cpp
@@ -1,7 +1,6 @@
 #include "Body.h"
 
-#include <Ext/House/Body.h>
-
+#include <Ext/Scenario/Body.h>
 
 DEFINE_HOOK_AGAIN(0x6FA33C, TechnoClass_ThreatEvals_OpenToppedOwner, 0x6) // TechnoClass::AI
 DEFINE_HOOK_AGAIN(0x6F89F4, TechnoClass_ThreatEvals_OpenToppedOwner, 0x6) // TechnoClass::EvaluateCell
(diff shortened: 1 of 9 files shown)


chaserli (changes requested):
if crash happens with ownership change, maybe we shouldn't store the vector in houseext at all. Just smash them all in ScenarioExt and update in logicclass::update (c.f. Misc/Hooks.VeinholeMonster.cpp DEFINE_HOOK(0x55B4E1, LogicClass_Update_Veinhole, 0x5))

src/Ext/Techno/Hooks.Transport.cpp
```diff
@@ -157,7 +160,7 @@ DEFINE_HOOK(0x51DF82, InfantryClass_FireAt_ReloadInTransport, 0x6)
 {
 	GET(InfantryClass* const, pThis, ESI);
 
-	if (pThis->Transporter)
+	if (pThis->Transporter && pThis->WhatAmI() != AbstractType::Aircraft && pThis->WhatAmI() != AbstractType::Building)
```
chaserli, line 163: it's an infantryclass so its rtti must be infantry

Coronia:
The previous fix doesn't work, so I took the idea of @chaserli and put those vectors into ScenarioExt.

Both ReloadInTransport and AutoDeath (including limbo buildings) worked properly in our intensive regression test, and the crash no longer happened.

Issue #1140, opened 2023-08-26
Singleplayer Mode Global variable cannot be saved


kyonkonyan:
### Description

the current level global variable is saved successfully, but the second level cannot read this global variable。

### Phobos Version

on [Build 22] and update to [Build 36] or Release v0.3.0.1

### Steps to reproduce

1.mission map level 1 set global variable and win this level
2.mission map level 2 read global variable
3.show global variable

### Expected behaviour

1.mission map level 1 set global variable and win this level
2.mission map level 2 read global variable
3.show global variable
4.it‘s 1

### Actual behaviour

1.mission map level 1 set global variable and win this level
2.mission map level 2 read global variable
3.show global variable
4.it‘s 0

### Checklist

- [X] The issue happens on the **latest official** version of Phobos and wasn't fixed yet.
- [X] I agree to elaborate the details if requested and provide thorough testing if the bugfix is implemented.
- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search and read the issue list to find a similar issue and didn't find it.
- [X] I have attached as much information as possible *(screenshots, gifs, videos, debug and exception logs, etc)*.

Issue #1108, opened 2023-07-19
Global variable set/read invalidation in singleplayer maps.Your bug report title here


AmethystCR:
### Description

Global variable defined in rulesmd.ini - [VariableNames] cannot set/read correctly in singleplayer maps. I don't know if the problem is on setting or reading of global variables. The current level global variable is set, but the second level cannot read this global variable correctly. I have tested the phobos version from B36 to B20，the problem is solved when version is below B20. The problem happened again in B22.
Maybe the same as problem B in #681 , but only solved problem A at that time.

### Phobos Version

B22 to B36

### Steps to reproduce

1. Get a mission defined Global Variables. Compete an object to set the global variable to 1.
2. Get the next mission, read this global variable to trigger events.
3. The event not happen.
...

### Expected behaviour

The event should be happen.

### Actual behaviour

The event not happen.

### Checklist

- [X] The issue happens on the **latest official** version of Phobos and wasn't fixed yet.
- [X] I agree to elaborate the details if requested and provide thorough testing if the bugfix is implemented.
- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search and read the issue list to find a similar issue and didn't find it.
- [X] I have attached as much information as possible *(screenshots, gifs, videos, debug and exception logs, etc)*.

PR #1324, opened 2024-07-22
[Minor] Fix AltNextScenario= not taking effect

a851903106:
An extension to “AltNextScenario=” has been made so that it can now be started with any variable with the name “<Alternate Next Scenario>”, but of course the original hardcoding still works.


chaserli:
I don't know if you are referring to issue #1108 and #1140, if yes then check if dde700f fixed it already

a851903106:
> I don't know if you are referring to issue #1108 and #1140, if yes then check if [dde700f](https://github.com/Phobos-developers/Phobos/commit/dde700fab27ade61e365e112a69fcde6c62b5195) fixed it already

确实有用，那我只留下AltNextScenario=这部分修复就行了。

Coronia (commented):
if the variables are not supposed to be editted in the function, we'd better avoid using reference here. Namely LocalVariables, GlobalVariables and their itr

a851903106:
> if the variables are not supposed to be editted in the function, we'd better avoid using reference here. Namely LocalVariables, GlobalVariables and their itr

I'm sorry, I don't quite understand.

Coronia:
> > if the variables are not supposed to be editted in the function, we'd better avoid using reference here. Namely LocalVariables, GlobalVariables and their itr
> 
> I'm sorry, I don't quite understand.

`auto const& LocalVariables = ScenarioExt::Global()->Variables[false];`
`auto const& GlobalVariables = ScenarioExt::Global()->Variables[true];`
I didn't notice any edit on these 2 variables in this pull request. It's just iterating them and reading their values, so maybe the reference sign `&` could be removed. Same for `auto const& itr`

a851903106:
> > > if the variables are not supposed to be editted in the function, we'd better avoid using reference here. Namely LocalVariables, GlobalVariables and their itr
> > 
> > 
> > I'm sorry, I don't quite understand.
> 
> `auto const& LocalVariables = ScenarioExt::Global()->Variables[false];` `auto const& GlobalVariables = ScenarioExt::Global()->Variables[true];` I didn't notice any edit on these 2 variables in this pull request. It's just iterating them and reading their values, so maybe the reference sign `&` could be removed. Same for `auto const& itr`



> > > if the variables are not supposed to be editted in the function, we'd better avoid using reference here. Namely LocalVariables, GlobalVariables and their itr
> > 
> > 
> > I'm sorry, I don't quite understand.
> 
> `auto const& LocalVariables = ScenarioExt::Global()->Variables[false];` `auto const& GlobalVariables = ScenarioExt::Global()->Variables[true];` I didn't notice any edit on these 2 variables in this pull request. It's just iterating them and reading their values, so maybe the reference sign `&` could be removed. Same for `auto const& itr`

got it.

Coronia (approved):

PR #1332, opened 2024-07-31
[Minor] - Check if it is the latest update.

Sovietianqi:
Check if the version in use is the latest version of Phobos.


src/Phobos.cpp
```diff
@@ -1,226 +1,294 @@
...
-
-		Point2D location { rect.X + 5,5 };
-
-		DSurface::Composite->FillRect(&rect, COLOR_BLACK);
-		DSurface::Composite->DrawText(Phobos::VersionDescription, &location, COLOR_RED);
-	}
-	return 0;
-}
-#endif
+#include "Phobos.h"
```
Sovietianqi, line 1: This thing is so strange; how could there be so many differences? I'm a bit confused.

Sovietianqi (commented):
It's amazing, there are so many differences. Could it be an issue with my editor? Probably, I'm a bit confused right now.

TaranDahl:
提交错地方了吧😂

Starkku:
EOL conversion. The files in your repo have CRLF line endings (Windows style) while the ones in main repo are LF (Unix style). The repository is configured to check out certain text files CRLF format but does not enforce pushing in LF (unsure if this is an oversight on our part). Git can locally be configured to do this, though (`core.autocrlf=true`).

Sovietianqi:
Thank you, sir, for your strong support. I will do my best to solve this problem. ^_^

PR #1334, opened 2024-07-31
[New Feature] More checks for DamageArea

TaranDahl:
### CellSpread enhancement

- In vanilla, the damage area of an AOE warhead is spherical. In some case, e.g. you want to make a warhead superweapon buff all units in an area, the affectted range for air units is always smaller than ground units. Now you can use a new flag `CellSpread.Cylinder` to overcome this problem.
- `AffectsAir` allow you to make a warhead only damage the units with height more than 208.
- `AffectsGround` allow you to make a warhead only damage the units with height less than 208.
- Noting that these features work independently with the ares flag `DamageAirThreshold`. A warhead with `CellSpread.Cylinder` detonating on floor will not affect units in air, unless it has `DamageAirThreshold = -1`.

In `rulesmd.ini`:
```ini
[SOMEWARHEAD]              ; WarheadType
CellSpread.Cylinder=false  ; boolean
AffectsAir=true            ; boolean
AffectsGround=true         ; boolean
```


Coronia (commented):
could simply name these 2 tags as AA and AG like projectile, so it's easy to understand and consistent

TaranDahl:
Added some code for the underground damage.
Still, there are some works that need to be done like the AU enemy finding and docs.
I'm getting busy now and may not be able to impl these functions recently, so I drafted this temporarily.

TaranDahl:
> could simply name these 2 tags as AA and AG like projectile, so it's easy to understand and consistent

As is well known, the vanilla flag `AG`'s effects are confusing and far apart from its name😅
So I think maybe we should avoid having the same name with its🤔

TaranDahl:
Reverted the underground-related commits and reopened this again.
Will open another pr to deal with the underground things.

Coronia (approved):
modified the codes a bit. Warhead in DamageArea should always exist according to IDA and actual testing, so a sanity check might not be needed. Also rename the tag to make it clearer and more inline with the already existed `AffectsUnderground` tag

PR #1335, opened 2024-08-04
Expand the Building Actions

CrimRecya:
I will update the docs soon and these GIFs will show you the main function.
building
building2


CrimRecya:
Building placing and deploying logic enhancement

- In vanilla games, buildings are always cannot placing or deploying on the cells that other infantries or units on. Now this can be  changed by setting `ExpandBuildingPlace` to true, when you try to place the building on these cells, it will check whether the occupiers can be scatter by yourself (include your own technos and allies non-player technos) and whether there are enough spaces to scatter. If can, it will record which building you are placing and show a preview to you and your allies, then start a timer to record this placement and order the occupiers to leave this building area. When the area is cleared, the building will be truly place down and the production queue will be restored to original state. But when the timer expires or an unexpected situation has occurred which make the building impossible be constructed here anymore, it will stop the action and play "cannot deploy here", then you should re-place or re-deploy the building in a valid space. Note that when the building has been recorded and is trying to place, unless the production queue has vanished (such as construction yard is no longer exist), it will continue to function normally until the conditions are not met.
- You can now set `CheckUnitBaseNormal` to true to use units to expand the construction scope of the base.
  - `UnitBaseNormal` controls whether our own buildings can be place around it like vanilla `BaseNormal` do.
  - `UnitBaseForAllyBuilding` controls whether ally buildings can be place around it like vanilla `EligibileForAllyBuilding` do.
- Now technos have `CanBeBuiltOn=true` can simply removed when building is placed on them.
- This feature is highly compatible with `ExpandBuildingPlace`. If set `CheckExpandPlaceGrid` to true, it will display the placing grids with `place.shp` and following corresponding frame number.
  - `ExpandLandGridFrames` controls the placing grids frames on non-water cell. The three numbers respectively represent "Some technos that can command departure have occupied this area", "This cell is actually beyond the scope, but there is still at least one cell inside the entire region" and "Here is no problem, everything is OK".
  - `ExpandWaterGridFrames` controls the placing grids frames on water cell. Each item corresponds to the same as above.
- `AlwaysExistTheCameo` is a setting that allows you to preview information. This feature can be used as before, playing "new construction options" and clearing the specific production queue when prerequisites loss.
  - `AlwaysExistTheCameo` controls whether you can see the cameo when the prerequisite have not satisfied (`TechnoLevel`, `Owner`, `RequiredHouses` and `ForbiddenHouses` should be satisfied). Defaults to `[General]` -> `AlwaysExistTheCameo`.
  - If `PrerequisiteForCameo` is not set, the grey cameo will only show when `AIBasePlanningSide` is satisfied. If set a techno type, the grey cameo will show if you have a techno in this type or this type's `TechnoLevel`, `Owner`, `RequiredHouses` and `ForbiddenHouses` is satisfied.
  - The `UIExtraDescription` is like `UIDescription`, but this only appearing when the techno is truly unbuildable.

In `rulesmd.ini`:
```ini
[General]
CheckUnitBaseNormal=false      ; boolean
ExpandBuildingPlace=false      ; boolean
AlwaysExistTheCameo=false      ; boolean

[AudioVisual]
CheckExpandPlaceGrid=false     ; boolean
ExpandLandGridFrames=1,0,0     ; integer, zero-based frame index - have technos, near boundary, is normal
ExpandWaterGridFrames=1,0,0    ; integer, zero-based frame index - have technos, near boundary, is normal

[SOMETECHNO]                   ; TechnoType
CanBeBuiltOn=false             ; boolean
AlwaysExistTheCameo=           ; boolean
PrerequisiteForCameo=          ; TechnoType
UIExtraDescription=            ; CSF entry key

[SOMEUNIT]                     ; UnitType
UnitBaseNormal=false           ; boolean
UnitBaseForAllyBuilding=false  ; boolean
```

8abs:
this should be merged asap

4SGX:
If I may suggest some more appropriate English for these tags:
`AlwaysExistTheCameo` ► `AlwaysShowCameo` or `AlwaysShowInSidebar`
`PrerequisiteForCameo` ► `SidebarPrerequisite` or `CameoPrerequisite`
`CheckExpandPlaceGrid` ► `UseNewGrid`
And possibly remove Expand from ~~Expand~~LandGridFrames and ~~Expand~~WaterGridFrames

And cannot `EligibleForAllyBuilding` be reused here instead of creating the new tag `UnitBaseForAllyBuilding`?

src/Ext/BuildingType/Body.cpp
```diff
@@ -97,6 +98,451 @@ int BuildingTypeExt::GetUpgradesAmount(BuildingTypeClass* pBuilding, HouseClass*
...
+								optionalTechnos.push_back(pCurTechno);
+							}
+
+							pCurObject = pCurObject->NextObject;
+						}
+
+						if (valid && pOptionalCell->IsClearToMove(pCheckedType->SpeedType, true, true, -1, pCheckedType->MovementZone, -1, false))
+						{
+							for (auto const& pOptionalTechno : optionalTechnos)
+								reCheckedTechnos.push_back(pOptionalTechno);
```
Coronia, line 444: for (auto const& pOptionalTechno : optionalTechnos)
							{
								reCheckedTechnos.push_back(pOptionalTechno);
							}

Coronia (changes requested):
I'm not quite sure about its logical correctness yet, more reviewing is definitely needed

also perhaps it should be separated to 4 different pull request for baseNormal, cameo display, building grid display and units auto moving/killed when building

Metadorius:
If the features are not very intertwined a split would be preferable. I understand that we're over our review capacity at the moment though and because of that maybe it's not necessary at least yet.

CrimRecya:
Get the idea. The placing grids and units base normal are hard to split, but I think it can still be divided into three parts, although it will take me some time. So later I will make modifications to the review that was just submitted, then do a split when I have time.

src/Ext/BuildingType/Body.cpp
CrimRecya: All fixed.

CrimRecya:
cameo
Add new functions for cameos. And will split this pr in a few days.

Metadorius:
> I have been really busy lately, so I will wait until I have some free time to do the splitting. Even if it is split, I won't have the energy to make any modifications, so wait for now. Here may be no need for review at the moment, this can be focused on other good places. If you are interested in this, maybe you can give it a try, at least so far the feedback I've received says it's good.

Take your time! It's not a big priority to split the features at all, the pull request isn't as big. More of a "nice to have" than a requirement in this case.

src/Ext/BuildingType/Body.cpp
```diff
@@ -97,6 +103,733 @@ int BuildingTypeExt::GetUpgradesAmount(BuildingTypeClass* pBuilding, HouseClass*
...
+					if (pUpgrade)
+						upgradeBuildings.push_back(pUpgrade);
+				}
+			}
+
+			for (const auto& pOwned : pHouse->Buildings)
+			{
+				for (const auto& pUpgradeType : upgradeBuildings)
+				{
+					if (pOwned->Type == pUpgradeType && pOwned->IsAlive && pOwned->Health > 0 && pOwned->IsOnMap && !pOwned->InLimbo && pOwned->CurrentMission != Mission::Selling)
```
NetsuNegi, line 681: you don't need check type because it will be checked in BuildingClass::CanUpgrade()

CrimRecya:
I‘m going to repull the split request in two weeks, so close this now.

PR #1338, opened 2024-08-06
Delayed fire weapons

Starkku:
New implementation of the same idea as #768

### Delayed firing

- It is possible to have any weapon fire with a delay by setting `DelayedFire.Duration` on a WeaponType - it supports a single integer or two comma-separated ones for a random range to pick value from.
  - If `DelayedFire.SkipInTransport` is set to true and firer is in a transport, no delay is applied to firing.
  - `DelayedFire.Animation` can be used to define animation to create when the delay timer starts. `DelayedFire.OpenToppedAnimation` is used instead if set if the firer is in a transport.
    - If `DelayedFire.AnimIsAttached` is set to true, the animation is attached to the firing TechnoType. If `DelayedFire.RemoveAnimOnNoDelay` is also set to true the animation is removed when the duration expires or firing is interrupted regardless of its remaining lifetime.
    - `DelayedFire.AnimOffset` can be used to override the weapon's firing coordinates / FLH for the animation's position.
    - `DelayedFire.AnimOnTurret` determines whether or not the animation's position is calculated relative to firer's body or turret (only if it has one).
    - If `DelayedFire.CenterAnimOnFirer` is set the animation is created at the firer's center rather than at the firing coordinates.
  - If the weapon was fired by InfantryType and `DelayedFire.PauseFiringSequence` is set to true, the infantry's firing sequence animation is paused when it hits the firing frame defined by `FireUp/Prone` or `SecondaryFire/Prone` in its `artmd.ini` entry until the delay timer has expired.
  - If the weapon has `Burst` > 1 and `DelayedFire.OnlyOnInitialBurst` set to true, the delay occurs only before the initial burst shot. Note that if using Ares, `Burst` index does not reset if firing is interrupted or the firer loses target, meaning it will be able to resume firing without waiting for the delay.
  
In `rulesmd.ini`:
```ini
[SOMEWEAPON]                           ; WeaponType
DelayedFire.Duration=                  ; integer - single or comma-sep. range (game frames)
DelayedFire.SkipInTransport=false      ; boolean
DelayedFire.Animation=                 ; Animation
DelayedFire.OpenToppedAnimation=       ; Animation
DelayedFire.AnimIsAttached=true        ; boolean
DelayedFire.AnimOffset=                ; integer - Forward,Lateral,Height
DelayedFire.AnimOnTurret=true          ; boolean
DelayedFire.CenterAnimOnFirer=false    ; boolean
DelayedFire.RemoveAnimOnNoDelay=false  ; boolean
DelayedFire.PauseFiringSequence=false  ; boolean
DelayedFire.OnlyOnInitialBurst=false   ; boolean
```

NOTE: AircraftTypes, due to their different attack patterns, will not wait for the delay to expire before attempting to fire and will instead continue without firing if the delay is too long.


mevitar:
DelayedFire.PauseFiringSequence=yes breaks the weapon if used with Burst.FireWithinSequence=yes, as the infantry will be unable to fire. Seems to work fine otherwise, for infantry and units.
Does not work most of the time for AircraftTypes, and even if it works, the animation doesn't play.

And i think this could use an option to play the charge animation only before the first shot if the weapon has Burst= above 1.

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -507,6 +508,70 @@ DEFINE_HOOK(0x6FF43F, TechnoClass_FireAt_FeedbackWeapon, 0x6)
...
+static void __forceinline CreateDelayedFireAnim(TechnoClass* pThis, AnimTypeClass* pAnimType, int weaponIndex, bool attach, bool center)
+{
+	if (pAnimType)
+	{
+		auto coords = pThis->GetCenterCoords();
+
+		if (!center)
+			coords = pThis->GetFLH(weaponIndex, coords);
+
+		if (auto const pAnim = GameCreate<AnimClass>(pAnimType, coords))
```
chaserli, line 520: the convention is that allocation failure leads to crash directly. Even if WW did similar sanity check all the time doesn't mean it makes sense

Starkku:
Updated with fixes and some new features.

- `DelayedFire.Duration` now supports two-comma separated values for random range.
- `DelayedFire.AnimOffset` can be used to override FLH position for the animation.
- `DelayedFire.RemoveAnimOnNoDelay` if set to true makes the animation disapper if delay is inactive.
- `DelayedFire.OnlyOnInitialBurst` if set to true makes it so that only first burst shot has delay. Do mind that with Ares burst index does not reset if unit stops firing midway through.

The interaction between `DelayedFire.PauseFiringSequence` and `Burst.FireWithinSequence` should now also be fixed.

mevitar:
Still doesn't work for AircraftTypes, so it should be noted in the documentation.
Other than that, everything seems to be working fine now. Didn't test in multiplayer.

Starkku:
Rebased, improved the handling of animation offsets. This is ready for another testing and review pass if needed.

mevitar:
This build crashes when building one specific unit, but that unit works fine with other builds so i don't know what's the problem. Here's the except file.
[except.zip](https://github.com/user-attachments/files/20194990/except.zip)

mevitar:
> This build crashes when building one specific unit, but that unit works fine with other builds so i don't know what's the problem.

Putting vehicle inside an open-topped vehicle crashes the game. This crash in particular happens for a VehicleType that has another VehicleType inside trough InitialPayload. Infantry does not cause this crash, only vehicles.

Starkku:
> Putting vehicle inside an open-topped vehicle crashes the game. This crash in particular happens for a VehicleType that has another VehicleType inside trough InitialPayload. Infantry does not cause this crash, only vehicles.

I can't reproduce this crash. There might be some parameters in play that I am missing.

mevitar:
Try building it from a naval yard.

Starkku:
> Try building it from a naval yard. Or preplace an open topped naval unit with a vehicle inside, that also crashes for me.

Still doesn't crash. Any chance you could perhaps provide full crash log(s), or atleast both except+debug?

Starkku:
Any chance this could get reviewed, as well as maybe an another testing pass or so given that I had to rebase and refactor this to fit in with the infantry gattling / SHP unit `FireUp` functionality *hopefully* not breaking anything - something I do not want to do again.

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -586,6 +586,63 @@ DEFINE_HOOK(0x6FF905, TechnoClass_FireAt_FireOnce, 0x6)
...
+	auto const pWeaponExt = WeaponTypeExt::ExtMap.Find(pWeapon);
+	auto const pExt = TechnoExt::ExtMap.Find(pThis);
+	auto& timer = pExt->DelayedFireTimer;
+
+	if (pExt->DelayedFireWeaponIndex >= 0 && pExt->DelayedFireWeaponIndex != weaponIndex)
+		pExt->ResetDelayedFireTimer();
+
+	if (pWeaponExt->DelayedFire_Duration.isset() && (!pThis->Transporter || !pWeaponExt->DelayedFire_SkipInTransport))
+	{
+		if (pWeaponExt->DelayedFire_PauseFiringSequence && (pThis->WhatAmI() == AbstractType::Infantry
```
NetsuNegi: Please use a variable to receive the return value of `pThis ->WhatAmI()`, do not call it every time

Coronia:
if primary fire during secondary's charging, it'll stop the charging process. Guess this is an intended behavior

otherwise good in testing

Starkku:
Per the current design charging is intentionally interrupted if firing is stopped or weapon changes.

Coronia:
now that this one is tested and reviewed, gonna merge it first since they're other PRs for weapons in the pipeline, which needs to be resolved based on this one (e.g. https://github.com/Phobos-developers/Phobos/pull/1467)

PR #1340, opened 2024-08-06
EMPulse improvements

Starkku:
### EMPulse settings

- It is possible to customize which weapon a building with `EMPulseCannon=true` fires when an associated `Type=EMPulse` superweapon (**only** if `EMPulse.TargetSelf=false` or omitted) is fired by setting `EMPulse.WeaponIndex`.
  - Note that if you fire another `Type=EMPulse` superweapon with different weapon index that the same building is capable of launching before the first weapon was fired, the latter superweapon's settings will take precedence.
  - Additionally, due to technical limitations the targeting constraints will always default to primary weapon's `Range/MinimumRange` unless `SW.RangeMinimum/SW.RangeMaximum` are explicitly set.
- Is is now also possible to have all other `Type=EMPulse` superweapons that can be fired by same buildings as current one be put on hold until first of the buildings currently set to fire goes off if the firing superweapon has `EMPulse.SuspendOthers=true`.

In `rulesmd.ini`:
```ini
[SOMESW]                     ; SuperWeapon
EMPulse.WeaponIndex=0        ; integer, weapon slot index
EMPulse.SuspendOthers=false  ; boolean
```

Note:
`Type=EMPulse` superweapon and any associated keys are [Ares features](https://ares-developers.github.io/Ares-docs/new/superweapons/types/empulse.html).

Additionally fixed a bug that caused such weapons to not fully respect the building's FLH.

### DIFF
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -64,7 +64,6 @@ This page describes all ingame logics that are fixed or improved in Phobos witho
 - Fixed Engineers being able to enter `Grinding` buildings even when they shouldn't (such as ally building at full HP).
 - Aircraft & jumpjet units are now affected by speed modifiers such as `SpeedAircraft/Infantry/UnitsMult` on `Countries`, `VeteranSpeed` and Crates / AttachEffect (Ares feature).
 - Both voxel and SHP vehicle units should now correctly respect custom palette set through `Palette`.
-- Weapons fired by EMPulse superweapons without `EMPulse.TargetSelf=true` *(Ares feature)* can now create radiation.
 - Setting `RadarInvisible` to true on TerrainTypes now hides them from minimap display.
 - Mind control indicator animations will now correctly restore on mind controlled objects when uncloaked.
 - Animations from Warhead `AnimList` & `SplashList` etc. as well as animations created through map trigger `41 Play Anim At` now have the appropriate house set as owner of the animation by default.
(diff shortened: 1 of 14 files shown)


docs/Fixed-or-Improved-Logics.md
```diff
@@ -165,7 +163,10 @@ This page describes all ingame logics that are fixed or improved in Phobos witho
 
 ## Fixes / interactions with other extensions
 
-- All forms of type conversion (including Ares') now correctly update `MoveSound` if a moving unit has their type changed.
+- Weapons fired by EMPulse superweapons *(Ares feature)* now fully respect the firing building's FLH.
+- Weapons fired by EMPulse superweapons *(Ares feature)* without `EMPulse.TargetSelf=true` can now create radiation.
+- Weapons fired by EMPulse superweapons *(Ares feature)* now respect `Floater` and Phobos-added `Gravity` setting.
+- - All forms of type conversion (including Ares') now correctly update `MoveSound` if a moving unit has their type changed.
```
Metadorius, line 169:
```suggestion
- All forms of type conversion (including Ares') now correctly update `MoveSound` if a moving unit has their type changed.
```

docs/New-or-Enhanced-Logics.md
```diff
@@ -776,6 +776,25 @@ Convert.To=                     ; TechnoType
...
+```ini
+[SOMEBUILDING]                          ; BuildingType
+EMPulseCannon.UseWeaponSelection=false  ; boolean
+
+[SOMESW]                                ; SuperWeapon
+EMPulse.WeaponIndex=-1                  ; integer, weapon slot index
+```
+
+```{note}
+`Type=EMPulse` superweapon and any associated keys are Ares features.
```
Metadorius, line 795: could be good to add a link

Starkku: > could be good to add a link

Link to where? Ares docs?

Starkku:
Reworked the feature. There's now only one setting controlling weapon selection on the superweapon type. Added additional logic to optionally allow temporarily putting on hold all other Type=EMPulse SW's that share cannons with the current one until current one has finished firing. Complex and/or shared cannon lists across multiple SW's might produce unusual results in cases of overlapping launches like SW's coming off suspension prematurely but there's little I can do about that.

PR #1341, opened 2024-08-08
[Dehardcode]Infantry firing while moving

TaranDahl:
### Infantry firing while moving
- In vanilla, there is a hardcoded behavior that the infantries can not fire until they stop, even if they have `OpportunityFire=yes` set. Now you can bypass this restriction by using the following flag.
  - Mind that you still need `OpportunityFire=yes` to make them acquire target when moving. However, if `OpportunityFire=yes` is set, they can still do that if you use the "ctrl+shift" command, just like the units on the ground do.
  - Additionally, the behavior that "rocketeers can not fire when they have buildings beneath them" is also caused by this hardcode. You can use the flag to bypass this behavior as well.
  - The vanilla flag `JumpJetTurn` will affect the visual behavior when the infantry is firing while moving. You need to set it to **no** if you want to make the infantry always facing the target during the attack.

In `rulesmd.ini`:
```ini
[SOMETECHNO]                 ; InfantryType
FiringByPassMovingCheck=true ; boolean
```


src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -392,6 +392,44 @@ DEFINE_HOOK(0x6FC689, TechnoClass_CanFire_LandNavalTarget, 0x6)
...
+DEFINE_HOOK(0x51C9B8, InfantryClass_CanFire_HitAndRun1, 0x17)
+{
+	enum { CheckPass = 0x51C9CF, CheckNotPass = 0x51CAFA };
+
+	GET(InfantryClass*, pThis, EBX);
+
+	const auto pType = pThis->GetTechnoType();
+	const auto pTypeExt = TechnoTypeExt::ExtMap.Find(pType);
+	if (pTypeExt && pTypeExt->FiringByPassMovingCheck
+		|| pThis->SpeedPercentage <= 0.1 ) // vanilla check
```
Coronia, line 404: const auto pType = pThis->GetTechnoType();
	const auto pTypeExt = TechnoTypeExt::ExtMap.Find(pType);

	if (pTypeExt && pTypeExt->FiringByPassMovingCheck
		|| pThis->SpeedPercentage <= 0.1 ) // vanilla check

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -392,6 +392,44 @@ DEFINE_HOOK(0x6FC689, TechnoClass_CanFire_LandNavalTarget, 0x6)
...
+DEFINE_HOOK(0x51CAAC, InfantryClass_CanFire_HitAndRun2, 0x13)
+{
+	enum { CheckPass = 0x51CACD, CheckNotPass = 0x51CABF };
+
+	GET(InfantryClass*, pThis, EBX);
+
+	const auto pType = pThis->GetTechnoType();
+	const auto pTypeExt = TechnoTypeExt::ExtMap.Find(pType);
+	if (pTypeExt && pTypeExt->FiringByPassMovingCheck
+		|| !pThis->Locomotor.GetInterfacePtr()->Is_Moving_Now() ) // vanilla check
```
Coronia, line 423: const auto pType = pThis->GetTechnoType();
	const auto pTypeExt = TechnoTypeExt::ExtMap.Find(pType);

	if (pTypeExt && pTypeExt->FiringByPassMovingCheck
		|| !pThis->Locomotor.GetInterfacePtr()->Is_Moving_Now() ) // vanilla check

Coronia (changes requested):
again the tag name is not very clear. Suggest to change FiringByPassMovingCheck to something like OpportunityFire.AllowInfantry to indicate it's only for infantry and requires previous OpportunityFire to work

also suggest to change the hook names from InfantryClass_CanFire_HitAndRun to InfantryClass_CanFire_OnMoving for better clarity

mevitar:
It's actually possible to have infantry firing on move in unmodified YR, it just needs to have its firing sequence set to something like 0,1,1 and thus making its firing sequence not play at all.

I did tests with this build, and it seems that for ground infantry, there is no difference between having a "clear" firing sequence or adding FiringByPassMovingCheck=yes. The firing sequence does not play on move in either case.
However, FiringByPassMovingCheck=yes does allow the infantry to fire on the move, and still have and use a regular firing sequence while stationary.

The biggest difference is for jumpjet infantry. Using FiringByPassMovingCheck=yes and JumpJetTurn=no like stated in the description makes the infantry able to fire on the fly and still face the target as its attack sequence plays.
Didn't notice any issues here, but i did no multiplayer testing.

Metadorius:
It could be cool to add a new sequence of firing on the move, but it would probably be hard to meaningfully do so because in the ideal case you would need up to `8 movement directions x 8 firing directions = 64 sequences`, I think.

mevitar:
It's not just 64 sequences, it's 64 sequences for fire-move on ground, 64 sequences for fire-move on water, and 64 for fire-move in air.

Starkku:
That's not how this works. Single sequence can contain all directions.

Starkku:
Don't forget that secondary weapon slot can have its own firing sequence on land.

TaranDahl:
True enough. At present this pr is mainly serving jumpjets. Additional sequences are needed, but I think it is hard to impl in a short time.

TaranDahl:
> again the tag name is not very clear. Suggest to change FiringByPassMovingCheck to something like OpportunityFire.AllowInfantry to indicate it's only for infantry and requires previous OpportunityFire to work
> 
> also suggest to change the hook names from InfantryClass_CanFire_HitAndRun to InfantryClass_CanFire_OnMoving for better clarity

I think the flag name should not mention `OpportunityFire` directly, considering the real behavior. At first, I bound it to `OpportunityFire=yes`, but later discovered that it could run independently without `OpportunityFire`.
Using the "ctrl+shift" command on infantries with this flag will let them fire while moving, just like you can use that command to make Apocalypses fire while moving, though they have no `OpportunityFire=yes` set.

src/Ext/Techno/Hooks.Firing.cpp
TaranDahl: Fixed.

src/Ext/Techno/Hooks.Firing.cpp
TaranDahl: Fixed.

TaranDahl:
> again the tag name is not very clear. Suggest to change FiringByPassMovingCheck to something like OpportunityFire.AllowInfantry to indicate it's only for infantry and requires previous OpportunityFire to work
> 
> also suggest to change the hook names from InfantryClass_CanFire_HitAndRun to InfantryClass_CanFire_OnMoving for better clarity

The problems with the hook name and code style are fixed.

mevitar:
> True enough. At present this pr is mainly serving jumpjets. Additional sequences are needed, but I think it is hard to impl in a short time.

**If** any new sequences are added, i think just a single FireMove= is sufficient. Making separate sequence for each direction is too complicated and i think not worth the effort (how many people will be making use of it anyway?).
Not to mention that, if infantry is expanded, then people will soon after start demanding the same be done for vehicles, and if that is not a feature creep, i don't know what is.

So i think that adding new sequences is out of scope for this feature anyway. Let's keep it simple, and worry about new sequences in the future.

CrimRecya (changes requested):

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -392,6 +392,38 @@ DEFINE_HOOK(0x6FC689, TechnoClass_CanFire_LandNavalTarget, 0x6)
...
+	enum { CheckPass = 0x51C9CF, CheckNotPass = 0x51CAFA };
+
+	GET(InfantryClass*, pThis, EBX);
+
+	const auto pType = pThis->GetTechnoType();
+	const auto pTypeExt = TechnoTypeExt::ExtMap.Find(pType);
+
+	if (pTypeExt && pTypeExt->FiringByPassMovingCheck
+		|| pThis->SpeedPercentage <= 0.1 ) // vanilla check
+		return CheckPass;
```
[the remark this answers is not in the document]
TaranDahl: fixed

PR #1342, opened 2024-08-08
[Vanilla Fix]Terror drone return fix

TaranDahl:
Fix the bug that parasite will vanish if it missed its target when its previous cell is occupied.


src/Misc/Hooks.BugFixes.cpp
```diff
@@ -45,6 +45,33 @@ DEFINE_JUMP(LJMP, 0x545CE2, 0x545CE9) //Phobos_BugFixes_Tileset255_RemoveNonMMAr
...
+		return 0;
+	}
+	auto pType = pParasiteTechno->GetTechnoType();
+	if (!pType)
+	{
+		return 0;
+	}
+	auto cell = MapClass::Instance->NearByLocation(pParasiteTechno->LastMapCoords,
+				pType->SpeedType, -1, pType->MovementZone, false, 1, 1, false,
+				false, false, true, CellStruct::Empty, false, false);
```
Coronia, line 67: auto pParasiteTechno = pParasite->Owner;

	if (bIsReturnSuccess || !pParasiteTechno)
		return 0;

	const auto pType = pParasiteTechno->GetTechnoType();

	if (!pType)
		return 0;

	auto cell = MapClass::Instance->NearByLocation(pParasiteTechno->LastMapCoords,
				pType->SpeedType, -1, pType->MovementZone, false, 1, 1, false,
				false, false, true, CellStruct::Empty, false, false);

Coronia (changes requested):
remember the following code styles:
- leave one blank line before and after if/switch sections, same for loop sections
- use a bracket {} to cover the loop sections, even if there's only 1 line in it
- add a const to variables that won't be modified

src/Misc/Hooks.BugFixes.cpp
TaranDahl: Fixed.

mevitar:
Without this fix, Terror Drones are destroyed if they miss and the previous space is occupied. After this fix, they respawn in the nearest free spot, unoccupied by buildings or units.
It seems to work fine, didn't notice any issues. Didn't test in multiplayer, though.

src/Misc/Hooks.BugFixes.cpp
```diff
@@ -45,6 +45,33 @@ DEFINE_JUMP(LJMP, 0x545CE2, 0x545CE9) //Phobos_BugFixes_Tileset255_RemoveNonMMAr
 DEFINE_JUMP(LJMP, 0x546C23, 0x546C8B) //Phobos_BugFixes_Tileset255_RefNonMMArray
 
 
+//Fix the bug that parasite will vanish if it missed its target when its previous cell is occupied.
+DEFINE_HOOK(0x62AA32, ParasiteClass_TryInfect_MissBehaviorFix, 0x5)
+{
+	GET(DWORD, dwdIsReturnSuccess, EAX);
+	bool bIsReturnSuccess = (BYTE)((WORD)(dwdIsReturnSuccess));
```
CrimRecya, line 52: I think this can simply change to `GET(bool, bIsReturnSuccess, EAX);` ?

CrimRecya (changes requested):

src/Misc/Hooks.BugFixes.cpp
TaranDahl: Changed. Didn't tested, though.

CrimRecya (approved):

PR #1344, opened 2024-08-08
Voxel light source position/tilting fix

Metadorius:
- Vanilla game applies some weird unnecessary math which resulted in light being applied incorrectly on tilted voxels and the light source being "nudged" up by a bit. It is now possible to fix that.

In `rulesmd.ini`:
```ini
[AudioVisual]
UseFixedVoxelLighting=false ; boolean, whether to fix the lighting
```


Starkku (approved):
LGTM.

PR #1345, opened 2024-08-09
[Customize] Weapon range finding in cylinder

TaranDahl:
### Range finding in cylinder

- In vanilla, technos in air will ignore the distance in Z axis when checking if the target is in range. Now you can use the following flags to make technos always range finding like that.
- `[General]->CylinderRangefinding` controls this globally, and can be customized per weapon type.
- Mind that set the flags to `false` meaning "use default" rather than "disable". Technos in air will always range finding in cylinder like vanilla, despite what you set.

In `rulesmd.ini`:
```ini
[General]                         ; WeaponType
CylinderRangefinding=false  ; boolean

[SOMEWEAPON]                      ; WeaponType
CylinderRangefinding=       ; boolean
```


src/Ext/Techno/Hooks.Misc.cpp
```diff
@@ -98,3 +99,27 @@ DEFINE_HOOK(0x6B7600, SpawnManagerClass_AI_InitDestination, 0x6)
...
+	}
+	if (bAlwaysCylinderRangefinding)
+	{
+		R->AL(true);
+	}
+	else
+	{
+		R->AL(pThis->IsInAir()); // vanilla check
+	}
+	return ret;
```
Coronia, line 124: auto const pWeaponExt = WeaponTypeExt::ExtMap.Find(pWeaponType);
	bool bAlwaysCylinderRangefinding = RulesExt::Global()->AlwaysCylinderRangefinding;

	if (pWeaponExt)
		bAlwaysCylinderRangefinding = pWeaponExt->AlwaysCylinderRangefinding.Get(bAlwaysCylinderRangefinding);

	if (bAlwaysCylinderRangefinding)
		R->AL(true);
	else
		R->AL(pThis->IsInAir()); // vanilla check

	return ret;

Coronia (commented):
Perhaps we don't need to add an Always to the tag name. CylinderRangefinding would be clear enough I think

Coronia (changes requested):
forget to mark it as request changes

Fryone (commented):
Looks fine logically. Also, can be named `UseCylinderRangefinding` for weapontypes

TaranDahl:
Todo: check if there is any Phobos function that has its own range-finding impl.
I was working on the DamageArea-related works last day and found that Phobos has its own cellspread function for some effects.
Need to check if this is happening on weapon range finding.

TaranDahl:
> Perhaps we don't need to add an Always to the tag name. CylinderRangefinding would be clear enough I think

Changed.

TaranDahl:
> Looks fine logically. Also, can be named `UseCylinderRangefinding` for weapontypes

Renamed to `CylinderRangefinding`

CrimRecya (approved):

TaranDahl:
Only handles the vanilla weapon rangefinding, at least for now.
Other rangefinding operations should be considered later, if necessary.

TaranDahl:
There is currently a bug suspected to be caused by this feature, which can cause misbehavior in judging the distance when the firing unit is located on the bridge.

CrimRecya (approved):

TaranDahl:
Going to merge this in 24h if no one has further issues.

PR #1347, opened 2024-08-10
Fix unit get stuck in factory

CrimRecya:
- Units are now unable to kick out from a factory that is in construction process, and will not always stuck in the factory. If the unit is not able to kick out from the factory, the unit will be removed and the cost will be given back to its house.

### DIFF
--- a/src/Ext/Building/Body.cpp
+++ b/src/Ext/Building/Body.cpp
@@ -334,13 +334,63 @@ bool BuildingExt::ExtData::HandleInfiltrate(HouseClass* pInfiltratorHouse,int mo
 		int idx = this->TypeExtData->SpyEffect_VictimSuperWeapon;
 		if (idx >= 0)
 			launchTheSWHere(pVictimHouse->Supers.Items[idx], pVictimHouse);
-		
+
 		idx = this->TypeExtData->SpyEffect_InfiltratorSuperWeapon;
 		if (idx >= 0)
 			launchTheSWHere(pInfiltratorHouse->Supers.Items[idx], pInfiltratorHouse);
 	}
 
 	return true;
+}
+
+void BuildingExt::ExtData::KickOutStuckUnits()
+{
+	if (Unsorted::CurrentFrame % 15)
+		return;
+
+	BuildingClass* const pThis = this->OwnerObject();
+
+	if (pThis->GetCurrentMission() == Mission::Unload)
+		return;
+
+	BuildingTypeClass* const pType = pThis->Type;
+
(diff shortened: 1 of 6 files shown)


src/Ext/Building/Body.cpp
```diff
@@ -343,6 +343,33 @@ bool BuildingExt::ExtData::HandleInfiltrate(HouseClass* pInfiltratorHouse,int mo
...
+			if (pObject->WhatAmI() == AbstractType::Unit)
+			{
+				UnitClass* const pUnit = static_cast<UnitClass*>(pObject);
+
+				if (pThis->Owner != pUnit->Owner)
+					continue;
+
+				const int height = pUnit->GetHeight();
+
+				if (height < 0 || height > 208)
```
Metadorius, line 361: Wasn't there a constant for 208? Somewhere in `CellClass` IIRC

src/Ext/Building/Hooks.cpp
```diff
@@ -324,6 +324,34 @@ DEFINE_HOOK(0x443CCA, BuildingClass_KickOutUnit_AircraftType_Phobos, 0xA)
 	return 0;
 }
 
+// Kick out stuck units when the factory building is not busy
+DEFINE_HOOK(0x450248, BuildingClass_UpdateFactory_KickOutStuckUnits, 0x6)
+{
+	GET(BuildingClass*, pThis, ESI);
+
+	if (!(Unsorted::CurrentFrame % 15) && pThis->GetCurrentMission() == Mission::Guard)
```
Metadorius, line 332: Could you explain what is the point of `if (!(Unsorted::CurrentFrame % 15)...` here? It's not necessarily bad, I just don't understand the point yet.

src/Ext/Building/Body.cpp
Starkku: `Unsorted::CellHeight`

CrimRecya: Get it. thx :)

src/Ext/Building/Hooks.cpp
CrimRecya: > Could you explain what is the point of `if (!(Unsorted::CurrentFrame % 15)...` here? It's not necessarily bad, I just don't understand the point yet.

Because this is in the factory update and I think this is no need to check this in every frames, so I took the initiative to have it checked every 15 frames.

Coronia (approved):
other than the 15 frames checking interval might be a bit questionable, the rest are good

PR #1349, opened 2024-08-10
Spawns Enhancement

Coronia:
Help TwinkleStar to upload these new features

- It is now possible to set the initial amount of spawnees for a spawner, instead of always being filled. Won't work if it's larger than `SpawnsNumber`.

In `rulesmd.ini`:
```ini
InitialSpawnsNumber=      ; integer
```

- It is now possible to spawn multiple types of spawnees from a spawner with `Spawns.Queue`. The order of spawnees in this queue is the order of their respawn.
  - `Spawns` still needs to be set to enable the spawner logic and act as a default spawnee.
  - `SpawnsNumber` still needs to be set to determine the amount of spawnee slots.
  - If the length of the queue is longer than the spawner's `SpawnsNumber`, spawnee after this length will be omitted. If the length is shorter however, the rest of the positions will be filled by the spawnee defined by `Spawns`. Hence, it's recommended to make them the same.

In `rulesmd.ini`:
```ini
[SOMETECHNO]        ; TechnoType
Spawns.Queue=       ; list of aircrafts in order
```

An example of `Spawns.Queue`:
```ini
[SOMETECHNO]
Spawns=AIRCRAFT1
SpawnsNumber=5
Spawns.Queue=AIRCRAFT1,AIRCRAFT2,AIRCRAFT1,AIRCRAFT2,AIRCRAFT3
```
In this case, it'll be built with 5 spawnees in the order of AIRCRAFT1->AIRCRAFT2->AIRCRAFT1->AIRCRAFT2->AIRCRAFT3, and so will be its respawn order if multiple of them being destroyed


MortonPL (changes requested):
The current implementation might contain a bug.

MortonPL (approved):

PR #1350, opened 2024-08-12
Two skirmish AI behavior dehardcoded and a minor improvement

TaranDahl:
### Skirmish AI behavior dehardcode

- In vanilla, there is a hardcoded behavior that when an skirmish AI player has no factory and has not taken damage for a while, it will sell its buildings and set its units to hunt. This can be customized now.
  - `[General]->AISellAllOnLastLegs` and `[General]->AIAllInOnLastLegs` control whether the AI will act selling and hunting respectively.
  - `[General]->AISellAllDelay` defines a timer, it will only work if `[General]->AISellAllOnLastLegs` is set to `true`. When the first time the AI reaches the trigger condition of vanilla behavior, the timer starts and prevents the selling behavior from happening until the timer is expired.
  - You can use these flags to make the AIs "all in" before they are defeated.
- Another hardcoded behavior is that, when the AI deploys a MCV, it will regroup all of its forces to that place. This can be toggle off now.
  - `[General]->RegroupWhenMCVDeploy` controls this behavior.

In `rulesmd.ini`:
```ini
[General]
AISellAllOnLastLegs=true   ; boolean
AISellAllDelay=0           ; integer, number of frames
AIAllInOnLastLegs=true     ; boolean
RegroupWhenMCVDeploy=true  ; boolean
```
### Global setting for `[Country House]->RepairBaseNodes` and `[Basic]->MCVRedeploys`

- Self-evident.

### DIFF
--- a/docs/AI-Scripting-and-Mapping.md
+++ b/docs/AI-Scripting-and-Mapping.md
@@ -52,12 +52,18 @@ DefaultLS800BkgdPal=     ; filename - including the .pal extension
 
 ### MCV redeploying
 
-- You can now decide whether MCV can redeploy in singleplayer campaign missions by setting `MCVRedeploys`. Overrides `[MultiplayerDialogSettings]`->`MCVRedeploys` only in singleplayer campaign missions.
+- You can now decide whether MCV can redeploy in singleplayer campaign missions by setting `[Basic]->MCVRedeploys` in map file. Overrides `[MultiplayerDialogSettings]`->`MCVRedeploys` only in singleplayer campaign missions. You can also define this globally by setting `[General]->MCVRedeploysInCampaign` in rulesmd.ini. The flag defined in map file has higher priority.
+
+In rulesmd.ini:
+```ini
+[General]
+MCVRedeploysInCampaign=false  ; boolean
+```
 
 In map file:
 ```ini
 [Basic]
-MCVRedeploys=false  ; boolean
+MCVRedeploys=                 ; boolean
 ```
 
 ### Set par times and related string labels in missionmd.ini
--- a/docs/AI-Scripting-and-Mapping.md
+++ b/docs/AI-Scripting-and-Mapping.md
@@ -18,12 +18,19 @@ This page describes all AI scripting and mapping related additions and changes i
 
 ### Base node repairing
 
-- In singleplayer campaign missions you can now decide whether AI can repair the base nodes / buildings delivered by SW (Ares) by setting `RepairBaseNodes`.
+- In singleplayer campaign missions you can now decide whether AI can repair the base nodes / buildings delivered by SW (Ares).
+  - You can control it globally by setting `[General]->RepairBaseNodes` in rulesmd.ini, or locally by setting the flag with same name in `[Some House]` in certain map file. The global one will be overriden if the local one is set.
+
+In rulesmd.ini:
+```ini
+[General]
+RepairBaseNodes=false,false,false  ; list of 3 booleans indicating whether AI repair basenodes in Easy / Normal / Difficult game diffculty.
+```
 
 In map file:
 ```ini
 [Country House]
-RepairBaseNodes=false,false,false  ; list of 3 booleans indicating whether AI repair basenodes in Easy / Normal / Difficult game diffculty.
+RepairBaseNodes=                   ; list of 3 booleans indicating whether AI repair basenodes in Easy / Normal / Difficult game diffculty.
 ```
 
 ### Default loading screen and briefing offsets
--- a/docs/Whats-New.md
+++ b/docs/Whats-New.md
@@ -687,8 +687,8 @@ New:
 - Animated (non-tiberium spawning) TerrainTypes (by Starkku)
 - Toggleable passenger killing for Explodes=true units (by Starkku)
 - New condition for automatic self-destruction logic when TechnoTypes exist/don't exist (by FlyStar)
-- Skirmish AI "sell all and all in" behavior dehardcode (by 航味麻酱)
-- Skirmish AI "regroup when mcv deploy" behavior dehardcode (by 航味麻酱)
+- Skirmish AI "sell all buildings and set all technos to hunt" behavior dehardcode (by 航味麻酱)
+- Skirmish AI "gather when mcv deploy" behavior dehardcode (by 航味麻酱)
 
 Vanilla fixes:
 - Fixed laser drawing code to allow for thicker lasers in house color draw mode (by Kerbiter, ChrisLv_CN)
--- a/docs/Whats-New.md
+++ b/docs/Whats-New.md
@@ -714,7 +714,7 @@ New:
 - Toggleable passenger killing for Explodes=true units (by Starkku)
 - New condition for automatic self-destruction logic when TechnoTypes exist/don't exist (by FlyStar)
 - Skirmish AI "sell all buildings and set all technos to hunt" behavior dehardcode (by 航味麻酱)
-- Skirmish AI "gather when mcv deploy" behavior dehardcode (by 航味麻酱)
+- Skirmish AI "gather when MCV deploy" behavior dehardcode (by 航味麻酱)
 
 Vanilla fixes:
 - Fixed laser drawing code to allow for thicker lasers in house color draw mode (by Kerbiter, ChrisLv_CN)
--- a/docs/Whats-New.md
+++ b/docs/Whats-New.md
@@ -713,8 +713,8 @@ New:
 - Animated (non-tiberium spawning) TerrainTypes (by Starkku)
 - Toggleable passenger killing for Explodes=true units (by Starkku)
 - New condition for automatic self-destruction logic when TechnoTypes exist/don't exist (by FlyStar)
-- Skirmish AI "sell all buildings and set all technos to hunt" behavior dehardcode (by 航味麻酱)
-- Skirmish AI "gather when MCV deploy" behavior dehardcode (by 航味麻酱)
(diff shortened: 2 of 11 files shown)


Coronia (changes requested):
The tags are not very understandable, suggest to change them to the names that being universally accepted by modders

AISellAllOnLastLegs -> AIFireSale
AISellAllDelay -> AIFireSaleDelay
AIAllInOnLastLegs -> AIAllToHunt
RegroupWhenMCVDeploy -> GatherWhenMCVDeploy

TaranDahl:
> The tags are not very understandable, suggest to change them to the names that being universally accepted by modders
> 
> AISellAllOnLastLegs -> AIFireSale AISellAllDelay -> AIFireSaleDelay AIAllInOnLastLegs -> AIAllToHunt RegroupWhenMCVDeploy -> GatherWhenMCVDeploy

Changed.

src/Ext/House/Hooks.cpp
```diff
@@ -271,3 +271,34 @@ DEFINE_HOOK(0x50B669, HouseClass_ShouldDisableCameo, 0x5)
...
+
+// Sell all and all in.
+DEFINE_HOOK(0x4FD8F7, HouseClass_UpdateAI_OnLastLegs, 0x10)
+{
+	enum { ret = 0x4FD907 };
+
+	GET(HouseClass*, pThis, EBX);
+
+	auto const pRules = RulesExt::Global();
+	auto const pExt = HouseExt::ExtMap.Find(pThis);
```
CrimRecya, line 283: this `pExt` seems like you can put it inside the next `if` after check `pRules->AIFireSale` is true.

src/Ext/Techno/Hooks.Misc.cpp
```diff
@@ -98,3 +98,15 @@ DEFINE_HOOK(0x6B7600, SpawnManagerClass_AI_InitDestination, 0x6)
...
 	return R->Origin() == 0x6B7600 ? SkipGameCode1 : SkipGameCode2;
 }
+
+// I must not regroup my forces.
+DEFINE_HOOK(0x739920, UnitClass_TryToDeploy_DisableRegroupAtNewConYard, 0x6)
+{
+	enum { SkipRegroup = 0x73992B, DoNotSkipRegroup = 0 };
+
+	auto const pRules = RulesExt::Global();
+	if (!pRules->GatherWhenMCVDeploy)
```
CrimRecya, line 108: I think we need a blank line before this `if`.
In addition, you can just merge these into one line `return RulesExt::Global()->GatherWhenMCVDeploy ? 0 : SkipRegroup;`

CrimRecya (changes requested):

src/Ext/House/Hooks.cpp
TaranDahl: Both have been fixed.

CrimRecya (approved):

src/Ext/Techno/Hooks.Misc.cpp
```diff
@@ -98,3 +98,16 @@ DEFINE_HOOK(0x6B7600, SpawnManagerClass_AI_InitDestination, 0x6)
...
+DEFINE_HOOK(0x739920, UnitClass_TryToDeploy_DisableRegroupAtNewConYard, 0x6)
+{
+	enum { SkipRegroup = 0x73992B, DoNotSkipRegroup = 0 };
+
+	auto const pRules = RulesExt::Global();
+
+	if (!pRules->GatherWhenMCVDeploy)
+		return SkipRegroup;
+	else
+		return DoNotSkipRegroup;
```
MortonPL, line 112:
```suggestion
	 return pRules->GatherWhenMCVDeploy ? DoNotSkipRegroup : SkipRegroup;
```

MortonPL (approved):

docs/Whats-New.md
```diff
@@ -713,6 +713,8 @@ New:
 - Animated (non-tiberium spawning) TerrainTypes (by Starkku)
 - Toggleable passenger killing for Explodes=true units (by Starkku)
 - New condition for automatic self-destruction logic when TechnoTypes exist/don't exist (by FlyStar)
+- Skirmish AI "sell all buildings and set all technos to hunt" behavior dehardcode (by 航味麻酱)
+- Skirmish AI "gather when MCV deploy" behavior dehardcode (by 航味麻酱)
```
MortonPL, line 717:
```suggestion
- Skirmish AI "sell all buildings and set all technos to hunt" behavior dehardcode (by TaranDahl/航味麻酱)
- Skirmish AI "gather when MCV deploy" behavior dehardcode (by TaranDahl/航味麻酱)
```

MortonPL (approved):

docs/AI-Scripting-and-Mapping.md
```diff
@@ -18,12 +18,19 @@ This page describes all AI scripting and mapping related additions and changes i
...
 ### Base node repairing
 
-- In singleplayer campaign missions you can now decide whether AI can repair the base nodes / buildings delivered by SW (Ares) by setting `RepairBaseNodes`.
+- In singleplayer campaign missions you can now decide whether AI can repair the base nodes / buildings delivered by SW (Ares).
+  - You can control it globally by setting `[General]->RepairBaseNodes` in rulesmd.ini, or locally by setting the flag with same name in `[Some House]` in certain map file. The global one will be overriden if the local one is set.
+
+In rulesmd.ini:
+```ini
+[General]
+RepairBaseNodes=false,false,false  ; list of 3 booleans indicating whether AI repair basenodes in Easy / Normal / Difficult game diffculty.
```
chaserli, line 27: I don't know how practical would it be to still have 3 here if you have to create a global one.  I think 1 should be sufficient and just let countries' default to that.

If not then the way you shouldn't have parsed them there with such "method"

docs/AI-Scripting-and-Mapping.md
```diff
@@ -45,12 +52,18 @@ DefaultLS800BkgdPal=     ; filename - including the .pal extension
 
 ### MCV redeploying
 
-- You can now decide whether MCV can redeploy in singleplayer campaign missions by setting `MCVRedeploys`. Overrides `[MultiplayerDialogSettings]`->`MCVRedeploys` only in singleplayer campaign missions.
+- You can now decide whether MCV can redeploy in singleplayer campaign missions by setting `[Basic]->MCVRedeploys` in map file. Overrides `[MultiplayerDialogSettings]`->`MCVRedeploys` only in singleplayer campaign missions. You can also define this globally by setting `[General]->MCVRedeploysInCampaign` in rulesmd.ini. The flag defined in map file has higher priority.
+
+In rulesmd.ini:
+```ini
+[General]
+MCVRedeploysInCampaign=false  ; boolean
```
chaserli, line 60: do you really need to make a 3rd new tag doing the exact same thing? wouldn't `[Basic]` section is parsed in rules as well as the included files? One of these 2 should be removed.

docs/Whats-New.md
```diff
@@ -713,6 +713,8 @@ New:
 - Animated (non-tiberium spawning) TerrainTypes (by Starkku)
 - Toggleable passenger killing for Explodes=true units (by Starkku)
 - New condition for automatic self-destruction logic when TechnoTypes exist/don't exist (by FlyStar)
+- Skirmish AI "sell all buildings and set all technos to hunt" behavior dehardcode (by TaranDahl/航味麻酱)
```
chaserli, line 716: you and your avatar?

chaserli (commented):
On the inconsistency of modifying the existing logic

docs/Whats-New.md
Metadorius: > you and your avatar?

elaborate

docs/AI-Scripting-and-Mapping.md
Metadorius: what's wrong with having this togglable per difficulty?

docs/Whats-New.md
chaserli: it looks like 2 people

docs/AI-Scripting-and-Mapping.md
chaserli: > what's wrong with having this togglable per difficulty?

it explictly ruined my poka-yoke back then. If they were to differ, under different difficulties, allies and enemies houses should have different values

Metadorius: > allies and enemies houses should have different values
you do realize "ally" and "enemy" for different players mean different things, leading to a desync? this can't be physically done unless you're in SP, which would be an exception that is kinda pointless to have. besides, params per difficulty is a normal thing to do

chaserli: > you do realize "ally" and "enemy" for different players mean different things, leading to a desync? this can't be physically done unless you're in SP, which would be an exception that is kinda pointless to have. besides, params per difficulty is a normal thing to do

It's overriding my existing tags, written right below that. And yes, it's for *campaign mode only*.
It has nothing to do with difficulty.

docs/AI-Scripting-and-Mapping.md
TaranDahl: The document states that the tag needs to be written in the map file. I haven't tested whether it can be written in rulesmd. If it works,  the new tag is unnecessary then.

docs/AI-Scripting-and-Mapping.md
TaranDahl: > I don't know how practical would it be to still have 3 here if you have to create a global one. I think 1 should be sufficient and just let countries' default to that.
> 
> If not then the way you shouldn't have parsed them there with such "method"

I just don't want to write it on every map and every house, so I added the same global tag to control it.
Reduce to 1 value is sufficient for me.

PR #1354, opened 2024-08-13
[Minor] Reflect Damage Tweak

Coronia:
added a few more tags to further customize ReflectDamage

- `ReflectDamage.Chance` determines the chance of reflection.
- `ReflectDamage.Override` directly overrides the damage.
- `SuppressReflectDamage.Groups` does the same thing as `SuppressReflectDamage.Types` but for all AttachEffectTypes in the listed groups.

In `rulesmd.ini`:
```ini
[SOMEATTACHEFFECT]                             ; AttachEffectType
ReflectDamage.Chance=1.0                       ; floating point value
ReflectDamage.Override=                        ; integer

[SOMEWARHEAD]
SuppressReflectDamage.Groups=                  ; comma-separated list of strings (group IDs)
```


Starkku (changes requested):
Looks mostly fine, little bit unsure about clarity of `ReflectDamage.Override` as a key name but I can't think of a suitable replacement currently. There's one small thing I'd change though.

src/Ext/Techno/Hooks.AttachEffect.cpp
```diff
@@ -154,11 +155,14 @@ DEFINE_HOOK(0x701E18, TechnoClass_ReceiveDamage_ReflectDamage, 0x7)
 			if (!pType->ReflectDamage)
 				continue;
 
-			if (pWHExt->SuppressReflectDamage && pWHExt->SuppressReflectDamage_Types.Contains(pType))
+			if (pWHExt->SuppressReflectDamage && (pWHExt->SuppressReflectDamage_Types.Contains(pType) || pType->HasGroups(pWHExt->SuppressReflectDamage_Groups, false)))
+				continue;
+
+			if (pType->ReflectDamage_Chance < ScenarioClass::Instance->Random.RandomDouble())
 				continue;
```
Starkku, line 163: Move the chance check before suppression, it is faster to check it first and bail out early before checking types and groups if it fails.

Coronia: changed

PR #1355, opened 2024-08-14
[Minor] Attach & Discard AE by Health Percentage

Coronia:
- If `DiscardOn.AbovePercent` or `DiscardOn.BelowPercent` is set, the effect is discarded when the object's health percentage is above/below that value.
  - If `AffectAbovePercent` or `AffectBelowPercent` is set, the effect can be applied only when the object's health percentage is above/below that value.

In `rulesmd.ini`:
```ini
[SOMEATTACHEFFECT]                             ; AttachEffectType
DiscardOn.AbovePercent=                        ; floating point value, percents or absolute (0.0-1.0)
DiscardOn.BelowPercent=                        ; floating point value, percents or absolute (0.0-1.0)
AffectAbovePercent=                            ; floating point value, percents or absolute (0.0-1.0)
AffectBelowPercent=                            ; floating point value, percents or absolute (0.0-1.0)
```


Coronia:
Close it for now, will use condition group (https://github.com/Phobos-developers/Phobos/pull/1346) to reimplement it and add a few more conditions

PR #1365, opened 2024-08-25
Insignia.Passengers & InsigniaTypes

Coronia:
1 - added insignia based on current passenger size: Normal insignia can be overridden when its current passenger size reaches a certain amount by setting `Insignia(.Frame/.Frames).PassengersN` where `N` stands for the current passenger size amount. If not set, defaults to non-passenger specific insignia settings. Will be overridden by weapon mode insignia settings, if set.

In `rulesmd.ini`:
```ini
Insignia.PassengersN=                    ; filename - excluding the .shp extension
Insignia.PassengersN.Rookie=             ; filename - excluding the .shp extension
Insignia.PassengersN.Veteran=            ; filename - excluding the .shp extension
Insignia.PassengersN.Elite=              ; filename - excluding the .shp extension
InsigniaFrame.PassengersN=-1             ; int, frame of insignia shp (zero-based) or -1 for default
InsigniaFrame.PassengersN.Rookie=-1      ; int, frame of insignia shp (zero-based) or -1 for default
InsigniaFrame.PassengersN.Veteran=-1     ; int, frame of insignia shp (zero-based) or -1 for default
InsigniaFrame.PassengersN.Elite=-1       ; int, frame of insignia shp (zero-based) or -1 for default
InsigniaFrames.PassengersN=-1,-1,-1      ; int, frames of insignia shp (zero-based) or -1 for default
```

2 - This isn't a real logic, but more like an interface to make ini edit easier. Previously we have to define insignia by the techno's veterancy, which requires a lot of tags especially if it's for Gunner or using `InsigniaFrame(s)`. This pull request allows you to define an `InsigniaType` that stores all these data together, so that you can reuse them in different places with one line of code.

In `rulesmd.ini`:
```ini
[InsigniaTypes]
0=SOMEINSIGNIATYPE

[SOMEINSIGNIATYPE]						 ; InsigniaType
Insignia=                                ; filename - excluding the .shp extension
Insignia.Rookie=                         ; filename - excluding the .shp extension
Insignia.Veteran=                        ; filename - excluding the .shp extension
Insignia.Elite=                          ; filename - excluding the .shp extension
InsigniaFrame=-1                         ; int, frame of insignia shp (zero-based) or -1 for default
InsigniaFrame.Rookie=-1                  ; int, frame of insignia shp (zero-based) or -1 for default
InsigniaFrame.Veteran=-1                 ; int, frame of insignia shp (zero-based) or -1 for default
InsigniaFrame.Elite=-1                   ; int, frame of insignia shp (zero-based) or -1 for default

InsigniaType=                            ; InsigniaType
InsigniaType.WeaponN=                    ; InsigniaType
InsigniaType.PassengersN=                ; InsigniaType
```

An example:
```ini
[InsigniaTypes]
0=InsigniaHero

[InsigniaHero]
Insignia.Rookie=herostcr
Insignia.Veteran=herostbr
Insignia.Elite=herostar

[TANY]
InsigniaType=InsigniaHero
IFVMode=15

[VOLKOV]
InsigniaType=InsigniaHero
IFVMode=35

[FV]
InsigniaType.Weapon15=InsigniaHero
InsigniaType.Weapon35=InsigniaHero
```


src/Ext/TechnoType/Body.cpp
```diff
@@ -358,6 +362,37 @@ void TechnoTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
...
+			this->InsigniaFrame_Passengers.resize(passengers);
+			this->InsigniaFrames_Passengers.resize(passengers);
+		}
+
+		for (size_t i = 0; i < passengers; i++)
+		{
+			Promotable<SHPStruct*> insignia;
+			_snprintf_s(tempBuffer, sizeof(tempBuffer), "Insignia.Passengers%d.%s", i, "%s");
+			insignia.Read(exINI, pSection, tempBuffer);
+			this->Insignia_Passengers[i] = insignia;
```
Starkku, line 381: Not sure what happens here currently if the value is not found but ideally it shouldn't be reset if you redefine only some of the keys. Same for weapon insignia as well.

docs/Fixed-or-Improved-Logics.md
```diff
@@ -472,6 +472,7 @@ ChronoSparkleBuildingDisplayPositions=occupantslots  ; list of chrono sparkle po
   - `InsigniaFrame(.Rookie|Veteran|Elite)` can be used to set (zero-based) frame index of the insignia to display, optionally for each veterancy stage. Using -1 uses the default setting. Default settings are -1 (none) for rookie, 14 for veteran and 15 for elite.
     - A shorthand `InsigniaFrames` can be used to list them in order from rookie, veteran and elite instead as well. `InsigniaFrame(.Rookie|Veteran|Elite)` takes priority over this.
   - Normal insignia can be overridden for specific weapon modes of `Gunner=true` units by setting `Insignia(.Frame/.Frames).WeaponN` where `N` stands for 1-based weapon mode index. If not set, defaults to non-mode specific insignia settings.
+  - Normal insignia can be overridden when its current passenger size reaches a certain amount by setting `Insignia(.Frame/.Frames).PassengersN` where `N` stands for the current passenger size amount. If not set, defaults to non-passenger specific insignia settings. Will be overridden by weapon mode insignia settings, if set.
```
Starkku, line 475: >  where `N` stands for the current passenger size amount

Is this actually the case given that the parsing code starts from 0 and loops until passenger count?

CREDITS.md
```diff
@@ -350,6 +350,7 @@ This page lists all the individual contributions to the project by their author.
 - **Ollerus**
   - Build limit group enhancement
   - Customizable rocker amplitude
+  - Custom insignias
```
Starkku, line 353: Could maybe mention it is the passenger-based insignia here but it is kinda inconsistent currently across any multi-author features so not a big deal.

Starkku (changes requested):
Mostly fine, a couple of questions about some things though.

docs/Fixed-or-Improved-Logics.md
Coronia: It is, it could work with passenger amount from 0 (empty) to its upper limit. Clarified this a bit in the doc

src/Ext/TechnoType/Body.cpp
Coronia: changed InsigniaFrames into Nullable<Vector3D<int>> and use isset() for condition instead of equal to (-1, -1, -1). Works fine for now but not sure if it'll accidentally break something

Coronia:
image
Since there're tons of tags for Insignia already, I'm thinking of removing the old tags for Passengers while only keeping its InsigniaType, would like some feedbacks from others to see if that's proper tho

Also @Metadorius could you list what should be modified for Insignia Type at here again, now that the old pull request has been closed?

src/Ext/Techno/Body.Visuals.cpp
```diff
@@ -134,6 +134,28 @@ void TechnoExt::DrawInsignia(TechnoClass* pThis, Point2D* pLocation, RectangleSt
...
+		}
+
+		int frame = pTechnoTypeExt->InsigniaFrame_Passengers[passengersIndex].Get(pThis);
+
+		if (frame != -1)
+			frameIndex = frame;
+
+		auto& frames = pTechnoTypeExt->InsigniaFrames_Passengers[passengersIndex];
+
+		if (frames!= Vector3D<int>(-1, -1, -1))
```
CrimRecya, line 155: `if (frames != Vector3D<int>(-1, -1, -1))`

CrimRecya (approved):
Overall, LGTM. Only some minor things.

PR #1369, opened 2024-08-30
Rewrite frame CRC generation

Starkku:
Reimplements game's frame CRC generation function, only change logic-wise is that it now excludes `AnimClass` and `ParticleClass` objects that are purely visual.

### DIFF
--- a/docs/Whats-New.md
+++ b/docs/Whats-New.md
@@ -652,6 +652,7 @@ Vanilla fixes:
 - Enabled playing ingame movie in non-campaign modes (i.e. trigger action `100 Play Sidebar Movie...` and `117 Play Sidebar Movie and pause...`) (by TaranDahl)
 - `ElectricAssault` weapons can now auto acquire allies' overpowerable defenses (by Ollerus)
 - Fixed the issue that the time for units in the area guard mission to reacquire targets after eliminating the target is significantly longer than that in other missions (by TaranDahl)
+- Purely visual animations and particles excluded from sync checks (by Starkku)
 
 Phobos fixes:
 - Fixed the bug that `AllowAirstrike=no` cannot completely prevent air strikes from being launched against it (by NetsuNegi)
--- a/src/Misc/SyncLogging.cpp
+++ b/src/Misc/SyncLogging.cpp
@@ -543,24 +543,28 @@ DEFINE_HOOK(0x7013A0, TechnoClass_OverrideMission_SyncLog, 0x5)
 	return 0;
 }
 
-}
-
 #pragma region FrameCRC
 
-static bool IsHashable(ObjectClass* pObj)
+class ObjectFake : public ObjectClass
 {
-	auto const rtti = pObj->WhatAmI();
+public:
+	inline bool _IsCRCHashable();
+};
+
+bool ObjectFake::_IsCRCHashable()
+{
+	auto const rtti = this->WhatAmI();
 
 	if (rtti == AbstractType::Anim)
 	{
-		if (pObj->UniqueID == -2)
+		if (this->UniqueID == -2)
 			return false;
 
-		auto const pAnim = static_cast<AnimClass*>(pObj);
+		auto const pAnim = static_cast<AnimClass*>((ObjectClass*)this);
 		auto const pType = pAnim->Type;
 
 		if (pType->Damage != 0.0 || pType->Bouncer || pType->IsMeteor || pType->IsTiberium || pType->TiberiumChainReaction
-			|| pType->IsAnimatedTiberium || pType->MakeInfantry != -1 || AnimTypeExt::ExtMap.Find(pType)->CreateUnit)
+			|| pType->IsAnimatedTiberium || pType->MakeInfantry != -1 || AnimTypeExt::ExtMap.Find(pType)->CreateUnitType.get())
 		{
 			return true;
 		}
@@ -569,7 +573,7 @@ static bool IsHashable(ObjectClass* pObj)
 	}
 	else if (rtti == AbstractType::Particle)
 	{
-		auto const pParticle = static_cast<ParticleClass*>(pObj);
+		auto const pParticle = static_cast<ParticleClass*>((ObjectClass*)this);
 		auto const pType = pParticle->Type;
 
 		if (pType->Damage)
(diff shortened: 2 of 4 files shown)


FS-21:
What kind of tests should be done here if you want feedback?

src/Misc/SyncLogging.cpp
```diff
@@ -527,3 +530,96 @@ DEFINE_HOOK(0x683AB0, ScenarioClass_Start_DisableSyncLog, 0x6)
...
+static bool __forceinline DisallowObject(ObjectClass* pObj)
+{
+	auto const rtti = pObj->WhatAmI();
+
+	if (rtti == AbstractType::Anim)
+	{
+		if (pObj->UniqueID == -2)
+			return true;
+
+		auto const pAnim = abstract_cast<AnimClass*>(pObj);
```
chaserli, line 543:
```suggestion
		auto const pAnim = static_cast<AnimClass*>(pObj);
```

src/Misc/SyncLogging.cpp
```diff
@@ -527,3 +530,96 @@ DEFINE_HOOK(0x683AB0, ScenarioClass_Start_DisableSyncLog, 0x6)
...
+			|| pType->IsAnimatedTiberium || pType->MakeInfantry != -1 || AnimTypeExt::ExtMap.Find(pType)->CreateUnit)
+		{
+			return false;
+		}
+
+		return true;
+	}
+	else if (rtti == AbstractType::Particle)
+	{
+		auto const pParticle = abstract_cast<ParticleClass*>(pObj);
```
chaserli, line 556:
```suggestion
		auto const pParticle = static_cast<ParticleClass*>(pObj);
```

src/Misc/SyncLogging.cpp
```diff
@@ -527,3 +530,96 @@ DEFINE_HOOK(0x683AB0, ScenarioClass_Start_DisableSyncLog, 0x6)
...
+	}
+
+	return false;
+}
+
+DEFINE_HOOK(0x64DAB0, ComputeFrameCRC, 0x6)
+{
+	enum { SkipGameCode = 0x64DE7C };
+
+	auto& GameCRC = EventClass::CurrentFrameCRC.get();
```
chaserli, line 572:
```suggestion
	auto& GameCRC = EventClass::CurrentFrameCRC();
```

chaserli (changes requested):
If you're rewriting these completely, it would be better to hash these in a more modern way for better readability and extensibility

src/Misc/SyncLogging.cpp
```diff
@@ -527,3 +530,141 @@ DEFINE_HOOK(0x683AB0, ScenarioClass_Start_DisableSyncLog, 0x6)
 
 	return 0;
 }
+
+#pragma region mp_frame_crc
+
+static bool __forceinline ObjectNonCRCHashable(ObjectClass const* pObj)
```
ZivDero, line 536:
```cpp
class ObjectClassFake : public ObjectClass
{
	inline bool _IsCRCHashable();
}

bool ObjectClassFake::_IsCRCHashable
{
...
}
```

src/Misc/SyncLogging.cpp
```diff
@@ -527,3 +530,141 @@ DEFINE_HOOK(0x683AB0, ScenarioClass_Start_DisableSyncLog, 0x6)
...
+	{
+		if constexpr (Alt)
+			value ^= 0x9e3779b9 + (value << 6) + (value >> 2) + other;
+		else
+			value += (value >> 31) + (value << 1) + other;
+		return *this;
+	}
+};
+
+DEFINE_HOOK(0x64DAB0, ComputeFrameCRC, 0x6)
```
ZivDero, line 613: Please reimplement the whole function as a normal (non-hook) function and patch it with a simple jump for cleanliness.

src/Misc/SyncLogging.cpp
```diff
@@ -527,3 +530,141 @@ DEFINE_HOOK(0x683AB0, ScenarioClass_Start_DisableSyncLog, 0x6)
...
+			continue;
+
+		GameCRC->Add(pObj);
+	}
+
+	GameCRC->Add(ScenarioClass::Instance->Random.Random());
+
+#define FRAMECRC_HASH_INI
+
+#ifdef FRAMECRC_HASH_INI
```
ZivDero, line 656: No.

ZivDero (changes requested):

src/Misc/SyncLogging.cpp
ZivDero: Instead of creating global functions where one would normally create a class method, you could create a "fake" class descending from ObjectClass, allowing you to effectively add new (non-virtual, of course) class methods, with access to `this`.

src/Misc/SyncLogging.cpp
Metadorius: No.

Starkku:
The wrapper may be good idea in theory but the current implementation mostly just obfuscates what gets hashed and for what type of object. The benefits would be more obvious if there was a realistic chance for new properties that need hashing to be added for, say, TechnoClass and derivatives in the future but as it stands this is *very* unlikely. If you can think of a way to do it without extra overhead or the obfuscating nature while reducing repeated code be my guest, but the previous implementation was not it so I've elected to revert it. I do hold a local copy of it for now still just in case though.

I'm not even going to repeat the point about the rules hash check in Phobos anymore, should be obvious by now that it does not belong here.

I did do some minor cleanup to the first iteration.

docs/Whats-New.md
```diff
@@ -451,6 +451,7 @@ New:
 - Nonprovocative Warheads (by Starkku)
 - Option to restore `PowerSurplus` setting for AI (by Starkku)
 - `FireOnce` infantry sequence reset toggle (by Starkku)
+- Purely visual animations and particles excluded from sync checks (by Starkku)
```
secsome, line 454: I feel like this PR is more like a `Vanilla Fix` instead of a `New Feature`

chaserli:
> If you can think of a way to do it without extra overhead or the obfuscating nature while reducing repeated code be my guest
It only wrapped a DWORD and serves for hash the object on creation and do the hash combine operation. I bet it makes less overhead than your current method. It also allows you to know the "sum" at compile time. But anyway these are small things
 and make literally no difference

Starkku:
Rebased on develop and applied some of the feedback. This is ready for further testing and another review pass.

Coronia:
no desync or other issue occured in our testing so far

src/Misc/SyncLogging.cpp
```diff
@@ -541,3 +542,128 @@ DEFINE_HOOK(0x7013A0, TechnoClass_OverrideMission_SyncLog, 0x5)
...
+bool ObjectFake::_IsCRCHashable()
+{
+	auto const rtti = this->WhatAmI();
+
+	if (rtti == AbstractType::Anim)
+	{
+		if (this->UniqueID == -2)
+			return false;
+
+		auto const pAnim = static_cast<AnimClass*>((ObjectClass*)this);
```
ZivDero: Uh.... use `reinterpret_cast`? Or two `static_cast`, not a C cast.

src/Misc/SyncLogging.cpp
```diff
@@ -541,3 +542,128 @@ DEFINE_HOOK(0x7013A0, TechnoClass_OverrideMission_SyncLog, 0x5)
...
+static void __forceinline AddCRC(DWORD* crc, unsigned int val)
+{
+	*crc = val + (*crc >> 31) + (*crc << 1);
+}
+
+static int __forceinline GetCoordHash(CoordStruct location)
+{
+	return location.X / 10 + ((location.Y / 10) << 16);
+}
+
```
ZivDero: Any real need for `__forceinline`? Just mark it `inline`, should be more than enough. High chance it will get inlined even without either.

ZivDero (approved):
A couple nitpicks but looks good, see no issues.

src/Misc/SyncLogging.cpp
Metadorius: doesn't `inline` have different semantcis nowadays?

ZivDero: Like what? Not that I'm aware.

Metadorius: > The original intent of the inline keyword was to serve as an indicator to the optimizer that inline substitution of a function is preferred over function call, that is, instead of executing the function call CPU instruction to transfer control to the function body, a copy of the function body is executed without generating the call. This avoids overhead created by the function call (passing the arguments and retrieving the result) but it may result in a larger executable as the code for the function has to be repeated multiple times.

> Since inline substitution is unobservable in the standard semantics, compilers are free to use inline substitution for any function that's not marked inline, and are free to generate function calls to any function marked inline. Those optimization choices do not change the rules regarding multiple definitions and shared statics listed above.

> Because the meaning of the keyword inline for functions came to mean "multiple definitions are permitted" rather than "inlining is preferred" since C++98, that meaning was extended to variables.

Starkku: Realistically could probably remove both keywords and let the compiler do its thing. As if I know better than it.

PR #1370, opened 2024-08-30
Penetration damage on garrisonable structures

FS-21:
### Penetration damage on garrisonable structures

- Warheads can now damage garrisoned infantry at impact.
- `PenetratesGarrison` Enables the logic.
  - `PenetratesGarrison.RandomTarget` specifies if the damage will go at some random garrisoned soldier or if all infantry should be damaged at the same time.
  - `PenetratesGarrison.DamageMultiplier` can be used to modify the damage applied against the garrisoned infantry. A random percentage value will be picked between the specified range.
  - `PenetratesGarrison.CleanSound` can be used to specify a sound to play when the structure lost all the garrisoned soldiers with this logic.
  - `PenetratesGarrison.Allowed` can be set on garrisonable buildings to protect the garrisoned infantry, or used on garrisoned infantry to make them not be affected by this logic.

In `rulesmd.ini`:
```ini
[SOMEWARHEAD]                                 ; WarheadType
PenetratesGarrison=false                      ; boolean
PenetratesGarrison.RandomTarget=true          ; boolean
PenetratesGarrison.DamageMultiplier=1.0,1.0   ; floating point value - single or comma-sep. range (percentages)
PenetratesGarrison.CleanSound=                ; sound entry

[SOMETECHNO]                                  ; TechnoType
PenetratesGarrison.Allowed=true               ; boolean
```


FS-21:
Reopened because I finally tested it online and is free of desyncs.

FS-21:
garrison-penetration-01

Fryone:
Isn't it similar to Ares Pass Through? https://ares-developers.github.io/Ares-docs/new/buildings/urbancombattrenches.html
though, Ares did it through projectiles and buildings...

mevitar:
There were some issues with that garrison logic that i do not remember right now, and this feature here might allow for better control over what can get cleared and what not.
If anything, this one is much more intuitive.

src/Ext/WarheadType/Detonate.cpp
```diff
@@ -501,3 +505,76 @@ double WarheadTypeExt::ExtData::GetCritChance(TechnoClass* pFirer) const
...
+	if (!pBuilding || !pBuilding->Occupants.Count || !pBulletExt)
+		return;
+
+	auto const pTargetTypeExt = TechnoTypeExt::ExtMap.Find(pTarget->GetTechnoType());
+	if (pTargetTypeExt->ImmuneToGarrisonPenetration)
+		return;
+
+	auto const pBullet = pBulletExt->OwnerObject();
+	double multiplierPercentage = GeneralUtils::GetRangedRandomOrSingleValue(this->GarrisonPenetration_DamageMultiplier);
+	int damage = static_cast<int>(std::round(pBullet->WeaponType->Damage * multiplierPercentage));
```
Coronia, line 521: should multiply the invoker's FirepowerMultiplier and AE.FirepowerMultiplier

src/Ext/WarheadType/Detonate.cpp
```diff
@@ -501,3 +505,76 @@ double WarheadTypeExt::ExtData::GetCritChance(TechnoClass* pFirer) const
...
+	auto DoDamage = [=](InfantryClass* pPassenger, int damage)
+	{
+		auto const pTypeExt = TechnoTypeExt::ExtMap.Find(pPassenger->GetTechnoType());
+		if (pTypeExt->ImmuneToGarrisonPenetration)
+			return;
+
+		damage = MapClass::GetTotalDamage(damage, pWH, pPassenger->Type->Armor, 0);
+
+		pPassenger->Health = pPassenger->Health - damage;
+		pPassenger->Health = pPassenger->Health < 0 ? 0 : pPassenger->Health;
```
Coronia, line 538: merging these 2 lines into 1:
```
pPassenger->Health = Math::clamp(0, pPassenger->Health - damage, pPassenger->Type->Strength);
```

Coronia (changes requested):
suggest to change these tags' names into `PenetratesGarrison`, to keep consistent with the already exists `PenetratesBunker` tag

src/Ext/WarheadType/Detonate.cpp
```diff
@@ -501,3 +505,73 @@ double WarheadTypeExt::ExtData::GetCritChance(TechnoClass* pFirer) const
...
+
+	int occupantIndex = this->PenetratesGarrison_RandomTarget ? ScenarioClass::Instance->Random.RandomRanged(0, pBuilding->Occupants.Count - 1) : -1;
+
+	auto DoDamage = [=](InfantryClass* pPassenger, int damage)
+	{
+		auto const pTypeExt = TechnoTypeExt::ExtMap.Find(pPassenger->GetTechnoType());
+		if (pTypeExt->ImmuneToPenetratesGarrison)
+			return;
+
+		damage = MapClass::GetTotalDamage(damage, pWH, pPassenger->Type->Armor, 0);
```
Coronia, line 532: should consider distance between epicenter as well

src/Ext/WarheadType/Detonate.cpp
FS-21: > pPassenger->Health = Math::clamp(pPassenger->Health - damage, 0, pPassenger->Type->Strength);

I don't understand the use of pPassenger->Type->Strength here.

src/Ext/WarheadType/Detonate.cpp
```diff
@@ -501,3 +505,73 @@ double WarheadTypeExt::ExtData::GetCritChance(TechnoClass* pFirer) const
...
+		for (int i = 0; i < pBuilding->Occupants.Count; i++)
+		{
+			InfantryClass* pPassenger = pBuilding->Occupants.GetItem(i);
+			DoDamage(pPassenger, damage);
+		}
+	}
+	else
+	{
+		InfantryClass* pPassenger = pBuilding->Occupants.GetItem(occupantIndex);
+		DoDamage(pPassenger, damage);
```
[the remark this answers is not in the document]
FS-21: Don't make me remember why 10 months ago I saw something that forced me to do this, I don't have that much memory sorry :-/

src/Ext/WarheadType/Detonate.cpp
FS-21: They are infantry gathered together inside a structure, I don't understand the point :-/

src/Ext/WarheadType/Detonate.cpp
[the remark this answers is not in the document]
TaranDahl: > Don't make me remember why 10 months ago I saw something that forced me to do this, I don't have that much memory sorry :-/

Remember to leave enough comments when coding next time 😝

src/Ext/WarheadType/Detonate.cpp
Coronia: since we're now supporting negative damage for PenetrateGarrison, this is for preventing it from gaining more health than its limit. It can be ignored if there're already some other sorts of check elsewhere tho

src/Ext/WarheadType/Detonate.cpp
Coronia: it's about judging the distance between the warhead's detonation point and the garrisoned building, which is still necessary. Imagine the case when using a neutron bomb with large cellspread on these buildings, it's expected that infantry garrisoned in the edge should take less damage than those in the center

src/Ext/WarheadType/Detonate.cpp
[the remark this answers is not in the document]
FS-21: Yes that would prevent future questions of the chosen choices in some very old mantained PRs, I'll take note :-S

src/Ext/WarheadType/Detonate.cpp
FS-21: What I mean, since I never used that clamp method before:

Example:
Passenger->Health = 50
damage = -25 (healer example)
pPassenger->Type->Strength = 300

Current code:
Passenger->Health = 50 - (-25) = 75;
pPassenger->Health = 75 < 0 ? 0 : 75;

That new line for merging code:
pPassenger->Health = Math::clamp(75, 0, 300;
It does return "pPassenger->Health = 75" ? Just ignorant question

Coronia: It does, clamp will return a value that's between minimum and maximum values, which are 0 and 300 in this case

src/Ext/WarheadType/Detonate.cpp
Coronia: we can simply calculate the distance between the warhead's detonation point and the garrisoned building in this case

src/Ext/WarheadType/Detonate.cpp
FS-21: But the right answer of the example is 75, not 300 :-/

src/Ext/WarheadType/Detonate.cpp
FS-21: Not sure how should be done since I see no similar examples in Phobos.
Is this?

int distanceFromCenter = pTarget->DistanceFrom(pPassenger);
damage = MapClass::GetTotalDamage(damage, pWH, pPassenger->Type->Armor, distanceFromCenter);

or:

int distanceFromCenter = pBullet->DistanceFrom(pTarget);
damage = MapClass::GetTotalDamage(damage, pWH, pPassenger->Type->Armor, distanceFromCenter);

or:

int distanceFromCenter = pBullet->DistanceFrom(pPassenger);
damage = MapClass::GetTotalDamage(damage, pWH, pPassenger->Type->Armor, distanceFromCenter);

or none of the above?

Coronia: should be the second one

Coronia:
the damage part should be done by a proper `ReceiveDamage` function, or it'll create various of issues in edge cases

FS-21:
> the damage part should be done by a proper `ReceiveDamage` function, or it'll create various of issues in edge cases

I still don't understand what is suggested, I need to hook inside ReceiveDamage and apply this damage to kill infantry?

Coronia:
You need to do the damage through `ReceiveDamage`, instead of doing so by just reducing HP. Otherwise there'll be leftover things to handle manually which can be risky

FS-21:
probably is done

fagonghaiwo:
[SOMETECHNO]                                  ; TechnoType
PenetratesGarrison.Allowed=false              ; boolean

maybe default value is true

FS-21:
> [SOMETECHNO] ; TechnoType PenetratesGarrison.Allowed=false ; boolean
> 
> maybe default value is true

Yes, the first message is very old. I'll update the first comment but always go to the documentation that must be updated to reflect what does the game tags

PR #1371, opened 2024-08-31
Drop Crate on Death

FS-21:
- Ìf `DropCrate` is declared then the specified crate will be spawned when the object is destroyed.

In `rulesmd.ini`:
```ini
[SOMETECHNO]  ; TechnoType
DropCrate=    ; Powerup crate type enum (money|unit|healbase|cloak|explosion|napalm|squad|reveal|armor|speed|firepower|icbm|invulnerability|veteran|ionstorm|gas|tiberium|pod)
```

Also, a new map action (see the documentation) allows the overwrite of this behaviour (disbale, enable or set a new crate) if the trigger is attached to technos.

Tested in online games so this doesn't desync.


MortonPL (commented):
Won't collecting this new crate spawn a new one according to random crate logic?

src/Ext/Techno/Hooks.cpp
```diff
@@ -360,6 +360,31 @@ DEFINE_HOOK(0x4DB218, FootClass_GetMovementSpeed_SpeedMultiplier, 0x6)
...
+			nSelectedPowerup = pTypeExt->DropCrate.isset() ? static_cast<int>(pTypeExt->DropCrate.Get()) : -1;
+		}
+
+		if (nSelectedPowerup < 0)
+			return 0;
+
+		Powerup selectedPowerup = static_cast<Powerup>(nSelectedPowerup);
+		TechnoExt::TryToCreateCrate(pThis->Location, selectedPowerup);
+	}
+
```
Coronia, line 387: the entire section should be put under a proper techno death handling hook in `Hooks.ReceiveDamage`, otherwise it might have the chance to miss or duplicate the crate

Coronia (changes requested):

src/Ext/Techno/Hooks.cpp
FS-21: First of all, why all the code calling this feature was moved from the hook "DEFINE_HOOK(0x701DFF, TechnoClass_ReceiveDamage_FlyingStrings, 0x7)" (initial commit I created) to "DEFINE_HOOK(0x4DB218, FootClass_GetMovementSpeed_SpeedMultiplier, 0x6)" (current commit)...

Now I understand the code started to drop random instead of a specific one stuff some time ago.
Originally it was as you mentioned Coronia. I'll push a fix moving the code to the original hook.

FS-21: Now should be fixed

TaranDahl (changes requested):
Two concerns.

src/Ext/Techno/Body.h
```diff
@@ -71,6 +71,9 @@ class TechnoExt
 
 		AirstrikeClass* AirstrikeTargetingMe;
 
+		int DropCrate; // Drop crate on death, modified by map action
```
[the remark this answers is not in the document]
FS-21: Ehh, this PR has map actions that modify the behaviour of drop crates enabling/disabling and modifying the current value

Starkku:
Is there any existing vanilla function that is suitable for finding available cells for crate spawning instead of implementing your own code?

TaranDahl (approved):

Coronia (approved):

PR #1374, opened 2024-09-04
[General Purpose] New Parabola trajectory

CrimRecya:
- As the name says, this is a completely reset `Arcing` with different enhanced functions. Without doubt, It supported linkage with `Trajectory=Disperse`.
  - `Trajectory.Parabola.DetonationDistance` controls the maximum distance in cells from intended target (checked at start of each game frame, before the projectile moves) at which the projectile will be forced to detonate. Set to 0 to disable forced detonation.
  - `Trajectory.Parabola.TargetSnapDistance` controls the maximum distance in cells from intended target the projectile can be at moment of detonation to make the projectile 'snap' on the intended target. Set to 0 to disable snapping.
  - `Trajectory.Parabola.OpenFireMode` controls how should the projectile be launched. This has the following 6 modes.
    - Speed - Automatic calculation mode with fixed horizontal velocity, using `Trajectory.Speed` and target coordinates as calculation conditions, i.e. the flight time of the projectile is permanently fixed.
    - Height - Automatic calculation mode with fixed maximum height, useing `Trajectory.Parabola.ThrowHeight` and target coordinates as calculation conditions, i.e. the detonation time of the projectile is relatively fixed.
    - Angle - Automatic calculation mode with fixed fire angle, useing `Trajectory.Parabola.LaunchAngle` and target coordinates as calculation conditions. In this mode, the performance consumption is high, and may have no solution. It is not recommended to enable `SubjectToCliffs` or enable `AA` with a smaller `MinimumRange` when using this mode.
    - SpeedAndHeight - Fixed horizontal velocity and maximum height mode, using `Trajectory.Speed` and `Trajectory.Parabola.ThrowHeight` as calculation conditions, i.e. the trajectory will only undergo altitude changes with the height of the target.
    - HeightAndAngle - Fixed maximum height and fire angle mode, using `Trajectory.Parabola.ThrowHeight` and `Trajectory.Parabola.LaunchAngle` as calculation conditions, i.e. the trajectory will change horizontally with the height of the target.
    - SpeedAndAngle - Fixed horizontal velocity and fire angle mode, using `Trajectory.Speed` and `Trajectory.Parabola.LaunchAngle` as calculation conditions, i.e. the trajectory will be permanently fixed.
  - `Trajectory.Parabola.ThrowHeight` controls the maximum height of the projectile and is only used for modes 1, 3, or 4. The specific height will be determined by taking the larger of the launch height and the target height then increasing this value. Non positive numbers are not supported.
  - `Trajectory.Parabola.LaunchAngle` controls the fire angle of the projectile and is only used for modes 2, 4, or 5. Only supports -90.0 ~ 90.0 (Cannot use boundary values) in Mode 2 or 5, and 0.0 ~ 90.0 (Cannot use boundary values) in Mode 4.
  - `Trajectory.Parabola.LeadTimeCalculate` controls whether the projectile need to calculate the lead time of the target when firing. Note that this will not affect the facing of the turret.
  - `Trajectory.Parabola.DetonationAngle` controls when the angle between the projectile in the current velocity direction and the horizontal plane is less than this value, it will detonate prematurely. Taking effect when the value is at -90.0 ~ 90.0 (Cannot use boundary values).
  - `Trajectory.Parabola.DetonationHeight` controls when the projectile is in a descending state and below the height of the launch position plus this value, it will detonate prematurely. Taking effect when it is set to non negative value.
  - `Trajectory.Parabola.BounceTimes` controls how many times can it bounce back when the projectile hits the ground or cliff. Be aware that excessive projectile speed may cause abnormal operation. And `Trajectory.Parabola.DetonationDistance` do not conflict with this and will take effect simultaneously. So if you want to explode the bullet only after the times of bounces is exhausted, you should set `Trajectory.Parabola.DetonationDistance` to a non positive value.
    - `Trajectory.Parabola.BounceOnWater` controls whether it can bounce on the water surface.
    - `Trajectory.Parabola.BounceDetonate` controls whether it detonates the warhead once extra during each bounce.
    - `Trajectory.Parabola.BounceAttenuation` controls the attenuation coefficient of projectile bounce damage, that is, how many times the next damage after each bounce is the damage just caused. This will also affect the damage of the final detonation.
    - `Trajectory.Parabola.BounceCoefficient` controls the attenuation coefficient of projectile bounce elasticity, that is, how many times the speed after each bounce is the speed before bouncing.
  - `Trajectory.Parabola.OffsetCoord` controls the offsets of the target. Projectile will aim at this position to attack. It also supports `Inaccurate=yes` and `Trajectory.Parabola.LeadTimeCalculate=true` on this basis.
    - `Trajectory.Parabola.RotateCoord` controls whether to rotate the projectile's firing direction within the angle bisector of `Trajectory.Parabola.OffsetCoord` according to the weapon's `Burst`. Set to 0 to disable this function.
    - `Trajectory.Parabola.MirrorCoord` controls whether `Trajectory.Parabola.OffsetCoord` need to mirror the lateral value to adapt to the current burst index. At the same time, the rotation direction calculated by `Trajectory.Parabola.RotateCoord` will also be reversed, and the rotation angle between each adjacent projectile on each side will not change as a result.
    - `Trajectory.Parabola.UseDisperseBurst` controls whether the calculation of `Trajectory.Parabola.RotateCoord` is based on its superior's `Trajectory.Disperse.WeaponBurst` of the dispersed trajectory, rather than `Burst` of the weapon. If this value is not appropriate, it will result in unsatisfactory visual displays.
    - `Trajectory.Parabola.AxisOfRotation` controls the rotation axis when calculating `Trajectory.Parabola.RotateCoord`. The axis will rotates with the unit orientation or the vector that from target position to the source position.

In `rulesmd.ini`:
```ini
Trajectory=Parabola                             ; Trajectory type
Trajectory.Parabola.DetonationDistance=0.4      ; floating point value
Trajectory.Parabola.TargetSnapDistance=0.5      ; floating point value
Trajectory.Parabola.OpenFireMode=speed          ; ParabolaFireMode value enumeration (speed|height|angle|speedandheight|heightandangle|speedandangle)
Trajectory.Parabola.ThrowHeight=600             ; integer
Trajectory.Parabola.LaunchAngle=30              ; floating point value
Trajectory.Parabola.LeadTimeCalculate=no        ; boolean
Trajectory.Parabola.DetonationAngle=-90.0       ; floating point value
Trajectory.Parabola.DetonationHeight=-1         ; integer
Trajectory.Parabola.BounceTimes=0               ; integer
Trajectory.Parabola.BounceOnWater=no            ; boolean
Trajectory.Parabola.BounceDetonate=no           ; boolean
Trajectory.Parabola.BounceAttenuation=0.8       ; floating point value
Trajectory.Parabola.BounceCoefficient=0.8       ; floating point value
Trajectory.Parabola.OffsetCoord=0,0,0           ; integer - Forward,Lateral,Height
Trajectory.Parabola.RotateCoord=0               ; floating point value
Trajectory.Parabola.MirrorCoord=yes             ; boolean
Trajectory.Parabola.UseDisperseBurst=no         ; boolean
Trajectory.Parabola.AxisOfRotation=0,0,1        ; integer - Forward,Lateral,Height
```

```{note}
- Compared to vanilla `Arcing`, this can also be used for aircrafts and airburst weapon.
- Certainly, `Gravity` can also affect the trajectory.
```

What is `Trajectory=Disperse` -> #1295 

`Trajectory=Disperse` + `Trajectory=Parabola`
parabola
bounce


src/Ext/Bullet/Trajectories/ParabolaTrajectory.cpp
```diff
@@ -0,0 +1,1043 @@
...
+	if (this->WaitOneFrame.HasTimeLeft())
+		return true;
+
+	this->PrepareForOpenFire(pBullet);
+	this->WaitOneFrame.Stop();
+
+	return false;
+}
+
+void ParabolaTrajectory::CalculateBulletVelocityLeadTime(BulletClass* pBullet, CoordStruct* pSourceCoords, double gravity)
```
secsome, line 377: Maybe this function needs an explanation about how all modes are working here, about the logical behind all modes.

src/Ext/Bullet/Trajectories/ParabolaTrajectory.cpp
```diff
@@ -0,0 +1,1043 @@
...
+		this->CheckIfNeedExtraCheck(pBullet);
+		return;
+	}
+	}
+
+	pBullet->TargetCoords = targetCoords + offsetCoords;
+	this->CalculateBulletVelocityRightNow(pBullet, pSourceCoords, gravity);
+}
+
+void ParabolaTrajectory::CalculateBulletVelocityRightNow(BulletClass* pBullet, CoordStruct* pSourceCoords, double gravity)
```
secsome, line 564: So does this function needs the explanation, or you could put all explanations at the beginning of this file.

src/Ext/Bullet/Trajectories/ParabolaTrajectory.cpp
```diff
@@ -0,0 +1,1043 @@
...
+}
+
+BulletVelocity ParabolaTrajectory::GetGroundNormalVector(BulletClass* pBullet, CellClass* pCell)
+{
+	if (const unsigned char index = pCell->SlopeIndex)
+	{
+		Vector2D<double> factor { 0.0, 0.0 };
+
+		if (index <= 4)
+			factor = Vector2D<double>{ 0.3763770469559380854890894443664, 0.9264665771223091335116047861327 };
```
secsome, line 845: The magic numbers here needs explanation.

src/Ext/Bullet/Trajectories/ParabolaTrajectory.cpp
```diff
@@ -0,0 +1,1043 @@
...
+			{
+				if (this->CheckBulletHitCliff(cell.X, cell.Y + reverseSgnY, bulletHeight, lastCellHeight))
+					return BulletVelocity{ static_cast<double>(reverseSgnX), 0.0, 0.0 };
+				else if (this->CheckBulletHitCliff(cell.X - reverseSgnX, cell.Y + reverseSgnY, bulletHeight, lastCellHeight))
+					index = 1;
+			}
+		}
+
+		if (index == 1)
+			return BulletVelocity{ 0.8944271909999158785636694674925 * reverseSgnX, 0.4472135954999579392818347337463 * reverseSgnY, 0.0 };
```
secsome, line 926: Magic numbers need explanation

src/Ext/Bullet/Trajectories/ParabolaTrajectory.cpp
```diff
@@ -0,0 +1,1043 @@
+#include "ParabolaTrajectory.h"
+#include <Ext/Bullet/Body.h>
+#include <Ext/BulletType/Body.h>
+#include <Ext/WarheadType/Body.h>
+#include <OverlayTypeClass.h>
+#include <ScenarioClass.h>
+
```
secsome, line 7: It's a good idea to put all logic explanation stuffs here. For details check the comment below.

Also for magic numbers you could use `constexpr` to give them a meaningful name.

secsome: The explanation is required so we can understand how does this trajectory work easier.

secsome (changes requested):
Looks good in general, but additional code comments are needed. The `mode` related code can be considered for switching to enums

src/Ext/Bullet/Trajectories/ParabolaTrajectory.cpp
CrimRecya: All changed.

src/Ext/Bullet/Trajectories/PhobosTrajectory.h
```diff
@@ -13,6 +13,7 @@ enum class TrajectoryFlag : int
 	Invalid = -1,
 	Straight = 0,
 	Bombard = 1,
+	Parabola = 4
```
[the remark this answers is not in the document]
Metadorius: @secsome is not aware of the other two trajectories pull requests. That is a valid reason to skip enum numbers.

docs/New-or-Enhanced-Logics.md
```diff
@@ -699,6 +702,66 @@ Trajectory=Bombard             ; Trajectory type
...
+    - `Trajectory.Parabola.UseDisperseBurst` controls whether the calculation of `Trajectory.Parabola.RotateCoord` is based on its superior's `Trajectory.Disperse.WeaponBurst` of the dispersed trajectory, rather than `Burst` of the weapon. If this value is not appropriate, it will result in unsatisfactory visual displays.
+    - `Trajectory.Parabola.AxisOfRotation` controls the rotation axis when calculating `Trajectory.Parabola.RotateCoord`. The axis will rotates with the unit orientation or the vector that from target position to the source position.
+
+In `rulesmd.ini`:
+```ini
+Trajectory=Parabola                             ; Trajectory type
+Trajectory.Parabola.DetonationDistance=0.4      ; floating point value
+Trajectory.Parabola.TargetSnapDistance=0.5      ; floating point value
+Trajectory.Parabola.OpenFireMode=0              ; ParabolaFireMode value enumeration (speed|height|angle|speedandheight|heightandangle|speedandangle)
```
secsome, line 740: Just a suggestion: Would it be better to change the naming convention for enumeration to start with capital letters?

secsome: **CHANGE REQUESTED**:
The doc here should be fixed, `OpenFireMode=speed` or any other mode instead of `=0` here

secsome (changes requested):
Just fix the document.

docs/New-or-Enhanced-Logics.md
CrimRecya: Get this. Fixed now. As for the issue of capitalization, although I also like to capitalize the first letter, in reality, capitalization does not affect the reading of the game. :)

secsome:
Resolve the conflict and we can merge it

src/Ext/Bullet/Trajectories/ParabolaTrajectory.cpp
```diff
@@ -0,0 +1,1231 @@
...
+bool ParabolaTrajectoryType::Load(PhobosStreamReader& Stm, bool RegisterForChange)
+{
+	this->PhobosTrajectoryType::Load(Stm, false);
+
+	Stm
+		.Process(this->DetonationDistance, false)
```
chaserli, line 13: Is it possible to find a way to reduce code duplication for Load/Save? Like what other classes do.

src/Utilities/TemplateDef.h
```diff
@@ -1369,6 +1369,44 @@ if(_strcmpi(parser.value(), #name) == 0){ value = __uuidof(name ## LocomotionCla
 		return false;
 	}
 
+	template <>
+	inline bool read<ParabolaFireMode>(ParabolaFireMode &value, INI_EX &parser, const char *pSection, const char *pKey)
```
chaserli, line 1373: I doubt if the enum will be used elsewhere, maybe the enum and its parsing template specialization should be defined in-place (I mean in ParabolaTrajectory.h/cpp)

chaserli (commented):
not quite related with the pr itself, 2 general questions for the maintainers

src/Ext/Bullet/Trajectories/ParabolaTrajectory.cpp
CrimRecya: I think this is not necessary. You can see that the type is saved using false, which is different from the default situation. This was initially informed to me by the sample trajectory template, and I am not sure what problems modifying this item may cause, so I chose to keep it as it is.

src/Utilities/TemplateDef.h
CrimRecya: Get it.

src/Ext/Bullet/Trajectories/ParabolaTrajectory.cpp
CrimRecya: I tried using template <typename T> void Serialize(T& Stm); like other ext classes, Although there were no issues with using it in Load, there were red wavy lines that I couldn't solve when using it in Save. :(

chaserli: Have you checked why the`RegisterForChange` was explicitly set to false in Load? I think it might be unnecessary. Say if you have a pointer in your class you'll still need to swizzle it, don't you?Just like what you did in your disperse one for the weapons. If it's normal POD nothing gets "swizzled" as expected.

Metadorius: POD = Plain Old Data, I don't think this fits though. @chaserli meant a type without pointers I think.

CrimRecya: Seems like the `bool Traj::Save(PhobosStreamWriter& Stm) const`'s `const` makes me cannot use the `void Serialize(T& Stm)`. :(

Metadorius: > Seems like the `bool Traj::Save(PhobosStreamWriter& Stm) const`'s `const` makes me cannot use the `void Serialize(T& Stm)`. :(

@CrimRecya please see the new commit:  https://github.com/Phobos-developers/Phobos/commit/29efd60311a3be770eb46d4a49295c54eba7e615

PR #1375, opened 2024-09-06
[Simple fix] InitialPayload was invalid in TeamClass

a851903106:
This is a very simple fix to help TechnoClasses created via a trigger operation get the InitialPayload correctly.

If the Ares.dll in use is 3.0 or 3.0p1, it will call the function from Ares itself, otherwise it will use another function.

Thanks to @NetsuNegi  and @ststl-s  for helping when I was in a bind.

PS: It will only trigger if the carrier has no passengers.


NetsuNegi:
It should be clear for now :)

chaserli:
If I understand it correctly, this aims to do the exact same thing as #1283 .
And I've yet which part in your impl prevented deadlock. 
In fact, I would like to know why the mutex was used for creation (https://github.com/Phobos-developers/Phobos/commit/82793929db630415af548decfb19710d8a85b947)
If you have an answer to that or to how to reproduce the deadlock in case the mutex handling is removed, I would love to learn about that.
Nevertheless #1283 no longer touches the mutex and is idempotent, theoretically no deadlock.

PR #1380, opened 2024-09-16
[Minor] Randomized anims for several behaviors

Coronia:
Allow several Phobos anim logic to display one of the random anims in the list, instead of one fixed anim.
Also fixed wrong definition of customized `WarpAway` anim.

In `rulesmd.ini`:
```ini
[AudioVisual]
Promote.VeteranAnimation=         ; list of Animation
Promote.EliteAnimation=           ; list of Animation

[SOMESHIELDTYPE]                            ; ShieldType name
BreakAnim=                                  ; list of Animation
HitAnim=                                    ; list of Animation

Shield.BreakAnim=                           ; list of Animation
Shield.HitAnim=                             ; list of Animation

[SOMETECHNO]                                   ; TechnoType
PassengerDeletion.Anim=                        ; list of Animation
AutoDeath.VanishAnimation=                     ; list of Animation
Spawner.RecycleAnim=                           ; list of Animation
Promote.VeteranAnimation=         ; list of Animation, default to Promote.VeteranAnimation in [AudioVisual]
Promote.EliteAnimation=           ; list of Animation, default to Promote.EliteAnimation in [AudioVisual]
WarpOut=                ; list of Animation (played when Techno warping out), default to [General] WarpOut
WarpIn=                 ; list of Animation (played when Techno warping in), default to [General] WarpIn
WarpAway=               ; list of Animation (played when Techno being erased by `Temporal=yes` warhead), default to [General] WarpAway

[SOMEANIM]                             ; AnimationType
WakeAnim=                              ; list of Animation
CreateUnit.SpawnAnim=                  ; list of Animation

[SOMETERRAINTYPE]  ; TerrainType
DestroyAnim=       ; list of Animation
```


Starkku:
Add a function to create random anim from `std::vector<AnimTypeClass*>` in `AnimExt` or something instead of copypasting the same thing across the place.

src/Ext/Anim/Body.cpp
```diff
@@ -239,6 +239,43 @@ void AnimExt::HandleDebrisImpact(AnimTypeClass* pExpireAnim, AnimTypeClass* pWak
 	}
 }
 
+AnimClass* AnimExt::CreateRandomAnim(std::vector<AnimTypeClass*> AnimList, CoordStruct coords, TechnoClass* pTechno, HouseClass* pHouse, bool invoker)
```
MortonPL, line 242:
```suggestion
AnimClass* AnimExt::CreateRandomAnim(const std::vector<AnimTypeClass*>& AnimList, CoordStruct coords, TechnoClass* pTechno, HouseClass* pHouse, bool invoker)
```

Since we're only accessing elements, this vector should be passed as an immutable reference.

MortonPL (changes requested):
Please add an entry in the changelog and credits as well.

Coronia:
remove those duplicated 'Anims' tags, despite it might be a bit weird to not using plural for a list, it should remain backward compatibility and clearness. Also add random anim support for `WakeAnim` and `Spawner.RecycleAnim`

src/New/Entity/ShieldClass.cpp
```diff
@@ -322,19 +322,15 @@ void ShieldClass::ResponseAttack()
...
+		pHitAnim = this->Type->HitAnim;
 
-	if (pAnimType)
-	{
-		auto const pAnim = GameCreate<AnimClass>(pAnimType, this->Techno->GetCoords());
-		AnimExt::SetAnimOwnerHouseKind(pAnim, this->Techno->Owner, nullptr, false, true);
-		AnimExt::ExtMap.Find(pAnim)->SetInvoker(this->Techno);
-	}
+	AnimExt::CreateRandomAnim(pHitAnim, this->Techno->GetCoords(), this->Techno, nullptr, true, true);
 }
```
CrimRecya, line 334: This will result in data being modified. If you only need to replace the animation to be played when it is empty, then you should use `const` like this:
```
void ShieldClass::WeaponNullifyAnim(const std::vector<AnimTypeClass*>& pHitAnim)
{
	if (this->AreAnimsHidden)
		return;

	AnimExt::CreateRandomAnim((pHitAnim.empty() ? this->Type->HitAnim : pHitAnim), this->Techno->GetCoords(), this->Techno, nullptr, true, true);
}
```

src/New/Entity/ShieldClass.cpp
```diff
@@ -687,16 +684,10 @@ void ShieldClass::BreakShield(AnimTypeClass* pBreakAnim, WeaponTypeClass* pBreak
...
 
-		if (pAnimType)
-		{
-			auto const pAnim = GameCreate<AnimClass>(pAnimType, this->Techno->Location);
-
-			pAnim->SetOwnerObject(this->Techno);
-			AnimExt::SetAnimOwnerHouseKind(pAnim, this->Techno->Owner, nullptr, false, true);
-			AnimExt::ExtMap.Find(pAnim)->SetInvoker(this->Techno);
-		}
+		AnimExt::CreateRandomAnim(pBreakAnim, this->Techno->Location, this->Techno, nullptr, true, true);
```
CrimRecya, line 690: Same as above.

src/New/Entity/ShieldClass.h
```diff
@@ -20,7 +20,7 @@ class ShieldClass
 	int ReceiveDamage(args_ReceiveDamage* args);
 	bool CanBeTargeted(WeaponTypeClass* pWeapon) const;
 	bool CanBePenetrated(WarheadTypeClass* pWarhead) const;
-	void BreakShield(AnimTypeClass* pBreakAnim = nullptr, WeaponTypeClass* pBreakWeapon = nullptr);
+	void BreakShield(std::vector<AnimTypeClass*>& pBreakAnim, WeaponTypeClass* pBreakWeapon = nullptr);
```
CrimRecya, line 23: Remember to modify several parts of the header file as well.

CrimRecya (changes requested):

Coronia:
The definition of customizable `WarpAway` in Phobos seems to be wrong, as it should be the anim that the techno is erased by `Temporal` warheads rather than the anim used by Chronoshift, which should still use `WarpIn` and `WarpOut`. But here it hooks out Chronoshift's WarpOut anim and make it become the customize WarpAway.

Not sure if these tag definitions should be changed now that they might be made use by mods already, and I'm thinking if the WarpIn/Out of Chronoshift still deserves a customization. As of now I might keep them as they are while adding the random anim feature for them, but if there's an idea about how to sort out these tags then I could also try to implement them in this pull request

Coronia:
> The definition of customizable `WarpAway` in Phobos seems to be wrong, as it should be the anim that the techno is erased by `Temporal` warheads rather than the anim used by Chronoshift, which should still use `WarpIn` and `WarpOut`. But here it hooks out Chronoshift's WarpOut anim and make it become the customize WarpAway.
> 
> Not sure if these tag definitions should be changed now that they might be made use by mods already, and I'm thinking if the WarpIn/Out of Chronoshift still deserves a customization. As of now I might keep them as they are while adding the random anim feature for them, but if there's an idea about how to sort out these tags then I could also try to implement them in this pull request

Fixed this in this PR already

PR #1381, opened 2024-09-16
[Minor] Shield respawn anim and weapon

Coronia:
- `Respawn.Anims`, if set, will be played when the shield respawns. If more than one animation is listed, a random one is selected.
- `Respawn.Weapon`, if set, will be fired at the TechnoType once the shield respawns.
- `Shield.Respawn.Anims` and `Shield.Respawn.Weapon` override `Respawn.Anims` and `Respawn.Weapon` for duration of `Shield.Respawn.Duration` amount of frames.

In `rulesmd.ini`:
```ini
[SOMESHIELDTYPE]                            ; ShieldType name
Respawn.Anims=                              ; list of Animation
Respawn.Weapon=                             ; WeaponType

Shield.Respawn.Anims=                       ; list of Animation
Shield.Respawn.Weapon=                      ; WeaponType
```

### DIFF
--- a/src/New/Entity/ShieldClass.cpp
+++ b/src/New/Entity/ShieldClass.cpp
@@ -93,6 +93,8 @@ bool ShieldClass::Serialize(T& Stm)
 		.Process(this->SelfHealing_RestartInCombatDelay_Warhead)
 		.Process(this->Respawn_Warhead)
 		.Process(this->Respawn_Rate_Warhead)
+		.Process(this->Respawn_Anims_Warhead)
+		.Process(this->Respawn_Weapon_Warhead)
 		.Process(this->LastBreakFrame)
 		.Process(this->LastTechnoHealthRatio)
 		.Success();
@@ -711,6 +713,23 @@ void ShieldClass::RespawnShield()
 		double amount = timerWHModifier->InProgress() ? Respawn_Warhead : this->Type->Respawn;
 		this->HP = this->GetPercentageAmount(amount);
 		this->UpdateTint();
+		const auto pAnimList = timerWHModifier->InProgress() ? this->Respawn_Anims_Warhead : this->Type->Respawn_Anims;
+		const auto pWeapon = timerWHModifier->InProgress() ? this->Respawn_Weapon_Warhead : this->Type->Respawn_Weapon;
+
+		if (!pAnimList.empty())
+		{
+			if (const auto pAnimType = pAnimList[ScenarioClass::Instance->Random.RandomRanged(0, pAnimList.size() - 1)])
+			{
+				if (auto const pAnim = GameCreate<AnimClass>(pAnimType, this->Techno->Location))
+				{
+					pAnim->SetOwnerObject(this->Techno);
+					pAnim->Owner = this->Techno->Owner;
+				}
+			}
+		}
+
+		if (pWeapon)
+			TechnoExt::FireWeaponAtSelf(this->Techno, pWeapon);
 	}
 	else if (timerWHModifier->Completed() && timer->InProgress())
 	{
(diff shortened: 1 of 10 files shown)


Metadorius (changes requested):
I propose to add a common template/composite tag like `Promotable` to which people can assign whatever they want - anim, weapon or whatever. The point is that we can reuse this template anywhere and easily add a lot of events, all consistently.

Starkku:
> I propose to add a common template/composite tag like `Promotable` to which people can assign whatever they want - anim, weapon or whatever. The point is that we can reuse this template anywhere and easily add a lot of events, all consistently.

And this template does what exactly?

Metadorius:
> > I propose to add a common template/composite tag like `Promotable` to which people can assign whatever they want - anim, weapon or whatever. The point is that we can reuse this template anywhere and easily add a lot of events, all consistently.
> 
> And this template does what exactly?

Almost-one-liner for modders to add whatever amounts of event handling possibilities for any "event" in a consistent fashion, like I do have those in techno attachment and I can just use the template. Further on it can be expanded to add whatever event handling possibilities if needed.
image

Metadorius:
@Coronia do you need any assistance on what I proposed? You can refer to #1304 for an example of the new template and bother me here or in Discord to help.

Coronia:
Yeah I would be appreciated if there's a demo for this. I've looked into the pull request but not quite got it yet. I take it that we should define MultiFlag for weapon and use it here, but still not sure how to do it correctly. If there's an existed template for weapon it would be better, and if my understanding is wrong please correct me.

src/New/Entity/ShieldClass.cpp
```diff
@@ -743,15 +750,44 @@ void ShieldClass::BreakShield(AnimTypeClass* pBreakAnim, WeaponTypeClass* pBreak
...
+
+	const auto pType = this->Type;
 	const auto timer = &this->Timers.Respawn;
 	const auto timerWHModifier = &this->Timers.Respawn_WHModifier;
 
 	if (this->HP <= 0 && timer->Completed())
 	{
 		timer->Stop();
-		double amount = timerWHModifier->InProgress() ? Respawn_Warhead : this->Type->Respawn;
+		double amount = timerWHModifier->InProgress() ? Respawn_Warhead : pType->Respawn;
```
CrimRecya, line 771: `this->Respawn_Warhead`

src/New/Entity/ShieldClass.cpp
```diff
@@ -743,15 +750,44 @@ void ShieldClass::BreakShield(AnimTypeClass* pBreakAnim, WeaponTypeClass* pBreak
...
 		this->HP = this->GetPercentageAmount(amount);
 		this->UpdateTint();
+		const auto pAnimList = timerWHModifier->InProgress() ? this->Respawn_Anim_Warhead : pType->Respawn_Anim;
+		const auto pWeapon = timerWHModifier->InProgress() ? this->Respawn_Weapon_Warhead : pType->Respawn_Weapon;
```
CrimRecya, line 775: You can put it at the back and take it when using it.

src/New/Entity/ShieldClass.cpp
```diff
@@ -760,18 +796,25 @@ void ShieldClass::RespawnShield()
...
+	this->Respawn_Anim_Warhead = anim;
+	this->Respawn_Weapon_Warhead = weapon ? weapon : Type->Respawn_Weapon;
 
 	timerWHModifier->Start(duration);
 
-	if (this->HP <= 0 && Respawn_Rate_Warhead >= 0 && resetTimer)
+	if (this->HP > 0)
+		return;
+
+	if (resetTimer)
```
CrimRecya, line 817: There's no problem here, I was just thinking:
I have seen that the timer is calculated separately in two states, so in `SetSelfHealing`, does a similar judgment, like `if (this->HP <= 0) return;`, also need to be made.

CrimRecya (commented):
Looks good, I only found some minor things. Then require some testing.

PR #1385, opened 2024-09-19
[BugFix]Fix CanBeHidden tag

NetsuNegi:
Fixed logical confusion and read section in CanBeHidden tags

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -327,6 +327,7 @@ This page lists all the individual contributions to the project by their author.
    - Forbidding parallel AI queues by type
    - Jumpjet crash speed fix when crashing onto building
    - Disguised units not using the correct palette if target has custom palette bugfix
+   - Fix logical confusion and read section in CanBeHidden tags
 - **Apollo** - Translucent SHP drawing patches
 - **ststl**
    - Customizable ShowTimer priority of superweapons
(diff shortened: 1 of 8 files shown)


Starkku:
The descriptions are very vague and should be clarified. What exactly does this fix?

MortonPL (changes requested):
Looks like this PR changes CanBeHidden to respect type's Image when reading from artmd.ini. Additionally, it inverts both the default value and the condition it affects!

Vanilla:
* if I CanBeHidden, check other conditions for the hiding animation
* if I can't, never show the hiding animation
This PR:
* if I CanBeHidden, never show the hiding animation
* if I can't, check other conditions for the hiding animation

It is a **breaking** change and I don't understand what problem does it fix exactly.

NetsuNegi:
> Looks like this PR changes CanBeHidden to respect type's Image when reading from artmd.ini. Additionally, it inverts both the default value and the condition it affects!
> 
> Vanilla:
> 
> * if I CanBeHidden, check other conditions for the hiding animation
> * if I can't, never show the hiding animation
>   This PR:
> * if I CanBeHidden, never show the hiding animation
> * if I can't, check other conditions for the hiding animation
> 
> It is a **breaking** change and I don't understand what problem does it fix exactly.

Based on my actual testing results, CanBeHidden=yes will actually display animations

PR #1386, opened 2024-09-19
[Minor] Observer sidebar in skirmish game mode

CrimRecya:
Enable the observer data panel, which could only be displayed in multiplayer games, to also be displayed in skirmish games.
observer


chaserli:
I think this should be for the spawner's side, maybe Belonit already had something similar. Perhaps you can check there and migrate in case it's not implemented yet.

Starkku:
[The YRpp-based spawner](https://github.com/CnCNet/yrpp-spawner) already has this implemented. There's already some overlap between it and some Phobos features, I would recommend to not make it worse.

PR #1388, opened 2024-09-19
[BugFix]Tunnel/Walk/Mech locomotor being stuck when moving too fast bugfix

NetsuNegi:
it is just a bugfix

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -327,6 +327,7 @@ This page lists all the individual contributions to the project by their author.
    - Forbidding parallel AI queues by type
    - Jumpjet crash speed fix when crashing onto building
    - Disguised units not using the correct palette if target has custom palette bugfix
+   - Fix a bug that infantry get stuck when move too fast
 - **Apollo** - Translucent SHP drawing patches
 - **ststl**
    - Customizable ShowTimer priority of superweapons
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -327,7 +327,7 @@ This page lists all the individual contributions to the project by their author.
    - Forbidding parallel AI queues by type
    - Jumpjet crash speed fix when crashing onto building
    - Disguised units not using the correct palette if target has custom palette bugfix
-   - Fix a bug that infantry get stuck when move too fast
+   - Fix a bug that some loco (Tunnel,Walk,Mech) get stuck when move too fast
 - **Apollo** - Translucent SHP drawing patches
 - **ststl**
    - Customizable ShowTimer priority of superweapons
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -167,7 +167,7 @@ This page describes all ingame logics that are fixed or improved in Phobos witho
 - The otherwise unused setting `[AI]` -> `PowerSurplus` (defaults to 50) which determines how much surplus power AI players will strive to have can be restored by setting `[AI]` -> `EnablePowerSurplus` to true.
 - Planning paths are now shown for all units under player control or when `[GlobalControls]->DebugPlanningPaths=yes` in singleplayer game modes.
 - Fixed `Temporal=true` Warheads potentially crashing game if used to attack `Slaved=true` infantry.
-- Fixed a bug that infantry get stuck when move too fast.
+- Fixed that some loco (Tunnel,Walk,Mech) get stuck when move too fast.
 
 ## Fixes / interactions with other extensions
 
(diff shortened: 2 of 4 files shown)


TaranDahl:
Mech locomotor also has this bug. Will be great if you can handle it too.
It is at 5B11DD.

NetsuNegi:
> Mech locomotor also has this bug. Will be great if you can handle it too. It is at 5B11DD.

got it

NetsuNegi:
> Mech locomotor also has this bug. Will be great if you can handle it too. It is at 5B11DD.

it should be fixed for now.

mevitar:
Does not glitch anymore whether the units/infantry have a very high speed from the start, or it gets increased and then lowered back.

The actual speed will not be exactly as high as the value indicates, but that's due to how these locomotors work (they force the infantry/unit to pause for a moment when transitioning between cells).

NetsuNegi:
> Does not glitch anymore whether the units/infantry have a very high speed from the start, or it gets increased and then lowered back.
> 
> The actual speed will not be exactly as high as the value indicates, but that's due to how these locomotors work (they force the infantry/unit to pause for a moment when transitioning between cells).

There is indeed this problem, and it is not clear what methods can be used to optimize it, but at least it is better than before

CREDITS.md
```diff
@@ -327,6 +327,7 @@ This page lists all the individual contributions to the project by their author.
    - Forbidding parallel AI queues by type
    - Jumpjet crash speed fix when crashing onto building
    - Disguised units not using the correct palette if target has custom palette bugfix
+   - Fix a bug that some loco (Tunnel,Walk,Mech) get stuck when move too fast
```
Metadorius, line 330:
```suggestion
   - Tunnel/Walk/Mech locomotor being stuck when moving too fast bugfix
```

docs/Fixed-or-Improved-Logics.md
```diff
@@ -167,6 +167,7 @@ This page describes all ingame logics that are fixed or improved in Phobos witho
 - The otherwise unused setting `[AI]` -> `PowerSurplus` (defaults to 50) which determines how much surplus power AI players will strive to have can be restored by setting `[AI]` -> `EnablePowerSurplus` to true.
 - Planning paths are now shown for all units under player control or when `[GlobalControls]->DebugPlanningPaths=yes` in singleplayer game modes.
 - Fixed `Temporal=true` Warheads potentially crashing game if used to attack `Slaved=true` infantry.
+- Fixed that some loco (Tunnel,Walk,Mech) get stuck when move too fast.
```
Metadorius, line 170:
```suggestion
- Fixed some locomotors (Tunnel, Walk, Mech) getting stuck when moving too fast.
```

docs/Whats-New.md
```diff
@@ -523,6 +523,7 @@ Vanilla fixes:
 - Building upgrades now consistently use building's `PowerUpN` animation settings corresponding to the upgrade's `PowersUpToLevel` where possible (by Starkku)
 - Subterranean units are no longer allowed to perform deploy functions like firing weapons or `IsSimpleDeployer` while burrowed or burrowing, they will instead emerge first like they do for transport unloading (by Starkku)
 - Fixed `Temporal=true` Warheads potentially crashing game if used to attack `Slaved=true` infantry (by Starkku)
+- Fixed that some loco (Tunnel,Walk,Mech) get stuck when move too fast (by NetsuNegi)
```
Metadorius, line 526:
```suggestion
- Fixed some locomotors (Tunnel, Walk, Mech) getting stuck when moving too fast (by NetsuNegi)
```

Metadorius (changes requested):
Code looks good apart from a few style nitpicks, but it would be cool if you also added a brief explanation in the code comments of what was the issue in the vanilla game and what you did with those hooks to fix it, so in future people will immediately have the idea of what the hooks are for.

NetsuNegi:
> Code looks good apart from a few style nitpicks, but it would be cool if you also added a brief explanation in the code comments of what was the issue in the vanilla game and what you did with those hooks to fix it, so in future people will immediately have the idea of what the hooks are for.

comments has been added

Metadorius (approved):

Metadorius (approved):

mevitar:
> There is indeed this problem, and it is not clear what methods can be used to optimize it, but at least it is better than before
I don't think that's necessarily a problem, i think it might be an intentional design choice there (at least for the walker locomotor).

NetsuNegi:
> > There is indeed this problem, and it is not clear what methods can be used to optimize it, but at least it is better than before
> 
> I don't think that's necessarily a problem, i think it might be an intentional design choice there (at least for the walker locomotor).

If it is really intentionally designed to avoid displaying defects, then I think a more appropriate approach is to limit current speed of the walk locomotor.

NetsuNegi:
> > > There is indeed this problem, and it is not clear what methods can be used to optimize it, but at least it is better than before
> > 
> > 
> > I don't think that's necessarily a problem, i think it might be an intentional design choice there (at least for the walker locomotor).
> 
> If it is really intentionally designed to avoid logic defects, then I think a more appropriate approach is to limit current speed of the walk locomotor.

mevitar:
I mean the pausing before moving to a new cell for walker locomotor might be intentional design choice by WestWood to simulate walker wobbling effect. Not the best way to do it, but it would make sense. I think better not to change that (people can use a different locomotor if they don't want the wobbling).

Locomotors breaking on higher speed is definitely a bug (or result of laziness), though.

NetsuNegi:
> I mean the pausing before moving to a new cell for walker locomotor might be intentional design choice by WestWood to simulate walker wobbling effect. Not the best way to do it, but it would make sense. I think better not to change that (people can use a different locomotor if they don't want the wobbling).
> 
> Locomotors breaking on higher speed is definitely a bug (or result of laziness), though.

Okay, I understand what you mean

PR #1389, opened 2024-09-21
[BugFix] Select allied country units that have been cloaked

a851903106:
Phobos allow the player to observe friendly units that have been cloaked, but they cannot be selected unless a unit with Sensors=yes is present nearby.
bug

So I made a short fix to select units with Sensors=yes even if they don't exist nearby.
fix

### DIFF
--- a/src/Ext/Techno/Hooks.Cloak.cpp
+++ b/src/Ext/Techno/Hooks.Cloak.cpp
@@ -1,4 +1,5 @@
 #include "Body.h"
+#include <Ext/House/Body.h>
 
 namespace CloakTemp
 {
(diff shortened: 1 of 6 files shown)


src/Ext/Techno/Hooks.Cloak.cpp
```diff
@@ -148,3 +149,57 @@ DEFINE_HOOK(0x4579A5, BuildingClass_ShouldNotCloak_Sensors, 0x6)
...
+
+	return HouseExt::CanSelectOwner(pThis->Owner, HouseClass::CurrentPlayer()) ?
+		SkipGameCode : 0;
+}
+
+// Block out Deselect() when a friendly unit goes cloakd.
+DEFINE_HOOK(0x6F4F10, TechnoClass_Sensed_DisableDSelect, 0x5)
+{
+	GET(TechnoClass* const, pThis, ESI);
+	enum { SkipGameCode = 0x6F4F3A, Continue = 0x6F4F21 };
```
MortonPL, line 189: Remove `Continue = 0x6F4F21` if it's not used.

MortonPL (changes requested):
Looks good. I've run it and it looks like it also fixes a campaign bug with multiple player houses and uncontrollable stealth units.

From modified all01md.map:
image

Remember to add an entry in `docs/Whats-New.md`, `CREDITS.md` and `docs/Fixed-or-Improved-Logics.md` (bugfixes section). Once that's done, this PR can be merged.

chaserli:
Partially already implemented, partially in another PR 

https://github.com/CnCNet/yrpp-spawner/pull/20
https://github.com/CnCNet/yrpp-spawner/blob/main/src/Misc/Observers.Visibility.cpp

MortonPL (approved):

Metadorius:
So uhh what are we doing with this one?

PR #1392, opened 2024-09-23
Append Ares' SW.Shots usage to sw tooltip

chaserli:
uimd.ini`[ToolTips]->SWShotsFormat` csf entry default to `Shots : %d`

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -330,6 +330,7 @@ This page lists all the individual contributions to the project by their author.
    - Skippable game save on scenario start
    - `InfDeath=9` versus jumpjet infantry 0 damage fix
    - Skip rally point line drawing when undeploying a building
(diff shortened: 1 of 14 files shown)


Metadorius:
Can you show a pic?

chaserli:
> Can you show a pic how it looks right now with default settings?

image

Metadorius:
Hmm what if we wrote `Name (7/10)` instead? seems intuitive enough and not overloaded much to me

chaserli:
> Hmm what if we wrote `Name (7/10)` instead? seems intuitive enough and not overloaded much to me

Then you can simply overwrite it in csf. You can use only 1 number if you want.
The c-style parsing as well as the default str was suggested by others.
I understand what you were thinking. You have a better word for it?

PR #1398, opened 2024-09-27
Warhead activation target health thresholds

Metadorius:
### Customizable Warhead trigger conditions

- It is now possible to make warheads only trigger when target's HP is above and/or below certain percentage.
  - Both conditions need to evaluate to true in order for the warhead to trigger.

In `rulesmd.ini`:
```ini
```


Metadorius:
This PR is supposed to be a reimpl. of #627. @FS-21 please verify that it can achieve the same functionality, keeping in mind what we commented earlier.

FS-21:
So basically we have to set multiple warheads in the same weapon and depending the threshold of the damage produced by the warhead it does any damage (or not)? Is for preparing the tests

Metadorius:
> depending the threshold of the damage produced

No, depending on the target health at the moment of hit.

Metadorius:
Otherwise you're correct.

FS-21:
So you said:

1º) The weapon hits the target.
2º) Checks the target health
3º) checks if there are the 2 new tags
4º) if meets the criteria the weapon works or does nothing if doesn't meet the criteria of the new 2 tags.

I tried With a Yuri mind control weapon (as the ones used in #627 ):


[EMindControl]
Damage=1 ;Number of mind control links
ROF=20
Range=7
Projectile=PsychicControl
Speed=100
Warhead=PsyController
Anim=YURICNTL
FireOnce=yes
CanTargetHouses=enemies

[PsyController] ;Mind control warhead.  Will skip normal damage like EMP did
Verses=100%,100%,100%,100%,0%,100%,0%,0%,0%,100%,100%
MindControl=yes
AnimList=YURICNTL
; ...
PenetratesBunker=no
AffectsBelowPercent=0.5     ; floating point value, percents or absolute

It ignores this new logic and mind controls the target. Same with ExtraWarheads logic.

Anotehr test in a Chrono legionnaire temporal weapon
[ChronoBeam]
Verses=100%,100%,100%,100%,100%,100%,100%,100%,100%,100%,0%
Temporal=yes
Versus.ImmuneToTemporal_concrete_armor=0%
Versus.defender_armor=0%
Versus.defender_armor.ForceFire=no
Versus.defender_armor.Retaliate=no
Versus.defender_armor.PassiveAcquire=no
AffectsBelowPercent=0.5     ; floating point value, percents or absolute

These tags are ignored and attacks & erase the target.
I also tried the Ares KillDriver=yes logic in WH and the same, these tags doesn't work

FS-21:
Wait, I checked in the code that exists a EffectsRequireVerses=false, I want re-test these cases sicne they should be considered effects.

Metadorius:
Meh I know what's the issue, currently it only applies to Phobos warhead effects... will see later if it's possible to fix this

FS-21:
0x4690B0 -> void __thiscall BulletClass::Logics(BulletClass *this, CoordStruct *coords) <- controls some WH effects (or all, I don't remember. Is what I was using in #627 for replacing the original warhead by another one if the threshold was reached)

Metadorius:
> If helps : 0x4690B0 -> void __thiscall BulletClass::Logics(BulletClass *this, CoordStruct *coords) <- controls some WH effects (or all, I don't remember. Is what I was using in #627 for replacing the original warhead by another one if the threshold was reached)

Yeah I am looking at this one too. I figured I'll ask other devs since for some reason it isn't hooked in order to be able to control all effects, might be cause of Ares. I might just move that check in the beginning of the function.

FS-21:
I updated the code with the suggested changes and I did basic tests and looks that it just works.

FS-21:
Applied feedback, pending that question between devs.

src/Ext/Bullet/Hooks.DetonateLogics.cpp
```diff
@@ -10,18 +10,41 @@
...
+			double hp = pTarget->GetHealthPercentage();
+			bool hpBelowPercent = hp <= pExt->AffectsBelowPercent;
+			bool hpAbovePercent = hp > pExt->AffectsAbovePercent;
+
+			if (!hpBelowPercent || !hpAbovePercent)
+			{
+				double versus = GeneralUtils::GetWarheadVersusArmor(pWarhead, pTarget->GetTechnoType()->Armor);
+
+				// Allow WH effects if under a valid health threshold ? Specially ExtraWarheads
+				if (pExt->EffectsRequireVerses && versus != 0.0)
```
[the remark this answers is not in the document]
Starkku: `ExtraWarheads` is not a warhead effect so no, it shouldn't be under domain of `EffectsRequireVerses`. Conditional detonation of `ExtraWarheads` should be its own thing.

src/Ext/WarheadType/Body.cpp
```diff
@@ -35,8 +35,13 @@ bool WarheadTypeExt::ExtData::CanAffectTarget(TechnoClass* pTarget, TechnoExt::E
...
 
-	if (!this->EffectsRequireVerses)
-		return true;
+	if (!IsHealthInThreshold(pTarget))
+		return false;
+
+	double versus = GeneralUtils::GetWarheadVersusArmor(this->OwnerObject(), pTarget->GetTechnoType()->Armor);
+
+	if (this->EffectsRequireVerses && versus == 0.0)
+		return false;
```
Starkku, line 44: Is there any particular reason why you're checking `Verses` twice in this function now instead of simply returning true if `EffectsRequireVerses=false`, bypassing all armor type checks that follow just as it should and used to do?

FS-21: "twice" because I didn't notice how the rest of the code was, I just readed it after your comment x'D

EffectsRequireVerses was something I deducted with the current code before I started, nothing I planned so I'll leave it as it was before but adding the threshold check before that "if".

For me  with the if (!IsHealthInThreshold(pTarget)) I'm happy. Thanks for noticing such incongruence.

src/Ext/Bullet/Hooks.DetonateLogics.cpp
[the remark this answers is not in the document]
FS-21: Ok so the simple "if" check for that HP threshold and done.

src/Ext/WarheadType/Body.cpp
FS-21: By the way, that
```
if (!this->EffectsRequireVerses)
		return true;
```
was supposed to be true o false to be right?

Starkku: Correct is `return true`. It skips verses/armor checks and allows WH effects to affect targets regardless of them.

docs/Whats-New.md
```diff
@@ -494,6 +494,7 @@ New:
 - Allow customizing charge turret delays per burst on a weapon (by Starkku)
 - Unit `Speed` setting now accepts floating point values (by Starkku)
 - Extending `Power` to all TechnoTypes (by Morton)
+- Warhead activation target health thresholds (by Kerbiter)
```
Metadorius, line 497:
```suggestion
- Warhead activation target health thresholds (by FS-21, Kerbiter)
```

src/Ext/WarheadType/Body.cpp
```diff
@@ -58,6 +61,19 @@ bool WarheadTypeExt::ExtData::CanAffectTarget(TechnoClass* pTarget, TechnoExt::E
...
+{
+	// Check if the WH should affect the techno target or skip it
+	double hp = pTarget->GetHealthPercentage();
+	bool hpBelowPercent = hp <= this->AffectsBelowPercent;
+	bool hpAbovePercent = hp > this->AffectsAbovePercent;
+
+	if (hpBelowPercent && hpAbovePercent)
+		return true;
+
+	return false;
```
Metadorius, line 74:
```suggestion
	return hpBelowPercent && hpAbovePercent;
```

src/Ext/Bullet/Hooks.DetonateLogics.cpp
```diff
@@ -309,6 +319,23 @@ DEFINE_HOOK(0x469AA4, BulletClass_Logics_Extras, 0x5)
...
+				// Allow WH effects if under a valid health threshold ? Specially ExtraWarheads
+				if (pWHExt->EffectsRequireVerses && versus == 0.0)
+					continue;
+
+				double hp = pTarget->GetHealthPercentage();
+				bool hpBelowPercent = hp <= pWHExt->AffectsBelowPercent;
+				bool hpAbovePercent = hp > pWHExt->AffectsAbovePercent;
+
+				if (!hpBelowPercent || !hpAbovePercent)
+					continue;
```
Metadorius, line 337:
```suggestion
				if (!pWHExt->IsHealthInThreshold(pTarget))
					continue;
```

FS-21:
Finally I tested this online and no desyncs. Then should this be merged?

FS-21:
I just noticed EffectsRequireVerses is true by default and the documentation mentions it is false by default.

I'll change the docs later

Coronia (approved):

Coronia:
tested, and it can do the effect we've expected: setting a series of `ExtraWarheads` with each of them having a warhead activation threshold, then it can detonate different warhead based on the target's strength

src/Ext/Bullet/Hooks.DetonateLogics.cpp
```diff
@@ -10,17 +10,27 @@
 #include <AircraftClass.h>
 #include <TacticalClass.h>
 
-DEFINE_HOOK(0x4690D4, BulletClass_Logics_ScreenShake, 0x6)
+DEFINE_HOOK(0x4690D4, BulletClass_Logics_NewChecks, 0x6)
 {
-	enum { SkipShaking = 0x469130 };
+	enum { SkipShaking = 0x469130, SkipFunction = 0x46A290, GoToExtras = 0x469AA4 };
```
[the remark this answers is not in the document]
FS-21: Residual address from one of the latest changes that removed a GoToExtras or a SkipFunction jump.

image

I'll delete it now

TaranDahl (approved):

Starkku (approved):
LGTM.

PR #1404, opened 2024-10-12
Enhanced Bombard trajectory

Coronia:
This is a sequel of https://github.com/Phobos-developers/Phobos/pull/1246, which makes use of logic and tag standard from the other trajectory pull requests (https://github.com/Phobos-developers/Phobos/pull/1293, https://github.com/Phobos-developers/Phobos/pull/1294, https://github.com/Phobos-developers/Phobos/pull/1295 and https://github.com/Phobos-developers/Phobos/pull/1374), and also utilizes the generic random anim function from (https://github.com/Phobos-developers/Phobos/pull/1380). Many thanks to @CrimRecya for making these projectiles as well as helping with improvement of enhanced Bombard trajectory. Also thanks **NaotoYuuki** for providing the prototypes of vertical and meteor projectiles, which are used as the base of these additions.

1 - Bombard Trajectory now support `Inaccurate=true, BallisticScatter.Max and BallisticScatter.Min` keys, which allows them to scatter when picking targets.

2 - Bombard Trajectory, instead of simply flying to the top of the target and free fall, is now extended into a series of trajectories which flying to a 'turning point' first, and then turning to the targets or respawning on their top. The behaviors of the projectile and the position of the turning point could be customized by the following tags:

  - `Trajectory.Bombard.Height` controls the height of the turning point.
  - `Trajectory.Bombard.FallPercent` controls the distance of the turning point by its percentage of the total distance between attacker and intended target. If set to 0%, then it'll fly up vertically. If set to 100%, then it'll travel to the top of the intended target.
    - For each launch the turning point percentage could add or minus a random value, which is not greater than `Trajectory.Bombard.FallPercentShift`. If set to 0%, random shift will be disabled.
    - You can also makes the turning point scatter randomly in a circle with `Trajectory.Bombard.FallScatter.Max` as its radius. If set to 0, random scatter will be disabled. `Trajectory.Bombard.FallScatter.Min` can be used to determine the minimum radius of the circle. If `Trajectory.Bombard.FallScatter.Linear` set to true, the random scatter will be limited to the line that is vertical to the original direction of the projectile.
  - `Trajectory.Bombard.FreeFallOnTarget` controls how it'll hit the intended target. If set to true, the projectile will be respawned above the intended target and free fall. If set to false, the projectile will travel to the intended target from the turning point.
  - `Trajectory.Bombard.NoLaunch` controls whether the attacker will fire the projectile by itself. If set to true, projectile will directly fall from the turning point.
  - `Trajectory.Bombard.FallSpeed` controls the initial speed of the projectile after it turns. If set to 0.0, then it'll use `Trajectory.Speed`. Can't work when `Trajectory.Bombard.FreeFallOnTarget` set to true.
  - `Trajectory.Bombard.DetonationDistance` controls the maximum distance in cells from intended target (checked at start of each game frame, before the projectile moves) at which the projectile will be forced to detonate. Set to 0 to disable forced detonation (note that this can cause the projectile to overshoot the target).
  - `Trajectory.Bombard.DetonationHeight` controls when the projectile is in a descending state and below the height of the launch position plus this value, it will detonate prematurely. Taking effect when it is set to non negative value. If `Trajectory.Bombard.EarlyDetonation` is set to true, it'll take effect during the ascending stage instead, which makes it detonate when its height is above the launch position plus this value.
  - `Trajectory.Bombard.TargetSnapDistance` controls the maximum distance in cells from intended target the projectile can be at moment of detonation to make the projectile 'snap' on the intended target. Set to 0 to disable snapping.
  - `Trajectory.Bombard.TurningPointAnims`, if set, will play an anim when the projectile reaches the turning point. If `Trajectory.Bombard.FreeFallOnTarget` is set to true, it'll be spawned above the target with the projectile together. If `Trajectory.Bombard.NoLaunch` is set to true, it'll be played at where the projectile falls, no matter if it's free fall or not. If more than one animation is listed, a random one is selected.
  - `Trajectory.Bombard.LeadTimeCalculate` controls whether the projectile need to calculate the lead time of the target when firing. Note that this will not affect the facing of the turret.
  - The following tags further customize the projectile's descending behaviors when `Trajectory.Bombard.FreeFallOnTarget` set to false.
    - `Trajectory.Bombard.OffsetCoord` controls the offsets of the target. Projectile will aim at this position to attack. It also supports `Inaccurate=yes` and `Trajectory.Bombard.LeadTimeCalculate=true` on this basis.
    - `Trajectory.Bombard.RotateCoord` controls whether to rotate the projectile's firing direction within the angle bisector of `Trajectory.Bombard.OffsetCoord` according to the weapon's `Burst`. Set to 0 to disable this function.
    - `Trajectory.Bombard.MirrorCoord` controls whether `Trajectory.Bombard.OffsetCoord` need to mirror the lateral value to adapt to the current burst index. At the same time, the rotation direction calculated by `Trajectory.Bombard.RotateCoord` will also be reversed, and the rotation angle between each adjacent projectile on each side will not change as a result.
    - `Trajectory.Bombard.UseDisperseBurst` controls whether the calculation of `Trajectory.Bombard.RotateCoord` is based on its superior's `Trajectory.Disperse.WeaponBurst` of the dispersed trajectory, rather than `Burst` of the weapon. If this value is not appropriate, it will result in unsatisfactory visual displays.
    - `Trajectory.Bombard.AxisOfRotation` controls the rotation axis when calculating `Trajectory.Bombard.RotateCoord`. The axis will rotates with the unit orientation or the vector that from target position to the source position.
  - `Trajectory.Bombard.SubjectToGround` controls whether the projectile should explode when it hits the ground. Note that this will not make AI search for suitable attack locations.

In `rulesmd.ini`:
```ini
[SOMEPROJECTILE]                              ; Projectile
Trajectory=Bombard                            ; Trajectory type
Trajectory.Bombard.Height=0.0                 ; double
Trajectory.Bombard.FallPercent=1.0            ; double
Trajectory.Bombard.FallPercentShift=0.0       ; double
Trajectory.Bombard.FallScatter.Max=0.0        ; floating point value
Trajectory.Bombard.FallScatter.Min=0.0        ; floating point value
Trajectory.Bombard.FallScatter.Linear=false   ; boolean
Trajectory.Bombard.FreeFallOnTarget=true      ; boolean
Trajectory.Bombard.NoLaunch=false             ; boolean
Trajectory.Bombard.FallSpeed=0.0              ; double
Trajectory.Bombard.DetonationDistance=0.4     ; floating point value
Trajectory.Bombard.DetonationHeight=-1        ; integer
Trajectory.Bombard.EarlyDetonation=false      ; boolean
Trajectory.Bombard.TargetSnapDistance=0.5     ; floating point value
Trajectory.Bombard.TurningPointAnims=         ; list of Animation
Trajectory.Bombard.LeadTimeCalculate=false    ; boolean
Trajectory.Bombard.OffsetCoord=0,0,0          ; integer - Forward,Lateral,Height
Trajectory.Bombard.RotateCoord=0              ; floating point value
Trajectory.Bombard.MirrorCoord=true           ; boolean
Trajectory.Bombard.UseDisperseBurst=false     ; boolean
Trajectory.Bombard.AxisOfRotation=0,0,1       ; integer - Forward,Lateral,Height
Trajectory.Bombard.SubjectToGround=false      ; boolean
```

If all values are set to default, it'll perform as the old Bombard Trajectory.

**NOTE:** Both the old and new Bombard trajectories might trigger a fatal error when firing a Voxel=true projectile (EIP: 007564B4), which is unresolved yet. Avoid using it for now, and if someone could make it work I'll be really grateful.


src/Ext/Bullet/Trajectories/PhobosTrajectory.cpp
```diff
@@ -567,7 +567,7 @@ DEFINE_HOOK(0x468B72, BulletClass_Unlimbo_Trajectories, 0x5)
 	auto const pExt = BulletExt::ExtMap.Find(pThis);
 	auto const pTypeExt = pExt->TypeExtData;
 
-	if (pTypeExt && pTypeExt->TrajectoryType)
+	if (pTypeExt && pTypeExt->TrajectoryType && !pExt->Trajectory)
```
CrimRecya, line 570: Is this unnecessary?

CrimRecya (approved):

src/Ext/Bullet/Trajectories/PhobosTrajectory.cpp
Coronia: It's an oversight when merging. Already removed

PR #1406, opened 2024-10-13
Allow customizing subterranean movement speed and allow parsing floating point Speed values

Starkku:
This should fix an issue introduced in 890eca0cd7eeb9298e876386cb2c3787bfb3e72b for tunnel locomotor where a distance check was made against unit's current speed without ensuring that `SpeedPercentage` which is normally not used by `TunnelLocomotionClass` is 1.0 instead of 0.0 and unhardcoding the subterranean movement speed.

### Subterranean unit travel speed

- Subterranean movement speed is now customizable, both globally and per TechnoType. If per-TechnoType value is negative, global value is used. This does not affect the speed at which the unit moves vertically when burrowing which is determined by `Speed` multiplied by `[General]`->`TunnelSpeed`.

In `rulesmd.ini`:
```ini
[General]
SubterraneanSpeed=7.5    ; floating point value
                         
SubterraneanSpeed=-1     ; floating point value
```

-----

Unit `Speed` setting now accepts floating-point values. Internally parsed values are clamped down to maximum of 100, multiplied by 256 and divided by 100, the result (which at this point is converted to an integer) then clamped down to maximum of 255 giving effective internal speed value range of 0 to 255, e.g leptons traveled per game frame.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -235,7 +235,7 @@ This page lists all the individual contributions to the project by their author.
   - Air unit tracking fix for large range / `CellSpread`
   - Extra tint intensity for Iron Curtain & Force Shield
   - Option to enable parsing 8-bit RGB values from `[ColorAdd]` instead of RGB565
-  - Customizing height at which subterranean units travel
+  - Customizing height and speed at which subterranean units travel
   - AI superweapon delay timer customization
   - Disabling `MultipleFactory` bonus from specific BuildingType
   - Customizable ChronoSphere teleport delays for units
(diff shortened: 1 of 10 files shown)


mevitar:
Fuckery fixed.

But there is one small problem...
tunnellolco
This happens only when the unit doesn't burrow. If it the unit will burrow after turning, the graphic doesn't bug out.
This also does not happen if SubterraneanUseSpeed= tags have default value.

mevitar:
> This also does not happen if SubterraneanUseSpeed= tags have default value.

It seems it was not this. I cannot reproduce this issue again no matter what SubterraneanUseSpeed= values i use,
I have no clue why it happened in the first place, even though all of these units on that animations were affected.

TaranDahl:
Maybe we need another flag like `SubterraneanSpeed` to control the tunnel speed instead of using the `Speed` flag directly. Just like `JumpjetSpeed`.

Starkku:
Changed the movement speed to be customizable individually from regular `Speed`. The burrowing speed remains tied to regular `Speed`.

PR #1407, opened 2024-10-17
Forcing specific weapon by range

Coronia:
- `ForceWeapon.InRange` forces specified a list of weapons to be used once the target is within their `Range`. The first weapon in the listed order satisfied will be selected. Can be applied to both ground and air target if `ForceAAWeapon.InRange` is not set.
    - `ForceAAWeapon.InRange` does the same thing but only for air target. Taking priority to `ForceWeapon.InRange`, which means that it can only be applied to ground target when they're both set.
    - `Force(AA)Weapon.InRange.Overrides` overrides the range when decides which weapon to use. Value from position matching the position from `Force(AA)Weapon.InRange` is used if found, or the weapon's own `Range` if not found or set to a value below 0. Specifically, if a position has `Force(AA)Weapon.InRange` set to -1 and `Force(AA)Weapon.InRange.Overrides` set to a positive value, it'll use default weapon selection logic once satisfied.
    - If `Force(AA)Weapon.InRange.ApplyRangeModifiers` is set to true, any applicable weapon range modifiers from the firer are applied to the decision range.

In `rulesmd.ini`:
```ini
ForceWeapon.InRange=                            ; list of integer. 0 for primary weapon, 1 for secondary weapon, -1 to disable
ForceWeapon.InRange.Overrides=                  ; list of floating point value
ForceWeapon.InRange.ApplyRangeModifiers=false   ; boolean
ForceAAWeapon.InRange=                          ; list of integer. 0 for primary weapon, 1 for secondary weapon, -1 to disable
ForceAAWeapon.InRange.Overrides=                ; list of floating point value
ForceAAWeapon.InRange.ApplyRangeModifiers=false ; boolean
```

An example:
```ini
[SOMETECHNO]
ForceWeapon.InRange=0,1,0,-1,1
ForceWeapon.InRange.Overrides=1,2,3,4,0
```
when the target distance is less than 1: use primary.
when the target distance is between 1 and 2: use secondary.
when the target distance is between 2 and 3: use primary.
when the target distance is between 3 and 4: use default weapon selection rule.
when the target distance is between 4 and secondary's range: use secondary.
when the target distance is greater than secondary's range: use default weapon selection rule.

This pattern can natually support WeaponX as well.


src/Ext/Techno/Body.h
```diff
@@ -180,4 +180,5 @@ class TechnoExt
 	static WeaponTypeClass* GetCurrentWeapon(TechnoClass* pThis, int& weaponIndex, bool getSecondary = false);
 	static WeaponTypeClass* GetCurrentWeapon(TechnoClass* pThis, bool getSecondary = false);
 	static int GetWeaponIndexAgainstWall(TechnoClass* pThis, OverlayTypeClass* pWallOverlayType);
+	static int ApplyForceWeaponInRange(TechnoClass* pThis, std::vector<int> weaponIndices, std::vector<double> rangeOverrides, bool applyRangeModifiers, int currentDistance);
```
CrimRecya, line 183: I think it's not so appropriate to use static functions in this situation.

src/Ext/Techno/WeaponHelpers.cpp
```diff
@@ -198,3 +198,44 @@ int TechnoExt::GetWeaponIndexAgainstWall(TechnoClass* pThis, OverlayTypeClass* p
...
+{
+	int forceWeaponIndex = -1;
+
+	for (size_t i = 0; i < weaponIndices.size(); i++)
+	{
+		int distance = 0;
+
+		// Value below 0 means Range won't be overriden
+		if (i < rangeOverrides.size() && rangeOverrides[i] > 0)
+			distance = static_cast<int>(rangeOverrides[i] * 256.0);
```
CrimRecya, line 212: Can be `Unsorted::LeptonsPerCell`.

src/Ext/Techno/WeaponHelpers.cpp
```diff
@@ -198,3 +198,44 @@ int TechnoExt::GetWeaponIndexAgainstWall(TechnoClass* pThis, OverlayTypeClass* p
...
 	return weaponIndex;
 }
+
+int TechnoExt::ApplyForceWeaponInRange(TechnoClass* pThis, std::vector<int> weaponIndices, std::vector<double> rangeOverrides, bool applyRangeModifiers, int currentDistance)
+{
+	int forceWeaponIndex = -1;
+
+	for (size_t i = 0; i < weaponIndices.size(); i++)
+	{
+		int distance = 0;
```
CrimRecya, line 208: Changing the variable name to `range` should be more intuitive.

src/Ext/Techno/WeaponHelpers.cpp
```diff
@@ -198,3 +198,44 @@ int TechnoExt::GetWeaponIndexAgainstWall(TechnoClass* pThis, OverlayTypeClass* p
...
+			{
+				forceWeaponIndex = weaponIndices[i];
+				break;
+			}
+		}
+		else
+		{
+			// Apply range modifiers regardless of weapon
+			if (applyRangeModifiers)
+				distance = WeaponTypeExt::GetRangeWithModifiers(nullptr, pThis, distance);
```
CrimRecya, line 232: This is indeed simple, but from the perspective of functional correctness, it is uncertain whether it is appropriate to do so.
Perhaps we can use a global mutex to avoid force weapon selection then call `SelectWeapon()` again.

src/Ext/Techno/Body.h
Coronia: this is a generic calculation logic instead of on some specific TechnoExt. If it's not a static TechnoExt function, then it needs to convert the Techno to TechnoExt first, then convert it back to acquire its weapon. Not sure if this would be better

src/Ext/Techno/WeaponHelpers.cpp
Coronia: these force weapon processes will return to default logic if the index is set to -1, so I think it should be correct

src/Ext/Techno/Body.h
CrimRecya: From my understanding, if you not use the static function, then you can simply use `TechnoExt`'s `TypeExtData` without passing these parameters except of `currentDistance`.

src/Ext/Techno/WeaponHelpers.cpp
CrimRecya: In fact, I don't quite understand why default weapons are not allowed to participate in the calculation. I think the range of the default weapon should be affected by the range bonus of a specific AE.

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -102,14 +106,19 @@ DEFINE_HOOK(0x6F3428, TechnoClass_WhatWeaponShouldIUse_ForceWeapon, 0x6)
...
 			forceWeaponIndex = pTypeExt->ForceWeapon_Disguised;
 		}
-
-		if (forceWeaponIndex >= 0)
+		else if (!pTypeExt->ForceWeapon_InRange.empty())
 		{
-			R->EAX(forceWeaponIndex);
-			return UseWeaponIndex;
+			int currentDistance = pThis->DistanceFrom(pTarget);
+			forceWeaponIndex = TechnoExt::ExtMap.Find(pThis)->ApplyForceWeaponInRange(pTarget, currentDistance);
```
CrimRecya, line 112: I see that this is not used except for selecting the default weapon. Perhaps you can directly input the weapon type.
It can even be more simplified because this target is the techno's target and the distance is calculated based on it, so no parameters need to be passed in.

src/Ext/Techno/WeaponHelpers.cpp
```diff
@@ -198,3 +198,60 @@ int TechnoExt::GetWeaponIndexAgainstWall(TechnoClass* pThis, OverlayTypeClass* p
...
+{
+	int forceWeaponIndex = -1;
+	auto const pThis = this->OwnerObject();
+	auto const pTypeExt = this->TypeExtData;
+	
+	auto const& weaponIndices = pTypeExt->ForceWeapon_InRange;
+	auto const& rangeOverrides = pTypeExt->ForceWeapon_InRange_Overrides;
+	const bool applyRangeModifiers = pTypeExt->ForceWeapon_InRange_ApplyRangeModifiers;
+
+	const int defaultWeaponIndex = pThis->SelectWeapon(pTarget);
```
CrimRecya, line 212: When using it this way, I mentioned a problem before, which is that if you call SelectWeapon() again in the hook of the SelectWeapon() function without using a lock, it may cause a dead loop.

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -75,16 +75,20 @@ DEFINE_HOOK(0x6F3428, TechnoClass_WhatWeaponShouldIUse_ForceWeapon, 0x6)
 
 	GET(TechnoClass*, pThis, ECX);
 
-	if (pThis && pThis->Target)
+	if (!pThis)
+		return 0;
```
CrimRecya, line 79: I think we can use a global variable here to determine whether to enter, switch before and after the second call, to skip this part if it is already locked when we reach this point during the call.

CrimRecya (commented):
There is a potential issue that could lead to a dead loop.

CrimRecya (approved):

PR #1408, opened 2024-10-17
[Minor] Clear jumpjets' redundant destination

CrimRecya:
Clear jumpjets' redundant `Destination` when they change state to hovering.
This is for improving `align jumpjet turret's facing with body's`, and adapt to  #1308 .

### DIFF
--- a/src/Ext/Unit/Hooks.Jumpjet.cpp
+++ b/src/Ext/Unit/Hooks.Jumpjet.cpp
@@ -120,10 +120,10 @@ DEFINE_HOOK(0x736990, UnitClass_UpdateRotation_TurretFacing_EMP, 0x6)
 }
 
 // Bugfix: Align jumpjet turret's facing with body's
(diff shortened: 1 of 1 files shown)


chaserli (commented):
I'm not sure about this, force queuing guard mission without target check seems a bit radical and I already observed something a bit weird.
That comment of mine was made 2+ years ago and I somehow doubt it now, I may need some explanation/confirmation.

CrimRecya:
> I'm not sure about this, force queuing guard mission without target check seems a bit radical and I already observed something a bit weird. That comment of mine was made 2+ years ago and I somehow doubt it now, I may need some explanation/confirmation.

I have removed unnecessary queue guard mission. Everything I have done here is just to clear the jumpjet's `Destination` when it has arrived at its destination, just like others.

Metadorius:
Why was the PR closed?

CrimRecya:
> Why was the PR closed?

Oh, actually, I just learned about a phenomenon about some jumpjets that cannot attack while moving will get stuck in the move mission without destination and allow enemies to pass underneath them. I haven't thought of any good solutions yet.

PR #1412, opened 2024-10-25
use SW deferment on Phobos logic

Coronia:
Previously several Phobos super weapon launching logic (e.g. `SW.Next` and `LaunchSW`) didn't support `SW.Deferment`. This pull request implements a workaround to enable this. `UseDeferment` tags delay the launch by a certain amount of frames determined by the superweapon's `SW.Deferment` if set to true, while `ExtraDeferment` tags add an extra value to this delay.

In `rulesmd.ini`:
```ini
[SOMEWARHEAD]                     ; Warhead
LaunchSW.UseDeferment=false       ; boolean
LaunchSW.ExtraDeferment=0         ; integer

[SOMESW]                        ; Super Weapon
SW.Next.UseDeferment=false      ; boolean
SW.Next.ExtraDeferment=0        ; integer

[SOMEBUILDING]                                        ; BuildingType
SpyEffect.VictimSuperWeapon.UseDeferment=false        ; boolean
SpyEffect.VictimSuperWeapon.ExtraDeferment=0          ; integer
SpyEffect.InfiltratorSuperWeapon.UseDeferment=false   ; boolean
SpyEffect.InfiltratorSuperWeapon.ExtraDeferment=0     ; integer
```

I was hoping to add this for the fire superweapon action as well, but it already has all the params used. Maybe this one could simply rely on elapse time action instead.

### DIFF
--- a/src/Ext/SWType/FireSuperWeapon.cpp
+++ b/src/Ext/SWType/FireSuperWeapon.cpp
@@ -250,16 +250,8 @@ void SWTypeExt::ExtData::ApplySWNext(SuperClass* pSW, const CellStruct& cell)
 					if ((this->SW_Next_IgnoreInhibitors || !pNextTypeExt->HasInhibitor(pHouse, cell))
 						&& (this->SW_Next_IgnoreDesignators || pNextTypeExt->HasDesignator(pHouse, cell)))
 					{
-						int oldstart = pSuper->RechargeTimer.StartTime;
-						int oldleft = pSuper->RechargeTimer.TimeLeft;
-						pSuper->SetReadiness(true);
-						pSuper->Launch(cell, pHouse->IsCurrentPlayer());
-						pSuper->Reset();
-						if (!this->SW_Next_RealLaunch)
-						{
-							pSuper->RechargeTimer.StartTime = oldstart;
-							pSuper->RechargeTimer.TimeLeft = oldleft;
-						}
+						int deferment = this->SW_Next_ExtraDeferment + (this->SW_Next_UseDeferment ? pNextTypeExt->SW_Deferment : 0);
+						ScenarioExt::Global()->LaunchSWs.push_back(SWFireState(pSuper, deferment, cell, pHouse->IsCurrentPlayer(), this->SW_Next_RealLaunch));
 					}
 				}
 			}
--- a/src/Ext/Scenario/Body.h
+++ b/src/Ext/Scenario/Body.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <ScenarioClass.h>
+#include <SuperClass.h>
 
 #include <Helpers/Macro.h>
 #include <Utilities/Container.h>
@@ -14,6 +15,51 @@ struct ExtendedVariable
 {
 	char Name[0x100];
 	int Value;
+};
+
+// Container for SW Deferment
+class SWFireState
+{
+public:
+	SuperClass* SW;
+	CDTimerClass deferment;
+	CellStruct cell;
+	bool playerControl;
+	int oldstart;
+	int oldleft;
+
+	bool Load(PhobosStreamReader& stm, bool registerForChange);
+	bool Save(PhobosStreamWriter& stm) const;
+
+	SWFireState() = default;
+
+	SWFireState(SuperClass* SW, int deferment, CellStruct cell, bool playerControl, bool realLaunch) :
+		SW { SW },
+		deferment {},
+		cell { cell },
+		playerControl { playerControl },
+		oldstart { SW->RechargeTimer.StartTime },
+		oldleft { SW->RechargeTimer.TimeLeft }
+	{
+		// reset the superweapon as if it's already launched
+		this->SW->Reset();
+
+		// if not real launch, save the timer states after reset
+		if (realLaunch)
+		{
+			this->oldstart = this->SW->RechargeTimer.StartTime;
+			this->oldleft = this->SW->RechargeTimer.TimeLeft;
+		}
+
+		this->oldleft = Math::max(oldleft - deferment, 0);
+		this->deferment.Start(deferment);
+	}
+
+	~SWFireState() = default;
+
+private:
+	template <typename T>
+	bool Serialize(T& stm);
 };
 
 class ScenarioExt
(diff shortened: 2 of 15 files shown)


src/Ext/SWType/FireSuperWeapon.cpp
```diff
@@ -252,13 +252,24 @@ void SWTypeExt::ExtData::ApplySWNext(SuperClass* pSW, const CellStruct& cell)
...
+						if (deferment > 0)
+						{
+							ScenarioExt::Global()->DefermentSWs.push_back(std::make_unique<SWFireTypeClass>(pSuper, deferment,
+								cell, pHouse->IsCurrentPlayer(), oldstart, oldleft, this->SW_Next_RealLaunch));
+						}
+						else
 						{
-							pSuper->RechargeTimer.StartTime = oldstart;
-							pSuper->RechargeTimer.TimeLeft = oldleft;
+							pSuper->SetReadiness(true);
```
chaserli, line 264: I think you can remove all in-place sw launches, and do that in your global launchsw state machines

src/Ext/Scenario/Body.h
```diff
@@ -16,6 +17,49 @@ struct ExtendedVariable
 	int Value;
 };
 
+// Container for SW Deferment
+class SWFireTypeClass
```
chaserli, line 21: The name is inappropriate, it's something like ares' sw state machines and shouldn't be named like xxTypeClass

src/Ext/WarheadType/Body.cpp
```diff
@@ -238,6 +238,8 @@ void WarheadTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 	this->LaunchSW_RealLaunch.Read(exINI, pSection, "LaunchSW.RealLaunch");
 	this->LaunchSW_IgnoreInhibitors.Read(exINI, pSection, "LaunchSW.IgnoreInhibitors");
 	this->LaunchSW_IgnoreDesignators.Read(exINI, pSection, "LaunchSW.IgnoreDesignators");
+	this->LaunchSW_UseDeferment.Read(exINI, pSection, "LaunchSW.UseDeferment");
```
chaserli, line 241: Not sure if such an additional xx.UseDeferment is necessary, deferment should have been respected and it was an oversight that I wasn't aware of back then.

src/Ext/Scenario/Body.h
```diff
@@ -35,6 +79,7 @@ class ScenarioExt
 
 		std::vector<TechnoExt::ExtData*> AutoDeathObjects;
 		std::vector<TechnoExt::ExtData*> TransportReloaders; // Objects that can reload ammo in limbo
+		std::vector<std::unique_ptr<SWFireTypeClass>> DefermentSWs;
```
chaserli, line 82: do you have any practical reason using a vector of unique_ptr rather than a vector of instances?

chaserli (changes requested):

src/Ext/Scenario/Body.h
Coronia: It's mostly for the convenience of constructing and destructing. Could change it to an instance if needed

src/Ext/WarheadType/Body.cpp
Coronia: I'm doing it with the assumption that someone already used what it was for their mods, while adjusting it directly might break those stuffs. So this is made into an optional thing.

src/Ext/SWType/FireSuperWeapon.cpp
Coronia: I'm not sure if making them all in the vector that being checked every frame is a good idea, might cost extra performance tax and make the SW unwantedly delay 1 frame sometime

NetsuNegi:
Have you test deferment at SW.Next? I had tried it and it cause desync by unknown reseaon.

Coronia:
Not really, only tested it in singleplayer so far. I do suspect changing SW timer tends to desync the game, but in that case there might also be similar problems with the original implementation of not-RealLaunch.

I'll be appreciated if someone could do some more multiplayer testing on it.

src/Ext/Scenario/Body.h
chaserli: which is also the reason why you don't need it in a vector. so far you don't have pimpl, inheritance or something else, there's no extra benefit but some minor runtime overhead

src/Ext/SWType/FireSuperWeapon.cpp
chaserli: you already checked this vector as well as the ReloadInTransport/AutoDeath ones on a per-frame basis, and they are most of the time empty.
You can investigate whether the updates of these vectors should happen before logics' (in your case bullets' and houses') or after. won't hurt to move the order.

Coronia:
Addressed the instance and global launch changes suggested above

Also we've done some multiplayer testing on it yesterday, and no desync so far

src/Ext/WarheadType/Body.cpp
chaserli: No I mean have you looked into how deferment works now? I don't think the underlying logic was ever changed since 0.A. Those 4 SWTypes that support SW.Deferment still use the old launch chain. 
In this context xx.UseDeferment as well as your additional parsing of SW.Deferment are unnecessary.
Mind if I ask which swtypes you want to use exactly?

src/Ext/Scenario/Body.h
```diff
@@ -16,6 +17,51 @@ struct ExtendedVariable
...
+// Container for SW Deferment
+class SWFireState
+{
+public:
+	SuperClass* SW;
+	CDTimerClass deferment;
+	CellStruct cell;
+	bool playerControl;
+	int oldstart;
+	int oldleft;
```
chaserli, line 29:
```suggestion
class PendingSW
{
public:
	SuperClass* SW;
	CDTimerClass Deferment;
	CellStruct Cell;
	bool ForReal;
```
It notes which sw to launch later so nothing should be done in ctor. You reset the sw and swap the timer more once.

Coronia:
It's confirmed that those SWs which supporting Deferment in Ares also work in Phobos. Therefore I'd like to close it for now since I don't want to add extra complexity and performance tax (this logic is checked per frame). But if there's a need to make Deferment support more SWs please let me know.

PR #1415, opened 2024-11-04
[Minor] Fix allies hospitals can not work properly

CrimRecya:
As the title says.

### DIFF
--- a/src/Ext/Building/Hooks.cpp
+++ b/src/Ext/Building/Hooks.cpp
@@ -474,6 +474,9 @@ DEFINE_HOOK(0x6F4D1A, TechnoClass_ReceiveCommand_Repair, 0x5)
 	GET(int, repairStep, EAX);
 	GET_STACK(TechnoClass*, pFrom, STACK_OFFSET(0x18, 0x4));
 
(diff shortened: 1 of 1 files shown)


Starkku:
Mind elaborating what exactly is broken here that this is supposed to fix?

CrimRecya:
> Mind elaborating what exactly is broken here that this is supposed to fix?
> 
> I know there's a bug where the hospital now requires credits to operate which I have already fixed properly (alongside Units.RepairRate not working for Hospitals when the other stuff did).

I found this problem in campaign, when infantry enter a hospital that is not owned by the player, they will not heal after coming out. Perhaps it's the issue you're referring to ?

Starkku:
> I found this problem in campaign, when infantry enter a hospital that is not owned by the player, they will not heal after coming out. Perhaps it's the issue you're referring to ?

If you could test it with latest develop nightly build that would help

CrimRecya:
> > I found this problem in campaign, when infantry enter a hospital that is not owned by the player, they will not heal after coming out. Perhaps it's the issue you're referring to ?
> 
> If you could test it with latest develop nightly build that would help

It works!

PR #1418, opened 2024-11-07
Faking Digital Display of Health

Coronia:
- If a TechnoType has a DigitalDisplayType with `InfoType=Health`, it's allowed to display fake values to enemy players.
    - `DigitalDisplay.Health.FakeStrength`, if set, will be used as the maximum value of health display. The current value will be displayed as the percentage of its current health multiplies the new maximum value.
    - `DigitalDisplay.Health.FakeAtDisguise`, if set to true on an InfantryType with Disguise, will use the disguised TechnoType's `Strength` value as the maximum value of health display. The current value will be displayed as the percentage of its current health multiplies the new maximum value. This will be applied before `DigitalDisplay.Health.FakeStrength`.

In `rulesmd.ini`:
```ini
[SOMETECHNOTYPE]
DigitalDisplay.Health.FakeStrength=-1       ; integer
DigitalDisplay.Health.FakeAtDisguise=true   ; boolean
```


src/Ext/Techno/Body.Visuals.cpp
```diff
@@ -465,3 +469,22 @@ void TechnoExt::GetValuesForDisplay(TechnoClass* pThis, DisplayInfoType infoType
 	}
 	}
 }
+
+void TechnoExt::GetDigitalDisplayFakeHealth(TechnoClass* pThis, int& value, int& maxValue)
+{
+	const auto pTypeExt = TechnoTypeExt::ExtMap.Find(pThis->GetTechnoType());
+	int newMaxValue = pTypeExt->DigitalDisplay_Health_FakeStrength;
+
+	if (pThis->Disguised && pTypeExt->DigitalDisplay_Health_FakeAtDisguise && pThis->WhatAmI() == AbstractType::Infantry)
```
Metadorius, line 478: Not sure why the check if it's infantry is necessary?

docs/User-Interface.md
```diff
@@ -46,50 +46,56 @@ IngameScore.LoseTheme= ; Soundtrack theme ID
   - Default `Shape.Spacing` for buildings is `4,-2`, `4,0` for others.
   - `ValueScaleDivisor` can be used to adjust scale of displayed values. Both the current & maximum value will be divided by the integer number given, if higher than 1.
 
+  - If a TechnoType has a DigitalDisplayType with `InfoType=Health`, it's allowed to display fake values to enemy players.
+    - `DigitalDisplay.Health.FakeStrength`, if set, will be used as the maximum value of health display. The current value will be displayed as the percentage of its current health multiplies the new maximum value.
```
Metadorius, line 50: I am not sure what is the use of this tag? IMO it makes sense to almost always have digital display to display the disguise target's strength.

Metadorius (changes requested):
Code looks mostly good, can be simplified though IMO.

src/Ext/Techno/Body.Visuals.cpp
Coronia: I was writing this in case of units which can disguise into TerrainType being affected, but perhaps it's not needed since there's another check about TechnoType later

docs/User-Interface.md
Coronia: This can be used in cases where you want to cheat your enemies. For example, there's a decoy type of units which only have half of the strength as the original one, so we can use this tag to fake it like the real ones.

src/Ext/Techno/Body.Visuals.cpp
Metadorius: Do TerrainTypes ever display health though? And do they use the same logic? I was under impression internally those are different things.

docs/User-Interface.md
Metadorius: Won't they just disguise as the main unit and automatically inherit it's display of strength?

Coronia: Not sure if I explained clearly, but I meant something like Nod's Decoy Army in CnC3, which is unit with same appearance but weaker in order to confuse enemies and attract firepower. Disguise as main unit won't work for decoys of VehicleTypes, also using value instead of reading from another TechnoType allows more customization.

Metadorius: Yeah I understood that. IMO it's better to fix disguise logic to work with vehicles and ensure disguising can work for decoy purposes (I know for sure people already use it like this using Phobos logics). If that is done, I feel like there would be no use case for such tag.

Metadorius: If I recall correctly, @Otamaa has vehicle disguising fixed in his fork.

Coronia:
there's another usage for `DigitalDisplay.Health.FakeStrength` now: when something has a lot of strength, like 9999999, you may want it to not show that many digits. This tag could scale down the number to a lower value. Ofc you can also do it by making a new techno and use `DigitalDisplay.Health.FakeAtDisguise` but that's way more complex

regardlessly, it's not bad to give a direct way to set this value since there'll always be cases that you want digital display to show something else from a given value, not to mention that only infantry can disguise as another so far

Metadorius:
> there's another usage for `DigitalDisplay.Health.FakeStrength` now: when something has a lot of strength, like 9999999, you may want it to not show that many digits. This tag could scale down the number to a lower value. Ofc you can also do it by making a new techno and use `DigitalDisplay.Health.FakeAtDisguise` but that's way more complex

I think that's better to be handled using a display coefficient, like when people use big numbers like 1 500 000 HP instead of 1 500 HP on *everything* just to have more precision on game's calculations, no?

Metadorius (approved):

PR #1420, opened 2024-11-10
[Minor] Fix Ammo features and tank bunkers.

Fryone:
- fixing ammo block inside tank bunkers;
 - partially fixing ammo adding/substracting inside tank bunkers

still problems with converters


Coronia (commented):
doc is needed for these fixes, otherwise looks fine

TaranDahl:
> doc is needed for these fixes, otherwise looks fine

Minor does not need doc, IIRC

Coronia:
> > doc is needed for these fixes, otherwise looks fine
> 
> Minor does not need doc, IIRC

if it's a vanilla bug or something that's broken by previous Phobos version, it's still recommended to note down what has been fixed

DeathFishAtEase:
This fix doesn't need to be written into Fixed / Improved Logics, but it should be mentioned in What's-New#version-tbd-develop-branch-nightly-builds - Phobos fixes:.

PR #1421, opened 2024-11-11
[Minor] Customizing bridge fall damageFalling down damage

a851903106:
坠落3
坠落4

Now you can customize the damage a unit receives when it falls from a bridge.
- `FallingDownDamage` customizes the damage a unit receives at the end of a fall. It can be a percentage or an integer.
- `FallingDownDamage.Water` customizes the damage a unit receives when it falls onto the water. Defaults to `FallingDownDamage`.
- If it is a negative percentage, corresponding damage will be dealt based on the current health of the unit.

In `rulesmd.ini`:
```ini
FallingDownDamage=              ; integer / percentage
FallingDownDamage.Water=        ; integer / percentage
```
---------
现在可以自定义单位从桥上坠落下来时受到的伤害。
- `FallingDownDamage`自定义单位坠落后受到的伤害。它可以是百分比，也可以是正数。
- `FallingDownDamage.Water`自定义自定义单位坠落到水面时受到的伤害。默认为`FallingDownDamage`。
- 如果是负百分比的话，它会根据单位的当前血量赋予对应百分比的伤害。


TaranDahl (approved):
Looking good.

src/Ext/Techno/Hooks.cpp
```diff
@@ -543,3 +543,62 @@ DEFINE_HOOK(0x70EFE0, TechnoClass_GetMaxSpeed, 0x6)
...
+
+					if (pCell->Tile_Is_Water() && pInf->SequenceAnim != Sequence::Swim)
+						pInf->PlayAnim(Sequence::Swim, true, false);
+					else if (!pCell->Tile_Is_Water() && pInf->SequenceAnim != Sequence::Guard)
+						pInf->PlayAnim(Sequence::Guard, true, false);
+				}
+			}
+			else
+			{
+				pTechno->UpdatePosition(PCPType::During);
```
CrimRecya, line 596: Can you explain why `PCPType::During` is used?

CrimRecya (approved):
Some small things. And remember to add the document.

src/Ext/Techno/Hooks.cpp
```diff
@@ -543,3 +543,62 @@ DEFINE_HOOK(0x70EFE0, TechnoClass_GetMaxSpeed, 0x6)
...
+				pThis->IsABomb = false;
+
+				if (pThis->WhatAmI() == AbstractType::Infantry)
+				{
+					auto pInf = abstract_cast<InfantryClass*>(pTechno);
+
+					if (pCell->Tile_Is_Water() && pInf->SequenceAnim != Sequence::Swim)
+						pInf->PlayAnim(Sequence::Swim, true, false);
+					else if (!pCell->Tile_Is_Water() && pInf->SequenceAnim != Sequence::Guard)
+						pInf->PlayAnim(Sequence::Guard, true, false);
```
CrimRecya, line 591: const auto pInf = static_cast<InfantryClass*>(pTechno);

				if (pCell->Tile_Is_Water())
				{
					if (pInf->SequenceAnim != Sequence::Swim)
						pInf->PlayAnim(Sequence::Swim, true, false);
				}
				else if (pInf->SequenceAnim != Sequence::Guard)
				{
					pInf->PlayAnim(Sequence::Guard, true, false);
				}
```suggestion
				const auto pInf = static_cast<InfantryClass*>(pTechno);

				if (pCell->Tile_Is_Water())
				{
					if (pInf->SequenceAnim != Sequence::Swim)
						pInf->PlayAnim(Sequence::Swim, true, false);
				}
				else if (pInf->SequenceAnim != Sequence::Guard)
				{
					pInf->PlayAnim(Sequence::Guard, true, false);
				}
```

CrimRecya:
Some possible questions are:
Do we need `UpdatePosition(PCPType::End)` to ensure effective compaction and other measures?
Do we need `CollectCrate()` to ensure that the box is picked up (maybe the box cannot be generated under the bridge)?
And do we need `Mark(MarkType::Down)` to ensure proper occupancy and avoid overlapping with teleport loco units?

Coronia (approved):

PR #1395, opened 2024-09-25
Customizable Bars (Health, Shield and much more)

Fryone:
Added customizable bars in form of BarTypes.

Example
```ini
[BarTypes] ; section in rulesmd.ini
0=HEALTHBAR1

[HEALTHBAR1]
InfoType= ;some infotype
PipBrd.Background.File= ; filename - without the .shp extension, image for use as background, first frame
PipBrd.Background.ShowWhenNotSelected=false ; boolean, should background be drawn, when ponting on techno
PipBrd.Background.Translucency= ; translucency level (0/25/50/75), background image translucency level
PipBrd.Foreground.File= ; filename - without the .shp extension, image for use as foreground (drawn after health sections), first frame
PipBrd.Foreground.ShowWhenNotSelected=false ; boolean, should foreground be drawn, when ponting on techno
PipBrd.Foreground.Translucency= ; translucency level (0/25/50/75), foreground image translucency level
PipBrd.Offset=0,0 ; X,Y, offset of background and foreground boards images
Bar.Offset=0,0 ; X,Y, offset of whole bar
Pips.File= ; filename - without the .shp extension, image to take frames from
Pips.Frames=16,17,18 ; list of 3 integers, frames of pips.shp to use (green, yellow, red states)
Pips.EmptyFrame=-1 ; integer, frame in pips.shp to use for empty section
Pips.Amount=17 ; integer, amount of sections, when bar is full
Pips.PositionDelta=2,0 ; X,Y, increment in pixels to next section
Pips.DrawBackwards=false ; boolean, changes order of drawing from last to first

[AudioVisual]
Buildings.DefaultBarTypes= ;list of bar types
Infantry.DefaultBarTypes= ;list of bar types
Vehicles.DefaultBarTypes= ;list of bar types
Aircraft.DefaultBarTypes= ;list of bar types

[SOMETECHNO]
BarTypes= ;list of bar types
```

No docs for now, needs testing.
image

PR #1287, opened 2024-06-23
New Infotypes and new display method

CrimRecya:
New InfoType: ROF timer, Reload timer, Spawn timer, Gattling timer, Produce cash timer, Passenger kill timer, Auto death timer, SW timer, IronCurtain timer, Temporal remainder, Factory process.
- `InfoIndex` defines the specific `InfoType`.
  - In `InfoType=Spawns`, 0 - alive spawns, 1 - docked spawns, 2 - launching spawns.
  - In `InfoType=Tiberium`, 0 - all, 1 - the first tiberium, 2 - the second tiberium, ...
  - In `InfoType=SpawnTimer`, 0 - the fastest spawnee, 1 - the first spawnee, 2 - the second spawnee, ...
  - In `InfoType=SuperWeapon`, 0 - the first SW of all, 1 - SW, 2 - SW2, 3 - the first SWs' SW, ...
  - In `InfoType=FactoryProcess`, 0 - the first factory in production, 1 - primary factory, 2 - secondary factory.
- If `Shape.PercentageFrame` set to true, it will only draw one frame that corresponds to total frames by percentage.
- `VisibleInSpecialState` controls whether this display type will show when the owner is in ironcurtain or is attacked by a temporal weapon.
- `ValueAsTimer` controls whether the value will be displayed in the form of a timer (i.e. 5:00, 25:00 or 1:00:00).

In `rulesmd.ini`:
```ini
InfoType=Health                         ; Displayed value enumeration (health|shield|ammo|mindcontrol|spawns|passengers|tiberium|experience|occupants|gattlingstage|ROF|Reload|SpawnTimer|GattlingTimer|ProduceCash|PassengerKill|AutoDeath|SuperWeapon|IronCurtain|TemporalLife|FactoryProcess)
InfoIndex=                              ; integer
Shape.PercentageFrame=false             ; boolean
VisibleInSpecialState=true              ; boolean
ValueAsTimer=false                      ; boolean
```

PR #1422, opened 2024-11-11
[Framework] Rework Extras draw 

chaserli:
Everyone (e.g. #1287  #1309  #1395  #1418 #1148) is doing their own work separately and creating overhead and incompatibilities.
As we've stated before the entire Healthbar drawing system needs to be overwritten completely.

This settles the framework for all these.

TODO: 
* BarType + DigitalsType
* For : HP, SP, IC timer, Temporal, ROF timer, Reload timer, Factory progress, SW progress, ...
* Condition: selected/always
* Location: Attached/ Global

### DIFF
--- a/src/Ext/Techno/Body.Visuals.cpp
+++ b/src/Ext/Techno/Body.Visuals.cpp
@@ -273,9 +273,6 @@ Point2D TechnoExt::GetBuildingSelectBracketPosition(TechnoClass* pThis, Building
 
 void TechnoExt::ProcessDigitalDisplays(TechnoClass* pThis)
 {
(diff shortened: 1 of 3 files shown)


Metadorius:
Why?

chaserli:
> Why?

To reduce PR number. 

I'm currently unavailable to work on this subject, more specifically properly dehardcoding stuff and design ini structure, which might require constant feedback and discussion. Meanwhile people might be impatient for things like #1287 #1309 etc. 

The common issue here and in the current digital display is the overhead, like the same draw point being calculated multiple times. Similar overhead also exists in vanilla code, if you don't use gdi ddraw you might not notice.

Will resume later.

PR #1426, opened 2024-11-17
[Minor] Make insignia respect PixelSelectionBracketDelta

Fryone:
```
[AudioVisual]
DrawInsignia.UsePixelSelectionBracketDelta=false  ;default
```


Coronia:
I'm not sure if this function is practical enough if it's a global setting. Maybe it'd be better to let it be set per techno? Something like `Insignia.AdjustPos` that could customize the offset for each techno.

Fryone:
It's already adjustable per type, this is just addition for sticking insignia with other top stuff like healthbar

Coronia (approved):

PR #550, opened 2022-03-10
Various Building properties Deharcoded

Otamaa:
### Customizable garrison and bunker properties

- You can now customize damage or ROF multipliers of a garrison or tank bunker building.
- You can now customize enter or exit sound of a tank bunker building.

In `rulesmd.ini`:
```ini
[SOMEBUILDING]              ; BuildingType
OccupyDamageMultiplier=     ; floating point value, default to [CombatDamage] -> OccupyDamageMultiplier
OccupyROFMultiplier=        ; floating point value, default to [CombatDamage] -> OccupyROFMultiplier
BunkerDamageMultiplier=     ; floating point value, default to [CombatDamage] -> BunkerDamageMultiplier
BunkerROFMultMultiplier=    ; floating point value, default to [CombatDamage] -> BunkerROFMultMultiplier
BunkerWallsUpSound=         ; Sound entry, default to [AudioVisual] -> BunkerWallsUpSound
BunkerWallsDownSound=       ; Sound entry, default to [AudioVisual] -> BunkerWallsDownSound
```

### Disable DamageSound

- Now you can disable `DamageSound` of a building.

In `rulesmd.ini`:
```ini
[SOMEBUILDING]            ; BuildingType
DisableDamageSound=false  ; boolean
```

### Power plant damage factor

- It is possible to customize the power decrement of a power plant when it's damaged. The actual power output for this plant will be: `Power` minuses the product of original power decrement and `Powerplant.DamageFactor`. Can't reduce power output lower than 0.
  - Specifically, if the factor is set to 0.0, power output won't be decreased by losing health for this power plant.

In `rulesmd.ini`:
```ini
[SOMEBUILDING]                     ; BuildingType
PowerPlant.DamageFactor=1.0        ; floating point value
```

PR #1433, opened 2024-11-29
[Draft] Customizing power output decrement when a power plant is damaged

Coronia:
This implementation worked in practice, but not sure if it's the proper approach. Also there might be missing check and cases, and its interaction with `Drain` (especially Ares's new options) still remains a question. For now I'd pull it as a draft to get some advice and testing. If anyone finds problem or has better idea please make me know, thank you very much.

- It is possible to customize the power decrement of a power plant when it's damaged. The actual power output for this plant will be: `Power` minuses the product of original power decrement and `Powerplant.DamageFactor`. Can't reduce power output lower than 0.
  - Specifically, if the factor is set to 0.0, power output won't be decreased by losing health for this power plant.

In `rulesmd.ini`:
```ini
PowerPlant.DamageFactor=1.0        ; floating point value
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -372,6 +372,7 @@ This page lists all the individual contributions to the project by their author.
 - **Ollerus**
   - Build limit group enhancement
   - Customizable rocker amplitude
+  - Customizing power output decrement when a power plant is damaged
 - **handama** - AI script action to jump back to previous script
 - **TaranDahl (航味麻酱)**
   - Skirmish AI "sell all buildings and set all technos to hunt" behavior dehardcode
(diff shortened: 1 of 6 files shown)


mevitar:
There was already an attempt to make power not degrade with powerplant's damage, in https://github.com/Phobos-developers/Phobos/pull/550 (but that one seems to be abandoned and this one is more customizable).

I think it would be better if the tag name was something like Power.DamageFactor= (or even PoweredSpecial.DamageFactor=, but this one is pretty long).

tyuah8:
Unless I am not understanding something, how come we can't just hook at `0x44E861` inside `BuildingClass::GetPowerOutput()` and change the call to `GetHealthPercentage()` to something else?  It is how the game now, even with Ares, is handling power loss on power plant damage.

Coronia:
reimplement it with https://github.com/Phobos-developers/Phobos/pull/550 as a base

PR #1435, opened 2024-12-01
Setting Sidebar tabs by selecting corresponded factory building

Fryone:
Factory=BuildingType -> switch between first and second
Factory=InfantryType -> third tab is set
Factory=UnitType/AircraftType -> fourth tab is set

[AudioVisual]
SetTabBySelectingFactory=false  ; boolean - disabled by default

[SOMEBUILDING]
SetTabBySelecting=-1                 ; integer, index of tab (1 - 4 for specific, 0 to disable for factories)


TaranDahl (approved):
LGTM

TaranDahl:
> I'm not sure if the current interaction is reasonable enough. There're many cases when you want to click a factory while not switching sidebars. E.g. set primary, set group etc. 

Maybe. But I don't think there's any harm in switching the sidebar in this situation.

> Also the 'switching to the other sidebar when clicking a ConYard' thing feels a bit unintuitive, maybe it would be better to just makes it always switch to building sidbar like CnC 3 and RA3 did

This can be disabled by ini.

DeathFishAtEase:
Since we already have `TabIndex`, I changed `1~4` to `0~3`, extended `-1` to all negative values, and changed `0` to *all other scenarios* (values greater than or equal to `4`).

PR #1439, opened 2024-12-01
[New Feature] Techno type conversion expanded

Aephiex:
1. Allows techno type conversion when unit swaps ownership into a side or country.
2. Allows techno type conversion when unit enters or leaves a transport.

This will be incorporated into an upcoming event trigger system, so I'm converting this draft.


Aephiex:
Can now also convert techno type on enter or exit a transport.

Such as we can convert a Rifle Squad into a rifle turret when loaded into MARV.

PR #1440, opened 2024-12-02
Fire weapon when kill

Coronia:
- `KillWeapon` will be fired at the target TechnoType's location once it's been killed by this Warhead.
  - `KillWeapon.AffectTargets` is used to filter which types of targets (TechnoTypes) are considered valid for KillWeapon. Only `none`, `all`, `aircraft`, `buildings`, `infantry` and `units` are valid values.
  - `KillWeapon.AffectHouses` is used to filter which houses targets can belong to be considered valid for KillWeapon.
  - `KillWeapon.AffectTypes` can be used to list specific TechnoTypes to be considered as valid targets for KillWeapon. If any valid TechnoTypes are listed, then only matching objects will be targeted.
  - `KillWeapon.IgnoreTypes` can be used to list specific TechnoTypes to be never considered as valid targets for KillWeapon.
- Ìf a TechnoType has `SuppressKillWeapons` set to true, it will not trigger KillWeapon upon being killed. `SuppressKillWeapons.Types` can be used to list WeaponTypes affected by this, if none are listed all WeaponTypes are affected.

 In `rulesmd.ini`:
```ini
[SOMEWARHEAD]                   ; Warhead
KillWeapon=                     ; WeaponType
KillWeapon.AffectTargets=all    ; list of Affected Target Enumeration (none|aircraft|buildings|infantry|units|all)
KillWeapon.AffectHouses=all     ; list of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
KillWeapon.AffectTypes=         ; list of TechnoTypes
KillWeapon.IgnoreTypes=         ; list of TechnoTypes

SuppressKillWeapons=false       ; boolean
SuppressKillWeapons.Types=      ; list of WeaponTypes
```


src/Ext/Techno/WeaponHelpers.cpp
```diff
@@ -198,3 +199,31 @@ int TechnoExt::GetWeaponIndexAgainstWall(TechnoClass* pThis, OverlayTypeClass* p
...
+		return;
+
+	if (pWHExt->KillWeapon_AffectTypes.size() > 0 && !pWHExt->KillWeapon_AffectTypes.Contains(pType))
+		return;
+
+	if (pWHExt->KillWeapon_IgnoreTypes.size() > 0 && pWHExt->KillWeapon_IgnoreTypes.Contains(pType))
+		return;
+
+	if (pTypeExt->SuppressKillWeapons_Types.size() > 0 && pTypeExt->SuppressKillWeapons_Types.Contains(pWHExt->KillWeapon))
+		return;
```
Starkku, line 219: Considering this is copypasted from revenge weapons or whatever, could consider moving the code to a function that returns a bool and takes the appropriate vectors (pass by reference ig) as params etc.

src/Ext/Techno/WeaponHelpers.cpp
```diff
@@ -198,3 +199,31 @@ int TechnoExt::GetWeaponIndexAgainstWall(TechnoClass* pThis, OverlayTypeClass* p
...
+
+	if (pType->WhatAmI() == AbstractType::AircraftType && (pWHExt->KillWeapon_AffectTargets & AffectedTarget::Aircraft) != AffectedTarget::None)
+		WeaponTypeExt::DetonateAt(pWHExt->KillWeapon, pThis, pSource);
+	else if (pType->WhatAmI() == AbstractType::BuildingType && (pWHExt->KillWeapon_AffectTargets & AffectedTarget::Building) != AffectedTarget::None)
+		WeaponTypeExt::DetonateAt(pWHExt->KillWeapon, pThis, pSource);
+	else if (pType->WhatAmI() == AbstractType::InfantryType && (pWHExt->KillWeapon_AffectTargets & AffectedTarget::Infantry) != AffectedTarget::None)
+		WeaponTypeExt::DetonateAt(pWHExt->KillWeapon, pThis, pSource);
+	else if (pType->WhatAmI() == AbstractType::UnitType && (pWHExt->KillWeapon_AffectTargets & AffectedTarget::Unit) != AffectedTarget::None)
+		WeaponTypeExt::DetonateAt(pWHExt->KillWeapon, pThis, pSource);
```
Starkku, line 228: Is there a good reason for this monstrosity as opposed to calling `EnumFunctions::IsTechnoEligible()`?

Starkku (changes requested):
Some stuff that probably needs to be checked but for most part it is the usual clean and basic stuff.

Metadorius:
I am once again asking for introduction of either `EventHandlerType` or a template to do uniform event handlers. Then instead of reinventing the same stuff with different features (like here you have Affects*) you could generalize that and have it available *everywhere* in a uniform fashion.

Coronia:
I'm still not sure what pattern should be used to group these conditions together. The current condition lines take reference from both `RevengeWeapon` and `DetonateOnAllMapObjects`, but these 3 things have some slight difference that I need to think of how to sort them out if making them into universal logic.

Besides, we're not just talking about something for my own pull requests, but also a shared template for the future. Think we'd better agree on a pattern that's acceptable by most first, and better makes a demo for that. I could help adding functions into it if the template is defined.

Metadorius:
> I'm still not sure what pattern should be used to group these conditions together. The current condition lines take reference from both `RevengeWeapon` and `DetonateOnAllMapObjects`, but these 3 things have some slight difference that I need to think of how to sort them out if making them into universal logic.
I think we're talking about different sides of the same problem. What you're talking about is the condition part, whereas I am talking about the sequence part.

I'll create a discussion/issue where we should document that then and where everyone could take part.

Coronia:
after some thought I think we don't need to add too many filters for it other than `AffectsHouses`, since it's not like the case of DetonateOnAllMapObjects which needs to filter targets for detonation. If anything using a series of `ExtraWarheads` with it would be enough

CrimRecya (commented):
lgtm about the logic and code. I prefer to use non static functions directly, but it's not a big deal.
However, there are some troubles with YRpp, remember to update it.

PR #1366, opened 2024-08-25
[Highly Customized] Optimize aircrafts actions

CrimRecya:
- Now, when a `stop` command (S) is issued to an aircraft, the aircraft will immediately return to the airport. When a `guard` command (G) is issued, the aircraft will search for targets around the current location and return immediately when target is not found, target is destroyed or ammos are depleted (Note that if the target is destroyed but ammos are not depleted yet, it will also return because the aircraft's command is one-time). When a `attack move` command (Ctrl+Shift) is issued, the aircraft will move towards the destination and search for nearby targets on the route for attack. Once ammos are depleted or the destination is reached, it will return (Note that if the  automatically selected target is destroyed but ammos are not depleted yet during the process, the aircraft will continue to go to the destination).

In `rulesmd.ini`:
```ini
[General]
ExpandAircraftMission=     ; boolean
```

PR #1449, opened 2024-12-10
Fix `Stop` command not working so well in some cases

CrimRecya:
When receive the stop(shortcut: s) command:
- Prevent jumpjets from falling into a state of standing idly by.
- Prevent technos from being unable to stop the attack move mission.
- Prevent aircrafts from briefly pausing in the air before returning.
- Prevent aircrafts from duplicatedly finding airport and overlapping.
- Prevent aircrafts with `AirportBound=no` continue moving forward.

The position of the hook and one of its functions overlap with #1366 .

### DIFF
--- a/src/Misc/Hooks.BugFixes.cpp
+++ b/src/Misc/Hooks.BugFixes.cpp
@@ -1090,7 +1090,7 @@ DEFINE_HOOK(0x4C75E6, EventClass_RespondToEvent_Stop, 0x5)
 		pTechno->vt_entry_4A8(); // pTechno->ClearMegaMissionData()
 
 	// To avoid aircrafts pausing in the air and let they returning to air base immediately
-	if (pTechno->WhatAmI() == AbstractType::Aircraft && !pTechno->Airstrike && !pTechno->Spawned && pTechno->GetHeight() > Unsorted::CellHeight)
+	if (pTechno->WhatAmI() == AbstractType::Aircraft && !pTechno->Airstrike && !pTechno->Spawned && static_cast<AircraftClass*>(pTechno)->Type->AirportBound && pTechno->GetHeight() > Unsorted::CellHeight)
 		pTechno->EnterIdleMode(false, true);
 
 	return SkipGameCode;
--- a/src/Misc/Hooks.BugFixes.cpp
+++ b/src/Misc/Hooks.BugFixes.cpp
@@ -1064,6 +1064,40 @@ DEFINE_HOOK(0x743664, UnitClass_ReadFromINI_Follower3, 0x6)
 
 #pragma endregion
 
+#pragma region StopEventFix
+
+DEFINE_HOOK(0x4C75E6, EventClass_RespondToEvent_Stop, 0x5)
+{
+	enum { SkipGameCode = 0x4C762A };
+
+	GET(TechnoClass* const, pTechno, ESI);
+
+	// Clearing the current target should still be necessary for all technos
+	pTechno->SetTarget(nullptr);
+	const auto pFoot = abstract_cast<FootClass*>(pTechno);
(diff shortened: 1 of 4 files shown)


mevitar:
As of yet, i didn't notice any problems with how units behave on stop order. I'll report it if i see problems.

However, there is now a visual glitch with jumpjet turrets where they change their angle while firing (but it happens only on certain firing angles).

jjetangle

I didn't have this issue with the devbuild or without Phobos at all, which means this build is causing it.
Furthermore, i also saw this happen in https://github.com/Phobos-developers/Phobos/pull/1366 , which means it's either the shared hook or the shared function that is causing it.

mevitar:
> I didn't have this issue with the devbuild or without Phobos at all, which means this build is causing it. Furthermore, i also saw this happen in #1366 , which means it's either the shared hook or the shared function that is causing it.

Nevermind, i retested it again several times and eventually got it to happen with the devbuild too. Clearly this one is not at fault.

src/Misc/Hooks.BugFixes.cpp
```diff
@@ -1064,6 +1064,40 @@ DEFINE_HOOK(0x743664, UnitClass_ReadFromINI_Follower3, 0x6)
...
+	const auto pFoot = abstract_cast<FootClass*>(pTechno);
+	const auto pJumpjetLoco = pFoot ? locomotion_cast<JumpjetLocomotionClass*>(pFoot->Locomotor) : nullptr;
+
+	// To avoid jumpjets falling into a state of standing idly by
+	if (!pJumpjetLoco) // If is not jumpjet, clear the destination is enough
+		pTechno->SetDestination(nullptr, true);
+	else if (!pFoot->Destination) // When in attack move and have had a target, the destination will be cleaned up, enter the guard mission can prevent the jumpjets stuck in a status of standing idly by
+		pTechno->QueueMission(Mission::Guard, true);
+	else if (static_cast<int>(CellClass::Coord2Cell(pFoot->Destination->GetCoords()).DistanceFromSquared(pTechno->GetMapCoords())) > 2) // If the jumpjet is moving, find the forward cell then stop in it
+		pTechno->SetDestination(pTechno->GetCell()->GetNeighbourCell(static_cast<FacingType>(((((pJumpjetLoco->LocomotionFacing.Current().Raw) >> 12) + 1) >> 1) & 7)), true);
```
NetsuNegi, line 1086: you can use `pJumpjetLoco->LocomotionFacing.Current().GetValue<3>()`

NetsuNegi:
> > 我在使用 devbuild 或根本没有 Phobos 时没有遇到这个问题，这意味着这个版本导致了它。此外，我在 #1366 中也看到了这种情况，这意味着导致它的要么是共享钩子，要么是共享函数。
> 
> 没关系，我又重新测试了几次，最终它也在 devbuild 中发生了。显然这个没有错，是其他原因导致了这个问题。

maybe [0449bdd](https://github.com/Phobos-developers/Phobos/commit/0449bdda117faae26509ac8b3606441c9b223f80)

src/Misc/Hooks.BugFixes.cpp
CrimRecya: Thx. I will try it later.

CrimRecya:
> As of yet, i didn't notice any problems with how units behave on stop order. I'll report it if i see problems.

For jumpjets, especially [JUMPJET], it will clearly appear after using the stop command. They will stand idly by in this state and will no longer attack the units under its feet. It can only be solved by moving again.
For attack move, especially when is attacking a target, it will continue to attack after a pause and may continue to move towards the previous destination after the attack ends. After expanding attack move behavior, the problem will become even more prominent.
As for aircrafts, they sometimes stay in place for a while before starting to return when executing the stop command.

Metadorius (approved):
see my latest review to #1366

CrimRecya:
Now this branch has been merged into #1366 , so close this separately.

PR #1450, opened 2024-12-11
Weapon target filtering by health percentage

NetsuNegi:
```
CanTarget.MaxHealth=1.0  ; double
CanTarget.MinHealth=0.0  ; double
```


src/Ext/WeaponType/Body.h
```diff
@@ -75,6 +77,8 @@ class WeaponTypeExt
 			, Strafing_UseAmmoPerShot { false }
 			, CanTarget { AffectedTarget::All }
 			, CanTargetHouses { AffectedHouse::All }
+			, CanTarget_MaxHealth { 1.0 }
+			, CanTarget_MinHealth { 0.0 }
```
TaranDahl, line 81: Maybe a wider default range is needed. Health ratio is often in 0 to 1, but ... not always. https://modenc.renegadeprojects.com/Strength

TaranDahl (approved):
Other codes lgtm.

src/Ext/WeaponType/Body.h
NetsuNegi: you mean negative strength ?

TaranDahl: Yes. But I think it is also acceptable to ignore them as bugs.

NetsuNegi: I think so

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -499,22 +503,29 @@ DEFINE_HOOK(0x6FE19A, TechnoClass_FireAt_AreaFire, 0x6)
...
 				int rand = ScenarioClass::Instance->Random.RandomRanged(0, size - 1);
 				unsigned int cellIndex = (i + rand) % size;
-				CellStruct tgtPos = pCell->MapCoords + adjacentCells[cellIndex];
+				CellStruct tgtPos = mapCoords + adjacentCells[cellIndex];
 				CellClass* tgtCell = MapClass::Instance.TryGetCellAt(tgtPos);
-				bool allowBridges = tgtCell && tgtCell->ContainsBridge() && (pThis->OnBridge || tgtCell->Level + CellClass::BridgeLevels == pThis->GetCell()->Level);
+				bool allowBridges = tgtCell && tgtCell->ContainsBridge() && (onBridge || tgtCell->Level + CellClass::BridgeLevels == level);
 
-				if (EnumFunctions::AreCellAndObjectsEligible(tgtCell, pExt->CanTarget, pExt->CanTargetHouses, pThis->Owner, true, false, allowBridges))
+				if (!skipWeaponPicking && EnumFunctions::AreCellAndObjectsEligible(tgtCell, canTarget, canTargetHouses, pOwner, true, false, allowBridges))
```
CrimRecya: `if (skipWeaponPicking || EnumFunctions::AreCellAndObjectsEligible(tgtCell, canTarget, canTargetHouses, pOwner, true, false, allowBridges))`
Additionally, it would be best to reimplement the original algorithm.

CrimRecya (commented):
Regarding the modifications to the `TechnoExt::PickWeaponIndex` section:
I think modifying the order will change the original logic.
For example, when both the primary and secondary weapons are unable to attack the target, according to the original logic, the primary weapon will be returned, while according to the current logic, the secondary weapon will be directly returned after the primary weapon cannot attack.

Coronia:
> Regarding the modifications to the `TechnoExt::PickWeaponIndex` section: I think modifying the order will change the original logic. For example, when both the primary and secondary weapons are unable to attack the target, according to the original logic, the primary weapon will be returned, while according to the current logic, the secondary weapon will be directly returned after the primary weapon cannot attack.

That’s true, but these conditions will be checked again during CanFire process. So whether primary or secondary is picked here, the result will be the same: the techno won’t fire since both weapons can’t target

Tho I’m not sure if there’re other cases that might go wrong. The purpose of this change is that the common use case of these conditions will be blocking primary and forcing secondary, so it’ll take less calculation if primary is being processed first. If this doesn’t go well, it could be reverted to the previous order

src/Ext/Techno/WeaponHelpers.cpp
```diff
@@ -30,33 +19,43 @@ int TechnoExt::PickWeaponIndex(TechnoClass* pThis, TechnoClass* pTargetTechno, A
 	else if (!pWeaponStructOne)
 		return weaponIndexTwo;
 
-	auto const pWeaponOne = pWeaponStructOne->WeaponType;
 	auto const pWeaponTwo = pWeaponStructTwo->WeaponType;
+	auto const pFirstExt = WeaponTypeExt::ExtMap.Find(pWeaponStructOne->WeaponType);
+	auto const pSecondExt = WeaponTypeExt::ExtMap.Find(pWeaponTwo);
+	const bool secondIsAA = pTargetTechno && pTargetTechno->IsInAir() && pWeaponTwo->Projectile->AA;
+	const bool skipPrimaryPicking = pFirstExt->SkipWeaponPicking;
```
CrimRecya: These two can be placed with `firstAllowedAE`.

CrimRecya (approved):

src/Ext/WeaponType/Body.cpp
```diff
@@ -41,6 +41,15 @@ bool WeaponTypeExt::ExtData::HasRequiredAttachedEffects(TechnoClass* pTarget, Te
...
 }
 
+bool WeaponTypeExt::ExtData::IsHealthRatioEligible(TechnoClass* const pTarget) const
+{
+	if (!pTarget)
+		return true;
+
+	const auto ratio = pTarget->GetHealthPercentage();
+	return ratio <= this->CanTarget_MaxHealth && ratio >= this->CanTarget_MinHealth;
+}
```
Metadorius, line 51: This one is not valid.

Imagine you want to make a weapon that targets units with health from 0% to 50% and another weapon from 50% to 100%. When unit health is 50%, both weapons would be valid.

Instead you should check `min < HP <= max`. Why this way and not vice versa? Because *normally* targets can have 100% HP but not 0% HP. However, targets *can* have 0% HP, but this is a special state. So, it makes sense to have a togglable switch, like `CanTarget.MinHealth.IncludeZero=true` which will make the threshold account for 0% HP units if it's set at 0%.

@Starkku is supposedly already working on a fix for that as well, we've found that out while discussing a bug with the similar logic for warheads. 

Also not sure why the target is TechnoClass when IIRC health is on ObjectClass.

Metadorius: Also why sanity check?

CrimRecya: I think `Health` should only have practical significance on `TechnoClass`. For example, bullet's `Health` is used to record its damage. Furthermore, the damage to wall is probabilistic and has a chain reaction, so its `Health` is actually not stable.

Metadorius: Trees also have health.

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -51,8 +51,8 @@ DEFINE_HOOK(0x6F33CD, TechnoClass_WhatWeaponShouldIUse_ForceFire, 0x6)
 		auto const pWeaponSecondary = pThis->GetWeapon(1)->WeaponType;
 		auto const pPrimaryExt = WeaponTypeExt::ExtMap.Find(pWeaponPrimary);
 
-		if (pWeaponSecondary && (!EnumFunctions::IsCellEligible(pCell, pPrimaryExt->CanTarget, true, true)
-			|| (pPrimaryExt->AttachEffect_CheckOnFirer && !pPrimaryExt->HasRequiredAttachedEffects(pThis, pThis))))
+		if (pWeaponSecondary && !pPrimaryExt->SkipWeaponPicking && (!EnumFunctions::IsCellEligible(pCell, pPrimaryExt->CanTarget, true, true)
+			|| !pPrimaryExt->IsHealthRatioEligible(pThis) || (pPrimaryExt->AttachEffect_CheckOnFirer && !pPrimaryExt->HasRequiredAttachedEffects(pThis, pThis))))
```
Metadorius, line 55: You're checking health of *yourself* when force firing on a cell. Why?

PR #1452, opened 2024-12-14
[New Feature] Autoload Command Shortcut

psi-cmd:
For selected units, pair the transport with infantry/vehicle.
1. If vehicle loadable transports selected, it won't be loaded if there are vehicles that can only be paired with infantry within selection.
2. It will always try to load the unit type with least amount of units first, eg. 2 GI and 3 GGI, it will match 2 GI to transport first.
3. For transport with multiple seats, it will try to fill those seats "diversely".


psi-cmd:
(Falsely include irrelevent file that belongs to another feature)

Aephiex:
Is it possible we can have a default shortcut key for the autoload command?

Aephiex:
I don't see any issue in my local build of this. Everything works as intended.

When I select an amphibious transport on water and units on land at a same time and press the auto load button, the amphibious transport will move ashore all the way to the land unit and stop there. The land unit doesn't get into the transport. However, this is exactly the same behavior when I call a land unit to get into an amphibious transport on water, so this isn't really an issue.

Aephiex (approved):

Aephiex:
Hello, made Ares tag support for you.

https://github.com/psi-cmd/Phobos/pull/1

Effects:
1. Transports with `NoManualEnter=true` will be viewed as if they didn't had passenger slots;
2. If transport has `Passengers.BySize=false` then every potential passenger are viewed as size 1;
3. If transport has `Passengers.Allowed=` and `Passengers.Disallowed=` defined then passengers unable to enter it will not try to get into it.

Known Issues:
1. If we make it so Jackal Racers only accept Knightframes, and we select 8 Jackal Racers, 4 Knightframes, 4 Lancers, and 2 ambiphious transports at the same time and press auto load button, the first activation moves all Knightframes into Jackal Racers which is fine, but the following activation of the auto load command does nothing at all. It is supposed to move the Jackal Racers and the remaining Lancers into the amphibious transports.
2. Maybe we have to detect if Ares is present before we take compatibility with Ares tags?

NetsuNegi:
have you test in multiplayer game?

psi-cmd:
> Hello, made Ares tag support for you, and more.
> 
> [psi-cmd#1](https://github.com/psi-cmd/Phobos/pull/1)
> 
> Effects:
> 
> 1. Transports with `NoManualEnter=true` will be viewed as if they didn't have passenger slots;
> 2. If a transport has `Passengers.BySize=false` then every potential passenger is viewed as size 1;
> 3. If a transport has `Passengers.Allowed=` and `Passengers.Disallowed=` defined then passengers unable to enter it will not try to get into it, as well as the auto assignment will never try to violent the `SizeLimit=`;
> 4. Vehicle types with size <= 2 and passenger <= 0 are viewed as standard passengers like infantry types, and infantry types with size >= 3 are viewed as larger passengers like tanks;
> 5. If a standard transport can't actually load anything due to its passenger filters, then just try to load it into larger transports;
> 6. Mind controlled units and units mind controlling something can't be passengers at all.
> 
> Maybe we have to detect if Ares is present before we take compatibility with Ares tags?

Wow, thanks for your work. I am not famillar with other features on transport loading and you help with the corner cases. 
And also thanks for a clear documentation on explaining things into details. 👍

For ares tags, maybe we could have the "original" default value for them. For example, `NoManualEnter=False`, `Passengers.BySize=True` are the default behavior without any MOD.

psi-cmd:
> have you test in multiplayer game?
Not yet, trying to find someone for help.

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,270 @@
...
+	//
+	// This array is only loaded into large vehicles if:
+	//   - either "passengerArray" is empty;
+	//   - or nothing in the "passengerArray" can actually be loaded into non-large vehicles.
+	std::vector<TechnoClass *> largePassengerArray;
+
+	// vehicles that can hold size <= 2
+	std::vector<std::pair<TechnoClass *, int>> vehicleIndexArray;
+	// vehicles that can hold size >= 3
+	std::vector<std::pair<TechnoClass *, int>> largeVehicleIndexArray;
```
Metadorius, line 177: Is there a specific reason that this is broken down into two halves exactly at size 2/size 3?

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+	// Get current selected units.
+	// The first iteration, we find units that can't be transports, and add them to the passenger arrays.
+	for (int i = 0; i < ObjectClass::CurrentObjects->Count; i++)
+	{
+		auto pUnit = ObjectClass::CurrentObjects->GetItem(i);
+		// try to cast to TechnoClass
+		TechnoClass *pTechno = abstract_cast<TechnoClass *>(pUnit);
```
Metadorius, line 185: We attach the  `*` and `&` to the type when declaring types, it makes more sense this way.

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,270 @@
...
+					auto pTypeExt = TechnoTypeExt::ExtMap.Find(pTransport->GetTechnoType());
+					bool bySize = true;
+					// Check passenger filter here.
+					if (pTypeExt)
+					{
+						if (!pTypeExt->CanLoadPassenger(pTransport, pPassenger))
+						{
+							continue;
+						}
+						bySize = pTypeExt->Passengers_BySize;
```
Metadorius, line 101:
```suggestion
						if (!pTypeExt->CanLoadPassenger(pTransport, pPassenger))
							continue;

						bySize = pTypeExt->Passengers_BySize;
```

Please check the style guide on this.

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,270 @@
...
+				if (pPassenger.size() < leastSize)
+				{
+					leastSize = pPassenger.size();
+					leastpID = pID;
+				}
+			}
+
+			{
+				unsigned int index = 0;
+				for (; index < leastSize && index < tTransports.size(); index++)
```
Metadorius, line 88:
```suggestion
				for ( ; index < leastSize && index < tTransports.size(); index++)
```

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,270 @@
...
+	std::unordered_map<const char *, std::vector<P>> passengerMap;
+	std::unordered_map<const char *, std::vector<P>> passengerMapIdle;
+
+	for (auto pPassenger : pPassengers)
+	{
+		auto pID = pPassenger->get_ID();
+		if (passengerMap.find(pID) == passengerMap.end())
+		{
+			passengerMap[pID] = std::vector<P>();
+		}
```
Metadorius, line 67:
```suggestion
		if (passengerMap.find(pID) == passengerMap.end())
			passengerMap[pID] = std::vector<P>();

```

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,270 @@
...
+{
+	Debug::Log("AutoLoadCommandClass::DebugPrintPassenger: Passenger count: %d\n", pPassengers.size());
+	// print address of each passenger
+	for (auto passenger : pPassengers)
+	{
+		Debug::Log("AutoLoadCommandClass::DebugPrintPassenger: Passenger address: %p\n", passenger);
+	}
+}
+
+template <typename P, typename T>
```
Metadorius, line 51: `P` and `T` is not that descriptive as typenames in this case, how about `TPassenger` and `TTransport`?

Code is generally more frequently read than written, so it's important to make code most easily readable.

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,270 @@
...
+			}
+		}
+		if (passengerMapIdle.size() != 0 && tTransports.size() != 0)
+		{
+			std::swap(passengerMap, passengerMapIdle);
+		}
+		else
+		{
+			break;
+		}
```
Metadorius, line 146:
```suggestion

		if (passengerMapIdle.size() != 0 && tTransports.size() != 0)
			std::swap(passengerMap, passengerMapIdle);
		else
			break;
```

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,270 @@
...
+	{
+		auto pUnit = ObjectClass::CurrentObjects->GetItem(i);
+		// try to cast to TechnoClass
+		TechnoClass *pTechno = abstract_cast<TechnoClass *>(pUnit);
+
+		// If not a techno, or is in air, then it can't be a vehicle.
+		if (!pTechno || pTechno->IsInAir())
+		{
+			continue;
+		}
```
Metadorius, line 220:
```suggestion
		if (!pTechno || pTechno->IsInAir())
			continue;
```

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,270 @@
...
+					vehicleIndexArray.push_back(std::make_pair(pTechno, transportTotalSize));
+					continue;
+				}
+			}
+
+			// If MCed, or MCs anything, then it can't be a passenger.
+			if (pTechno->IsMindControlled() || (pTechno->CaptureManager && pTechno->CaptureManager->IsControllingSomething()))
+			{
+				continue;
+			}
```
Metadorius, line 250:
```suggestion
				continue;

```

src/Ext/TechnoType/Body.cpp
```diff
@@ -32,6 +32,33 @@ void TechnoTypeExt::ExtData::ApplyTurretOffset(Matrix3D* mtx, double factor)
...
+
+bool TechnoTypeExt::ExtData::CanLoadAny(TechnoClass* pTransport, std::vector<TechnoClass*> pPassengerList) const
+{
+	for (auto pPassenger : pPassengerList)
+	{
+		if (this->CanLoadPassenger(pTransport, pPassenger))
+		{
+			return true;
+		}
+	}
```
Metadorius, line 56: Think this one can be allowed an exception like this:

```suggestion
	for (auto pPassenger : pPassengerList)
		if (this->CanLoadPassenger(pTransport, pPassenger))
			return true;

```

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,270 @@
...
+	Debug::Log("AutoLoadCommandClass::DebugPrintPassenger: Passenger count: %d\n", pPassengers.size());
+	// print address of each passenger
+	for (auto passenger : pPassengers)
+	{
+		Debug::Log("AutoLoadCommandClass::DebugPrintPassenger: Passenger address: %p\n", passenger);
+	}
+}
+
+template <typename P, typename T>
+void SpreadPassengersToTransports(std::vector<P> &pPassengers, std::vector<std::pair<T, int>> &tTransports)
```
Metadorius, line 52: Why not just `passengers` and `transports`? `p` means `pointer`, it's a part of Hungarian notation that was inherited by us from Ares and as a convenience when in hooks we need `this` but can't write it, so we write `pThis`.

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,270 @@
...
+}
+
+void DebugPrintTransport(std::vector<std::pair<TechnoClass *, int>> &tTransports)
+{
+	Debug::Log("AutoLoadCommandClass::DebugPrintTransport: Transport count: %d\n", tTransports.size());
+	// print address of each transport and its passengers
+	for (auto transport : tTransports)
+	{
+		Debug::Log("AutoLoadCommandClass::DebugPrintTransport: Transport address: %p, Now size: %d, Virtual size: %d\n", transport.first, transport.first->Passengers.GetTotalSize(), transport.second);
+	}
```
Metadorius, line 38:
```suggestion
	for (auto transport : tTransports)
		Debug::Log("AutoLoadCommandClass::DebugPrintTransport: Transport address: %p, Now size: %d, Virtual size: %d\n", transport.first, transport.first->Passengers.GetTotalSize(), transport.second);
```

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,270 @@
...
+}
+
+void DebugPrintPassenger(std::vector<TechnoClass *> &pPassengers)
+{
+	Debug::Log("AutoLoadCommandClass::DebugPrintPassenger: Passenger count: %d\n", pPassengers.size());
+	// print address of each passenger
+	for (auto passenger : pPassengers)
+	{
+		Debug::Log("AutoLoadCommandClass::DebugPrintPassenger: Passenger address: %p\n", passenger);
+	}
```
Metadorius, line 48:
```suggestion
	for (auto passenger : pPassengers)
		Debug::Log("AutoLoadCommandClass::DebugPrintPassenger: Passenger address: %p\n", passenger);
```

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,270 @@
...
+{
+	Debug::Log("AutoLoadCommandClass::DebugPrintTransport: Transport count: %d\n", tTransports.size());
+	// print address of each transport and its passengers
+	for (auto transport : tTransports)
+	{
+		Debug::Log("AutoLoadCommandClass::DebugPrintTransport: Transport address: %p, Now size: %d, Virtual size: %d\n", transport.first, transport.first->Passengers.GetTotalSize(), transport.second);
+	}
+}
+
+void DebugPrintPassenger(std::vector<TechnoClass *> &pPassengers)
```
Metadorius, line 41: I don't think that's a pointer either, so could use just `passengers` here and in above.

src/Commands/AutoLoad.cpp
psi-cmd: I may not quite understand this, if you mean `auto *` instead of `auto`?
From my observation, if I search `auto p` globally there is a lot result. However if I search `auto *p` or `auto *` or `auto*` ··· there is seldom cases, so I was like, wondering which is more suitable for this project.

Metadorius: I mean `Type* varName` instead of `Type *varName`

Metadorius: I think I see the  confusion though, I meant to comment on the line that appears right above the comment, not the first displayed line, it's just GitHub showing it like this.

Aephiex:
> For ares tags, maybe we could have the "original" default value for them. For example, `NoManualEnter=False`, `Passengers.BySize=True` are the default behavior without any MOD.

Yes, I defaulted every vehicle to allow maual enter and to count passengers by size.

Aephiex:
https://github.com/psi-cmd/Phobos/pull/2

I forgot to consider the passenger slot budget in my first pull request. So I'm making another pull request to fix it.

Previously:
If you select an amphibious transport with 2 passenger slots remaining and a vehicle that is size 3, and press autoload command, it will try to send the vehicle into the amphibious transport.

Now:
It will not try to send the vehicle into the amphibious transport.

psi-cmd:
(Better to discuss here)
Oh, you repaired that faster than me.

I think originally there will be a strange corner case. If you have a "size=3" passenger and "1 remainsize=2" transport, 1 remainsize=3"​ transport. Unfortunately we select the small remain transport first, and we crossout that only passenger.

One possible solution is to do the sort according to size. However size is not the only restriction, we may have other conditions to decide weather we should load that passenger. As I said in chat, it seems to be a task assignment problem if we want our solution to be cleaver enough.

Aephiex:
@Metadorius 

> Is there a specific reason that this is broken down into two halves exactly at size 2/size 3?

I'm leaving a reply here, because I somehow can't leave a reply into a file.

Why passengers and transports are split by size? In actual play you may want to load your Conscripts into Flak Tracks first, then load the Flak Tracks and any remaining infantry into amphibious transports, to make the most of passenger slots. It's a hard-coded way but it helps for most practical cases. A mod can break it by the introduction of the 4th layer of transport, that even amphibious transports can be loaded, then the logic will not be able to handle that thing, but such a giga transport doesn't really make sense.

To avoid making it too complicated to do, we make a several assumptions:
1. A transport for standard passengers (size <=2) has size >= 3 itself, and if it somehow has size <= 2, it is treated as though it were larger and never try to load it into other standard transports;
2. A transport that can handle size >= 3 is an amphibious transport and never try to load it into anything;
3. The player wants to load standard passengers into standard vehicles first to make the most of passenger slots.

Aephiex:
Also maybe we can consider adding support for bio reactors, too. If the user selects infantry, vehicles, and bio reactors at a same time, the vehicles take precedence. If no infantry can be loaded into transports, then send the infantry into bio reactors. We make a several assumptions to avoid complexity:
1. A Bio Reactor is a building type with `Passengers=<greater_than_zero>`, `InfantryAbsorb=yes` , and `Grinding=no`;
2. Only infantry types can be loaded into a Bio Reactor.

Regarding garrisonable structures, we can't select our own troops and a neutral structure at a same time, and if we add support for garrisonable structures here, we can't tell the user's intention when the user selects 10 Conscripts and 2 Flak Tracks just close to a garrisonable structure and presses autoload command, meaning we may need another command to tell the infantry units to find a garrisonable structure nearby, but again we can't read the user's intention and how long is "nearby". So I suggest to not add garrisonable structure support here.

TaranDahl:
I did not test it in game, but as far as I understand, it will cause desync.
Maybe you should use `ClickedMission`.

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,266 @@
...
+						// so a transport not by size will not need to check against the passenger budget.
+						if (tTransports[index].second + passengerSize > pTransport->GetTechnoType()->Passengers)
+							continue;
+					}
+
+					if (pPassenger->GetCurrentMission() != Mission::Enter)
+					{
+						pPassenger->QueueMission(Mission::Enter, false);
+						pPassenger->SetTarget(nullptr);
+						pPassenger->SetDestination(pTransport, true);
```
TaranDahl, line 117: Use `ClickedMission` here. `RespondToEvent` will handle the rest.

Aephiex:
https://github.com/user-attachments/assets/3e96e258-4570-489e-a902-a650f7d83cbc

Added Bio Reactor and Tank Bunker support.

https://github.com/psi-cmd/Phobos/pull/3

psi-cmd:
That is good, after this feature is complete. I will make another pull request that implement a new T selection. At that time besides select screen/all units with same passengers, select screen/all Tank Bunkers or Bio Reactors may also be implemented.

Aephiex:
> I did not test it in game, but as far as I understand, it will cause desync. Maybe you should use `ClickedMission`.

Just tried that in my local build, and my Conscript attacks the Flak Track instead of getting into it. Maybe we should investigate.

Aephiex:
> That is good, after this feature is complete. I will make another pull request that implement the new T selection. At that time besides select screen/all units with same passengers, select screen/all Tank Bunkers or Bio Reactors may also be implemented.

That's a great quality of life, I'm looking forward.

Metadorius:
> @Metadorius 
> 
> > Is there a specific reason that this is broken down into two halves exactly at size 2/size 3?
> 
> Why passengers and transports are split by size? In actual play you may want to load your Conscripts into Flak Tracks first, then load the Flak Tracks and any remaining infantry into amphibious transports, to make the most of passenger slots. It's a hard-coded way but it helps for most practical cases. A mod can break it by the introduction of the 4th layer of transport, that even amphibious transports can be loaded, then the logic will not be able to handle that thing, but such a giga transport doesn't really make sense.

Well, the greatest algorithms are the most robust. Imagine if a mod makes everything bigger for more precise control, and normal infantry will take 3 slots, then this algorithm fails.

Or imagine a campaign mega unit that is such a transport. A player clicks our new hotkey, expecting it to work as intended, and it doesn't work properly, causing frustration for the player.

I think it is necessary to not make such assumptions about the size here, and I am not sure if this split is necessary, can't you do a sorted vector instead?

Aephiex:
> Well, the greatest algorithms are the most robust. Imagine if a mod makes everything bigger for more precise control, and normal infantry will take 3 slots, then this algorithm fails.
> 
> Or imagine a campaign mega unit that is such a transport. A player clicks our new hotkey, expecting it to work as intended, and it doesn't work properly, causing frustration for the player.
> 
> I think it is necessary to not make such assumptions about the size here, and I am not sure if this split is necessary, can't you do a sorted vector instead?

I have a pseudo code in mind.

```
Map<int, List<Unit>> passengerMap; // unit size is the key
Map<int, List<Unit>> transportMap; // size limit is the key
Map<int, List<Unit>> ambiguousMap; // unit size is the key

for (Unit a : selectedUnits) {
	if (a is an infantry, or is a vehicle that is either Passengers <= 0, no manual entry, or is fully loaded) {
		passengerMap[a.size].add(a);
	}
}

for (Unit a : selectedUnits) {
	if (a is a vehicle with Passengers >= 1, manual entry allowed, and is not already fully loaded) {
		if (passengerMap.isNotEmpty() && a.canLoadAnyAmong(passengerMap)) {
			transportMap[a.sizeLimit].add(a);
		} else {
			ambiguousMap[a.size].add(a);
		}
	}
}

if (ambiguousMap is not empty) {
	List<Unit> passengersNew;

	for (Entry<int, List<Unit>> entry : ambiguousMap) {
		for (Unit b : entry.getValue()) {
			if (passengersNew.isEmpty() || !b.canLoadAnyAmong(passengersNew)) {
				passengerMap[b.size].add(b);
				passengersNew.add(b);
			} else {
				transportMap[b.sizeLimit].add(b);
			}
		}
	}

	// At this point we have no ambiguousity and every unit has its position as a passenger or transport.
}
```

1. Units that are clearly not transports are viewed as passengers;
2. Units that are transports are viewed as transports, however if they have nothing to load then they are put into ambiguousity;
3. Units in ambiguousity are added to the passenger list from the smallest to the largest, and if a unit in ambiguousity find itself able to load any previous one added to the passenger list from ambiguousity, it goes to the transports list instead;
4. At this point every unit is supposed to have been added to either the passengers or the transports, so we just search from the smallest passenger to the largest, from the smallest transport size limit to the largest.

This program should be able to handle such a case you mentioned, and should be able to support any layers of transports and any mixed selections of them.

Aephiex:
Couldn't help to work on it, so I just work on it. :-D

https://github.com/psi-cmd/Phobos/pull/3

Local tests passed, it is now working based on the pseudo code posted above.

@psi-mod  Maybe you will take a look?

Aephiex:
> I did not test it in game, but as far as I understand, it will cause desync. Maybe you should use `ClickedMission`.

Just tried the following implementation.

```
auto crd = pTransport->GetCoords();
auto crdNear = MapClass::GetRandomCoordsNear(crd, 1, false);
pPassenger->ClickedMission(Mission::Enter, pTransport, MapClass::Instance->GetCellAt(crd), MapClass::Instance->GetCellAt(crdNear));
```

And it still desynced. Maybe I shouldn't use the `GetRandomCoordsNear`?

Aephiex:
Okay, let's summarize the status of this pull request.

The features of auto load command:
1. Can handle any layers of cascading transports and any mixed selection of them;
2. Can handle Ares's passenger whitelist and blacklist feature;
3. Passengers are loaded from larger to smaller;
4. Transports are used from smaller size limit to larger size limit;
5. If no auto loading can happen between selected units, then try to load selected units into selected Bio Reactors and Tank Bunkers.

The only thing we left behind is garrisonable structures, but we can't select our own troops and a neutral garrisonable structure at a same time, meaning we need other approaches to do it.

Known issues that must be solved:
1. `pPassenger->QueueMission(Mission::Enter, false);` causes multiplayer desync;
2. `pPassenger->ClickedMission(Mission::Enter, pTransport, ...` makes passengers attack the transport as well as it causes multiplayer desync.

We must find a way to give directives to passengers without causing multiplayer desync.

TaranDahl:
> Okay, let's summarize the status of this pull request.
> 
> The features of auto load command:
> 
> 1. Can handle any layers of cascading transports and any mixed selection of them;
> 2. Can handle Ares's passenger whitelist and blacklist feature;
> 3. Passengers are loaded from larger to smaller;
> 4. Transports are used from smaller size limit to larger size limit;
> 5. If no auto loading can happen between selected units, then try to load selected units into selected Bio Reactors and Tank Bunkers.
> 
> The only thing we left behind is garrisonable structures, but we can't select our own troops and a neutral garrisonable structure at a same time, meaning we need other approaches to do it.
> 
> Known issues that must be solved:
> 
> 1. `pPassenger->QueueMission(Mission::Enter, false);` causes multiplayer desync;
> 2. `pPassenger->ClickedMission(Mission::Enter, pTransport, ...` makes passengers attack the transport as well as it causes multiplayer desync.
> 
> We must find a way to give directives to passengers without causing multiplayer desync.

You may try using `ObjectClickedAction`, just like [https://github.com/Phobos-developers/Phobos/pull/1453](url) did.
By the way, as far as I understand, is what this pr wants to do similar to pr1453?

Aephiex:
> You may try using `ObjectClickedAction`, just like [https://github.com/Phobos-developers/Phobos/pull/1453](url) did. By the way, as far as I understand, is what this pr wants to do similar to pr1453?

Thank you, `ObjectClickedAction` worked.

The said pull request is more generalist that it can handle many stuff, more than just loading units into vehicles, but it isn't quite clear what are "similar" to the initial target, and I don't know how it is going to handle passenger whitelists and blacklists.

Aephiex:
@psi-cmd Sorry to disturb you again. Multiplayer desync is solved.

https://github.com/psi-cmd/Phobos/pull/4

I didn't find the function myself, but someone else did and is already using it, clearly far before I get to know such a function existed. My feeling like a dumb student, was only told the correct answer but have no idea how to get it.

psi-cmd:
I was doing final project these days and sorry for delayed cache up.
Nevermind, there are really some people knows too much about how everything should works. Just not be depressed about that, what you have done is faster and more enthusiatic than I think.

psi-cmd:
> You may try using `ObjectClickedAction`, just like [https://github.com/Phobos-developers/Phobos/pull/1453](url) did. By the way, as far as I understand, is what this pr wants to do similar to pr1453?

https://github.com/Phobos-developers/Phobos/commit/325989ff2df243b632da2275d7eef337e8db4eec#diff-66722f09d599809b82cabd274dc71ebc0717bc008a437b77d1833c1605564ed6R158-R169

The pr 1453 implement a more general batch operation by intercept the target selection. Our PR spend a lot of code on equally and recursive matching of selected units. So there are differences.

psi-cmd:
> @psi-cmd Sorry to disturb you again. Multiplayer desync is solved.
> 
> [psi-cmd#4](https://github.com/psi-cmd/Phobos/pull/4)
> 
> I didn't find the function myself, but someone else did and is already using it, clearly far before I get to know such a function existed. My feeling like a dumb student, was only told the correct answer but have no idea how to get it.

Oh we really need investigate, the troops are attacking transports now.

Aephiex:
https://github.com/psi-cmd/Phobos/pull/5

Documentized.

Aephiex:
> Oh we really need investigate, the troops are attacking transports now.
> 
> Edit: the reason for attacking is because we set our shortcut to ctrl+sth, so the virtual click leads to force attack.
> 
> So if we change `pPassenger->ObjectClickedAction(Action::Enter, pTransport, false);` to true, and bind hotkey to ctrl+sth, then everything works.

Just tried in my local build, if the 3rd parameter was true, then nothing happens when I click the hotkey, no matter the hotkey was V or Ctrl+V.

And if the 3rd parameter was false, V hotkey works as normal and Ctrl+V hotkey turns into a force attack command.

Maybe we simply note in the document that the hotkey shouldn't be Ctrl+Sth.

psi-cmd:
That is different from my test may be, I set ctrl+a as shortcut and force attack is not triggered after the last parameter set to true.

Aephiex:
> That is different from my test may be, I set ctrl+a as shortcut and force attack is not triggered after the ignoreForce set to true.
> 
> And shift+a also works now since ignoreForce set to true.

Can confirm your words when I git merge your last change into my local branch. It's odd, maybe something was wrong on my last local test. Anyway, for now it is solved.

Also, maybe `MakeCommand<AutoLoadCommandClass>();` shouldn't be in the developer commands, that's only available to debug builds, and is never available to release builds. It requires `[GlobalControls] DebugKeysEnabled=yes` in `rulesmd.ini`, and it requires current Phobos build to be debug build, as we can see in `Phobos\src\Phobos.INI.cpp` at around line 240. If it was intentional because it is still WIP, remember to make it available to release builds when everything seem ready.

Aephiex (approved):

Aephiex:
@chaserli Hello, this pull request seems ready for review. Maybe take a look when you have time?

docs/User-Interface.md
```diff
@@ -317,6 +317,18 @@ SelectionFlashDuration=0  ; integer, number of frames
 - Switches on/off [frame by frame mode](Miscellanous.html#frame-step-in).
 - For localization add `TXT_FRAME_BY_FRAME` and `TXT_FRAME_BY_FRAME_DESC` into your `.csf` file.
 
+### `[ ]` Auto Load
+- A shortcut to quickly command multiple units to board multiple transports at a same time. Select units and valid transports, then press the auto load hotkey, the units will be distributed among the transports and will be ordered to board them.
+  - Transports can be considered passengers if fully loaded, or `NoManualEnter=yes`, or it can't actually load anything from the selected passengers.
+  - Larger passengers are loaded first, and transports with smaller size limits are used first.
+  - At a given unit size and size limit, passengers will be diversely distributed into transports if possible.
+  - Ares `Passengers.Allowed=` and `Passengers.Disallowed=` are taken into account.
+- It also supports Bio Reactors and Tank Bunkers. Select valid candidates and multiple said buildings while pressing Shift, then press the auto load hotkey, the units will be distributed among these buildings and will be ordered to enter them.
```
Metadorius, line 326: What about Soviet Battle Bunkers and other structures that you own and can select and can have garrisons?

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,380 @@
...
+				transports.end());
+
+			if (passengerMap[leastpID].size() == 0)
+			{
+				passengerMap.erase(leastpID);
+			}
+			else
+			{
+				passengerMapIdle[leastpID] = passengerMap[leastpID];
+			}
```
Metadorius, line 151:
```suggestion
			if (passengerMap[leastpID].size() == 0)
				passengerMap.erase(leastpID);
			else
				passengerMapIdle[leastpID] = passengerMap[leastpID];
```

CREDITS.md
```diff
@@ -408,3 +408,4 @@ This page lists all the individual contributions to the project by their author.
 - **Damfoos** - extensive and thorough testing
 - **Dmitry Volkov** - extensive and thorough testing
 - **Rise of the East community** - extensive playtesting of in-dev features
+- **psi-cmd**, **Aephiex** - autoload hotkey command
```
Metadorius, line 411: We do one line per contributor, also place yourself closer to the developers specifically.

Metadorius (approved):
Looks good, lemme fix some of the style stuff quickly.

docs/User-Interface.md
Aephiex: Garrisonable structures are currently not supported. Neutral structures and owned troops can't be selected at a same time. Also can't read the user's intention when a bio reactor, a battle bunker, and a group of conscripts are selected at a same time. Maybe we need another shortcut key for that.

Metadorius: > Also can't read the user's intention when a bio reactor, a battle bunker, and a group of conscripts are selected at a same time.

Everyone who can fit in bio reactor but not the garrison goes to bio reactor, everyone else is spread between bunkers and reactors. I think the situation when battle bunkers and reactors will be selected is rather rare so there's no need to do much thought about it

Aephiex: You are right. I will add such support later. However since neutral structures still can't be selected with owned troops at a same time, this will only ease the loading of battle bunkers, which is always owned by a player even when not garrisoned.

chaserli (changes requested):
Redundant sanity checks such as misuse of abstract_cast and all kinds of sanity check on xxExt are not the most significant code style issues in the current code. 

I have the impression that the entire selection-filtering is somehow repeated. I can think of 2 ways: either you just let the game tell which cursor to use and select the "load onto vehicle" cursors among them. Or just handle the network event directly. 

If you don't fully understand how network events are handled, you can check the  implementation of the virtual function you've chosen. Methods in EventClass.h in yrpp is also updated to handle them directly.

Metadorius:
@chaserli it will be good to see the actual explanations of why do you think something is wrong (like the cast thing) and how and it should be done properly. Code reviews should be done so the person can learn and grow, saying "it is wrong" without explaining when you know the answer is not productive.

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,484 @@
...
+{
+	if (pPassenger->WhatAmI() == AbstractType::Infantry)
+	{
+		// Bio Reactor & garrisonable structure
+		return !IsMindControlling(pPassenger);
+	}
+	else if (pPassenger->WhatAmI() == AbstractType::Unit)
+	{
+		// Tank Bunker
+		return pPassenger->GetTechnoType()->Turret
```
TaranDahl, line 87: Missed a lot of checks. See sub_70FB50.

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,484 @@
...
+}
+
+// Oddly, a Tank Bunker can't load anything when selected by the player.
+// Therefore it has to be unselected.
+inline static void DeselectMe(TechnoClass* pTransport)
+{
+	if (pTransport->WhatAmI() == AbstractType::Building)
+	{
+		auto pBuilding = abstract_cast<BuildingClass*>(pTransport);
+		auto pBuildingType = abstract_cast<BuildingTypeClass*>(pBuilding->GetTechnoType());
```
TaranDahl, line 190: pBuilding->Type

src/Commands/AutoLoad.cpp
TaranDahl: 如果你不想仔细检查每个可能的进入关系是否成立的话，可以试试使用MouseOverObject。它会返回一个Action，只要返回的Action是Enter那应该就是可以进入的。不过这个函数会进行很多很多判断，如果你要用的话最好实际测试一下会不会造成卡顿。
If you don't want to carefully check whether each possible entry relationship is established, you can try using MouseOverObject. It will return an Action. As long as the returned Action is Enter, it should be possible to enter. However, this function will make many, many checks. If you decide to use it, it is best to actually test whether it will cause lag.

Aephiex: > Missed a lot of checks. See sub_70FB50.
As far as I read from the assembly, it seems hovered units, parasited units, and those without a "turret changer" weapon, are never eligible to enter a Tank Bunker, right?
I don't know what is a turret changer. The rest can be added to the filter.

Aephiex:
@chaserli Hello, redundancy have been migrated to inline functions, so it doesn't look as redundant as before. Maybe you can take a look again when you have time. Also I don't know what's wrong with the "misuse" of `abstract_cast`, maybe you will elaborate.

src/Commands/AutoLoad.cpp
TaranDahl: You can directly reference that function by using `reinterpret_cast`. The use cases can be found in the project.

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,490 @@
...
+	if (pTransport->WhatAmI() == AbstractType::Unit)
+	{
+		auto pTechnoType = pTransport->GetTechnoType();
+		auto pTypeExt = TechnoTypeExt::ExtMap.Find(pTechnoType);
+		// the check of MCing or MCed are redundant here, see inline function "CanBeVehiclePassenger".
+		return pTypeExt->CanLoadPassenger(pTransport, pPassenger);
+	}
+	else if (pTransport->WhatAmI() == AbstractType::Building)
+	{
+		auto pBuilding = abstract_cast<BuildingClass*>(pTransport);
```
CrimRecya, line 140: `else if (auto pBuilding = abstract_cast<BuildingClass*>(pTransport))`

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,490 @@
...
+			return pPassenger->WhatAmI() == AbstractType::Infantry
+				&& pTypeExt->CanLoadPassenger(pTransport, pPassenger);
+		}
+		else if (pBuildingType->CanBeOccupied)
+		{
+			// garrisonable structure
+			// the check of MCing is redundant here, see inline function "CanBeBuildingPassenger".
+			return pPassenger->WhatAmI() == AbstractType::Infantry
+				&& !pPassenger->IsMindControlled()
+				&& abstract_cast<InfantryTypeClass*>(pPassenger->GetTechnoType())->Occupier
```
CrimRecya, line 156: `static_cast<InfantryTypeClass*>(pPassenger->GetTechnoType())->Occupier`

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,490 @@
...
+	}
+	return false;
+}
+
+// Checks if this transport substrats passenger budget by passenger size.
+inline static const bool IsBySize(TechnoClass* pTransport)
+{
+	if (pTransport->WhatAmI() == AbstractType::Unit)
+	{
+		auto pType = pTransport->GetTechnoType();
```
CrimRecya, line 174:
```
if (auto pUnit = abstract_cast<UnitClass*>(pTransport))
{
  auto pType = pUnit->Type;
  ...
}
```

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,490 @@
...
+	return true;
+}
+
+// Oddly, a Tank Bunker can't load anything when selected by the player.
+// Therefore it has to be unselected.
+inline static bool DeselectMe(TechnoClass* pTransport)
+{
+	if (pTransport->WhatAmI() == AbstractType::Building)
+	{
+		auto pBuilding = abstract_cast<BuildingClass*>(pTransport);
```
CrimRecya, line 191: `if (auto pBuilding = abstract_cast<BuildingClass*>(pTransport))`

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,490 @@
...
+		TechnoClass* pTechno = abstract_cast<TechnoClass*>(pUnit);
+
+		// If not a techno, or is in air, then it can't be a passenger, and it can't be a Bio Reactor or Tank Bunker.
+		if (!pTechno || pTechno->IsInAir())
+			continue;
+
+		// Detect enterable buildings.
+		if (pTechno->WhatAmI() == AbstractType::Building)
+		{
+			auto pBuilding = abstract_cast<BuildingClass*>(pTechno);
```
CrimRecya, line 347: `if (auto pBuilding = abstract_cast<BuildingClass*>(pTechno))`

src/Commands/AutoLoad.cpp
```diff
@@ -0,0 +1,490 @@
...
+					if (!CanHoldPassenger(pTransport, pPassenger))
+						goto seeNextTransport; // "continue" causes to skip the passenger, not to skip the transport, so we shouldn't use "continue" here
+
+					// Gets the passenger slot budget that would be substracted.
+					int passengerSize;
+					if (IsBySize(pTransport))
+					{
+						// If by size then get the actual size of the potential passenger.
+						// Otherwise, every potential passenger counts as size 1.
+						passengerSize = static_cast<int>(abstract_cast<TechnoClass*>(pPassenger)->GetTechnoType()->Size);
```
CrimRecya, line 254: If it is determined that `pPassenger` is a `TechnoClass*`, you can replace the redundant `abstract_cast` with `reinterpret_cast`

CrimRecya (commented):
It seems that you understand the `abstract_cast`. But I found some parts that can be omitted to avoid repeating check.

Aephiex:
> It seems that you understand the `abstract_cast`. But I found some parts that can be omitted to avoid repeating check.

Thank you for your detailed review. I appreciate.

src/Commands/AutoLoad.cpp
Aephiex: Wow, didn't know that, thank you.

PR #1453, opened 2024-12-15
[Highly Customized] Distribution click action mode

CrimRecya:
`[ ]` Distribution Mode Spread / Filter / Enable

- Now you can change the click action by using `AllowSwitchNoMoveCommand` hotkey. If the behavior to be executed by the current techno is different from the behavior displayed by the mouse, and the behavior to be executed will make the techno move near the target, the behavior will be replaced with area guard. Regardless of whether or not switch hotkey is used, default behavior can be changed through `DefaultApplyNoMoveCommand`.
- Now you can also change the click action when hold down the specific hotkey if enabled `AllowDistributionCommand`. The new behavior is like using the selected objects one by one to click on each target within the range.
- `AllowDistributionCommand.SpreadMode` & `AllowDistributionCommand.FilterMode` allow you to set spread range and target filter by hotkeys, which default to `DefaultDistributionSpreadMode` and `DefaultDistributionFilterMode`.
  - When the range is 0, it is the original default behavior of the game. The range can be adjusted to 4, 8 or 16 cells by another shortcut key. You can also adjust this by using the mouse wheel while holding down the specific hotkey if `AllowDistributionCommand.SpreadModeScroll` set to true;
    - The targets within the range will be allocated equally to the selected technos. Only when the behavior to be performed by the current techno is the same as that displayed by the mouse will it be allocated. Otherwise, it will return to the original default behavior of the game (it will not be effective for technos in the air). This will display a range ring.
  - When the filter is `None`, it is the default behavior of the game. If the range is not zero at this time, a green ring will be displayed. You can adjust the filter mode to:
    - `Like` - only targets with the same armor type (Completely identical `Armor`) will be selected among the targets allocated in the range. At this time, a blue ring will be displayed.
    - `Type` - only targets of the same type (like infantries, vehicles or buildings) will be selected among the targets allocated in the range. At this time, a yellow ring will be displayed.
    - `Name` - only targets of the same name (or with the same `GroupAs`) will be selected among the targets allocated in the range. At this time, a red ring will be displayed.
- `AllowDistributionCommand.AffectsAllies` & `AllowDistributionCommand.AffectsEnemies` allow the distribution command to work on allies (including owner) or enemies target. If picking a target that's not eligible, it'll fallback to vanilla command.
- It's possible to add a button for distribution mode in the bottom bar by adding `DistributionMode` in the `ButtonList` of `AdvancedCommandBar` and `MultiplayerAdvancedCommandBar`.
  - The positions of each button are hardcoded, so it'll only decide whether enable this button or not. Distribute Mode button is now always listed after all the vanilla ones.
  - The asset of these buttons should be added in `sidec0x.mix` files which correspond to different sides, with the name `button12.shp`.
- For localization add `TXT_SWITCH_NOMOVE`, `TXT_DISTR_SPREAD`, `TXT_DISTR_FILTER`, `TXT_DISTR_HOLDDOWN`, `TXT_SWITCH_NOMOVE_DESC`, `TXT_DISTR_SPREAD_DESC`, `TXT_DISTR_FILTER_DESC`, `TXT_DISTR_HOLDDOWN_DESC`, `MSG:DistributionModeOn`, `MSG:DistributionModeOff`, `TIP:DistributionMode` into your `.csf` file.

In `rulesmd.ini`:
```ini
[GlobalControls]
AllowSwitchNoMoveCommand=false                      ; boolean
AllowDistributionCommand=false                      ; boolean
AllowDistributionCommand.SpreadMode=true            ; boolean
AllowDistributionCommand.SpreadModeScroll=true      ; boolean
AllowDistributionCommand.FilterMode=true            ; boolean
AllowDistributionCommand.AffectsAllies=true         ; boolean
AllowDistributionCommand.AffectsEnemies=true        ; boolean

[AudioVisual]
StartDistributionModeSound=                         ; sound entry
EndDistributionModeSound=                           ; sound entry
AddDistributionModeCommandSound=                    ; sound entry
```

In `ra2md.ini`:
```ini
[Phobos]
DefaultApplyNoMoveCommand=true                      ; boolean
DefaultDistributionSpreadMode=2                     ; integer, 0 - r=0 , 1 - r=4 , 2 - r=8 , 3 - r=16
DefaultDistributionFilterMode=2                     ; integer, 0 - None , 1 - Like , 2 - Type , 3 - Name
```

In `uimd.ini`:
```ini
[AdvancedCommandBar]
ButtonList=[Button1],DistributionMode,[ButtonX]     ; List of button entry

[MultiplayerAdvancedCommandBar]
ButtonList=[Button1],DistributionMode,[ButtonX]     ; List of button entry
```

https://github.com/user-attachments/assets/6f343fef-3003-408f-8703-2f2ff22dc072


TaranDahl (approved):
lgtm

CrimRecya:
In theory, it should work normally under the condition of multiple people, but I haven't tested it yet. It would be great if it could work as expected.
In addition, as for the display of the current mode, the scheme has not been finalized yet. At present, I want it to be displayed directly at the mouse position, such as the display range of SW, instead of using print message like in video. If anyone has any good ideas, PLEASE comment.

TaranDahl:
Tested locally. Doing good.

TaranDahl:
> I am hesitant on how to deal with this feature. It removes a huge amount of microcontrol from the game, essentially automating half of whatever required skill before and now just requires a mouse click.-----------谷歌翻译-----------我对如何处理此功能感到犹豫。它从游戏中删除了大量的微控制，从本质上讲是在此之前所需的任何技能的一半，现在只需单击鼠标即可。

I think all you want is a toggle?

Metadorius:
I would say that if we will merge it, a toggle is a must, because the whole existence of such a hotkey alters gameplay significantly, removing a huge amount of microcontrol requireemtn.

The bigger question is where do we draw the line on what should be included in mainline Phobos and what not? In other words, what is the scope of the project?

This is an exaggeration, but just to show the point that I am trying to make -- imagine suddenly a pull request appears that is intended specifically to turn YR into a whole different genre of game (say, a visual novel with no RTS stuff at all) and is not usable for purposes of regular RTS players or modders. Should we merge it at all? I don't think we should, it is, as we say, out of scope.

The main question is where do we draw the line between such features that are out of scope and the features that should be merged and are in the scope? And what potential drawbacks are there if we merge it?

TaranDahl:
> The main question is where do we draw the line between such features that are out of scope and the features that should be merged and are in the scope?

Personally, I came to Phobos because everyone knows Phobos. If I get something done here, then people will know it's done and other people don't need to spend time on what I've already done. I think it is meaningful.
So for me, as long as I think this feature is likely to be used by others, then it is not out of scope.

> And what potential drawbacks are there if we merge it?

Negligible performance impact.
Careless devs can introduce issues that can cause crashes even if features are not enabled.

Metadorius:
> as for the display of the current mode, the scheme has not been finalized yet. At present, I want it to be displayed directly at the mouse position, such as the display range of SW, instead of using print message like in video. If anyone has any good ideas, PLEASE comment.

I have an idea.

First, the radius can be controlled not via hotkeys, but with a mouse. How that would work? You press a hotkey to activate distribution mode, and when you press right mouse button on the target and drag your mouse without releasing the mouse button - it displays the circle and changes the radius depending on how far you drag the cursor, so you regulate the radius with the mouse drag. Then when you release the mouse - the orders are given to every selected unit. You do that each time and effectively draw circles to give orders.

This is inspired by Call to Arms: Gates of Hell game. There you can spread your infantry among covers this way, click behind covers and drag to spread people better.

CrimRecya:
> I would say that if we will merge it, a toggle is a must, because the whole existence of such a hotkey alters gameplay significantly, removing a huge amount of microcontrol requireemtn.

Yes, I think we still need a toggle to enable this feature, not just defining buttons. I will add later.

> The bigger question is where do we draw the line on what should be included in mainline Phobos and what not? In other words, what is the scope of the project?

Personally, if someone (whether they are modders or task package authors) needs a certain feature and I happen to be interested in it, I will try to do it. Whether to merge or not is actually related to whether more people will use it. But no matter what, these features will only be pushed if there is a demand.

> This is an exaggeration, but just to show the point that I am trying to make -- imagine suddenly a pull request appears that is intended specifically to turn YR into a whole different genre of game (say, a visual novel with no RTS stuff at all) and is not usable for purposes of regular RTS players or modders. Should we merge it at all? I don't think we should, it is, as we say, out of scope.

If it is assessed that this should not be merged, then I will close it.

> The main question is where do we draw the line between such features that are out of scope and the features that should be merged and are in the scope? And what potential drawbacks are there if we merge it?

As TaranDahl said, for example, if this feature does not have any impact on performance and can be disabled (assuming I have already added the switch), then I think it can be merged.

CrimRecya:
The opening video is based on the initial design, which is now outdated. I will share another one when I have time.

CrimRecya:
> > but I'd prefer it to be another optional manifestation of this feature.
> 
> Do you mean toggle between hotkey radius selection and mouse adjustment? I am not sure if the hotkey radius selection is intuitive enough to be left in.
> 
> Compare "a single long enough drag to give and customize the order" against "toggle a hotkey or multiple, give an order, toggle hotkeys again". It is also far more intuitive to adjust the radius with a mouse, just like you do in a 2D or 3D graphics editor when you create a circle or, say, a cylinder, and change the size with a mouse. Hotkey doesn't have immediate visual feedback, you have to read the number and translate it into a circle in your mind, or, alternatively, you'll have a constantly floating circle around, and a toggle to enable/disable the feature altogether. If you implement it like I suggested - there will be much less friction for the user.

I currently have a new solution that uses a mouse wheel to assist. Still testing and waiting for feedback.

Metadorius:
> I currently have a new solution that uses the mouse wheel to assist. Still testing and waiting for feedback.

Sounds like an interesting option, is that what you mentioned earlier?

DeathFishAtEase:
> Do you mean toggle between hotkey radius selection and mouse adjustment? I am not sure if the hotkey radius selection is intuitive enough to be left in.

不，我认为它们都应当是可选的，它们的具体效果都非常有趣，它们也同样会有人希望使用，所以只要给足 mod 创作者是否使用这个效果的自由，再确保它们不会拖累 FPS，那么它们就是可以保留的。

---
No, I mean they should all be optional.
Their specific effects are very interesting, and there are also always have people who would want to use them.
So as long as **mod creators are given the freedom to choose** whether to use these effects and ensure that they don't drag down the FPS, then they can be kept.

CrimRecya:
> Sounds like an interesting option, is that what you mentioned earlier?

Yes, it's based on #1522 .

DeathFishAtEase:
> This is an exaggeration, but just to show the point that I am trying to make -- imagine suddenly a pull request appears that is intended specifically to turn YR into a whole different genre of game (say, a visual novel with no RTS stuff at all) and is not usable for purposes of regular RTS players or modders. Should we merge it at all? I don't think we should, it is, as we say, out of scope.

Have you heard of *TS: Another World*? It is a YR MOD that is set in the world of TS, retains RTS elements, and **features visual novel-style missions**. We might not need to focus on requirements that may deviate from RTS content, but if someone provides a feature, accepting it with the option to disable it is not an issue.

---
> 这有点夸张，但只是为了说明我想要表达的一点——想象一下突然出现一个拉取请求，它专门用于将 YR 变成一个完全不同类型的游戏（比如，一部根本没有 RTS 内容的视觉小说），并且不适用于普通 RTS 玩家或 Mod 作者。我们应该合并它吗？我认为我们不应该这样做，正如我们所说，这超出了范围。

你听说过《TS:AnotherWorld》吗，它是一个以TS为世界背景、保留RTS内容，任务内容属于视觉小说的YR MOD。我们或许不必要将可能有些偏离 RTS 内容的需求作为重点，但如果有人提供了功能那么在允许关闭的情况下接收它并不是什么问题。

Metadorius:
Well, after a bit of thinking and input from you all I think that if we gate it behind a toggle - it's fine. The feature is isolated and doesn't make a big impact neither at modders, nor at users, nor at developers if it's off. Of course, it's a maintenance cost, but hopefully there won't be much problems with it.

Metadorius:
GIF 20 02 2025 17-23-00

Another example for the approach I suggested. This GIF is a recording of an RTS game Zero-K, based on Spring RTS engine. It has one of the best in class user interface feature, they make playing the game a delight. You can draw paths or areas with the mouse and it's all very intuitive. I recommend you to give a look and maybe even try it.
Can be downloaded for free on [Steam](https://store.steampowered.com/app/334920/ZeroK) or without it from the [official site](https://zero-k.info/).

Metadorius:
(If someone is not familiar with Supreme Commander/Total Annihilation style gameplay: the GIF shows designating metal extractors to be built at all spots that are available in the area designated by drag and click.)

DeathFishAtEase:
Personally, I might prefer to use two separate clicks to determine the center and radius (the second click on a point on the arc) rather than long-pressing and dragging, unless it's for weapons like bows in a first-person perspective that require a special feel :P

Coronia:
think this is a better user interaction approach than https://github.com/Phobos-developers/Phobos/pull/1452, which is using mouse click to trigger the following command. The other one has the drawbacks of 1) having to select both passengers and transport then uses hotkey, which is kinda against players' habit 2) the hotkey setting visible issue in https://github.com/Phobos-developers/Phobos/issues/1496

However, Autoload shortcut PR has the advantage of calculating whether a passenger should get into a transport more accurately. If we're thinking a way to unify these 2 PRs, I would suggest use Distribution click as a base, while adding some passenger picking logic from Autoload shortcut

Coronia:
one thing I'm concerning is whether `DefaultDistributionFilterMode` should be allowed to customize by Players, since it requires much learning cost. The filter rule is fairly complicated to Players who don't know about ini, and there's no place to explain it clearly in game, so they might not be able to tell what these modes are exactly doing, which could result in confusion and misclick. Meanwhile, the current interaction is 'click a hotkey and rotate between 4 modes', which is also kinda unintuitive since Players can only know what they're currently selected by the color of ring, which again couldn't be explained to Players well. While `DefaultDistributionSpreadMode` uses the same method, the difference of ring size is at least self-explanatory

IMO `DefaultDistributionFilterMode` should be a [General] thing that's controlled by rules, while `DefaultDistributionSpreadMode` could only use mouse wheel to control its size, no need to add more hotkeys here other than the one to enable distribution mode

CrimRecya:
> The opening video is based on the initial design, which is now outdated. I will share another one when I have time.

A late video.

https://github.com/user-attachments/assets/e2e3b5f3-f472-43ac-96be-4e42a40df2a6

CrimRecya:
> think this is a better user interaction approach than #1452, which is using mouse click to trigger the following command. The other one has the drawbacks of 1) having to select both passengers and transport then uses hotkey, which is kinda against players' habit 2) the hotkey setting visible issue in #1496

The visible issue was solved a long time ago.

> However, Autoload shortcut PR has the advantage of calculating whether a passenger should get into a transport more accurately. If we're thinking a way to unify these 2 PRs, I would suggest use Distribution click as a base, while adding some passenger picking logic from Autoload shortcut

I don't plan to do that at the moment, that's a specially designed feature.

> IMO DefaultDistributionFilterMode should be a [General] thing that's controlled by rules, while DefaultDistributionSpreadMode could only use mouse wheel to control its size, no need to add more hotkeys here other than the one to enable distribution mode

I think this is a very simple classification method. Players only need a little time and try using it a few times, and they can clearly understand the difference between them. The display of different colors may seem hasty, but I don't think it's a reason to prevent players from freely choosing.

Metadorius (changes requested):
I consulted with @ZivDero and we agree that currently the feature is good in the concept but rather clunky to use in details. I haven't had the chance to review the code just yet, so pardon if I repeat something  that is already implemented

1. Different modes and customisations like whether to affect allies seem to be an overcomplication. What I think has to be done and that is more intuitive is replicating the mouse cursor action over all eligible targets in the range.

   I imagine the following algorithm:
   1) get currently shown action on the cursor (like attack, enter, capture) (What_Action/MouseOverObject/MouseOverCell call)
   2) use this to enumerate all eligible targets in the circle (so, enter would enumerate all enterable buildings, capture would enumerate all capturable buildings, repair would enumerate all repairable buildings, attack -- all enemies that are legit for attack like the vanilla cursor, move/guard/attack move -- units get random destinations to unblocked cells in the circle, force attack civilians -- all civilians in the circle, heal allies -- all allies in the circle, etc etc etc)
   3) uniformly randomly and non-repetitively assign each individual target, simulating a mouse click on every unit or cell. Non-repetitively means that unless there are more units than targets -- no target should get 2 or more units assigned to it; if there are -- just repeat the process until there are no unassigned units left.
   
   So, to recap, this essentially means that we use the cursor to pick eligible targets, and then simulate mouse click (without voice) for each separate unit from our selection to a randomly chosen not-yet-clicked target. I've did this in TA branch here: https://github.com/Phobos-developers/Phobos/pull/352/files#diff-9698616732fcbee5d98549d09c43f973aa6e3c3a2bba245f1a2427c9972894f2R492-R549
   
2. The hotkeys and the circle control could be brought in-line with existing implementations:
   - I would still advocate that it would be more useful to have the user specify radius with mouse drag, like on the image I've shown. There are multiple RTS and RTT games that use drag-distribution or drag-positioning (Zero-K, Beyond All Reason, Gates of Hell, IIRC C&C3 also uses drag-select; Total War uses drag for spreading units), so it won't be unnatural.
   - The hotkey could work like `[Z] Waypoint mode`: hold to enter it and then dragging with something selected would produce a circle.
   - That being said, I don't object for optionally (user togglable) using mouse wheel to select the circle radius as alternative option. However, I am not sure on the exact implementation.
   
3. Maybe an extra, but it would be extra cool to show the targets which are going to be affected in the circle by highlighting it with a white-ish color, the same flash as the target flash. It would be highlighted like the target flashing when you press on it, but not flash until you release it. This would immediately tell the player what is considered a target.
   - For "move" or "guard" command showing specific unit positions as faint silhouttes for the units would be cherry on top, but I realise this is a big workload, so it should be left for later. Just something to keep in mind. I know AutoGavy has such implemented for his formations in Tiberium Crisis, maybe he could share the code for that later.

Coronia:
The affect allies/enemies setting is needed cuz there's indeed a demand that some people would only like to use it for auto loading instead of auto targeting, since the latter one might break existing stuffs.

I can agree that extra options for Spread Mode and Filter Mode can be confusing and unfriendly to players, hence the other toggles are made to disable it. Not sure if directly removing them would be a good idea tho, also the scroll of Spread Mode is fine and intuitive as I illustrated above so this one could be kept at least.

Other settings are mostly worked like what you've proposed already, except the 'selection flash' thing.

Metadorius:
> also the scroll of Spread Mode is fine and intuitive as I illustrated above so this one could be kept at least.

Yeah I don't mind it being an option instead of "drag" mode, and I agree it is intuitive too. But I think while both modes are convenient to use, that "press and drag to specify radius" behavior should be the default, because from my observation it is far more common in RTS/RTT games (wouldn't mind to hear any examples of games where scroll is used like this). Also, while niche, some people may play on laptop and click-drag is much easier to execute on a trackpad.

CrimRecya:
> I guess could use green/yellow/red to be in-line with ConditionYellow and ConditionRed, but I am not sure on the colors. Not sure on what could be the best UI for the hotkeys here, but having this as "level" could mean there would be "make filter more specific" and "make filter less specific" hotkeys.

At present, the filter mode uses a single hotkey cycle switching modes. Replacing with two hotkeys is indeed more straightforward. But I think it may not be very convenient to operate.
There are not many modes available for filtering now, and a single hotkey should be sufficient. In addition, I have received feedback from players that too many hotkeys are not conducive to operation, especially when the keys on the left-hand side of the keyboard are already heavily occupied.
So I don't quite agree with the suggestion of splitting it into two hotkeys.

CrimRecya:
I haven't thought of any good methods for dragging and flashing for now. I will consider it later.

Metadorius:
Same should be done to disguised units I think

CrimRecya:
> Same should be done to disguised units I think

The mouse behavior on disguised units will be different, so there is no need for special adaptation

Coronia (approved):
as per discussed with CrimRecya, it already has most of the feature done, and the requested things can be added without needing to alter the existing stuffs. Better merge this first and do further tweak if needed

DeathFishAtEase:
Since this PR is labeled as `T2 maintainer review` is sufficient, it is indeed better for another T2 maintainer to review the code first and confirm approval. Especially it was previously in the Request changed state. Especially it was previously in the Request changed state.

TaranDahl:
> No, the fundamental reason is the lack of active T2 maintainers and slow "growth" of new T2 maintainers, which is the cause of slow review process. 

我去看了Teams，我们有8位T2维护者。即使现在有4位因为各种原因而无法审查PR，还有另外4位。他们在做什么？
I checked Teams. We have 8 T2 maintainers. Even if 4 of them are unable to review PRs for various reasons at the moment, there are still 4 others. What are they doing?

Metadorius:
> 我去看了Teams，我们有8位T2维护者。即使现在有4位因为各种原因而无法审查PR，还有另外4位。他们在做什么？ I checked Teams. We have 8 T2 maintainers. Even if 4 of them are unable to review PRs for various reasons at the moment, there are still 4 others. What are they doing?

The teams are not representative of *active* maintainers. For the past years I've been executing the policy of keeping historical roles of people and not removing them from the team out of courtesy for their past work. It's a hobby after all, maybe they'll want to return some day and it would be mentally easier for them to do so.

See active maintainers in the README.md instead.

Metadorius:
Maybe, however, let's not stray away from the core issue.

ZivDero:
@TaranDahl 
> there are still 4 others. What are they doing?

I dunno? Did you ask them? Yes, in an ideal world reviewers come by themselves. But as you've said, we're very short on manpower. I have TS, Starkku has PP. However, if the authors asked either of us, I'm sure they would've received feedback. Maybe not immediately. Maybe it'd take a few weeks even. But within the lifespan of this PR, even that seems pretty short.

So don't hesitate to ask directly. We're a tightly knit community, and the situation dictates that it's appropriate and even necessary.

Metadorius:
This PR has to be reopened to facilitate proper process, as we, of course, don't want to stop it from being merged eventually.

PR #1455, opened 2024-12-17
[New Feature] Crusher Level System

Aephiex:
Extracted crusher level related stuff from draft pull request.


NetsuNegi:
is there any effect when crush a building?

Aephiex:
> is there any effect when crush a building?

In an unmodded game, 1x1 buildings can be made crushable by `Crushable=yes`, however they seem to have some innate immunity to `OmniCrusher=yes`, they resist omni crushers without `OmniCrushResistant=yes`. When crushed, a building silently disappears from the map. No explosion, no sound. Buildings are apparently not made to be crushed.

With crusher level system enabled for 1x1 buildings, they can be safely crushed by anything with `CrusherLevel` >= 3, which might not be a desired behavior of a modder, so I'm making that an option. When crusher level system is not turned on for buildings, crush check will fallback to unmodded behavior.

TaranDahl:
> > is there any effect when crush a building?
> 
> In an unmodded game, 1x1 buildings can be made crushable by `Crushable=yes`. Buildings seem to have some innate immunity to `OmniCrusher=yes`, they resist omni crushers without explicit `OmniCrushResistant=yes`. When crushed, a building silently disappears from the map. No explosion, no sound. Buildings are apparently not made to be crushed.
> 
> With crusher level system enabled for 1x1 buildings, they can be "safely" crushed by anything with `CrusherLevel` >= 3, which might not be a desired behavior of a modder, so I'm making that an option. When crusher level system is not turned on for buildings, crush check will fallback to unmodded behavior.

In truth, you must enter the middle cell of the building to trigger a crushing. For the buildings with no middle cell, there is no way to trigger crushing with no other's help. Still, you can place another techno above the building and crush that techno, and the building will be crushed at the same time.

Aephiex:
> In truth, you must enter the middle cell of the building to trigger a crushing. For the buildings with no middle cell, there is no way to trigger crushing with no other's help.

I suppose so. Maybe a 3x3 structure whose middle cell is somehow passible can be crushed as well, such as the allied service depot. But that thing doesn't exist in Mental Omega, so I don't know anything good enough to do the test.

> Still, you can place another techno above the building and crush that techno, and the building will be crushed at the same time.

Yes I watched that video, and it frightened me about my pull request "Crush Expanded", that one of the features is units can fire a death rattle weapon when crushed. What if 3 infantries were crushed at a same time, does it mean only 1 infantry gets to fire the death rattle and the rest are just silently deleted by the game? Then I did a quick test and in fact all 3 get to fire the death rattle.

mevitar:
For buildings, there is no explosion(s) like when they are blown up, crushing simply removes them. Also, buildings that do not have a middle cell (like for example the Refinery or ConYard) cannot be crushed, units will simply go through them.
But it's not that it's impossible to properly crush buildings, as there is already a TS tag (IsTrain=yes) that allows units to crush buildings if they enter any cell occupied by that building. The problem with that tag is that the vehicle will blow up its own War Factory when it leaves it.

Against units, it works mostly fine, except when giving force-crush order. If multiple units at once are given a force-crush order against an immobile vehicle, they will then all stack up on the cell of their target. Not sure if it's something caused by this build, by Phobos or Ares, or a general YR issue.

PR #1456, opened 2024-12-18
[Minor] [Ares Fix] Ares academy fix

Aephiex:
Bug Description:
This is a fix to the Ares bug: Academy feature doesn't apply to the initial payload of vehicles built off a War Factory. Curiously, Academy applies to the initial payloads of vehicles under any other circumstances, even when built off a Naval Shipyard.

Cause:
It is "Unsorted::IKnowWhatImDoing" prevented the "HouseExt::ApplyAcademy" from taking effect.

Fix:
The fix is simple, when Ares is supposed to have initialized the initial payload, and Academy is prevented from taking any effect, temporarily turn off "Unsorted::IKnowWhatImDoing", invoke "HouseExt::ApplyAcademy", then turn on "Unsorted::IKnowWhatImDoing" again.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -408,3 +408,4 @@ This page lists all the individual contributions to the project by their author.
 - **Damfoos** - extensive and thorough testing
 - **Dmitry Volkov** - extensive and thorough testing
 - **Rise of the East community** - extensive playtesting of in-dev features
+- **Aephiex** - fixed Ares academy not working on the initial payloads of vehicles built from a war factory
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -191,6 +191,7 @@ This page describes all ingame logics that are fixed or improved in Phobos witho
 - Fixed an issue introduced by Ares that caused `Grinding=true` building `ActiveAnim` to be incorrectly restored while `SpecialAnim` was playing and the building was sold, erased or destroyed.
 - Fixed Ares' Abductor weapon leaves permanent placement stats when abducting moving vehicles.
 - Suppressed Ares' swizzle warning when parsing `Tags` and `TaskForces` (typically begin with `[Developer fatal]Pointer 00000000 declared change to both`).
+- Fixed Academy *(Ares feature)* not working on the initial payloads *(Ares feature)* of vehicles built from a war factory.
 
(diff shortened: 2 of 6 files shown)


Metadorius (approved):
Looks good. Do docs, maybe split out the abstractType for academy determination into a separate function (not necessary but IMO it is reusable) and it's good to merge.

Aephiex:
@Metadorius Docs added and inline function added.

Metadorius (approved):

chaserli:
Ares fixes should be in another file. 
The coupling with Ares should stay at the minimum level. Namely if no known version of Ares is in use the "fix" should not be invoked.
Also HouseExt should not be fetched like that. Please do not commit what you haven't fully investigated.

Aephiex:
> Ares fixes should be in another file. The coupling with Ares should stay at the minimum level. Namely if no known version of Ares is in use the "fix" should not be invoked. Also HouseExt should not be fetched like that. Please do not commit what you haven't fully investigated.

The fix isn't invoked at all because it checks if ares func exists first.

Regarding how to fetch ares ext, I copied that from user NetsuNegi's work of super weapon sidebar. Needs explain why it shouldn't be fetched like that.

Aephiex:
> Ares fixes should be in another file. The coupling with Ares should stay at the minimum level. Namely if no known version of Ares is in use the "fix" should not be invoked. Also HouseExt should not be fetched like that. Please do not commit what you haven't fully investigated.

Also, I fully understand your decision to force-push the dev branch overwritting my immature code, but please note the fix in the doc. Do not forget the doc update are overwritten by your force-push as well, and to the users it is something done without saying. What do you think?

chaserli:
> Also, I fully understand your decision to force-push the dev branch overwritting my immature code, but please note the fix in the doc. Do not forget the doc update are overwritten by your force-push as well. Without docs, to the users, it is something done without saying. What do you think?

 Will do later

Metadorius:
I do not think force pushing should have been done at all. Just add commits on top. Do mot abuse force pushes without need.

Metadorius:
I have fixed up the doc. Please do not do force pushes like this, it is disrespectful to others.

Metadorius:
> Also HouseExt should not be fetched like that.

@chaserli Please elaborate why.

PR #1463, opened 2024-12-26
[Minor] Update Trajectory Template and Uniform Trajectory Modifications

CrimRecya:
Update the old template of Trajectory=Sample, add some comments. And make some uniform changes for Trajectories: 1. Avoid inertia effect on trajectories fired by aircrafts; 2. Draw laser trails with the actual moving direction of the projectile.

### DIFF
--- a/src/Ext/Bullet/Hooks.cpp
+++ b/src/Ext/Bullet/Hooks.cpp
@@ -455,11 +455,8 @@ DEFINE_HOOK(0x415F25, AircraftClass_Fire_TrajectorySkipInertiaEffect, 0x6)
 
 	GET(BulletClass*, pThis, ESI);
 
-	if (auto const pExt = BulletExt::ExtMap.Find(pThis))
-	{
-		if (pExt->Trajectory)
-			return SkipCheck;
-	}
+	if (BulletExt::ExtMap.Find(pThis)->Trajectory)
+		return SkipCheck;
 
 	return 0;
 }
--- a/src/Ext/Bullet/Hooks.cpp
+++ b/src/Ext/Bullet/Hooks.cpp
@@ -290,6 +290,11 @@ DEFINE_HOOK(0x46902C, BulletClass_Explode_Cluster, 0x6)
 	return SkipGameCode;
 }
 
+constexpr bool CheckTrajectoryCanSnap(const TrajectoryFlag flag)
+{
+	return flag == TrajectoryFlag::Straight;
+}
+
 DEFINE_HOOK(0x467CCA, BulletClass_AI_TargetSnapChecks, 0x6)
 {
 	enum { SkipChecks = 0x467CDE };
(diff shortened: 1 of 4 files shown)


src/Ext/Bullet/Hooks.cpp
```diff
@@ -447,3 +447,19 @@ DEFINE_HOOK(0x44D23C, BuildingClass_Mission_Missile_ArcingFix, 0x7)
...
+
+// Vanilla inertia effect only for bullets with ROT=0
+DEFINE_HOOK(0x415F25, AircraftClass_Fire_TrajectorySkipInertiaEffect, 0x6)
+{
+	enum { SkipCheck = 0x4160BC };
+
+	GET(BulletClass*, pThis, ESI);
+
+	if (auto const pExt = BulletExt::ExtMap.Find(pThis))
+	{
```
Metadorius, line 459: Isn't this sanity check redundant? If the extension is missing - we got way bigger problems than this.

src/Ext/Bullet/Trajectories/SampleTrajectory.h
```diff
@@ -2,51 +2,49 @@
...
 public:
-	// This constructor is for Save & Load
-	SampleTrajectory() : PhobosTrajectory(TrajectoryFlag::Sample)
-		, IsFalling { false }
-	{ }
+	SampleTrajectory(noinit_t) { }
 
-	SampleTrajectory(PhobosTrajectoryType* pType) : PhobosTrajectory(TrajectoryFlag::Sample)
-		, IsFalling { false }
+	SampleTrajectory(SampleTrajectoryType const* trajType) : Type { trajType }
```
Metadorius, line 30: I barely touched trajectories code, so take this with a grain of salt: aren't you supposed to explicitly declare which constructor you are "inheriting" from here as well?

src/Ext/Bullet/Hooks.cpp
CrimRecya: Get it.

src/Ext/Bullet/Trajectories/SampleTrajectory.h
CrimRecya: I don't know much about this question. Maybe you'd better ask Trsdy. I just reset the old Trajectory=Sample (header files are copied from Trajectory=Straight) with the new trajectory template remade by Trsdy, and added my thought in comments.

chaserli: PhobosTrajectory is now a pure virtual interface. There's no constructor.

chaserli (changes requested):
You also have some other common changes in your trajectory prs. Copy them here so that we can process them together.

CrimRecya:
> You also have some other common changes in your trajectory prs. Copy them here so that we can process them together.

I have picked out the most common part. Please check it out.

chaserli (approved):

Coronia (approved):
since this isn't a real function, think it could be merged safely

PR #1464, opened 2024-12-28
[New Feature] Aggressive Stance (auto target buildings)

Aephiex:
Works the same as RA3 Aggressive Stance. When toggled on for a unit, it will target unarmed enemy buildings on its own. You can also make some techno types start as Aggressive Stance.


Aephiex:
It's done! Looks very good in my local build.

src/Misc/Network.cpp
```diff
@@ -0,0 +1,54 @@
...
+{
+	auto target = TargetClass(pTechno);
+	auto eventType = static_cast<EventType>(PhobosNetEvent::Events::ToggleAggressiveStance);
+	EventClass Event = EventClass(pTechno->Owner->ArrayIndex, target.m_ID);
+	Event.Type = eventType;
+	Event.Idle.Whom.m_ID = target.m_ID;
+	Event.Idle.Whom.m_RTTI = target.m_RTTI;
+	EventClass::AddEvent(Event);
+}
+
```
Aephiex, line 43: This is the only constructor I find to instantize EventType without error.

src/Misc/Network.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include <EventClass.h>
+#include <vector>
+
+class PhobosNetEvent
+{
+public:
+	enum class Events : unsigned char
+	{
+		ToggleAggressiveStance = 0x80,
```
Aephiex, line 11: Ares custom events are 0x60 and 0x61, not sure if there are more of it in 0.a. The event code is restricted to the range of unsigned char.

docs/Whats-New.md
```diff
@@ -470,6 +470,7 @@ New:
 - `<Player @ X>` can now be used as owner for pre-placed objects on skirmish and multiplayer maps (by Starkku)
 - Allow customizing charge turret delays per burst on a weapon (by Starkku)
 - Unit `Speed` setting now accepts floating point values (by Starkku)
+- aggressive stance toggle command (auto target buildings) (by Aephiex)
```
TaranDahl, line 473:
```suggestion
- Aggressive stance toggle command (auto target buildings) (by Aephiex)
```

src/Ext/Techno/Body.h
```diff
@@ -59,6 +59,8 @@ class TechnoExt
 		int LastWarpInDelay;                   // Last-warp in delay for this unit, used by HasCarryoverWarpInDelay.
 		bool IsBeingChronoSphered;             // Set to true on units currently being ChronoSphered, does not apply to Ares-ChronoSphere'd buildings or Chrono reinforcements.
 
+		int AggressiveStance;               // Aggressive stance that will auto target buildings; -1 = not defined, 0 = will not, 1 = will
```
TaranDahl, line 62: `bool` is sufficient. No need for `int`.

src/Ext/Techno/Body.cpp
```diff
@@ -517,6 +518,42 @@ void TechnoExt::ExtData::SaveToStream(PhobosStreamWriter& Stm)
...
+bool TechnoExt::ExtData::GetAutoTargetBuildings(TechnoClass* pThis)
+{
+	// if this is a passenger then obey the configuration of the transport
+	if (auto pTransport = pThis->Transporter)
+	{
+		auto pTransportExt = TechnoExt::ExtMap.Find(pTransport);
+		return pTransportExt->GetAutoTargetBuildings(pTransport);
+	}
+
+	if (this->AggressiveStance < 0)
```
TaranDahl, line 530: You can set the flag in the initialization of the techno, according to the flag of the technotype.

src/Ext/Techno/Body.cpp
```diff
@@ -517,6 +518,42 @@ void TechnoExt::ExtData::SaveToStream(PhobosStreamWriter& Stm)
 	this->Serialize(Stm);
 }
 
+bool TechnoExt::ExtData::GetAutoTargetBuildings(TechnoClass* pThis)
```
TaranDahl, line 521: This function name sounds like it should return a `BuildingClass` instead of a `bool`. "CanAutoTargetHarmlessBuildings" might be a better function name.

src/Ext/Techno/Body.cpp
```diff
@@ -517,6 +518,42 @@ void TechnoExt::ExtData::SaveToStream(PhobosStreamWriter& Stm)
...
+	{
+		if (auto pTechnoTypeExt = TechnoTypeExt::ExtMap.Find(pThis->GetTechnoType()))
+			this->AggressiveStance = pTechnoTypeExt->AggressiveStance.Get() ? 1 : 0;
+		else
+			this->AggressiveStance = 0;
+	}
+	return this->AggressiveStance > 0;
+}
+
+void TechnoExt::ExtData::TogggleAutoTargetBuildings(TechnoClass* pThis)
```
TaranDahl, line 540: "ToggleAggressiveStance"

src/Misc/Network.cpp
TaranDahl, line 1: 你的处理应该是有问题的。按照我的理解，gamemd在拷贝事件对象的时候通过事件类型确定事件对象的尺寸，所以你不仅需要在处理事件的地方考虑新事件，还得在拷贝的地方告诉gamemd正确的对象尺寸（否则事件数据那个Union不会被发送）。
Ares 0.A的处理是错误的。我们有一个新增事件的模板，它在`src/Ext/Event`里。
Your handling should be problematic. According to my understanding, gamemd determines the size of the event object by the event type and a hardcoded array when copying the event object. So you not only need to consider new events at the place where events are handled, but also need to tell gamemd the correct object size at the place where it is copied (otherwise the Union of event data will not be sent). 
The handling of Ares 0.A is wrong. We have a template for adding new events. It is in `src/Ext/Event`.

TaranDahl: 如果它实际上能工作的话那倒确实出乎我的意料...不过你还是应该查看一下现有的模板。
If it could actually work, it would indeed be beyond my expectation... However, you should still check the existing templates.

src/Commands/AggressiveStance.cpp
```diff
@@ -0,0 +1,130 @@
...
+	// Otherwise, we should turn them into aggressive stance.
+	bool isAnySelectedUnitArmed = false;
+	bool isAllSelectedUnitAggressiveStance = true;
+	for (const auto& pUnit : ObjectClass::CurrentObjects())
+	{
+		// try to cast to TechnoClass
+		TechnoClass* pTechno = abstract_cast<TechnoClass*>(pUnit);
+
+		// if not a techno or is in berserk or is not controlled by the local player then ignore it
+		if (!pTechno || pTechno->Berzerk || pTechno->Owner != HouseClass::CurrentPlayer)
```
TaranDahl, line 52: In some campaigns, the player can control units that belong to other houses.

src/Misc/Network.cpp
Aephiex: It in fact works. However I will take a look on the commented out event template.

src/Ext/Techno/Body.cpp
```diff
@@ -517,6 +518,41 @@ void TechnoExt::ExtData::SaveToStream(PhobosStreamWriter& Stm)
...
+		return pTransportExt->GetAggressiveStance(pTransport);
+	}
+
+	return this->AggressiveStance;
+}
+
+void TechnoExt::ExtData::ToggleAggressiveStance(TechnoClass* pThis)
+{
+	
+	if (this->GetAggressiveStance(pThis))
```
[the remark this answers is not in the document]
Aephiex: Thank you, changed.

src/Commands/AggressiveStance.cpp
```diff
@@ -0,0 +1,130 @@
...
+	// Otherwise, we should turn them into aggressive stance.
+	bool isAnySelectedUnitArmed = false;
+	bool isAllSelectedUnitAggressiveStance = true;
+	for (const auto& pUnit : ObjectClass::CurrentObjects())
+	{
+		// try to cast to TechnoClass
+		TechnoClass* pTechno = abstract_cast<TechnoClass*>(pUnit);
+
+		// if not a techno or is in berserk or is not controlled by the local player then ignore it
+		if (!pTechno || pTechno->Berzerk || !pTechno->Owner->IsInPlayerControl)
```
CrimRecya, line 52: `if (!pTechno || pTechno->Berzerk || !pTechno->Owner->IsControlledByCurrentPlayer())`

src/Commands/AggressiveStance.cpp
```diff
@@ -0,0 +1,130 @@
...
+		{
+			EventExt::RaiseToggleAggressiveStance(pTechno);
+
+			auto pTechnoType = pTechno->GetTechnoType();
+			auto pTechnoTypeExt = TechnoTypeExt::ExtMap.Find(pTechnoType);
+			int voiceIndex;
+
+			if (isAllSelectedUnitAggressiveStance)
+			{
+				voiceIndex = pTechnoTypeExt->VoiceExitAggressiveStance.Get(-1);
```
CrimRecya, line 104: `voiceIndex = pTechnoTypeExt->VoiceExitAggressiveStance.Get();` with `ValueableIdx<VocClass> VoiceExitAggressiveStance` and `VoiceExitAggressiveStance { -1 }`, the same below.

src/Commands/AggressiveStance.cpp
```diff
@@ -0,0 +1,130 @@
...
+					if (voiceList.Count)
+					{
+						unsigned int idxRandom = Randomizer::Global().Random();
+						auto index = idxRandom % voiceList.Count;
+						voiceIndex = voiceList.GetItem(index);
+					}
+				}
+			}
+
+			if (voiceIndex > 0)
```
CrimRecya, line 121: `QueueVoice(voiceIndex)` has checked for value `-1`.

src/Ext/Event/Body.cpp
```diff
@@ -13,18 +14,40 @@ void EventExt::RespondEvent()
...
+	TargetClass* pTarget = reinterpret_cast<TargetClass*>(this->ToggleAggressiveStance.DataBuffer);
+	if (auto pTechno = pTarget->As_Techno())
+	{
+		if (auto pTechnoExt = TechnoExt::ExtMap.Find(pTechno))
+		{
+			pTechnoExt->ToggleAggressiveStance();
+		}
+	}
+}
+
```
CrimRecya, line 44:
```
void EventExt::RaiseToggleAggressiveStance(TechnoClass* pTechno)
{
	EventExt eventExt {};
	eventExt.Type = EventTypeExt::ToggleAggressiveStance;
	eventExt.HouseIndex = static_cast<char>(pTechno->Owner->ArrayIndex);
	eventExt.Frame = Unsorted::CurrentFrame;
	eventExt.ToggleAggressiveStance.Who = TargetClass(pTechno);
	eventExt.AddEvent();
}

void EventExt::RespondToToggleAggressiveStance()
{
	if (const auto pTechnoExt = TechnoExt::ExtMap.Find(this->ToggleAggressiveStance.Who.As_Techno()))
		pTechnoExt->ToggleAggressiveStance();
}
```

src/Ext/Event/Body.h
```diff
@@ -27,20 +28,23 @@ class EventExt
 	{
 		char DataBuffer[104];
 
-		struct Sample
+		struct ToggleAggressiveStance
 		{
-			char DataBuffer[104];
-		} Sample;
+			char DataBuffer[5];
```
CrimRecya, line 33: `TargetClass Who;` is OK. And you may need `#include <TargetClass.h>`.

src/Ext/Techno/Hooks.TargetEvaluation.cpp
```diff
@@ -299,3 +299,21 @@ Action __fastcall InfantryClass__WhatAction_Wrapper(InfantryClass* pThis, void*
 DEFINE_JUMP(VTABLE, 0x7EB0CC, GET_OFFSET(InfantryClass__WhatAction_Wrapper))
 
 #pragma endregion
+
+DEFINE_HOOK(0x6F858F, TechnoClass_EvaluateObject_AggressiveStance, 0x7)
+{
+	GET(TechnoClass*, pThis, EDI);
+	GET(TechnoClass*, pTarget, ESI);
+	if (pThis && pThis->Owner->IsHumanPlayer
```
CrimRecya, line 307: `pThis->Owner->IsHumanPlayer` -> `pThis->Owner->IsControlledByHuman()`

src/Ext/TechnoType/Body.h
```diff
@@ -132,6 +132,11 @@ class TechnoTypeExt
 		Valueable<bool> AutoFire;
 		Valueable<bool> AutoFire_TargetSelf;
 
+		Valueable<bool> AggressiveStance;
+		Nullable<bool> AggressiveStance_Togglable;
+		NullableIdx<VocClass> VoiceEnterAggressiveStance;
```
CrimRecya, line 137: `ValueableIdx<VocClass> VoiceEnterAggressiveStance;`, the same below.

src/Ext/TechnoType/Body.h
```diff
@@ -322,6 +327,12 @@ class TechnoTypeExt
 
 			, AutoFire { false }
 			, AutoFire_TargetSelf { false }
+
+			, AggressiveStance { false }
+			, AggressiveStance_Togglable {}
+			, VoiceEnterAggressiveStance {}
```
CrimRecya, line 333: `, VoiceEnterAggressiveStance { -1 }`, the same below.

CrimRecya (commented):
I made some simplification and solved a potential problem.

Aephiex:
> I made some simplification and solved a potential problem.

Thank you for your help!

CrimRecya (approved):

TaranDahl (approved):

src/Ext/Techno/Body.cpp
```diff
@@ -552,6 +552,34 @@ void TechnoExt::ExtData::ToggleAggressiveStance()
 	}
 }
 
+bool TechnoExt::ExtData::CanToggleAggressiveStance()
+{
+	if (!this->TypeExtData->AggressiveStance_Togglable.isset())
+	{
+		// Only techno that are armed and open-topped can be aggressive stance.
+		if (!(this->OwnerObject()->IsArmed() || this->OwnerObject()->GetTechnoType()->OpenTopped))
```
CrimRecya, line 560: You can use `this->TypeExtData->OwnerObject()->OpenTopped` instead of `this->OwnerObject()->GetTechnoType()->OpenTopped`.

src/Ext/Techno/Body.cpp
```diff
@@ -552,6 +552,34 @@ void TechnoExt::ExtData::ToggleAggressiveStance()
...
+		{
+			this->TypeExtData->AggressiveStance_Togglable = false;
+			return false;
+		}
+
+		// Engineers and Agents are default to not allow aggressive stance.
+		if (auto pInfantryTypeClass = abstract_cast<InfantryTypeClass*>(this->OwnerObject()->GetTechnoType()))
```
CrimRecya, line 567: The same. I think `GetTechnoType()` is a bit more complex than `TypeExtData`.

Speederovsky:
Where and what is the tag to unlock this behavior in the first place? This hotkey shouldn't be available by default.

Aephiex:
> Where and what is the tag to unlock this behavior in the first place? This hotkey shouldn't be available by default.

If you don't want to use it, then you don't config the hotkey. As long as you don't config the hotkey it has no effect to your gameplay. There is no need to prepare a way to disable it other than to not config the hotkey.

Speederovsky:
It's not a matter of configuring a hotkey or not - you are making Phobos add a new type of unit behavior to the base game without an enabler, and this should not be the case.

Also why is this in 'Selection' and not 'Control'?

Aephiex:
> It's not a matter of configuring a hotkey or not - you are making Phobos add a new type of unit behavior to the base game without an enabler, and this should not be the case.

The enabler is to config a hotkey. You don't config a hotkey and it has no effect.

> Also, why is this in 'Selection' and not 'Control'?

Fair enough.

> Also, toggling Aggressive Stance off on a unit doesn't make them stop attacking a building.

Apparently you didn't try. When a unit attacking building is turned off aggressive stance it is automatically ordered to disengage and cease fire.

Starkku:
> The enabler is to config a hotkey. You don't config a hotkey and it has no effect.

The precedent to contrary is already here. If you have not enabled debug features (`[GlobalControls]` -> `DebugKeysEnabled`) Phobos does **not** initialize the debug keyboard commands. Ares does it like you say, we should not.

Speederovsky:
Configuring an always available hotkey is NOT EQUAL to enabling this feature in the game/mod. Right now any user can configure this hotkey and enable this behavior, which should not even be available by default in Red Alert 2 with Phobos, as on its own it should ONLY FIX BUGS.

I also did try, and the units I tested on did NOT stop attacking a building they began attacking after switching to this stance.

Aephiex:
> The precedent to contrary is already here. If you have not enabled debug features (`[GlobalControls]` -> `DebugKeysEnabled`) Phobos does **not** initialize the debug keyboard commands. Ares does it like you say, Phobos does not.

Aggressive stance is not a debug hotkey.

Aephiex:
> Configuring an always available hotkey is NOT EQUAL to enabling this feature in the game/mod. Right now any user can configure this hotkey and enable this behavior, which should not even be available by default in Red Alert 2 with Phobos, as on its own it should ONLY FIX BUGS.
Anyone asks you to disable aggressive stance hotkey on the mod's basis? Or you personally don't want your players to be able to use aggressive stance? If that's the case, add `AggressiveStance.Togglable=false` to every techno in the game. Don't forget garrisonable structures too, as they can be toggled as well. I don't get it why you hate it so much and what are you asking to be done, are you asking for the PR to be closed and never merged to Phobos?

> I also did try, and the units I tested on did NOT stop attacking a building they began attacking after switching to this stance.
I see, the nightly build is bit of outdated, it is generated first day this PR is opened, and the said behavior is only added later than that. Maybe grab the source code and build a DLL from this branch and test again?

Starkku:
> I see, the nightly build is bit of outdated, it is generated first day this PR is opened

For first time yes, the bot does update it if everything is going right though.

It has been updated today and timestamps on the files match that.

image

Aephiex:
Oops, seems the `this->OwnerObject()->SetTarget(nullptr);` can't be omitted. I'm sorry for this, I replaced it with a guard mission queue without testing.

TaranDahl:
Perhaps we need a `QoLKeysEnabled=boolean`.

Aephiex:
Now has the option to make some building types exempt of aggressive stance. Even though a building type may be exempt, it will still be passively attacked if it was a defensive structure or an occupied civilian structure. `AggressiveStance.Exempt` only cancels the effect of aggressive stance, will not make the building itself exempt of the passive target acquisition. Hopefully this can help some players to not accedently destroy some mission objects that must be infiltrated or captured.

Aephiex:
[https://www.nexusmods.com/commandandconquerredalert2/mods/143](https://www.nexusmods.com/commandandconquerredalert2/mods/143)
This is now available as an independant engine extension on moddb.

PR #1465, opened 2024-12-28
[Vanilla Improvement] No turret unit turn to the target when rearming

CrimRecya:
- Now vehicles (exclude jumpjets) without turret will attempt to turn to the target while the weapon is cooling down, rather than after the weapon has cooled down, by setting to `NoTurret.TrackTarget` true.

In `rulesmd.ini`:
```ini
[General]
NoTurret.TrackTarget=false   ; boolean

[SOMEUNIT]                   ; UnitType
NoTurret.TrackTarget=        ; boolean, defaults to [General]->NoTurret.TrackTarget
```


TaranDahl:
Tested locally.

ZivDero:
> In view of some recent events, I have to ask a question here: Does anyone feel that this should not be used as a vanilla fix and need to add a switch?

IMO this is impactful enough for it to require a switch.

Coronia (approved):
despite the code seems fine, it'd better to have a shorter tag name. Couldn't think of one so far, would like some suggestion

Coronia:
come up with a tag naming idea. Despite it's still long, it should be more extendible and allows customization for individual units:
```ini
[General]
NoTurretEarlyTurnToTarget=false

[SOMEUNIT]
NoTurret.EarlyTurnToTarget=(default to NoTurretEarlyTurnToTarget)
```
in this case we can somehow make it clearer by making a `NoTurret` prefix. If there're other functions for units without turret in the future, it can also use NoTurret.XXX tags

TaranDahl:
> allows customization for individual units:

Is it really necessary to set this per techno? I can't imagine any case other than the stupid "original flavor"(《原汁原味》) that would require turning this off.

Coronia:
> > allows customization for individual units:
> 
> Is it really necessary to set this per techno? I can't imagine any case other than the stupid "original flavor"(《原汁原味》) that would require turning this off.

1 - it's mostly for setting up a tag series that start with `NoTurret` in the future, so it's not just for customization, but also for a universal standard that could make users of dll easier to understand
2 - pretty sure there're cases when players don't want their units to turn frequently. E.g. if a unit has low ROT, and trying to auto engage enemies when players want it to do something else

PR #1466, opened 2024-12-28
[Simple Fix] Draw visual effects for airburst weapons

CrimRecya:
- `AirburstWeapon` now supports `IsLaser` (with Ares `LaserThickness`), `IsElectricBolt` (with Phobos `Bolt.Disable1`, `Bolt.Disable2`, `Bolt.Disable3`, `Bolt.Arcs`, without Ares `Bolt.Color1` 、`Bolt.Color2` 、`Bolt.Color3`), `IsRadBeam` (with Ares `Beam.Color` 、`Beam.Duration` 、`Beam.Amplitude` 、`Beam.IsHouseColor`), and `AttachedParticleSystem`.


Coronia (changes requested):
again I would like some comments about how the position in `SimulatedFiringAnim` is calculated. If someone else reading the codes, they might be confused by the algorithm or magic number in it

Coronia (commented):
the current code only used `SimulatedFiringEffects` for once with `firingEffect` set to false, which means that `SimulatedFiringAnim` and `SimulatedFiringReport` are actually omitted. Is this the intended way? If so, then where should these functions be used?

CrimRecya:
> the current code only used `SimulatedFiringEffects` for once with `firingEffect` set to false, which means that `SimulatedFiringAnim` and `SimulatedFiringReport` are actually omitted. Is this the intended way? If so, then where should these functions be used?

`Disperse` and `Tracing` trajectory.

Coronia (approved):
tested with laser, bolt and rad beam, they can all be shown/dealt damage/applied effects properly

PR #1467, opened 2024-12-28
[Highly Customized] Burst without delay

CrimRecya:
- In vanilla, vehicles and infantries will only fire once in one frame, even if their `ROF` or `BurstDelay` is set to 0. Now you can force units to fire all bursts in one frame by setting the `Burst.NoDelay` to true (useless for buildings yet).

In `rulesmd.ini`:
```ini
[SOMEWEAPON]          ; WeaponType
Burst.NoDelay=false   ; boolean
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -372,6 +372,7 @@ This page lists all the individual contributions to the project by their author.
   - Allow to change the speed of gas particles
 - **CrimRecya**
   - Fix `LimboKill` not working reliably
+  - Burst without delay
 - **Ollerus**
   - Build limit group enhancement
   - Customizable rocker amplitude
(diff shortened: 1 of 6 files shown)


TaranDahl:
Tested locally.

Starkku:
Does this play nice with `DelayedFire`?

CrimRecya:
> Does this play nice with `DelayedFire`?

`IsAnimDelayedFire` ? No, it's exclusive to buildings. @TaranDahl has done another implement that allows buildings to work like this with `IsAnimDelayedFire`, which is not included in this.

Starkku:
I was thinking of #1338 but forgot it was not merged yet

CrimRecya:
> I was thinking of #1338 but forgot it was not merged yet

I looked at it briefly and felt it might not fit perfectly. I used a simple loop to call the firing function here, and `DelayedFire` seems to need to complete each "charge" check in the firing function, so it may only fire once in the end.

PR #1468, opened 2024-12-28
[Customized] Undeploy building on impact

CrimRecya:
- Warheads with `BuildingUndeploy` can now undeploy buildings with `UndeploysInto`.
  - `BuildingUndeploy.Leave` controls whether need to let them move to low threat locations nearby. The threat degree here is calculated using the technos' cost. If a nearby techno has no primary weapons or is owned by your allies, it will not be included.

In `rulesmd.ini`:
```ini
[SOMEWARHEAD]                  ; Warhead
BuildingUndeploy=false         ; boolean
BuildingUndeploy.Leave=false   ; boolean
```

### DIFF
--- a/src/Ext/WarheadType/Detonate.cpp
+++ b/src/Ext/WarheadType/Detonate.cpp
@@ -170,6 +170,9 @@ void WarheadTypeExt::ExtData::DetonateOnOneUnit(HouseClass* pHouse, TechnoClass*
 	if (this->AttachEffects.AttachTypes.size() > 0 || this->AttachEffects.RemoveTypes.size() > 0 || this->AttachEffects.RemoveGroups.size() > 0)
 		this->ApplyAttachEffects(pTarget, pHouse, pOwner);
 
+	if (this->BuildingUndeploy)
+		this->ApplyBuildingUndeploy(pTarget);
+
 #ifdef LOCO_TEST_WARHEADS
 	if (this->InflictLocomotor)
 		this->ApplyLocomotorInfliction(pTarget);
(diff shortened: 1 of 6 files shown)


Coronia:
For more generic usage I would suggest to make it also support selling a building when impact. Shouldn't take much effort to edit with the base of current implementation.

CrimRecya:
> For more generic usage I would suggest to make it also support selling a building when impact. Shouldn't take much effort to edit with the base of current implementation.

Thank you for your suggestions. I will update it later.

src/Ext/WarheadType/Detonate.cpp
```diff
@@ -180,6 +183,86 @@ void WarheadTypeExt::ExtData::DetonateOnOneUnit(HouseClass* pHouse, TechnoClass*
...
+			}
+		}
+
+		if (!costs)
+			dir = ScenarioClass::Instance->Random.RandomRanged(0, 15);
+
+		const double radian = -(((dir - 4) / 16.0) * Math::TwoPi);
+
+		cell.X -= static_cast<short>(14 * cos(radian));
+		cell.Y += static_cast<short>(14 * sin(radian));
```
Coronia, line 251: again, some comments about the algorithm and magic number are recommended, for others to understand how the target cell is calculated here

Coronia (commented):
just curious, the only difference of sell and undeploy here is that the latter one has a SetArchiveTarget action. Is it really sufficient to make the building undeploy instead of being sold?

will run some test in the Tester Group about it later too

Coronia (approved):
tested as well, works properly

PR #1469, opened 2024-12-28
[Customized] Extended gattling rate down logic

CrimRecya:
- Now you can customize some effects of RateDown.
  - `RateDown.Delay` controls the delay before using `RateDown` to reduce the gattling value.
  - `RateDown.Reset` controls whether to reset the gattling value directly when the techno has no target or changes targets.
  - `RateDown.Cover.Value` replaces the original `RateDown` when techno's ammo is lower than `RateDown.Cover.AmmoBelow`.

In `rulesmd.ini`:
```ini
[SOMETECHNO]                  ; TechnoType, `IsGattling=yes`
RateDown.Delay=0              ; integer, game frames
RateDown.Reset=false          ; boolean
RateDown.Cover.Value=0        ; integer
RateDown.Cover.AmmoBelow=-2   ; integer
```


Speederovsky:
This feature sounds useful, because RateDown.Reset would help fix certain issues with MO Crycopters, ie. maintaining deep freeze effect (high Gattling Stage) upon switching targets without moving the Cryocopter. It seems RateDown.Reset=yes mostly works, but there is a chance for a non-moving Cryocopter to maintain its high Gattling Stage effects for a two-three seconds after changing its target before it gets resets properly. Not sure what this is related to, it mostly resets the Rate properly.

TaranDahl:
> This feature sounds useful, because RateDown.Reset would help fix certain issues with MO Crycopters, ie. maintaining deep freeze effect (high Gattling Stage) upon switching targets without moving the Cryocopter. It seems RateDown.Reset=yes mostly works, but there is a chance for a non-moving Cryocopter to maintain its high Gattling Stage effects for a two-three seconds after changing its target before it gets resets properly. Not sure what this is related to, it mostly resets the Rate properly.

看上去钩子被设置在`Update`的子函数`UpdateFiring`上，而尝试开火的操作在`UpdateFiring`之前被执行，这导致单位有可能在设置新目标后的一帧仍然以重置之前的加特林阶段开火。
It seems that the hook is set on the sub-function `UpdateFiring` of `Update`, and the operation of attempting to fire is executed before `UpdateFiring`. This causes the unit to possibly still fire in the Gatling stage before resetting in the frame after setting a new target.

CrimRecya:
> This feature sounds useful, because RateDown.Reset would help fix certain issues with MO Crycopters, ie. maintaining deep freeze effect (high Gattling Stage) upon switching targets without moving the Cryocopter. It seems RateDown.Reset=yes mostly works, but there is a chance for a non-moving Cryocopter to maintain its high Gattling Stage effects for a two-three seconds after changing its target before it gets resets properly. Not sure what this is related to, it mostly resets the Rate properly.

Thanks for the feedback, I will fix it later.

Coronia (approved):
tag naming suggestion:
RateDown.Ammo -> RateDown.Cover.AmmoBelow
RateDown.Cover -> RateDown.Cover.Value

this makes it clearer and inline with other condition tags, while making room for more RateDown cover conditions in the future

PR #1472, opened 2024-12-28
[Customized] Avoid weapon fire overflow in the same frame

CrimRecya:
- Now you can make the weapon declare that the target has been locked at the moment of firing, and all weapons with positive `NoRepeatFire` value will not attack the declared target again during this period of time. Made for seckill and control weapons.

In `rulesmd.ini`:
```ini
[SOMEWEAPON]       ; WeaponType
NoRepeatFire=0     ; integer, game frames
```


CrimRecya:
If the techno can attach effect to the target at the moment of firing, it can achieve a similar effect. But I am not sure whether it needs to be expanded. It seems that there are not many places suitable for this.

Aephiex:
> Now you can make the weapon declare that the target has been locked at the moment of firing, and all weapons **with positive NoRepeatFire value** will not attack the declared target again during this period of time. Made for seckill and control weapons.

Maybe change that to only the weapons in a specific group?

docs/New-or-Enhanced-Logics.md
```diff
@@ -1748,3 +1748,13 @@ In `rulesmd.ini`:
 CanTarget=all        ; list of Affected Target Enumeration (none|land|water|empty|infantry|units|buildings|all)
 CanTargetHouses=all  ; list of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
 ```
+
+### Avoid weapon fire overflow in the same frame
+
+- Now you can make the weapon declare that the target has been locked at the moment of firing, and all weapons with positive `NoRepeatFire` value will not attack the declared target again during this period of time. Made for seckill and control weapons.
```
Coronia, line 1754: Rename the tag to something like `PreventTargetDuration` for better clarification
- what it can stop is the targeting process, but not the real damage
- it's supposed to be an integer tag for frame amount, instead of a boolean one

Coronia (changes requested):

Coronia:
> > Now you can make the weapon declare that the target has been locked at the moment of firing, and all weapons **with positive NoRepeatFire value** will not attack the declared target again during this period of time. Made for seckill and control weapons.
> 
> Maybe change that to only the weapons in a specific group?

I think we can already use new AEs for that. This avoid fire logic should remain simple as it is imo

CrimRecya:
#1482 can complete this implement in a more customized way, so I decided to turn this off.

PR #1473, opened 2024-12-28
[Customized] Aggressive attack move mission

CrimRecya:
- `AttackMove.Aggressive` allows your technos to attack the enemy's unarmed buildings more aggressively when in attack move mission (Ctrl+Shift). Default to `[General]->AttackMove.Aggressive`.
- `AttackMove.UpdateTarget` allows your technos to automatically change and select a higher threat target when in attack move mission (Ctrl+Shift). Also default to `[General]->AttackMove.UpdateTarget`.

In `rulesmd.ini`:
```ini
[General]
AttackMove.Aggressive=false    ; boolean
AttackMove.UpdateTarget=false  ; boolean

[SOMETECHNO]
AttackMove.Aggressive=         ; boolean
AttackMove.UpdateTarget=       ; boolean
```


TaranDahl:
Tested locally.

Coronia (approved):

src/Ext/Techno/Hooks.TargetEvaluation.cpp
```diff
@@ -140,6 +140,75 @@ DEFINE_JUMP(CALL6, 0x6F8DD2, GET_OFFSET(TechnoClass_EvaluateCellGetWeaponRangeWr
...
+
+	const auto pPrimary = pThis->GetWeapon(0);
+
+	if (!pPrimary)
+		return false;
+
+	const auto pPrimaryWeapon = pPrimary->WeaponType;
+
+	if (!pPrimaryWeapon)
+		return false;
```
Starkku, line 168: This kind of sanity checking is and unnecessary, `GetWeapon()` never returns `nullptr`. You can simply do something like

```cpp
const auto pPrimaryWeapon = pThis->GetWeapon(0)->WeaponType;

if (!pPrimaryWeapon)
		return false;
```

instead and it will be fine.

Starkku (changes requested):
A little bit of feedback about weapon-related sanity checks.

Coronia:
should add an option on building to prevent it being the target of aggressive attack move

also it would be better to split the tag in [General] just like techno, or it would be weird and inconvenient to have both 2 tags default to the same one

PR #1475, opened 2024-12-28
[Customized] Fast access vehicle

CrimRecya:
- Now you can let passengers quickly enter or leave the transport vehicles (useless for buildings) without queuing. Defaults to `[General]` -> `NoQueueUpToEnter` or `NoQueueUpToUnload`.

In `rulesmd.ini`:
```ini
[General]
NoQueueUpToEnter=false    ; boolean
NoQueueUpToUnload=false   ; boolean

[SOMEUNIT]                ; UnitType
NoQueueUpToEnter=         ; boolean
NoQueueUpToUnload=        ; boolean
```


mevitar:
This happened when i tried to load PoweredBy= units into a NoQueueUpToEnter=yes and NoQueueUpToUnload=yes transport, and then i unpowered them mid load. They stayed unpowered this entire time.

fastloadbug

Will still test if it happens for regular load.

mevitar:
> I'll still recheck if it happens with regular transports.

Yes, it indeed happens with regular transports, and it's not an issue introduced by Phobos.

src/Ext/Techno/Hooks.Transport.cpp
```diff
@@ -220,3 +221,227 @@ DEFINE_HOOK(0x710552, TechnoClass_SetOpenTransportCargoTarget_ShareTarget, 0x6)
...
+				if (CanEnterNow(pDest, pThis)) // Replace send radio command: QueryCanEnter
+				{
+					if (const auto pTag = pDest->AttachedTag)
+						pTag->RaiseEvent(TriggerEvent::EnteredBy, pThis, CellStruct::Empty);
+
+					pThis->ArchiveTarget = nullptr;
+					pThis->OnBridge = false;
+					pThis->MissionAccumulateTime = 0;
+					pThis->GattlingValue = 0;
+					pThis->CurrentGattlingStage = 0;
```
Coronia, line 300: guess the gatling part is not needed unless we're making infantry support it
also I'm not sure if more variable reset is needed here, if we can look into how vanilla was handling it then it would be better

src/Ext/Techno/Hooks.Transport.cpp
```diff
@@ -220,3 +221,227 @@ DEFINE_HOOK(0x710552, TechnoClass_SetOpenTransportCargoTarget_ShareTarget, 0x6)
...
+				{
+					// I don't know why units have no trigger
+
+					pThis->ArchiveTarget = nullptr;
+					pThis->OnBridge = false;
+					pThis->MissionAccumulateTime = 0;
+					pThis->GattlingValue = 0;
+					pThis->CurrentGattlingStage = 0;
+
+					if (const auto pMind = pThis->MindControlledBy)
```
Coronia, line 361: the same 2 things as above

src/Ext/Techno/Hooks.Transport.cpp
```diff
@@ -220,3 +221,227 @@ DEFINE_HOOK(0x710552, TechnoClass_SetOpenTransportCargoTarget_ShareTarget, 0x6)
...
+
+DEFINE_HOOK(0x73DC1E, UnitClass_Mission_Unload_NoQueueUpToUnloadLoop, 0xA)
+{
+	enum { UnloadLoop = 0x73D8CB, UnloadReturn = 0x73E289 };
+
+	GET(UnitClass* const, pThis, ESI);
+
+	if (TechnoTypeExt::ExtMap.Find(pThis->Type)->NoQueueUpToUnload.Get(RulesExt::Global()->NoQueueUpToUnload))
+	{
+		if (pThis->Passengers.NumPassengers <= pThis->NonPassengerCount)
```
Coronia, line 431: would it be more accurate to just use size for this check? NonPassengerCount's comment mentioned it has interaction with Gunner but I'm not sure what it is

src/Ext/Techno/Hooks.Transport.cpp
```diff
@@ -220,3 +221,227 @@ DEFINE_HOOK(0x710552, TechnoClass_SetOpenTransportCargoTarget_ShareTarget, 0x6)
...
+					if (const auto pTag = pDest->AttachedTag)
+						pTag->RaiseEvent(TriggerEvent::EnteredBy, pThis, CellStruct::Empty);
+
+					pThis->ArchiveTarget = nullptr;
+					pThis->OnBridge = false;
+					pThis->MissionAccumulateTime = 0;
+					pThis->GattlingValue = 0;
+					pThis->CurrentGattlingStage = 0;
+
+					if (const auto pMind = pThis->MindControlledBy)
```
Coronia, line 302: mind control is already filtered in CanEnterNow it seems?

docs/New-or-Enhanced-Logics.md
```diff
@@ -1679,6 +1679,21 @@ WarpInWeapon.UseDistanceAsDamage=false  ; boolean
 WarpOutWeapon=                          ; WeaponType
 ```
 
+### Fast access vehicle
+
+- Now you can let passengers quickly enter or leave the transport vehicles (useless for buildings) without queuing. Defaults to `[General]` -> `NoQueueUpToEnter` or `NoQueueUpToUnload`.
```
Coronia, line 1684: technically aircraft could also have passenger, should we support it as well? If not, then it should be mentioned in the doc

Coronia (changes requested):
worked properly in test, the issue listed below are just some minor stuffs

also, maybe we don't have to follow the QueueUpToEnter variable name, but name this feature as `QuickEnter` and `QuickUnload` instead

src/Ext/Techno/Hooks.Transport.cpp
```diff
@@ -220,3 +221,227 @@ DEFINE_HOOK(0x710552, TechnoClass_SetOpenTransportCargoTarget_ShareTarget, 0x6)
 
 	return 0;
 }
+
+#pragma region NoQueueUpToEnterAndUnload
+
+bool __fastcall CanEnterNow(UnitClass* pTransport, FootClass* pPassenger)
+{
+	if (!pTransport->Owner->IsAlliedWith(pPassenger) || pTransport->IsBeingWarpedOut())
```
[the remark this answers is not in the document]
CrimRecya: Rewrite from 7375A6.

src/Ext/Techno/Hooks.Transport.cpp
```diff
@@ -220,3 +221,227 @@ DEFINE_HOOK(0x710552, TechnoClass_SetOpenTransportCargoTarget_ShareTarget, 0x6)
...
+		if (abs(delta.X) <= 384 && abs(delta.Y) <= 384)
+			return (predictSize <= (maxSize - (bySize ? static_cast<int>(pLink->GetTechnoType()->Size) : 1)));
+	}
+
+	const auto remain = maxSize - predictSize;
+
+	if (remain < 0)
+		return false;
+
+	if (needCalculate && remain < (bySize ? static_cast<int>(pLink->GetTechnoType()->Size) : 1))
```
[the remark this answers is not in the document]
CrimRecya: The passenger who is asking if they can enter may not necessarily be the link techno.

src/Ext/Techno/Hooks.Transport.cpp
CrimRecya: Rewrite from 51A2AD.

src/Ext/Techno/Hooks.Transport.cpp
CrimRecya: Rewrite from 73A6F1.

src/Ext/Techno/Hooks.Transport.cpp
CrimRecya: If your unit is a `Gunner`, it will stop before unloading to this passenger, which is the vanilla logic. See 73D8B7.

src/Ext/Techno/Hooks.Transport.cpp
CrimRecya: That's true, but this is how it was done in vanilla. See 51A2C3 and 73A714.

docs/New-or-Enhanced-Logics.md
CrimRecya: Passengers on airplanes are completely different from those on vehicles or even buildings.
Their boarding and unloading logic is completely different.

CrimRecya:
> worked properly in test, the issue listed below are just some minor stuffs
> 
> also, maybe we don't have to follow the QueueUpToEnter variable name, but name this feature as `QuickEnter` and `QuickUnload` instead

Wait a moment, there is currently an issue: when the vehicle to be loaded is surrounded by the transport vehicle within a circle, it cannot be loaded quickly because non moving units will not call `UpdatePosition()`.
Also, regarding the key names, I would like to emphasize that there is no need to wait in line when taking transportation vehicles. Simple key names may not be clear enough, so I would like to hear other people's opinions.

Coronia (approved):

PR #1384, opened 2024-09-18
Exclusive SuperWeapon Sidebar

NetsuNegi:
- It is possible to put sw cameos on the left of screen like C&C3 when `SuperWeaponSidebar` is true. Cameos arranged in a pyramid shape. In theory, it should be compatible with Ares.
- 当 `SuperWeaponSidebar` 为true时，可以把超级武器的图标排列在屏幕左侧了，就像命令与征服3中那样。图标按照金字塔形状排列。理论上它应该是兼容Ares的。
  - `SuperWeaponSidebar.Interval` controls the distance between two column cameos (excluding the background). When you need to make a background, the width of the background should be (`SuperWeaponSidebar.Interval` + cameo fixed width 60).
  - `SuperWeaponSidebar.Interval` 控制两列图标之间的横向间距（图标背景不包含在内）。如果你需要制作图标的背景，那么背景的宽度应该是 `SuperWeaponSidebar.Interval` + 图标的宽度（60像素）。
  - `SuperWeaponSidebar.LeftOffset` controls the distance between the left side of cameo and the left side of its column (background). This will not be greater than `SuperWeaponSidebar.Interval`.
  - `SuperWeaponSidebar.LeftOffset` 控制图标距离背景左边缘的距离，这将不会超过 `SuperWeaponSidebar.Interval` 的值。
  - `SuperWeaponSidebar.CameoHeight` controls the distance from the top of the previous cameo to the top of the next cameo. That is, the space between the upper and lower cameos is (`SuperWeaponSidebar.CameoHeight` - cameo fixed height 48). This will not be less than 48. When you need to make a background, this is the height of the background.
  - `SuperWeaponSidebar.CameoHeight` 控制图标顶部到下一个图标顶部的距离（也就是每两个背景之间的纵向间距），图标的顶部到其背景的上边缘的距离即为 `SuperWeaponSidebar.CameoHeight` - 图标高度（48）像素。如果你需要制作图标的背景，那么背景的高度应该是 `SuperWeaponSidebar.CameoHeight` 的值。
  - `SuperWeaponSidebar.Max` controls the maximum number of cameos on the leftmost column, which also depends on the current game resolution.
  - `SuperWeaponSidebar.Max` 控制最左侧一列最多能有多少个图标，这也会受到当前游戏分辨率的影响。
  - `SuperWeaponSidebar.MaxColumns` controls that maximum count of columns.
  - `SuperWeaponSidebar.MaxColumns` 控制最多能显示多少列。
[...]

PR #1477, opened 2024-12-28
[Highly Customized] Grey cameo preview and cameo overlays

CrimRecya:
- A setting that allows you to preview information. This feature can be used as before, playing "new construction options" and clearing the specific production queue when prerequisites loss.
  - `Cameo.AlwaysExist` controls whether you can see the cameo when the prerequisite have not satisfied (`TechnoLevel`, `Owner`, `RequiredHouses` and `ForbiddenHouses` should be satisfied). Defaults to `[AudioVisual]` -> `Cameo.AlwaysExist`.
  - `ShowBuildingStatistics` controls whether the number of buildings of this type that you currently own needs to be displayed in the upper right corner of the building cameo (requires the cameo exist).
  - `Cameo.OverlayShapes` controls the drawn image file.
  - `Cameo.OverlayFrames` controls which frame in `Cameo.OverlayShapes` to draw in three different situations: currently owned this building type, grey cameo and have its prerequisite, grey cameo but have no prerequisite (The last situation requires `Cameo.AlwaysExist` to be true). When set to a negative number, it means that there is no need to draw under the corresponding conditions.
  - `Cameo.OverlayPalette` the color palette used when drawing `Cameo.OverlayShapes`.
  - If `Cameo.AuxTechnos` is not set, in addition to basic conditions, the grey cameo will only show when `AIBasePlanningSide` condition is satisfied. Otherwise, the grey cameo will only show when at least one of these types is owned by you or its `TechnoLevel`, `Owner`, `RequiredHouses`, `ForbiddenHouses`, `Cameo.AuxTechnos` (use `AIBasePlanningSide` if not set) and `Cameo.NegTechnos` (if set) conditions are satisfied.
  - If `Cameo.NegTechnos` is set, the grey cameo will not show when you have a techno in one of these types.
  - The `UIDescription.Unbuildable` is like `UIDescription`, but this only appearing when the techno is truly unbuildable.

In `ra2md.ini`:
```ini
[Phobos]
ShowBuildingStatistics=false     ; boolean
```

In `rulesmd.ini`:
```ini
[AudioVisual]
Cameo.AlwaysExist=false          ; boolean
Cameo.OverlayShapes=pips.shp     ; filename - including the .shp extension
Cameo.OverlayFrames=-1,-1,-1     ; integer - owned this building, grey and have its prerequisite, grey but have no prerequisite
Cameo.OverlayPalette=palette.pal ; filename - including the .pal extension

[SOMETECHNO]                     ; TechnoType
Cameo.AlwaysExist=               ; boolean
Cameo.AuxTechnos=                ; List of TechnoTypes
Cameo.NegTechnos=                ; List of TechnoTypes
UIDescription.Unbuildable=       ; CSF entry key
```

In `artmd.ini`:
```ini
[SOMETECHNO]                     ; TechnoType
GreyCameoPCX=                    ; PCX filename - including the .pcx extension
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -372,6 +372,7 @@ This page lists all the individual contributions to the project by their author.
   - Allow to change the speed of gas particles
 - **CrimRecya**
   - Fix `LimboKill` not working reliably
+  - Grey cameo preview and cameo overlays
 - **Ollerus**
   - Build limit group enhancement
   - Customizable rocker amplitude
--- a/docs/User-Interface.md
+++ b/docs/User-Interface.md
@@ -367,6 +367,45 @@ In `rulesmd.ini`:
 MissingCameo=XXICON.SHP  ; filename - including the .shp/.pcx extension
 ```
 
+### Show cameo when unbuildable
+
+- A setting that allows you to preview information. This feature can be used as before, playing "new construction options" and clearing the specific production queue when prerequisites loss.
+  - `Cameo.AlwaysExist` controls whether you can see the cameo when the prerequisite have not satisfied (`TechnoLevel`, `Owner`, `RequiredHouses` and `ForbiddenHouses` should be satisfied). Defaults to `[AudioVisual]` -> `Cameo.AlwaysExist`.
+  - `ShowBuildingStatistics` controls whether the number of buildings of this type that you currently own needs to be displayed in the upper right corner of the building cameo (requires the cameo exist).
+  - `Cameo.OverlayShapes` controls the drawn image file.
+  - `Cameo.OverlayFrames` controls which frame in `Cameo.OverlayShapes` to draw in three different situations: currently owned this building type, grey cameo and have its prerequisite, grey cameo but have no prerequisite (The last situation requires `Cameo.AlwaysExist` to be true). When set to a negative number, it means that there is no need to draw under the corresponding conditions.
+  - `Cameo.OverlayPalette` the color palette used when drawing `Cameo.OverlayShapes`.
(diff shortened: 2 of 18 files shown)


CrimRecya:
Splits from #1335 .

Speederovsky:
FactoryOwners?

CrimRecya:
> FactoryOwners?

Not Supported. You can set `Cameo.AuxTechnos` to the critical technos.

Speederovsky:
Stolen tech?

Speederovsky:
Not supporting FactoryOwners might be a bit bold, it's not some obscure feature. Something for stolen tech?

Also, this doesn't seem to support building upgrades as prerequisites.

CrimRecya:
> * Not supporting FactoryOwners might be a bit bold, it's not some obscure feature.
> * Something for stolen tech? One might want to include their special settings.
> * The '_you own this building type_' overlay being drawn for defenses is quite a bit weird and it doesn't get drawn for walls, obviously. But visually that will give a weird final effect. It would be useful to let the Statistics only appear for buildings that are related to teching and not defenses.
> * "_the upper right corner of the building cameo_" You mean left. Right is for queues.

- Okay, I will consider it later.
- Can you tell me how you would like this to be displayed? In my opinion, not displaying these directly is the best.
- Great suggestion, I will separate their switches in the future.
- Thank you for the reminder. I will revise it again in the future.

Metadorius:
From earlier discussions in Discord, QQ and #1384: a suggestion on how to implement fixed position cameos (not necessary for this PR but it would complement it nicely).

Design-wise I propose every pickable player country/subfaction to have it's own set of static cameos which would be unique to this specific country/subfaction, and if the player has some tech that isn't specific to this faction/subfaction (for example, MCV crate, engineer capture, reverse-engineering) - it would be added as dynamic (regular) cameos. Fake cameos are not visible to user, they're simply transparent squares with no function.

Because cameo sorting is already controlled by Phobos (courtesy of @Thrifinesma), it should be trivial to add fake cameos and change the sorting function to always place the cameos in the fixed order.

image

PR #1478, opened 2024-12-28
[Highly Customized] Unit base normal and new placing grids

CrimRecya:
- You can now set `CheckUnitBaseNormal` to true to use units to expand the construction scope of the base.
  - `UnitBaseNormal` controls whether our own buildings can be place around it like vanilla `BaseNormal` do.
  - `UnitBaseForAllyBuilding` controls whether ally buildings can be place around it like vanilla `EligibileForAllyBuilding` do.

In `rulesmd.ini`:
```ini
[General]
CheckUnitBaseNormal=false      ; boolean

[SOMEUNIT]                     ; UnitType
UnitBaseNormal=false           ; boolean
UnitBaseForAllyBuilding=false  ; boolean
```

- This feature is highly compatible with `ExpandBuildingPlace`. If set `DrawAdjacentBoundary` to true, it will display the four corners of the `Adjacent` boundary. If set `PlacementGrid.Expand` to true, it will display the placing grids with `place.shp` and following corresponding frame number.
  - `PlacementGrid.LandFrames` controls the placing grids frames on non-water cell. The three numbers respectively represent "Some technos that can command departure have occupied this area", "This cell is actually beyond the scope, but there is still at least one cell inside the entire region" and "Here is no problem, everything is OK".
  - `PlacementGrid.WaterFrames` controls the placing grids frames on water cell. Each item corresponds to the same as above.

In `ra2md.ini`:
```ini
[Phobos]
DrawAdjacentBoundary=false       ; boolean
```

In `rulesmd.ini`:
```ini
[AudioVisual]
PlacementGrid.Expand=false       ; boolean
PlacementGrid.LandFrames=1,0,0   ; integer, zero-based frame index - have technos, near boundary, is normal
PlacementGrid.WaterFrames=1,0,0  ; integer, zero-based frame index - have technos, near boundary, is normal
```

Splits from #1335  .

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -372,6 +372,7 @@ This page lists all the individual contributions to the project by their author.
   - Allow to change the speed of gas particles
 - **CrimRecya**
   - Fix `LimboKill` not working reliably
+  - Unit base normal and new placing grids
 - **Ollerus**
   - Build limit group enhancement
   - Customizable rocker amplitude
--- a/Phobos.vcxproj
+++ b/Phobos.vcxproj
@@ -166,6 +166,7 @@
     <ClCompile Include="src\Misc\Hooks.SkirmishColors.cpp" />
     <ClCompile Include="src\New\Type\Affiliated\TypeConvertGroup.cpp" />
     <ClCompile Include="src\Ext\BuildingType\Hooks.Upgrade.cpp" />
+    <ClCompile Include="src\Ext\BuildingType\Hooks.Placing.cpp" />
     <ClCompile Include="src\Phobos.COM.cpp" />
     <ClCompile Include="src\Phobos.CRT.cpp" />
     <ClCompile Include="src\Phobos.Ext.cpp" />
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -1023,6 +1023,22 @@ ForbidParallelAIQueues.Building=no  ; boolean
 ForbidParallelAIQueues=false        ; boolean
 ```
 
+### Check building adjacent by using units
+
+- You can now set `CheckUnitBaseNormal` to true to use units to expand the construction scope of the base.
+  - `UnitBaseNormal` controls whether our own buildings can be place around it like vanilla `BaseNormal` do.
+  - `UnitBaseForAllyBuilding` controls whether ally buildings can be place around it like vanilla `EligibileForAllyBuilding` do.
+
+In `rulesmd.ini`:
+```ini
+[General]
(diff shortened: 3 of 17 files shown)


mevitar:
This ignores Adjacent.Allowed= , which means buildings that are restricted to be built only next to specific buildings will be buildable near a unit with UnitbaseNormal=yes.
Adjacent.Allowed= should either be expanded to allow VehicleTypes, or (i think better option) a similar tag should be added for buildings to check for specific VehicleTypes (like: Adjacent.AllowedUnits= ?)

Also, i see that only ground VehicleTypes appear to work, so there should be a note about that. I can understand if it's not meant for InfantryTypes and AircraftTypes, as supporting them might be too problematic, but there still should be a note that only grounded VehicleTypes will work.

CrimRecya:
> This ignores Adjacent.Allowed= , which means buildings that are restricted to be built only next to specific buildings will be buildable near a unit with UnitbaseNormal=yes. Adjacent.Allowed= should either be expanded to allow VehicleTypes, or (i think better option) a similar tag should be added for buildings to check for specific VehicleTypes (like: Adjacent.AllowedUnits= ?)

Get it. I will add it later.

> Also, i see that only ground VehicleTypes appear to work. I can understand if it's not meant for InfantryTypes and AircraftTypes, as supporting them might be too problematic, but there still should be a note that only grounded VehicleTypes will work.

Okay, but I would like to know if you think infantry/air force (jumpjet, aircraft) need this feature?
If it's not necessary, it's very simple. I'll add an explanation later;
If necessary, the logic of that part needs to be redone (it won't be very complicated, but it will take time).

mevitar:
> Okay, but I would like to know if you think infantry/air force (jumpjet, aircraft) need this feature? If it's not necessary, it's very simple. I'll add an explanation later; If necessary, the logic of that part needs to be redone (it won't be very complicated, but it will take time. In fact, infantry is also easy, but the air force cannot handle it like the original version).

Necessary or not is an incorrect question, because if it's there, then people will find a use for it. :P
(for example, an engineering helicopter that allows placing base defenses nearby, or a hero infantry that allows the same)
And I myself wanted to add it to an infantry, but i'll manage without it if i have to.

Still, i think adding it to air units might still be too much work for too little gain. Just a note that airborne objects are not supported should suffice.

Metadorius:
If it's not too complex, jumpjet vehicles could be cool too. For example, there was an old game called Urban Assault, I once wanted to replicate it in C&C. There a floating host station could build stuff around itself.

AircraftTypes I don't see much point in adding.

Also seeing that there's an UI improvement - it could be cool to show it in docs as an image.

PR #1480, opened 2024-12-28
[Customized] Customized Vehicle Turret Rotation

CrimRecya:
- When `ExpandTurretRotation` is set to true, the following functions will be enabled.
- Units with turret without `TurretSpins=true` can looks more vivid when it is in idle.
  - `Turret.IdleRotate` controls whether units can rotate their turrets when in idle. Defaults to `[AudioVisual]` -> `Turret.IdleRotate`.
  - `Turret.PointToMouse` controls whether units will turn their turrets to your mouse when in idle. Defaults to `[AudioVisual]` -> `Turret.PointToMouse`.
  - `Turret.IdleRestartMin` and `Turret.IdleRestartMax` control the delay from idle to action occurrence together.
  - `Turret.IdleIntervalMin` and `Turret.IdleIntervalMax` control the delay between every idle actions together.
- The turret and body of the units can now be rotated under control.
  - `Turret.Restriction` defines the angle at which the turret can be turned to both sides.
  - `Turret.ExtraAngle` defines the additional rotation angle of the turret, and `Turret.Restriction` will also rotate with this value. The positive number is clockwise and the negative number is counterclockwise, that is, what angle the turret should use by default or face the target.
  - `Turret.BodyOrientation` controls whether the body needs to aim at the target when firing.
  - `Turret.BodyOrientationAngle` defines the additional rotation angle of the body when aiming at the target. The positive number is clockwise and the negative number is counterclockwise, that is, what angle the body should use to face the target.
  - `Turret.BodyOrientationSymmetric` controls whether both sides of the additional body rotation angle can be used.

In `rulesmd.ini`:
```ini
[General]
ExpandTurretRotation=no             ; boolean

[AudioVisual]
Turret.IdleRotate=false             ; boolean
Turret.PointToMouse=false           ; boolean
Turret.IdleRestartMin=150           ; integer, number of frames
Turret.IdleRestartMax=300           ; integer, number of frames
Turret.IdleIntervalMin=150          ; integer, number of frames
Turret.IdleIntervalMax=450          ; integer, number of frames

[SOMEUNIT]                          ; VehicleType, `Turret=yes`
Turret.IdleRotate=                  ; boolean
Turret.PointToMouse=                ; boolean
Turret.Restriction=180.0            ; floating point value
Turret.ExtraAngle=0                 ; floating point value
Turret.BodyOrientation=no           ; boolean
Turret.BodyOrientationAngle=0       ; floating point value
Turret.BodyOrientationSymmetric=yes ; boolean
```

Update for #1308 .


src/Ext/Techno/Body.Update.cpp
```diff
@@ -588,6 +590,121 @@ void TechnoExt::ExtData::UpdateMindControlAnim()
...
+
+		if (pTypeExt->Turret_IdleRotate.Get(RulesExt::Global()->Turret_IdleRotate))
+			this->StopIdleAction();
+
+		this->UnitIdleIsSelected = true;
+		const auto mouseCoords = TacticalClass::Instance->ClientToCoords(WWMouseClass::Instance->XY1);
+
+		if (mouseCoords != CoordStruct::Empty) // Mouse in tactical
+		{
+			const auto offset = -static_cast<int>(pThis->GetCoords().Z * 1.2307692307692307692307692307692); // ((Unsorted::LeptonsPerCell / 2) / Unsorted::LevelHeight)
```
Metadorius, line 679: I'm pretty sure the compiler will optimize this for you, so no need to calculate manually.

src/Ext/TechnoType/Body.h
```diff
@@ -485,6 +507,8 @@ class TechnoTypeExt
 
 	static ExtContainer ExtMap;
 
+	static constexpr double AngleToRaw = (65536.0 / 360);
```
Metadorius, line 510: I am pretty sure this can be handled with the code @secsome has written for Binary Angle Measurement, see Dir.h and Facing.h in YRpp.

CrimRecya: I think I may have missed something? I did notice that it can use radian to raw, but I didn't see angle to raw.

Metadorius: Oh, I see where the issue is. I think most angle values parsed from INI are DirType, not degrees, so I would use this too. See `RotationAdjust` here for example: https://github.com/Phobos-developers/Phobos/pull/352/files

CrimRecya: Get it!

CrimRecya: I looked at it, but I thought it might be troublesome.
Because if you use DirType, INIer need to fill in the value of 0-255, which is not very intuitive.
But if you use the angle directly, it is very easy to understand.

Metadorius: There is already such a usage in vanilla game. `PoseDir` takes DirType. I think it's better to stay consistent in this case. Personally I don't feel much issue with using 0-255 instead od 0-360.

CrimRecya: OK, I will consider modifying it later.

Metadorius:
Would it be easy to also make it so that rotating the turret is done relative to the body rotation? Right now it is possible to order the unit to move right behind it, and the turret turns in one direction, while the body turns in another direction, which is completely unrealistic.

Also, it doesn't make a lot of sense for the turret to not be affected by body rotation. In #352 I found that if I make the attached unit ignore the movement order -- its turret just faces in fthe orward direction, which is alright, but when the attached unit is turned - the turret lags behind a frame or a few because there's no dependency between body orientation and turret orientation, and it didn't turn with the body automatically.

This will also allow for some interesting gameplay things like not allowing to fire when rotating the body if you set TurretROT to a value that can't compensate for ROT quickly enough, like older WW2 tanks which didn't have good stabilization

CrimRecya:
> Would it be easy to also make it so that rotating the turret is done relative to the body rotation? Right now it is possible to order the unit to move right behind it, and the turret turns in one direction, while the body turns in another direction, which is completely unrealistic.

Are you referring to allowing the rotation of the turret to overlap with the rotation of the vehicle body (when the vehicle body rotates by one angle, the turret can rotate its own angle plus the angle of the vehicle body)? I have indeed considered this before, but currently my thinking is limited. Apart from constantly forcibly resetting the orientation and allowing it to continue rotating, I haven't come up with any particularly good methods yet.

> Also, it doesn't make a lot of sense for the turret to not be affected by body rotation. In #352 I found that if I make the attached unit ignore the movement order -- its turret just faces in fthe orward direction, which is alright, but when the attached unit is turned - the turret lags behind a frame or a few because there's no dependency between body orientation and turret orientation, and it didn't turn with the body automatically.

At present, a delay of 1 frame can also be observed on ordinary units, and I am not sure if it is due to the update order at the moment. I'll mark it first so that I can make modifications when I have time later.

> This will also allow for some interesting gameplay things like not allowing to fire when rotating the body if you set TurretROT to a value that can't compensate for ROT quickly enough, like older WW2 tanks which didn't have good stabilization

lol

PR #1482, opened 2024-12-29
[Minor] Attach effect when weapon fire

CrimRecya:
`AttachEffect` will directly act on the target at the instant after firing.

```ini
[SOMEWEAPON]                                       ; WeaponType
AttachEffect.AttachTypes=                          ; List of AttachEffectTypes
AttachEffect.CumulativeRefreshAll=false            ; boolean
AttachEffect.CumulativeRefreshAll.OnAttach=false   ; boolean
AttachEffect.CumulativeRefreshSameSourceOnly=true  ; boolean
AttachEffect.RemoveTypes=                          ; List of AttachEffectTypes
AttachEffect.RemoveGroups=                         ; comma-separated list of strings (group IDs)
AttachEffect.CumulativeRemoveMinCounts=            ; integer - minimum required instance count (comma-separated) for cumulative types in order from first to last.
AttachEffect.CumulativeRemoveMaxCounts=            ; integer - maximum removed instance count (comma-separated) for cumulative types in order from first to last.
AttachEffect.DurationOverrides=                    ; integer - duration overrides (comma-separated) for AttachTypes in order from first to last.
```

More general expansion for #1472 .


Coronia:
I would say we should still keep AE for warhead only, especially when we can use ExtraWarheads and such to provide different effects for different technos in one shot. Adding a connection between weapon and AE would make the system overcomplicated, while not gaining much

CrimRecya:
> I would say we should still keep AE for warhead only, especially when we can use ExtraWarheads and such to provide different effects for different technos in one shot. Adding a connection between weapon and AE would make the system overcomplicated, while not gaining much

The biggest advantage of weapon AE over warhead AE is that it is completed in an instant. It can produce different effects with the order of unit update, and the warhead AE will not take effect until all bullets are updated and detonated. Sometimes, it is too late.

CrimRecya:
> I would say we should still keep AE for warhead only, especially when we can use ExtraWarheads and such to provide different effects for different technos in one shot. Adding a connection between weapon and AE would make the system overcomplicated, while not gaining much

The biggest advantage of weapon AE over warhead AE is that it is completed in an instant. It can produce different effects with the order of unit update, and the warhead AE will not take effect until all bullets are updated and detonated. Sometimes, it is too late.

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -426,6 +428,20 @@ DEFINE_HOOK(0x6FDDC0, TechnoClass_FireAt_DiscardAEOnFire, 0x6)
...
 	}
 
+	if (const auto pWeaponExt = WeaponTypeExt::ExtMap.Find(pWeapon))
+	{
+		if (pWeaponExt->AttachEffect_Enable)
+		{
+			if (const auto pTargetTechno = abstract_cast<TechnoClass*>(pTarget))
+			{
+				auto const& info = pWeaponExt->AttachEffects;
+				AttachEffectClass::Attach(pTargetTechno, pThis->Owner, pThis, pWeapon->Warhead, info);
```
Starkku, line 438: The source parameter should be just the weapon, not Warhead. It takes `AbstractClass` pointer and does not make assumptions about what type it is, it is there just to track where the AE instance came from.

Starkku (changes requested):
It is potentially niche but the code is not as complex as it could be since there's very little bespoke code (parsing is handled by `AEAttachInfoTypeClass` which can directly be passed to the attach/detach functions). Whether or not there's a performance overhead that outweighs potential benefits is hard to quantify, though.

There's one small thing that should be changed at the very least.

src/Ext/Techno/Hooks.Firing.cpp
CrimRecya: In this regard, I found it inappropriate to pass weapons as parameters in previous tests, because the warhead of the current weapon is used to determine whether the weapon can attack (`IgnoreFromSameSource`). If the incoming source here is a weapon, units cannot continue to attack the target.

Starkku: Right, forgot about that. Might be acceptable concession then in order to reduce complexity in the AE checks then, although it could cause some unusual results.

Speederovsky:
Why make this when you could design something like an InstantWarhead detonation that skips all these checks too and gives us more flexibility to what the instantly applied effects actually are? Not just AE, and you skip having to redo all nuAE tags on WeaponTypes.

CrimRecya:
> Why make this when you could design something like an InstantWarhead detonation that skips all these checks too and gives us more flexibility to what the instantly applied effects actually are? Not just AE, and you skip having to redo all nuAE tags on WeaponTypes.

Your idea touched me. If killing some units in the process of traversing all units will not cause problems in the iterator due to DTOR or quantity reduction (maybe, not sure), it is worth trying.

CrimRecya:
After evaluation, `InstantWarhead` is considered to be unhelpful to the content of this pr. The same warhead is required if AE needs to be used to search target by the weapon. That is, `InstantWarhead` should be used as the Boolean value of a projectile with `Inviso=true`, so that the projectile will detonate immediately after it is set in position, instead of waiting for a uniform update time.

Metadorius:
I didn't quite get why is `InstantWarhead` unsuitable, can you elaborate?

CrimRecya:
> I didn't quite get why is `InstantWarhead` unsuitable, can you elaborate?

As discussed with Starkku before, if the weapon needs to use AE to target (such as `IgnoreFromSameSource`), the warhead of the weapon must be the same. If it wants to replace the requirement that PR can solve, it should move to the target position immediately after the projectile is generated, and then detonate immediately (to ensure that the `Warhead` is the same). If we use a new logic similar to `ExtraWarhead` to carry a new warhead, we can not achieve the purpose of PR.
In short, it should be possible, but it is different from the goal of PR.

Coronia (approved):

PR #1484, opened 2024-12-30
More aircraft actions enhancement for #1366

CrimRecya:
1. Fix aircrafts planning waypoint problem - At the end of each task, all units with planning waypoints will generate an mega event to the next planning waypoint and respond after 1 frame. This is really no problem for most units, but for aircrafts, they will always return to the airport after each task. So they will try to return to the airport in the frame before it respond to event. There will be looked like a "shaking", if the `Speed` and `ROT` of the aircraft are fast. To solve this problem, I check the waypoints of aircrafts and let them wait before returning.
2. Fix a vanilla issue - Now aircrafts can fly towards the water surface normally.

### DIFF
--- a/src/Ext/Aircraft/Hooks.cpp
+++ b/src/Ext/Aircraft/Hooks.cpp
@@ -362,7 +362,7 @@ DEFINE_HOOK(0x416A0A, AircraftClass_Mission_Move_SmoothMoving, 0x5)
 	return EnterIdleAndReturn;
 }
 
-DEFINE_HOOK(0x4DDD66, FootClass_vt_entry_550_ReplaceHardcode, 0x5) // To avoid that the aircraft cannot fly towards the water surface normally
+DEFINE_HOOK(0x4DDD66, FootClass_IsLandZoneClear_ReplaceHardcode, 0x6) // To avoid that the aircraft cannot fly towards the water surface normally
 {
 	enum { SkipGameCode = 0x4DDD8A };
 
--- a/src/Ext/Aircraft/Hooks.cpp
+++ b/src/Ext/Aircraft/Hooks.cpp
@@ -356,7 +356,7 @@ DEFINE_HOOK(0x416A0A, AircraftClass_Mission_Move_SmoothMoving, 0x5)
 		return (R->Origin() == 0x4168C7 ? ContinueMoving1 : ContinueMoving2);
 
 	// Try next planning waypoint first, then return to air base if it does not exist or cannot be taken
-	if (!reinterpret_cast<bool(__thiscall*)(AircraftClass*)>(0x6385C0)(pThis))
+	if (!pThis->TryNextPlanningTokenNode())
 		pThis->EnterIdleMode(false, true);
 
 	return EnterIdleAndReturn;
--- a/src/Ext/Aircraft/Hooks.cpp
+++ b/src/Ext/Aircraft/Hooks.cpp
@@ -351,13 +351,30 @@ DEFINE_HOOK(0x416A0A, AircraftClass_Mission_Move_SmoothMoving, 0x5)
 	const int distance = Game::F2I(Point2D { pCoords->X, pCoords->Y }.DistanceFrom(Point2D { pThis->Location.X, pThis->Location.Y }));
 
 	// When the horizontal distance between the aircraft and its destination is greater than half of its deceleration distance
-	// or its turning radius, continue to move forward, otherwise return to airbase or execute the next planning path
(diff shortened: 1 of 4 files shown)


src/Ext/Aircraft/Hooks.cpp
```diff
@@ -351,15 +351,32 @@ DEFINE_HOOK(0x416A0A, AircraftClass_Mission_Move_SmoothMoving, 0x5)
...
 
 	// When the horizontal distance between the aircraft and its destination is greater than half of its deceleration distance
-	// or its turning radius, continue to move forward, otherwise return to airbase or execute the next planning path
+	// or its turning radius, continue to move forward, otherwise return to airbase or execute the next planning waypoint
 	if (distance > std::max((pType->SlowdownDistance >> 1), (2048 / pType->ROT)))
 		return (R->Origin() == 0x4168C7 ? ContinueMoving1 : ContinueMoving2);
 
-	pThis->EnterIdleMode(false, true);
+	// Try next planning waypoint first, then return to air base if it does not exist or cannot be taken
+	if (!reinterpret_cast<bool(__thiscall*)(AircraftClass*)>(0x6385C0)(pThis))
```
Metadorius, line 359: Let's add this to YRpp. Not sure of the name though, do you have any suggestions?

src/Ext/Aircraft/Hooks.cpp
```diff
@@ -351,15 +351,32 @@ DEFINE_HOOK(0x416A0A, AircraftClass_Mission_Move_SmoothMoving, 0x5)
...
+		pThis->EnterIdleMode(false, true);
 
 	return EnterIdleAndReturn;
 }
 
+DEFINE_HOOK(0x4DDD66, FootClass_vt_entry_550_ReplaceHardcode, 0x5) // To avoid that the aircraft cannot fly towards the water surface normally
```
Metadorius, line 365:
```suggestion
DEFINE_HOOK(0x4DDD66, FootClass_IsLZClear_ReplaceHardcode, 0x6) // To avoid that the aircraft cannot fly towards the water surface normally
```

src/Ext/Aircraft/Hooks.cpp
```diff
@@ -351,15 +351,32 @@ DEFINE_HOOK(0x416A0A, AircraftClass_Mission_Move_SmoothMoving, 0x5)
...
+	enum { SkipGameCode = 0x4DDD8A };
+
+	GET(FootClass* const, pThis, EBP);
+	GET_STACK(CellStruct, cell, STACK_OFFSET(0x20, 0x4));
+
+	const auto pType = pThis->GetTechnoType();
+
+	// In vanilla, only aircrafts or `foots with fly locomotion` will call this virtual function
+	// So I don't know why WW use hard-coded `SpeedType::Track` and `MovementZone::Normal` to check this
+	R->AL(MapClass::Instance->GetCellAt(cell)->IsClearToMove(pType->SpeedType, false, false, -1, pType->MovementZone, -1, true));
```
Metadorius, line 376: I think this change alone may change the vanilla behavior. Are `SpeedType::Track` and `MovementZone::Normal` default values?

Metadorius (changes requested):
Looking good so far, I'll fix the docs a bit soon.

src/Ext/Aircraft/Hooks.cpp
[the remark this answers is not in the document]
CrimRecya: Oh, sorry for the mistake.

src/Ext/Aircraft/Hooks.cpp
CrimRecya: Maybe `TryNextPlanningTokenNode` ?
Sorry, but I can't think of a good name.

src/Ext/Aircraft/Hooks.cpp
CrimRecya: This is to enable the aircraft to fly directly to above the water surface by clicking. It will definitely change the vanilla behavior.
These two were hard coded, which caused the aircraft to set the destination back to the shore like ordinary ground units.

src/Ext/Aircraft/Hooks.cpp
Metadorius: Sounds good. Can you submit a PR to YRpp?

CrimRecya: I just submitted a PR to YRpp.

Metadorius (dismissed):
Clarify docs

Metadorius (approved):

PR #1485, opened 2024-12-30
[Minor] Additionary toggle in settings for SelectionFlashDuration

Fryone:
in RA2MD.ini

[Phobos]
ShowFlashOnSelecting = true/false

### DIFF
--- a/src/Ext/Techno/Hooks.Misc.cpp
+++ b/src/Ext/Techno/Hooks.Misc.cpp
@@ -369,8 +369,9 @@ DEFINE_HOOK(0x5F46AE, ObjectClass_Select, 0x7)
 
 	pThis->IsSelected = true;
 
-	if (RulesExt::Global()->SelectionFlashDuration > 0 && pThis->GetOwningHouse()->IsControlledByCurrentPlayer())
-		pThis->Flash(RulesExt::Global()->SelectionFlashDuration);
+	if (Phobos::Config::ShowFlashOnSelecting && RulesExt::Global()->SelectionFlashDuration > 0)
+		if (pThis->GetOwningHouse()->IsControlledByCurrentPlayer())
+			pThis->Flash(RulesExt::Global()->SelectionFlashDuration);
 
 	return 0;
 }
--- a/docs/User-Interface.md
+++ b/docs/User-Interface.md
@@ -268,12 +268,19 @@ ShowTimer.Priority=0  ; integer
 
 ### Flashing Technos on selecting
 
-Selecting technos, controlled by player, now may show a flash effect by setting `SelectionFlashDuration` parameter. Set `SelectionFlashDuration=0` to disable it.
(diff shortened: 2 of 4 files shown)


src/Ext/Rules/Body.h
```diff
@@ -139,7 +139,7 @@ class RulesExt
 
 		Valueable<bool> ShowDesignatorRange;
 		Valueable<bool> IsVoiceCreatedGlobal;
-		Valueable<int> SelectionFlashDuration;
+		Nullable<int> SelectionFlashDuration;
 		AnimTypeClass* DropPodTrailer;
```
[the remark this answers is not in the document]
Fryone: ok, ok, I get it now

src/Ext/Rules/Body.cpp
```diff
@@ -170,7 +170,10 @@ void RulesExt::ExtData::LoadBeforeTypeData(RulesClass* pThis, CCINIClass* pINI)
 	this->DisplayIncome_AllowAI.Read(exINI, GameStrings::AudioVisual, "DisplayIncome.AllowAI");
 
 	this->IsVoiceCreatedGlobal.Read(exINI, GameStrings::AudioVisual, "IsVoiceCreatedGlobal");
-	this->SelectionFlashDuration.Read(exINI, GameStrings::AudioVisual, "SelectionFlashDuration");
+
+	if (Phobos::Config::ShowFlashOnSelecting)
```
Starkku, line 174: This has to be checked in-place e.g the hook in `ObjectClass::Select()` rather than in INI parsing to be consistent with how these user config settings are handled. There is currently no way to change them in-game if not tied to a keyboard command but it is not a reason to not future-proof the implementation.

Starkku (changes requested):
Needs one change, besides that it is good to go.

src/Ext/Rules/Body.cpp
Fryone: Reverted back

src/Ext/Techno/Hooks.Misc.cpp
```diff
@@ -369,8 +369,9 @@ DEFINE_HOOK(0x5F46AE, ObjectClass_Select, 0x7)
 
 	pThis->IsSelected = true;
 
-	if (RulesExt::Global()->SelectionFlashDuration > 0 && pThis->GetOwningHouse()->IsControlledByCurrentPlayer())
-		pThis->Flash(RulesExt::Global()->SelectionFlashDuration);
+	if (Phobos::Config::ShowFlashOnSelecting)
+		if (pThis->GetOwningHouse()->IsControlledByCurrentPlayer() && RulesExt::Global()->SelectionFlashDuration > 0)
```
CrimRecya, line 373: One line is ok.

CrimRecya (approved):
There is a minor issue with the code format, but I think it's okay.

src/Ext/Techno/Hooks.Misc.cpp
Fryone: changed

Aephiex (approved):

PR #1486, opened 2025-01-01
[New Feature] Basic Event Handler System

Aephiex:
The intentions:
1. To grasp a corner of the greatness of KratosPP, with the functionality as close to it as possible, but still stay a good performance and stability.
2. Allow many highly customized effects in a straightforward and comprehensible way. It must not be complicated or confusing, and the resulting ini code must be highly readible.
3. Various effects can be done with a simple weapon detonation. Such effects are not recommended to be added to the Event Handler system, unless doing in the other way is too complicated or not fairly straightforward, or is not able to affect some scopes that the modder may have been intended to (passengers for example).
4. More pre-defined event calls are welcome, as they open up more possibility for general modding.


Aephiex:
@Metadorius This is not fully ready yet, but what are your thoughts?

Aephiex:
Okay, it is now working as intended.

mevitar:
Is this the reason for why you closed your recent branches? Putting everything into one system?
And so i presume weapon firing on being crushed will be handled through events now?

Aephiex:
> Is this the reason for why you closed your recent branches? Putting everything into one system? And so i presume weapon firing on being crushed will be handled through events now?

Yes, this one is more versatile and can do more than just the crushed weapon trigger, infiltrate weapon trigger, and so on, and this one is fairly straightforward and easy to use.

Aephiex:
Alright, the todo list is cleared.

Aephiex:
A new house compare filter is here, it is used to compare a house's credits, power surplus, power output, or power drain with an integer. Such as the following house compare filter requires credits >= 1000. This paved the path to AE stack count filters, and even the upcoming player flag stack count filters.

Not tested if this works though, no more free time for today.

```ini
(actor).Filter.HouseCompare.Params=Credits
(actor).Filter.HouseCompare.Methods=ge
(actor).Filter.HouseCompare.Values=1000
```

<hr/>

By the say, a ValueableVector for an enumerator type causes build error.

```c
ValueableVector<HouseParameterType> HouseCompare_Params;
ValueableVector<ComparatorType> HouseCompare_Methods;
```

Had to modify `parser.h` to support enumerators. Not sure if it's right, at least it doesn't cause a build error. Will test it later.

Aephiex:
Now with Player Emblems. These are some attached effects directly given to the houses and can have their own event triggers. Can also modify the player's build options.

Aephiex:
1. Player Emblems can now enforce some Attached Effects on owned techno types.
2. Attached Effects can be automatically discarded when the house has or has no certain emblem types.
3. House filters now support emblems.

Aephiex:
New event types added.
1. Factory produces a unit: `WhenProduce`, `WhenProduced`
2. A unit is grinded: `WhenGrind`, `WhenGrinded`
3. A SW is launched / impacts a techno: `WhenLaunch`, `WhenImpact`
4. AE's `AttachEffect.InitialDelays` has passed: `WhenInitialize`

Aephiex:
Some projectile related events:
1. `WhenLaunch`
2. `WhenIntercept` / `WhenIntercepted`

Aephiex:
> This amount use of `std::map` is not really advisable , this can lead to performance issues , consider to replace it with `PhobosMap` since it more simple impementation , unless you need specific feature(s) from the `std::map` like hasing, etc .

Thought `std::map` was more basic and more performance friendly. If `PhobosMap` was more performance friendly then there is no need to abuse `std::map`. Thank you for telling me.

Aephiex:
New features:
1. New event types to be fired when the unit changes its layer, such as digging into the underground, or taking a flight.
2. Custom events can be fired through an user's hotkey. Up to 10 custom hotkeys can be enabled by the modder; in theory, an infinite amount of custom hotkeys can be supported, but I can't think of a case the modder needs more than just 2.

I made an open-topped `BalloonHover=yes` jumpjet able to land and take / release passengers through conventional means, and later stay flying in the air just like a normal `BalloonHover=yes` jumpjet.

动画

Metadorius:
Why did you close those PRs?

PR #1487, opened 2025-01-05
Use 2D distance to check whether team members have arrived destination

CrimRecya:
- Now the team members use the 2D distance instead of the 3D distance to judge whether to reach the mission destination, so as to prevent the problem that the mission is stuck and cannot continue in some cases (such as when the jumpjet stops on the building).

### DIFF
--- a/src/Misc/Hooks.BugFixes.cpp
+++ b/src/Misc/Hooks.BugFixes.cpp
@@ -1141,7 +1141,7 @@ DEFINE_HOOK(0x4C75DA, EventClass_RespondToEvent_Stop, 0x6)
 int __fastcall Check2DDistanceInsteadOf3D(ObjectClass* pSource, void* _, AbstractClass* pTarget)
 {
 	return (pSource->IsInAir() && pSource->WhatAmI() != AbstractType::Aircraft) // Jumpjets or sth in the air
-		? pSource->DistanceFrom(pTarget) // 2D distance
+		? (pSource->DistanceFrom(pTarget) * 2) // 2D distance (2x is the bonus to units in the air)
 		: reinterpret_cast<int(__thiscall*)(ObjectClass*, AbstractClass*)>(0x5F6360)(pSource, pTarget); // 3D distance (vanilla)
 }
 DEFINE_JUMP(CALL, 0x6EBCC9, GET_OFFSET(Check2DDistanceInsteadOf3D));
--- a/src/Misc/Hooks.BugFixes.cpp
+++ b/src/Misc/Hooks.BugFixes.cpp
@@ -1143,25 +1143,17 @@ int __fastcall Check2DDistanceInsteadOf3D(AbstractClass* pSource, void* _, Abstr
 	const auto sourceCoords = pSource->GetCoords();
 	const auto targetCoords = pTarget->GetCoords();
 
-	int distance = Game::F2I(Point2D { sourceCoords.X - targetCoords.X, sourceCoords.Y - targetCoords.Y }.Magnitude());
+	// Aircraft has its own unique treatment, and it will not be changed here
+	int distance = Game::F2I((pSource->IsInAir() && pSource->WhatAmI() != AbstractType::Aircraft) ? // Jumpjets or sth in the air ?
+		(Point2D { sourceCoords.X - targetCoords.X, sourceCoords.Y - targetCoords.Y }.Magnitude() * 2) : // bonus to units in the air
+		sourceCoords.DistanceFrom(targetCoords)); // Original
 
 	if (const auto pBuilding = abstract_cast<BuildingClass*>(pTarget)) // Vanilla bonus to building
 		distance -= ((pBuilding->Type->GetFoundationWidth() + pBuilding->Type->GetFoundationHeight(false)) << 6);
 
-	return distance;
+	return Math::max(0, distance);
 }
-DEFINE_JUMP(CALL, 0x6EB686, GET_OFFSET(Check2DDistanceInsteadOf3D)); // MoveToObject - TeamSpawnCell
-DEFINE_JUMP(CALL, 0x6EB8FF, GET_OFFSET(Check2DDistanceInsteadOf3D)); // TryToGuard - TeamSpawnCell1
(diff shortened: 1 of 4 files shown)


Aephiex (approved):

NetsuNegi:
The game has it's own DistanceFrom2D function, see 0x5F6440.
Maybe you can just redirect to that function.

CrimRecya:
> The game has it's own DistanceFrom2D function, see 0x5F6440. Maybe you can just redirect to that function.

Thank you for reminding me. But I found that this fix affect the aircraft, so I decided to specialize it.

src/Misc/Hooks.BugFixes.cpp
```diff
@@ -1136,6 +1136,27 @@ DEFINE_HOOK(0x4C75DA, EventClass_RespondToEvent_Stop, 0x6)
...
+
+int __fastcall Check2DDistanceInsteadOf3D(AbstractClass* pSource, void* _, AbstractClass* pTarget)
+{
+	const auto sourceCoords = pSource->GetCoords();
+	const auto targetCoords = pTarget->GetCoords();
+
+	// Aircraft has its own unique treatment, and it will not be changed here
+	int distance = Game::F2I((pSource->IsInAir() && pSource->WhatAmI() != AbstractType::Aircraft) ? // Jumpjets or sth in the air ?
+		(Point2D { sourceCoords.X - targetCoords.X, sourceCoords.Y - targetCoords.Y }.Magnitude() * 2) : // bonus to units in the air
+		sourceCoords.DistanceFrom(targetCoords)); // Original
```
Metadorius, line 1149:
```suggestion
	int distance = Game::F2I((pSource->IsInAir() && pSource->WhatAmI() != AbstractType::Aircraft) // Jumpjets or sth in the air ?
		? (Point2D { sourceCoords.X - targetCoords.X, sourceCoords.Y - targetCoords.Y }.Magnitude() * 2) // bonus to units in the air
		: sourceCoords.DistanceFrom(targetCoords)); // Original
```

The reason for this change is because we read code from left to right, so it's far easier to see the operator(s) at the beginning of the line than on the end.

src/Misc/Hooks.BugFixes.cpp
```diff
@@ -1136,6 +1136,27 @@ DEFINE_HOOK(0x4C75DA, EventClass_RespondToEvent_Stop, 0x6)
...
+
+	if (const auto pBuilding = abstract_cast<BuildingClass*>(pTarget)) // Vanilla bonus to building
+		distance -= ((pBuilding->Type->GetFoundationWidth() + pBuilding->Type->GetFoundationHeight(false)) << 6);
```
Metadorius, line 1152: 1. Does this account for custom foundations Ares introduces?
2. ~~What is the purpose of `<< 6`, if you know? Is it some kind of a magic number?~~ Oh nevermind, you're getting leptons per cell. I think the compiler will optimize this on it's own, so I suggest to instead use multiplication by `Unsorted::LeptonsPerCell`.

Also, not quite sure if this (decomp. from `ObjectClass::Distance`) translates to what you've written?
image

Metadorius (changes requested):

src/Misc/Hooks.BugFixes.cpp
CrimRecya: Get it.

src/Misc/Hooks.BugFixes.cpp
CrimRecya: 1. I really forgot this.
2. It is copied from the paragraph in the picture.
In general, maybe it would be better for me to use like `return jumpjet ? 2D() : 3D()` directly?

Metadorius: 1. I think it may work already as is, IIRC Ares just makes the custom foundations parse-able, just need to verify whether this is true.
2. I guess it's just compiler optimizations.
   It is an option, I think both current way and the way you suggested should work.

src/Misc/Hooks.BugFixes.cpp
```diff
@@ -1138,20 +1138,11 @@ DEFINE_HOOK(0x4C75DA, EventClass_RespondToEvent_Stop, 0x6)
...
-		(Point2D { sourceCoords.X - targetCoords.X, sourceCoords.Y - targetCoords.Y }.Magnitude() * 2) : // bonus to units in the air
-		sourceCoords.DistanceFrom(targetCoords)); // Original
-
-	if (const auto pBuilding = abstract_cast<BuildingClass*>(pTarget)) // Vanilla bonus to building
-		distance -= ((pBuilding->Type->GetFoundationWidth() + pBuilding->Type->GetFoundationHeight(false)) << 6);
-
-	return Math::max(0, distance);
+	return (pSource->IsInAir() && pSource->WhatAmI() != AbstractType::Aircraft) // Jumpjets or sth in the air
+		? (pSource->DistanceFrom(pTarget) * 2) // 2D distance (2x is the bonus to units in the air)
+		: reinterpret_cast<int(__thiscall*)(ObjectClass*, AbstractClass*)>(0x5F6360)(pSource, pTarget); // 3D distance (vanilla)
```
Metadorius: It seems it's called `AbstractClass::Distance` in my IDB. I'll put it in YRpp.

src/Misc/Hooks.BugFixes.cpp
```diff
@@ -1136,6 +1136,18 @@ DEFINE_HOOK(0x4C75DA, EventClass_RespondToEvent_Stop, 0x6)
 
 #pragma endregion
 
+#pragma region TeamCloseRangeFix
+
+int __fastcall Check2DDistanceInsteadOf3D(ObjectClass* pSource, void* _, AbstractClass* pTarget)
+{
+	return (pSource->IsInAir() && pSource->WhatAmI() != AbstractType::Aircraft) // Jumpjets or sth in the air
```
Metadorius, line 1143: Is there a reason you omit AircraftClass? Also does all the in air stuff need to be considered like so?

CrimRecya: At present, it seems that aircrafts use their own mapcoords and the team destination's mapcoords to check.
During the previous test, it was found that if the aircraft uses this and needs to return to the airport with the script first, it will interrupt the remaining tasks for unknown reasons.

Metadorius (approved):

PR #1490, opened 2025-01-07
[Minor] Fix a RE caused by `DeployToFire`

CrimRecya:
Fix the problem caused by the incorrect return address of the hook in b37 repair.
Skipping repeated check. And avoiding RE caused by uninitialized different data in EBX of different players.


Coronia:
With DeployToFire changes in b37, there's a high chance of desync when using this on unit with DeploysInto ([see attached vid](https://github.com/user-attachments/assets/538c212a-29f9-4fb2-997c-db215799829e))

This fix did prevent the said desync in our actual test

Coronia (approved):

Metadorius (approved):

PR #1493, opened 2025-01-15
Manually empty and reload ammo

CrimRecya:
- You can now use the shortcut key to manually empty and reload ammo for units. Aircraft still needs to return to the airport to reload.
  - `CanManualReload` controls whether this type of techno can manually empty and reload ammo.
  - For shortcut keys’ localization add `TXT_MANUAL_RELOAD` and `TXT_MANUAL_RELOAD_DESC` into your `.csf` file.

In `rulesmd.ini`:
```ini
[SOMETECHNO]            ; TechnoType
CanManualReload=false   ; boolean
```

You can also combine `CanManualReload`, `NoAmmoWeapon`, `FeedbackWeapon` and `AttachEffect` to make some special attack effects.

### DIFF
--- a/src/Ext/Event/Body.cpp
+++ b/src/Ext/Event/Body.cpp
@@ -1,8 +1,10 @@
-/*
 #include "Body.h"
 
 #include <Helpers/Macro.h>
 #include <EventClass.h>
+#include <HouseClass.h>
+
+#include <Ext/TechnoType/Body.h>
 
 bool EventExt::AddEvent()
 {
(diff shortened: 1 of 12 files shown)


Speederovsky:
What inspired this idea?

CrimRecya:
> What inspired this idea?

A unit that does not require ammunition and can overload (just like a skill), or a unit consumes a lot in the first round but the sustained firepower is insufficient.
The combination I first talked about was wave force artillery.

Coronia:
is it possible to add a shortcut to trigger this by deploy command? Think that'll be easier to be used by players than adding a new hotkey

src/Ext/Event/Body.cpp
```diff
@@ -13,18 +16,59 @@ void EventExt::RespondEvent()
...
+	eventExt.ManualReloadEvent.Who = TargetClass(pTechno);
+	eventExt.AddEvent();
+	Debug::LogGame("Adding event MANUAL_RELOAD\n");
+}
+
+void EventExt::RespondToManualReloadEvent()
+{
+	if (const auto pTechno = this->ManualReloadEvent.Who.As_Techno())
+	{
+		if (pTechno->Ammo > 0 && pTechno->IsAlive && !pTechno->Berzerk)
```
TaranDahl, line 40: I think there should be more checking for availability, such deactive and chrono freeze.

TaranDahl (approved):
Overall, there are no major issues.
However, Ammo may not be a commonly used springboard. Perhaps detonating warheads on units or something similar would be more universal?

CrimRecya:
> However, Ammo may not be a commonly used springboard. Perhaps detonating warheads on units or something similar would be more universal?

Yes, so there is `CanManualReload.DetonateWarhead`.

Issue #92, opened 2021-03-09
Issue about MassSelectable


Cpj636:
The Ares feature [BuildingType] MassSelectable= (boolean) seems not work when using Phobos.

Thrifinesma:
You can use LowSelectionPriority instead for now, but it is not a perfect solution.

PR #1495, opened 2025-01-16
[Dehardcode] Type select for buildings

TaranDahl:
### Type select for buildings
- In vanilla game, type select can only be used on 1x1 buildings with `UndeploysInto`. Now it's possible to use it on all buildings if `BuildingTypeSelectable` set to true.

In `rulesmd.ini`
```ini
[General]
BuildingTypeSelectable=false  ; boolean
```


Speederovsky:
Type Select already works on buildings, just differently to prevent misclicks: you press T and then you select the building. 1x1 buildings with UndeploysInto are considered vehicles so they use their version of Type Select, where the order of pressing T doesn't matter. So what this is is allowing that for buildings, not enabling Type Select for buildings as a whole.

TaranDahl:
> Type Select already works on buildings, just differently to prevent misclicks: you press T and then you select the building. 1x1 buildings with UndeploysInto are considered vehicles so they use their version of Type Select, where the order of pressing T doesn't matter. So what this is is allowing that for buildings, not enabling Type Select for buildings as a whole.

I am more inclined to believe that WW originally intended to ban type selection on buildings, but did not completely ban it. The basis is that WW provides a type selection button in the bottom bar, but it is not valid for buildings. If button clicks and shortcut keys produce different effects, I would consider it a bug.
I will revise the description in the document later to clarify what specific changes were made in this PR.

Aephiex (approved):

TaranDahl:
目前的代码可能会在没有Ares的情况下工作异常。之后我会仔细检查这个问题。
The current code may work abnormally without Ares. I will carefully check this problem later.

tyuah8:
Ares already has a flag called `MassSelectable`, that when set to `yes` for a building, allows box-selection and type selection to work with that building.

[https://ares-developers.github.io/Ares-docs/new/buildings/massselect.html](https://ares-developers.github.io/Ares-docs/new/buildings/massselect.html)

TaranDahl:
> Ares already has a flag called `MassSelectable`, that when set to `yes` for a building, allows box-selection and type-selection to work with that building.
> 
> https://ares-developers.github.io/Ares-docs/new/buildings/massselect.html

This is a bit less labor-intensive than writing `MassSelectable` on every building and setting selection priorities. 
However, it is indeed not so necessary. If many people think it is useless, I will close it.

Aephiex:
This is still useful, because I don't want to box select buildings.

Aephiex:
Oh, I'm sorry. Seems this also make all buildings box selectable.

Metadorius:
Is #92 relevant to this? I am not sure

cc @Speederovsky

tyuah8:
This PR seems to modify the method at `0x465D40`, which is `BuildingTypeClass::IsVehicle()`; however, since this method is called in other parts of the software, that may not be related to selection, there will likely be side-effects with this implementation.

TaranDahl:
> This PR modifies the method at `0x465D40`, which is `BuildingTypeClass::IsVehicle()`; however, since this method is called in other parts of the game, that may not be related to selection, there will likely be side-effects with this implementation.

我设置那个全局变量就是为了处理这个问题。程序在执行到type select的时候会设置全局变量，只有在全局变量已被设置的情况下才会影响`BuildingTypeClass::IsVehicle()`的结果。不过好像它没有完全按我想要的方式工作。
I set that global variable just to handle this problem. The program will set the global variable when it executes to "type select". Only when the global variable has been set will it affect the result of `BuildingTypeClass::IsVehicle()`. However, it seems that it doesn't work completely in the way I want.

TaranDahl:
> Oh, I'm sorry. Seems this also make all buildings box selectable.

你确定吗，我测试没有这个问题
Are you certain?  I didn't encounter this problem during testing.

Aephiex:
Sorry, that seems to be an issue with my play branch. I recently reconstructed my play branch and there is no such issue with it.

CrimRecya (approved):

Issue #1499, opened 2025-01-18
Trajectory.Straight.ProximityMedial=yes sometimes detonating warheads at structure


mevitar:
### Description

Trajectory.Straight.ProximityMedial=yes is supposed to make the projectile's Trajectory.Straight.ProximityWarhead= detonate at the projectile every time its proximity impact is triggered. Works fine for vehicles and infantry, but when the proximity fuse is triggered by a building, the warhead will be detonated on the central point of the building instead of the projectile itself.

This seems to always happen for buildings larger than 2x2. For 2x2, the warhead might be detonated properly when the projectile does not get close to its central point (or maybe top cell, not sure). For 1x1 buildings, it appears to be working fine, detonating at the projectile.

### Phobos Version

https://github.com/Phobos-developers/Phobos/commit/47c7751cb8f541e0dbf73b40d11c20badc9fecfd

### INI code

```Trajectory=Straight
Trajectory.Straight.DetonationDistance=20
Trajectory.Straight.TargetSnapDistance=.4
Trajectory.Straight.PassThrough=yes
Trajectory.Straight.ProximityImpact=-1
Trajectory.Straight.ProximityWarhead=WARHEAD
Trajectory.Straight.ProximityDamage=120
Trajectory.Straight.ProximityRadius=1.5
Trajectory.Straight.ProximityMedial=yes
```

### Steps to reproduce

1. Use a Trajectory=Straight projectile with Trajectory.Straight.ProximityImpact=-1 and Trajectory.Straight.ProximityMedial=yes.
2. Fire the projectile in a way that it touches the building, but doesn't go through its center.
3. Watch where the projectile's Trajectory.Straight.ProximityWarhead= is detonated.

### Expected behaviour

The warhead should always detonate on the projectile.

### Actual behaviour

The warhead is detonated on the central point of the building.

### Additional context
I suppose that since Trajectory.Straight.ThroughBuilding= entry specifically says that it makes it detonate at structure's center if Trajectory.Straight.ProximityImpact= is in effect, there were no checks added there for Trajectory.Straight.ProximityMedial=yes, and that's why this issue happens.

### Checklist

- [x] The issue happens on the **latest official** version of Phobos and wasn't fixed yet.
- [x] I agree to elaborate the details if requested and provide thorough testing if the bugfix is implemented.
- [x] I added a very descriptive title to this issue.
- [x] I used the GitHub search and read the issue list to find a similar issue and didn't find it.
[...]

PR #1500, opened 2025-01-19
[Minor] Fix two issues of enhanced straight trajectory

CrimRecya:
- Fix when `Trajectory=Straight` bullet with `Trajectory.Straight.ProximityMedial=true` is attacking a large building, the warhead may snap to the center of the building. #1499 
- Fix when `Trajectory=Straight` bullet with a large `Trajectory.Straight.ProximityRadius` and a low `Trajectory.Speed`, the final speed may suddenly increase and rush out of range.

### DIFF
--- a/src/Ext/Bullet/Trajectories/StraightTrajectory.cpp
+++ b/src/Ext/Bullet/Trajectories/StraightTrajectory.cpp
@@ -619,7 +619,7 @@ void StraightTrajectory::BulletDetonateLastCheck(BulletClass* pBullet, HouseClas
 		pBullet->SetTarget(pDetonateAt);
 		pBullet->TargetCoords = position;
 
-		if (std::abs(velocity) > 1e-10)
+		if (std::abs(velocity) > 1e-10 && distance < velocity)
 			pBullet->Velocity *= distance / velocity;
 
 		if (this->ProximityImpact != 0)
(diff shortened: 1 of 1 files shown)


mevitar:
Trajectory.Straight.ProximityMedial=yes now fires properly at the projectile instead of the building.

And as i tested it i noticed that previous builds might have had issues with reading Bright= from the Trajectory.Straight.ProximityWarhead=, because right now i'm getting bright flashes of light each time the proximity fuse is triggered, and i don't remember them being there earlier.

Whether i simply didn't notice it or there was an actual bug that got fixed... it works fine now (in fact, other warhead effects, like AttachEffect, appear to work too).

CrimRecya:
> Trajectory.Straight.ProximityMedial=yes now fires properly at the projectile instead of the building.
> 
> And as i tested it i noticed that previous builds might have had issues with reading Bright= from the Trajectory.Straight.ProximityWarhead=, because right now i'm getting bright flashes of light each time the proximity fuse is triggered, and i don't remember them being there earlier.
> 
> Whether i simply didn't notice it or there was an actual bug that got fixed... it works fine now (in fact, other warhead effects, like AttachEffect, appear to work too).

I didn't make any changes to the flash. Maybe it was not noticed at that time? Anyway, I will merge this repair. Thank you for your test.

PR #1501, opened 2025-01-20
[Minor] Deactivated check & allow passenger deletion under EMP

Coronia:
For consistency: if `IsUnderEMP` is checked, then the techno's `Deactivated` should be checked in the same time. Added the remaining few cases where it's not being checked.

A logic that might be influence by this is `PassengerDeletion`, and idk why it even needs to check if the transport is EMP'd in the first time. Hence, I've added a tag to customize if passenger deletion should still be processed in such case.

In `rulesmd.ini`:
```ini
PassengerDeletion.UnderEMP=false                ; boolean
```

### DIFF
--- a/src/Ext/Techno/Body.Update.cpp
+++ b/src/Ext/Techno/Body.Update.cpp
@@ -207,10 +207,13 @@ void TechnoExt::ExtData::EatPassengers()
 	auto const pThis = this->OwnerObject();
 	auto const pTypeExt = this->TypeExtData;
 
-	if (!pTypeExt->PassengerDeletionType || !TechnoExt::IsActive(pThis))
+	if (!pTypeExt->PassengerDeletionType || !TechnoExt::IsActiveIgnoreEMP(pThis))
 		return;
 
 	auto pDelType = pTypeExt->PassengerDeletionType.get();
+
+	if (!pDelType->UnderEMP && (pThis->Deactivated || pThis->IsUnderEMP()))
+		return;
 
 	if (pTypeExt && (pDelType->Rate > 0 || pDelType->UseCostAsRate))
 	{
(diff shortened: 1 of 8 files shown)


src/Ext/Techno/Body.Update.cpp
```diff
@@ -207,11 +207,17 @@ void TechnoExt::ExtData::EatPassengers()
 	auto const pThis = this->OwnerObject();
 	auto const pTypeExt = this->TypeExtData;
 
-	if (!pTypeExt->PassengerDeletionType || !TechnoExt::IsActive(pThis))
+	if (!pTypeExt->PassengerDeletionType)
```
CrimRecya, line 210: I think using `if (!pTypeExt->PassengerDeletionType || !TechnoExt::IsActiveIgnoreEMP(pThis))` here is sufficient to avoid duplicate judgments.

src/Ext/Techno/Body.Update.cpp
```diff
@@ -207,11 +207,17 @@ void TechnoExt::ExtData::EatPassengers()
...
 		return;
 
 	auto pDelType = pTypeExt->PassengerDeletionType.get();
 
+	if (!TechnoExt::IsActive(pThis))
+	{
+		if (!pDelType->UnderEMP || !TechnoExt::IsActiveIgnoreEMP(pThis))
+			return;
+	}
```
CrimRecya, line 219: In conjunction with the one above, `if (!pDelType->UnderEMP && (pThis->Deactivated || pThis->IsUnderEMP()))` can be directly used here to simplify the process.

CrimRecya (commented):
It looks good, but there is still one part that can be simplified.

CrimRecya (approved):

CrimRecya:
LGTM. But considering the newly added features, perhaps I should hear the original author's opinion.

PR #1502, opened 2025-01-20
[Minor] Use SetAnimOwnerHouseKind to set owner for anim

Coronia:
Currently, many anims are setting their ownership with `pAnim->Owner=` method, which actually misses several check and set logic of Phobos (mostly related to create unit). This is another consistency fix to make all of these cases use the proper function to set anim's ownership.


Starkku:
Shield and AE anims will never actually execute `MakeInfantry`/`CreateUnit` code because they play for infinite iterations, so for them this is superfluous even though it would appear that shield animation had the call prior to this PR even.

Coronia:
> Shield and AE anims will never actually execute `MakeInfantry`/`CreateUnit` code because they play for infinite iterations regardless of explicit `LoopCount` setting, so for them this is superfluous even though it would appear that shield animation had the call prior to this PR even.

The one I've added in ShieldClass is for `BreakAnim`, which isn't its attached looped anim, think it should be legit. Can revert this setting for AE and shield's IdleAnim if needed

Coronia:
added `SetInvoker` to the following cases as well
- deploy/undeploy anims
- promotion anim
- shield's hit/break/idle anims
- passenger deletion anim
- AE anim when `Animation.UseInvokerAsOwner` is set to false, which will set Invoker to the attached techno

btw, I've noticed that shield's `BreakAnim` is attached to the shielded object, while `HitAnim` doesn't. Should they follow the same scheme here?

src/New/Entity/AttachEffectClass.cpp
```diff
@@ -308,16 +308,15 @@ void AttachEffectClass::CreateAnim()
 		auto const pAnim = GameCreate<AnimClass>(pAnimType, this->Techno->Location);
 
 		pAnim->SetOwnerObject(this->Techno);
-		pAnim->Owner = this->Type->Animation_UseInvokerAsOwner ? InvokerHouse : this->Techno->Owner;
+		auto const pOwner = this->Type->Animation_UseInvokerAsOwner ? InvokerHouse : this->Techno->Owner;
```
CrimRecya, line 311:
```suggestion
		auto const pOwner = this->Type->Animation_UseInvokerAsOwner ? this->InvokerHouse : this->Techno->Owner;
```

src/New/Entity/AttachEffectClass.cpp
```diff
@@ -308,16 +308,15 @@ void AttachEffectClass::CreateAnim()
...
 		pAnim->RemainingIterations = 0xFFu;
 		this->Animation = pAnim;
 
 		if (this->Type->Animation_UseInvokerAsOwner)
-		{
-			auto const pAnimExt = AnimExt::ExtMap.Find(pAnim);
-			pAnimExt->SetInvoker(Invoker);
-		}
-
+			AnimExt::ExtMap.Find(pAnim)->SetInvoker(Invoker);
```
CrimRecya, line 317:
```suggestion
			AnimExt::ExtMap.Find(pAnim)->SetInvoker(this->Invoker, this->InvokerHouse);
```

CrimRecya (commented):
Seems like there is a minor issue.

CrimRecya (approved):

PR #1504, opened 2025-01-21
[Highly Customized] Allow merging AOE damage to buildings into one

CrimRecya:
- Warheads are now able to damage building only once by merging the AOE damage when setting `MergeBuildingDamage` to true, which default to `[CombatDamage]->MergeBuildingDamage`.

In `rulesmd.ini`:
```ini
[CombatDamage]
MergeBuildingDamage=false    ; boolean

[SOMEWARHEAD]                ; Warhead
MergeBuildingDamage=         ; boolean
```

```{note}
- This is different from `CellSpread.MaxAffect`.
- Due to the rounding of damage, there may be a slight increase in damage.
```

### DIFF
--- a/src/Ext/WarheadType/Hooks.cpp
+++ b/src/Ext/WarheadType/Hooks.cpp
@@ -262,6 +262,73 @@ DEFINE_HOOK(0x489B49, MapClass_DamageArea_Rocker, 0xA)
 	return 0x489B53;
 }
 
+#pragma region MergeBuildingDamage
+
+namespace DamageBuildingHelper
+{
+	std::map<BuildingClass*, double> Buildings;
+}
+
+DEFINE_HOOK(0x4899DA, DamageArea_DamageBuilding_SetContext, 0x7)
+{
+	GET_BASE(WarheadTypeClass* const, pWH, 0xC);
+
+	if (!WarheadTypeExt::ExtMap.Find(pWH)->MergeBuildingDamage.Get(RulesExt::Global()->MergeBuildingDamage))
+		return 0;
+
+	DamageBuildingHelper::Buildings.clear();
+
+	struct DamageGroup
+	{
+		ObjectClass* Target;
+		int Distance;
+	};
+
+	GET_STACK(DynamicVectorClass<DamageGroup*>, groups, STACK_OFFSET(0xE0, -0xA8));
+
+	for (auto& group : groups)
+	{
+		if (const auto pBuilding = abstract_cast<BuildingClass*>(group->Target))
+		{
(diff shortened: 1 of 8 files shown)


MortonPL:
In what cases would you use it instead of high MaxAffect?

Aephiex:
> In what cases would you use it instead of high MaxAffect?

There are 2 possible cases:
1. A building bears a similar ability of the SC2 Immortal will fail to resist the damage if it was muliple instances;
2. A building with SC2 style armor that each instance of the damage is substracted by a number will count multiple times if the damage was multiple instances.

There is no apparent difference for this PR alone, as far as I see.

CrimRecya:
> In what cases would you use it instead of high MaxAffect?

As Aephiex said, this is not related to high MaxAffect.
Here are a few simple applicable examples:
- Display of damage
- Limit maximum/minimum single damage
- Evasion probability or blocking probability of damage

src/Ext/WarheadType/Hooks.cpp
```diff
@@ -262,6 +262,79 @@ DEFINE_HOOK(0x489B49, MapClass_DamageArea_Rocker, 0xA)
...
+
+	GET_STACK(const DynamicVectorClass<DamageGroup*>, groups, STACK_OFFSET(0xE0, -0xA8));
+	GET_STACK(const bool, invincibleWithoutPenetrateAndCloseTo, STACK_OFFSET(0xE0, -0xC9));
+	GET_STACK(const int, baseDamage, STACK_OFFSET(0xE0, -0xBC));
+	GET_BASE(TechnoClass* const, pAttacker, 0x08);
+	GET_BASE(HouseClass* const, pAttackHouse, 0x14);
+
+	// Because during the process of causing damage, fragments may be generated that need to continue causing damage, resulting in nested calls
+	// to this function. Therefore, a single global variable cannot be used to store this data.
+	PhobosMap<BuildingClass*, double> MapBuildings;
```
Thrifinesma, line 288: I think std::map is more suggested. Since PhobosMap is only a std::vector<std::pair>, not really a map

CrimRecya: I mainly changed to using `PhobosMap` here for performance reasons, because I don't need a lot of `std::map` features, I just need to record these buildings.

Thrifinesma: That's right, I meant performance too. `std::map` or `std::unordered_map` is more suitable here. `PhobosMap` is just a vector.

CrimRecya:
What needs to be discussed about this?

PR #1508, opened 2025-01-25
[Minor] Fix building production queue issue

CrimRecya:
Fix an issue: When the defensive buildings in the queue lose their prerequisites, they will not be cancelled but will remain stuck forever. (This issue does not occur when using #1477 )
The problem still comes from hardcoded `BuildCat::DontCare`.

### DIFF
--- a/src/Ext/Sidebar/Body.cpp
+++ b/src/Ext/Sidebar/Body.cpp
@@ -18,6 +18,8 @@ void SidebarExt::Remove(SidebarClass* pThis)
 	Data = nullptr;
 }
 
+// Reversed from Ares source code (In fact, it's the same as Vanilla).
+// And compared to 0.A, it has been encapsulated. That's why here's such a simple way to make modifications.
 bool __stdcall SidebarExt::AresTabCameo_RemoveCameo(BuildType* pItem)
 {
 	const auto pTechnoType = TechnoTypeClass::GetByTypeAndIndex(pItem->ItemType, pItem->ItemIndex);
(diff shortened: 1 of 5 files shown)


CrimRecya:
Use Ares3.0p1 for testing.
Need someone to help me try using Ares 3.0 and without Ares.

CrimRecya:
After testing, it seems to work fine with Ares3.0 or Ares3.0p1. But it seems that starting without Ares is no longer supported since b27? So no test if without Ares here.

Metadorius:
> After testing, it seems to work fine with Ares3.0 or Ares3.0p1. But it seems that starting without Ares is no longer supported since b27? So no test if without Ares here.

I think that's a bug that just wasn't caught yet because almost everyone uses Ares.

CrimRecya:
> > After testing, it seems to work fine with Ares3.0 or Ares3.0p1. But it seems that starting without Ares is no longer supported since b27? So no test if without Ares here.
> 
> I think that's a bug that just wasn't caught yet because almost everyone uses Ares.

True.

src/Ext/Sidebar/Body.cpp
```diff
@@ -14,6 +18,56 @@ void SidebarExt::Remove(SidebarClass* pThis)
 	Data = nullptr;
 }
 
+bool __stdcall SidebarExt::AresTabCameo_RemoveCameo(BuildType* pItem)
```
Metadorius, line 21: Is this function a rewrite of some publicly available Ares function, was it reversed from Ares source code, or is this some custom code? A comment in code here would be good.

Metadorius (approved):
Let's add a comment explaining the purpose and what it solves, otherwise this is fine to merge.

src/Ext/Sidebar/Body.cpp
CrimRecya: Both. I'll add a comment.

PR #1510, opened 2025-01-27
[Vanilla Fix] Two AI anger related bug-fix

TaranDahl:
- Fixed an issue that the AI would set anger towards friendly houses, causing it to act stupidly.
- Fixed an issue that the AI would look for the first house in the array as an enemy instead of the nearest one when there were no enemies.


Aephiex (approved):

DeathFishAtEase:
Regarding AI anger-related, perhaps you can obtain some information by reading [this post](https://bbs.ra2diy.com/forum.php?mod=viewthread&tid=23550) and trying to ask handama.

Metadorius:
Is there a point in a toggle? I don't see a scenario when the bugged behavior is preferred.

DeathFishAtEase:
If extensive and diverse testing consistently reveals no side effects, perhaps a manually controlled switch is not necessary.

TaranDahl:
> If extensive and diverse testing consistently reveals no side effects, perhaps a manually controlled switch is not necessary. Perhaps it's just because the internal mechanisms aren't fully understood yet, so it can't be certain that there are no problems at present? Theoretically, features like `[AI]►Paranoid= (boolean)` have not worked since before YR. Have you tried declaring war on the AI during an online game to change the AI's list of allies and see if there are any abnormalities in the AI? Can it still normally handle allies who become enemies during the game?-----------谷歌翻译-----------如果广泛且多样化的测试始终显示没有副作用，则也许不需要手动控制开关。或许只是因为内部机制还没有完全搞清楚，所以目前还不能确定没有问题？理论上，像 [AI]►Paranoid=（布尔值）这样的功能自去年以来就不起作用了。你是否尝试过在网络游戏中向AI宣战，改变AI的盟友列表，看看AI是否有异常？它还能正常处理游戏中变成敌人的盟友吗？

May tested them later

FS-21:
Isn't much better if instead of this PR by default ALL houses anger values are initialized as 1 when the game is loaded? AI will ignore friendly anger values and since they have enemies with anger value > 0 the AI will work as expected...

TaranDahl:
> Perhaps it's just because the internal mechanisms aren't fully understood yet, so it can't be certain that there are no problems at present?

Indeed. In fact, I found the hook position by intuition.

MortonPL:
If it turns out to not a be a problem or that mapmakers are willing to account for this, we can drop the toggle.

TaranDahl:
> If it turns out to not a be a problem or that mapmakers are willing to account for this, we can drop the toggle.

I didn't do completely reverse engineering, so I lack confidence in whether there are side effects. It is better to retain the toggle.

Starkku:
Code looks fine, quick test shows it does atleast fix the obvious aggression issue with ally AI's before they gain anger towards enemies. Fix merge conflicts and it is good to go I'd say.

PR #1514, opened 2025-01-28
Add powered infantry/vehicle/aircraft logic

MortonPL:
### Power drain for units

- Infantry, vehicles and aircraft can now drain or provide `Power` if `UnitPowerDrain=true` is set.

In `rulesmd.ini`:
```ini
[General]
UnitPowerDrain=false                 ; boolean

[SOMETECHNO] ; TechnoType
Power=0                              ; integer, positive means output, negative means drain
```

image

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -271,8 +271,9 @@ This page lists all the individual contributions to the project by their author.
   - Show designator & inhibitor range
   - Dump variables to file on scenario end / hotkey
   - "House owns TechnoType" and "House doesn't own TechnoType" trigger events
-  - Help with docs
   - Voxel light source position customization
+  - Extending `Power` to all TechnoTypes
+  - Help with docs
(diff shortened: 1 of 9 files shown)


FS-21:
If the game isn't prepared to this change can you add support for powered=yes in all affected technos that aren't buildings?
So in case of low power these objects will be disabled like the robot tanks case in the original game.

Aephiex (approved):

Aephiex:
Units at low power suffer from no penalties. Besides I don't see anything apparently wrong.

Suggestions: since this makes the house recheck power every time they gain or lose a techno, even when this feature is not in use, there should be a global toggle for this, what do you think?

mevitar:
The check probably won't happen that often to affect performance, but still, i think it's better to make it something that has to be enabled first. Per-house or a global toggle, whatever is easier.

Metadorius (approved):

PR #1515, opened 2025-01-30
[Uhh...Minor Stuff?] Infantry can play die sound when crushed

Aephiex:
With this configuration, infantry will play the die sound when crushed.

```ini
[AudioVisual]
InfantryPlayDieSoundWhenCrushed=true
```


mevitar:
Why was this closed?
Appears to be working fine, although the tag looks too long.
Perhaps changing it to "PlayDieSoundOnCrushed=" and making it work for crushing other technos would be better?

PR #1516, opened 2025-02-02
NoSecondaryWeaponFallback can now cancel the weapon selection for walls.

DeathFishAtEase:
Phobos previously added a rule to select weapons based on `[Warhead] ► Wall= (bool)`
but `[TechnoType] ► NoSecondaryWeaponFallback= (bool)` cannot disable this rule
this causing the Soviet Siege Chopper to use the `160mm` weapon to attack walls even when not deployed.


Coronia (approved):

Starkku (approved):
This fix was not sufficient or at correct position. `TechnoExt::GetWeaponIndexAgainstWall()` already had a check for `NoSecondaryWeaponFallback` but due to wrongly positioned brackets it didn't work as intended. 

I have took a liberty of committing changes that address this as well as moving the entire weapon selection against wall logic behind a toggle to prevent issues from arising with existing INI configurations, with modders having to manually enable the logic if they wish to use it and documentation mentioning this appropriately.

PR #1518, opened 2025-02-05
Warhead that can not kill

FS-21:
- Warheads can now damage the enemy without killing them (minimum health will be 1).
- I tested this tag against conventional weapons and radiation types.

In `rulesmd.ini`:
```
[SOMEWARHEAD]  ; Warhead
CanKill=true; boolean
```

- `Unkillable` can be used to prevent the techno from being killed by taken damage (minimum health will be 1).
In `rulesmd.ini`:
```
Unkillable=false                                   ; boolean
```


Coronia (commented):
idk if setting the target's health to 1 is the correct way, since there's other logic that rely damage greater than 0 to work. Perhaps it would be safer to set the damage to current health - 1.

Also those check warning issues should be fixed as well.

FS-21:
Since this code was born from a split from another feature I forgot to delete the WH declaration of the posterior if-else :-D thanks!

And I'm against to set -1 since all the games's logic understand positive values but I don't know how are the checks with negative ones. Moreover lots of Phobos checks could break because they are against health=0 and not negative health.
The only reason of the feature is that in case of having a weapon that can kill the object it just leave only 1 point of life, anything else will be able to kill it.

Useful for debuff/plagues, auxiliary logics like the PR related to #1398 and both together fulfills the closed #627, etc.

FS-21:
> You misunderstood. @Coronia meant that instead of setting health to 1 it's better to modify the damage before actually applying anything to deal `current health - 1` amount of damage, this way it won't be possible for health to go lower than 1.

I see, sorry I though wrongly it was that different approach when I readed the message :-P
Anyways (unless a skilled one enlight me) you can't simply reduce the damage value (integer) until it matches health 1 because the armour (double) calculation not done here and probably other multipliers.
Try to get a value to find the health= 1 if the weapon does 125 and uses a warhead with a 12500% against that target's armour.

Isn't simplier if the final calculation surpasses the target's health set that minimum health and skip the checks for ending it?

FS-21:
> if you return that early , isnt that mean you will skip some of ares affect then , is it intended behavoiur ?.

I changed that behaviour so now it doesn't skip code. I don't see any side effects.

Coronia (approved):
tested, can work properly with the RadSite codes being removed, only ReceiveDamage would be enough

PR #1520, opened 2025-02-07
[Dehardcode?] Cache click if the event buffer is full

TaranDahl:
- In vanilla, players can only issue commands to up to 128 technos at a time. This is due to the fact that unit execution requires sending events, and the event buffer length is 128.
- Now you can use the following flag to make the unit cache received instructions if the event buffer is full and send them again when the buffer is empty.
  - If the unit tries to cache instructions when it has already cached another, then the last one will be overwritten. But usually you don't need to worry about this problem.
  - Waypoint mode is not currently supported. This might be changed in the future.


TaranDahl:
> Can't you increase the buffer size instead?-----------谷歌翻译-----------您不能增加缓冲尺寸吗？

i simply dunno how to do that XD

Metadorius:
> i simply dunno how to do that XD

There were already similar expansions, mainly made by @secsome, you can see how he did them or maybe even ask him directly. Some are not merged.

Generally we prefer proper fixes unless there are some considerable obstacles.

TaranDahl:
> > i simply dunno how to do that XD
> 
> There were already similar expansions, mainly made by @secsome, you can see how he did them or maybe even ask him directly. Some are not merged.-----------谷歌翻译-----------已经有类似的扩展，主要由@secsome进行，您可以看到他是如何做的，甚至直接问他。有些没有合并。
> 
> Generally we prefer proper fixes unless there are some considerable obstacles.-----------谷歌翻译-----------通常，除非存在相当大的障碍，否则我们更喜欢适当的修复。

Got it.

PR #1521, opened 2025-02-11
[Highly Customized] Jumpjet Tilts While Moving

CrimRecya:
- Now you can make jumpjets tilt forward when moving forward and sideways when turning by setting `JumpjetTilt` to true.
- The maximum tilt angle will not exceed 45 degrees.
  - The magnitude of the forward tilt is related to the current speed and acceleration. They are additive and have two coefficients that can be adjusted for details.
  - The magnitude of the sideways tilt is related to the current speed and rotation angle. They are multiplied and also have two coefficients that can be adjusted for details.

In `rulesmd.ini`:
```ini
[SOMEUNIT]                              ; UnitType, with Locomotor=Jumpjet
JumpjetTilt=false                       ; boolean
JumpjetTilt.ForwardAccelFactor=1.0      ; floating point value
JumpjetTilt.ForwardSpeedFactor=1.0      ; floating point value
JumpjetTilt.SidewaysRotationFactor=1.0  ; floating point value
JumpjetTilt.SidewaysSpeedFactor=1.0     ; floating point value
```

jjtilt


src/Ext/TechnoType/Hooks.MatrixOp.cpp
```diff
@@ -177,6 +178,33 @@ Matrix3D* __stdcall JumpjetLocomotionClass_Draw_Matrix(ILocomotion* iloco, Matri
...
+			arf += static_cast<float>(std::min(32.0, pThis->CurrentSpeed) * factor);
+		}
+
+		const auto& locoFace = pThis->LocomotionFacing;
+
+		if (locoFace.IsRotating())
+		{
+			constexpr auto factor = (Math::HalfPi / 4) / 32768 / 65536;
+			const auto leftRaw = locoFace.RotationTimer.GetTimeLeft() * locoFace.ROT.Raw;
+			const auto dirMult = (static_cast<short>(locoFace.Difference().Raw) * leftRaw);
```
Metadorius, line 196: It would be good to have an explanation in a comment about what is the calculation logic. Also magic numbers should be fixed.

src/Ext/TechnoType/Hooks.MatrixOp.cpp
```diff
@@ -177,6 +178,33 @@ Matrix3D* __stdcall JumpjetLocomotionClass_Draw_Matrix(ILocomotion* iloco, Matri
 			ret->RotateY(arf);
 		}
 	}
+	else if (TechnoTypeExt::ExtMap.Find(linked->GetTechnoType())->JumpjetTilt.Get(RulesExt::Global()->JumpjetTiltWhenMoving)
+		&& !onGround && linked->IsAlive && linked->Health > 0 && !linked->IsAttackedByLocomotor)
+	{
+		if (pThis->CurrentSpeed > 0.0)
+		{
+			constexpr auto factor = (Math::HalfPi / 4) / 32;
+			arf += static_cast<float>(std::min(32.0, pThis->CurrentSpeed) * factor);
```
Metadorius, line 187: I suggest limiting the final rotation instead, so it doesn't exceed 90°.

docs/New-or-Enhanced-Logics.md
```diff
@@ -1585,6 +1585,19 @@ Convert.HumanToComputer =   ; TechnoType
...
+- Now you can make jumpjet tilts forward when moving forward and tilts sideward when turning by set `JumpjetTilt` to true. Default to `[AudioVisual]->JumpjetTiltWhenMoving`.
+
+In `rulesmd.ini`:
+```ini
+[AudioVisual]
+JumpjetTiltWhenMoving=false   ; boolean
+
+[SOMEUNIT]                    ; UnitType, with Locomotor=Jumpjet
+JumpjetTilt=                  ; boolean
+```
```
Metadorius, line 1599:
````suggestion
- Now you can make jumpjets tilt forward when moving forward and sideways when turning by setting `JumpjetTilt` to true.

In `rulesmd.ini`:
```ini
[AudioVisual]
JumpjetTiltWhenMoving=false   ; boolean

[SOMEUNIT]                    ; UnitType, with Locomotor=Jumpjet
JumpjetTilt=                  ; boolean, defaults to [AudioVisual]->JumpjetTiltWhenMoving
```
````

**Review by Metadorius — CHANGES_REQUESTED** [review]
Looks good, few style and readability issues.

I have a few suggestions though:
1. Ability to control how much the jumpjet tilts, I imagine via separate factors for forward and sideway tilts.
2. Ability to account acceleration for directional tilt (when accelerating - tilt forwards, decelerating - tilt backwards to decelerate). This could work additionally to movement-dependant tilt.
3. Debatable: ability to account speed for rotational tilt, because static helicopter rotating via tilt looks weird, and to do sharp turns at high speed you must tilt a lot. But maybe could just remove rotation sideways tilt when not moving. 

```ini
JumpjetTilt.ForwardAccelFactor  ; FAF 
JumpjetTilt.ForwardSpeedFactor  ; FSF
JumpjetTilt.SidewaysRotationFactor  ; SRF
JumpjetTilt.SidewaysSpeedFactor  ; SSF
```

Pseudocode to show what I mean:
```cpp
arf += movement_acceleration * FAF + movement_speed * FSF
ars += rotation_speed * SRF * movement_speed * SSF
```

(By the way, I am not sure I understand the current sideways rotation logic correctly. What does it depend on?)

Coronia:
If a tilting jumpjet is affected by EMP, it'll gradually resume to normal stance while being EMP'd. Not sure if it should behave like that

Coronia (approved):
maybe there's no need to set it globally since you'll still have to customize those factors for each jumpjet. Otherwise LGTM

TaranDahl:
How about move `JumpjetTilt=` to `[JumpjetControls]` 🤔

CrimRecya:
I imagined that perhaps there is no need for global label?
It seems that only helicopters require this function, while spacecraft may only need to tilt sideways, but this definitely requires customization. Other UFO or balloons/airships may not need this.

Metadorius (approved):
I can imagine some quick/nimble helicopters tilting up to 90 degrees, so it would be good to have that as the limitation instead of 45. I doubt it will hamper usability, so why not?

Otherwise I have no complaints regarding this feature. Maybe it's worth adding a GIF to the docs? I suggest to use [GifCam](https://blog.bahraniapps.com/gifcam/#download), as it doesn't make the GIF lose the quality.

CrimRecya:
> I can imagine some quick/nimble helicopters tilting up to 90 degrees, so it would be good to have that as the limitation instead of 45. I doubt it will hamper usability, so why not?

That probably requires modifying the shadow matrix to make it more effective. I will add it later.

Metadorius:
> That probably also requires modifying the shadow matrix to make it more effective. I will add it later.

Oh, that's a fair comment, I didn't think of that. Feel free to merge as is then, the 90 degree adjustment can be made later. The shadow drawing doesn't make a lot of sense in the vanilla game so it needs a good rewrite.

PR #1522, opened 2025-02-11
[Customized] Change the scrolling action of the sidebar

CrimRecya:
- Allow players to decide for themselves:
  - Whether can use mouse wheel to scroll sidebar strip when the mouse is not on it by `ScrollSidebarStripInTactical` .
  - Whether can use mouse wheel to scroll sidebar strip when pressing Ctrl by `ScrollSidebarStripWhenHoldCtrl`, Alt by `ScrollSidebarStripWhenHoldAlt`, or Shift by `ScrollSidebarStripWhenHoldShift`.

In `RA2MD.ini`:

```ini
[Phobos]
ScrollSidebarStripInTactical=true     ; boolean
ScrollSidebarStripWhenHoldAlt=true    ; boolean
ScrollSidebarStripWhenHoldCtrl=true   ; boolean
ScrollSidebarStripWhenHoldShift=true  ; boolean
```

I also prepared a trigger hook to catch the state when mouse wheel scrolling.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -386,6 +386,7 @@ This page lists all the individual contributions to the project by their author.
   - Fix for sidebar not updating queued unit numbers when on hold
   - New Parabola trajectory
   - Enhanced Bombard trajectory
+  - Change the scrolling action of the sidebar and trigger hook of mouse wheel scroll
 - **Ollerus**
   - Build limit group enhancement
   - Customizable rocker amplitude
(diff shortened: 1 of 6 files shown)


Thrifinesma:
The feature tagged customized, but when `ScrollSidebarStripWhenHoldKey`, all three keys are avaliable. I suggest to change it to allow specify which keys (one or two or three) is usable to trigger scrolling action.

CrimRecya:
What needs to be discussed about this?

PR #1524, opened 2025-02-12
[Minor] Disallow other actions for building in planning mode

TaranDahl:

### DIFF
--- a/src/Ext/Building/Hooks.cpp
+++ b/src/Ext/Building/Hooks.cpp
@@ -719,13 +719,14 @@ DEFINE_HOOK(0x4AE95E, DisplayClass_sub_4AE750_DisallowBuildingNonAttackPlanning,
 	enum { SkipGameCode = 0x4AE982 };
 
 	GET(ObjectClass* const, pObject, ECX);
-	GET_STACK(CellStruct, cell, STACK_OFFSET(0x20, 0x8));
+	LEA_STACK(CellStruct*, pCell, STACK_OFFSET(0x20, 0x8));
 
-	auto action = pObject->MouseOverCell(&cell);
+	auto action = pObject->MouseOverCell(pCell);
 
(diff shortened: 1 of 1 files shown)


Metadorius:
Works good, can you also handle showing the no action map cursor instead of move/no move cursor?

Starkku:
I would use a more descriptive name for the hook if possible.

src/Ext/Building/Hooks.cpp
```diff
@@ -713,4 +714,22 @@ bool __fastcall BuildingTypeClass_CanUseWaypoint(BuildingTypeClass* pThis)
...
+	auto action = pObject->MouseOverCell(pCell);
+
+	bool shouldSkip = PlanningNodeClass::PlanningModeActive && pObject->WhatAmI() == AbstractType::Building && action != Action::Attack;
+
+	if (!shouldSkip)
+		pObject->CellClickedAction(action, pCell, pCell, false);
+
+	return Ret;
+}
+
```
CrimRecya, line 734:
```suggestion
DEFINE_HOOK(0x4AE95E, DisplayClass_sub_4AE750_DisallowBuildingNonAttackPlanning, 0x5)
{
	enum { SkipGameCode = 0x4AE982 };

	GET(ObjectClass* const, pObject, ECX);
	GET_STACK(CellStruct, cell, STACK_OFFSET(0x20, 0x8));

	auto action = pObject->MouseOverCell(&cell);

	if (!PlanningNodeClass::PlanningModeActive || pObject->WhatAmI() != AbstractType::Building || action == Action::Attack)
		pObject->CellClickedAction(action, &cell, &cell, false);

	return SkipGameCode;
}
```

CrimRecya (approved):
You can directly Get_Stack.

Metadorius:
Look into REF_STACK and LEA_STACK. They should be more suitable here.

CrimRecya:
> Look into REF_STACK and LEA_STACK. They should be more suitable here.

Yes, LEA_STACK should be the most direct, but considering that neither of these functions actually changes the cell struct, both can be used.

PR #1526, opened 2025-02-12
Better theme for docs and other doc improvements

Metadorius:


DeathFishAtEase:
Currently, this causes the generator of `CustomGS` to malfunction.

Metadorius:
Closing because most of the changes are not strictly better and cause issues over the doc. Furo theme is ill-suited for big docs like ours, copybutton breaks `CustomGS` target code block, tabs seem useless for our cause, and Sphinx Design which adds dropdowns/spoilers requires additional styling in order to be usable at our custom theme. Maybe opengraph plugin could still be added or some custom CSS could be added for Sphinx Design elements but so far it doesn't seem worthwile.

Issue #1528, opened 2025-02-13
[BUG]Problems With Customizable Veterancy Insignias


ahsy666:
### Description

I think this function is the source of the crash.

[FV];;;;;;;At first, I did not use that function.
TurretCount=4
WeaponCount=18;;;;;;;This is original YR's FV, only add a new weapon for Boris.
Weapon18=CRAKM
EliteWeapon18=CRAKM

WeaponTurretIndex18=1

[BORIS]
IFVMode=17

------
The code is OK, never crash in 0.3.0.1. But when I use B44(maybe early builds with this function could have this problem too), crashes happens.
First, Boris take his ride, then crash. After that, I use GGI take a ride first, then change to Boris. By this way game may not crash, until FV kills enemies then get promoted. 
And what I saw is, the insignia of FV gone when Boris take a ride. So I think that function MAY cause this problem, then I search codes and have another try. 

------
[FV]
TurretCount=4
WeaponCount=18
Weapon18=CRAKM
EliteWeapon18=CRAKM

WeaponTurretIndex18=1

Insignia.Weapon18=pips;;;;;;;;Tried pips.shp
InsigniaFrames.Weapon18=-1,-1,-1

;InsigniaFrame.Weapon18.Rookie=-1;;;;;;;another failure
;InsigniaFrame.Weapon18.Veteran=14
;InsigniaFrame.Weapon18.Elite=15

[BORIS]
IFVMode=17

[E1];only change IFVMode
IFVMode=17

[ADOG];only change IFVMode
IFVMode=17

------

Of course failed, so I post them here.

------
2025.2.14:I put codes into .map directly. I didn't modify rulesmd.ini.

------
2025.2.14:After another test, I think real reason is B44 can't combine INIs correctly. I created a new FV. Codes above worked and never crush.

[FV2]
......;;;;;;;;;other codes copy from [FV]

WeaponCount=18
Weapon18=CRAKM
EliteWeapon18=CRAKM
WeaponTurretIndex18=1

Insignia.Weapon18=pip-super
InsigniaFrames.Weapon18=0,1,2

-------
Now it works normally.

### Phobos Version

B44 and B45

### Conditions to reproduce

Anybody take a ride with IFVMode≥17 could crash, especially when FV take down something then get promoted.

### INI code

```ini
Codes are in description above.
2025.2.14:I put codes into .map directly. I didn't modify rulesmd.ini.

```

### Steps to reproduce

1.
2.
3.
...

[...]

PR #1530, opened 2025-02-14
[Minor] Fix Insignia.Weapon parsing

Coronia:
Separate the fix from https://github.com/Phobos-developers/Phobos/pull/1365 to a standalone pull request, which should fix the issue in https://github.com/Phobos-developers/Phobos/issues/1528.
- previous Insignia(.Frame/.Frames).WeaponN settings can't be read or overridden from maps, which is fixed now

### DIFF
--- a/src/Ext/Techno/Body.Visuals.cpp
+++ b/src/Ext/Techno/Body.Visuals.cpp
@@ -152,8 +152,8 @@ void TechnoExt::DrawInsignia(TechnoClass* pThis, Point2D* pLocation, RectangleSt
 
 		auto& frames = pTechnoTypeExt->InsigniaFrames_Weapon[weaponIndex];
 
-		if (frames != Vector3D<int>(-1, -1, -1))
-			insigniaFrames = frames;
+		if (frames.isset() && frames.Get() != Vector3D<int>(-1, -1, -1))
+			insigniaFrames = frames.Get();
 	}
 
 	if (pVeterancy->IsVeteran())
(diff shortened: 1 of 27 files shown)


src/Ext/Techno/Body.Visuals.cpp
```diff
@@ -152,8 +152,8 @@ void TechnoExt::DrawInsignia(TechnoClass* pThis, Point2D* pLocation, RectangleSt
 
 		auto& frames = pTechnoTypeExt->InsigniaFrames_Weapon[weaponIndex];
 
-		if (frames != Vector3D<int>(-1, -1, -1))
-			insigniaFrames = frames;
+		if (frames.isset() && frames.Get() != Vector3D<int>(-1, -1, -1))
```
Thrifinesma, line 155: Not sure if this will be better
```suggestion
		if (frames.Get(Vector3D<int>(-1, -1, -1)) != Vector3D<int>(-1, -1, -1))
```

src/Ext/TechnoType/Body.cpp
```diff
@@ -487,30 +487,27 @@ void TechnoTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 
 	char tempBuffer[32];
```
Thrifinesma, line 488: Just curious why there is a new buffer, why not use Phobos::readBuffer?

CrimRecya (approved):

PR #1531, opened 2025-02-14
[Minor] Make quick save happens after main loop

TaranDahl:
According to what ZivDero told me, saving game inside the main loop may cause corruption. So we need to move it out of the main loop.
This needs to be used together with the unmerged yrpp-spawner feature to work properly.
It needs to be discussed about how we should interact with spawner. Maybe a flag like `CanUseSpawner` or sth alike.
Open a draft PR to keep this in mind.

### DIFF
--- a/src/Phobos.INI.cpp
+++ b/src/Phobos.INI.cpp
@@ -5,6 +5,7 @@
 #include <SessionClass.h>
 #include <MessageListClass.h>
 #include <HouseClass.h>
+#include <EventClass.h>
 
 #include <Utilities/Parser.h>
 #include <Utilities/GeneralUtils.h>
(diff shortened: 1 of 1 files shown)


Metadorius:
Is there a reason this is a draft?

TaranDahl:
> Is there a reason this is a draft?

Currently, this one can not work correctly without the unmerged `AutoSave` feature of the spawner. Because the event handling of `SaveGame` in vanilla is somehow broken. As the spawner rewrote the event handling, I'm not sure what we are going to do with this one. Maybe we need a `CanUseSpawner` flag to check if the spawner is present, but I dunno how to add it for now.

Metadorius:
@TaranDahl We have two ways of dealing with that:
1. Copy the relevant parts from the spawner here so that if any of the spawner and Phobos are present - it works.
2. Move the quick save command to the spawner.

What do you and others think?

TaranDahl:
Will be impl in https://github.com/Phobos-developers/Phobos/pull/1790

PR #1533, opened 2025-02-14
Fix inconsistencies with Strafing.Shots logic

Starkku:
- Fixes an issue where last shot of strafing aircraft weapon with `Strafing.Shots` < 5 or if interrupted by ammo running out if `Strafing.UseAmmoPerShot=true` would not be followed by a similar delay before resuming strafing run or returning to base as there would otherwise be. The delay is also now explicitly customizable although I urge caution in using it in certain cases.

- `Strafing.EndDelay` can be used to override the delay after firing last shot in strafing run before aircraft resumes another strafing run or returns to base. Defaults to (Weapon `Range` * 256 + 1024) / Aircraft `Speed`. Note that using a short delay with aircraft that can do multiple strafing runs with their ammo can cause undesired behaviour.
 
In `rulesmd.ini`:
```ini
Strafing.EndDelay=   ; integer, game frames
```

### DIFF
--- a/src/Ext/Aircraft/Hooks.cpp
+++ b/src/Ext/Aircraft/Hooks.cpp
@@ -4,7 +4,6 @@
 #include <Ext/Aircraft/Body.h>
 #include <Ext/Techno/Body.h>
 #include <Ext/Anim/Body.h>
-#include <Ext/Techno/Body.h>
 #include <Ext/WeaponType/Body.h>
 #include <Utilities/Macro.h>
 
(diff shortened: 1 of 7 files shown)


src/Ext/Aircraft/Hooks.cpp
```diff
@@ -176,28 +154,92 @@ DEFINE_HOOK(0x418B1F, AircraftClass_Mission_Attack_FireAtTarget5Strafe_BurstFix,
...
+	int delay = pWeapon->ROF;
 
-	if (pExt->CurrentAircraftWeaponIndex >= 0)
-		return pThis->GetWeapon(pExt->CurrentAircraftWeaponIndex);
-	else
-		return pThis->GetWeapon(pThis->SelectWeapon(pThis->Target));
+	if (isLastShot || pExt->Strafe_BombsDroppedThisRound == pWeaponExt->Strafing_Shots || (pWeaponExt->Strafing_UseAmmoPerShot && !pThis->Ammo))
+	{
+		pThis->MissionStatus = (int)AirAttackStatus::FlyToPosition;
+		delay = pWeaponExt->Strafing_EndDelay.Get((pWeapon->Range + 1024) / pThis->Type->Speed);
```
Thrifinesma, line 175: Guess there needs a comment to explain what 1024 stands for?

Starkku: I could replace it with `Unsorted::LeptonsPerCell` * 4 I guess.

fagonghaiwo:
I adjusted the Strafing.EndDelay to 10. Which is the undesired behaviour?
bandicam 2025-03-23 03-12-20-809

Coronia:
tested and worked

PR #1535, opened 2025-02-14
Improve subterranean harvester pathfinding

Starkku:
This PR should allow `MovementZone=Subterannean` harvesters to access docks that are in areas closed off from normal pathfinding such as cliffs and islands. Whether or not it fixes all issues with such harvesters is another thing.


src/Ext/Unit/Hooks.Harvester.cpp
```diff
@@ -0,0 +1,94 @@
...
+	GET(UnitClass*, pThis, ESI);
+
+	auto const pExt = TechnoExt::ExtMap.Find(pThis);
+
+	if (pExt->SubterraneanHarvFreshFromFactory)
+	{
+		pThis->SetArchiveTarget(nullptr);
+		pThis->ClearNavigationList();
+
+		if (pThis->Destination && pExt->SubterraneanHarvRallyDest && pThis->Destination != pExt->SubterraneanHarvRallyDest && pThis->DistanceFrom(pThis->Destination) > Unsorted::LeptonsPerCell)
```
CrimRecya, line 80: I think `Unsorted::LeptonsPerCell` can be replaced with `RulesClass::Instance->CloseEnough`?

CrimRecya (commented):
It looks good. And I need to test whether it will be affected by the refinery on water surface.

Thrifinesma (changes requested):

src/Ext/Unit/Hooks.Harvester.cpp
Starkku: Too big of a margin for this.

CrimRecya: Seems like there is no problem according to the current test results.

CrimRecya: The only problem currently encountered is that when there is a closer `WaterBound=yes` refinery, the harvesters tend to return here and get stuck.
Of course, given that dev version has not fixed the issue with the floating refinery, I am not sure if this can be considered a problem.

Thrifinesma (approved):

CrimRecya (approved):

mevitar:
Did some games against the AI with this, and at some random point the game dropped to like 1 fps. This happened several times, but lasted only few seconds (except for one game, which didn't seem to fix itself so i had to quit). I also haven't encountered this in any build other than this one.

All these games where it happened were me and AI playing as sides with subterranean harvesters. I'll try investigating further if they are indeed the cause.

mevitar:
No fps drops this time, but game crashed with this build, can't tell the reason from except.txt
https://drive.google.com/file/d/1xd3PoLCVYi60-u0V2jbiQue97VGWkg5t

mevitar:
Haven't had fps drops anymore, but these games were played at like 40-45 fps. The fps drops happened only in games with 60 fps, so that might have been related to how fast the game was processing, or the drawer that was used.

Had only sporadic cases of miners getting stuck, only in the lategame when the resources are mined out on most of the map. If Ollerus didn't point them out i wouldn't even be paying attention.
I'm not sure how much of that was due to them being subterranean miners and some bug making them forget to mine, or simply stopping mining because they can't find anyting to mine anymore.

Coronia:
when we were testing it in PP it gave us this, probably due to the recent include change
image

and it still didn't work quite well in MO after making miners move subterraneanly. They would stuck like this
image

KuroNoSeiHai:
Hello. I found some issues in subterranean harvester.
1. **Phobs b44**: If the subterranean harvester is ordered to move immediately after completing ore unloading (during the unloading animation), it may freeze indefinitely (visually stuck in the unloading animation). A workaround is to press the **S key** or issue multiple movement commands to resolve the freeze.  
2. **Vanilla Behavior**: When manually ordering a subterranean harvester to return to the Ore Refinery, it unnecessarily circles around once before finally stopping. However, if using this nightly fix for subterranean harvester behavior , the harvester returns to the refinery but does not stop completely—it resumes mining instead.

mevitar:
But did you check the behavior in this build?
How harvesters behave in b44 is not indicative of how they behave here.

KuroNoSeiHai:
> But did you check the behavior in this build? How harvesters behave in b44 is not indicative of how they behave here.

I'm sorry, It's B47. I misspelled.

CrimRecya:
> Hello. Excuse me, I found some other issues in subterranean harvester.
> 
> 1. **Phobos b47**: If the subterranean harvester is ordered to move immediately after completing unloading (changing to the unload class), it may freeze indefinitely (visually stuck in the unload vxl). A workaround is to press the **S key** or issue multiple movement commands to resolve the freeze.
> 2. **Vanilla Behavior**: When manually ordering a subterranean harvester to return to the Ore Refinery, it unnecessarily circles around once and then finally stop. However, if using this nightly fix, the harvester returns to the refinery but does not stop completely—it resumes mining instead.

The first issue starts to occur after b44, and b43 does not have this issue.
The second issue only occurs at close range (when the unit does not enter underground).

PR #1537, opened 2025-02-18
[Minor] Interceptor.ApplyFirepowerMult

Coronia:
Interceptor is supposed to be affected by the firer's firepower multipliers, but the previous implementation is missing new attach effect's multiplier. While I once wanted to add it directly, I'm afraid that not everyone want to use AE to affect interceptor's damage. Hence I've added an `Interceptor.ApplyFirepowerMult` tag that'll determine whether or not the intercepting weapon's damage should multiply the TechnoType's firepower multipliers.


CrimRecya (approved):
LGTM.

PR #1538, opened 2025-02-19
[Minor] Fix `KeepTargetOnMove` will stuck unit status

CrimRecya:
The problem arises in:

When a unit with `KeepTargetOnMove` moves after attacking a target, it will enter `Mission::Guard` instead of `Mission::Attack` when it stops, which will result in it not only not pursuing but also not resetting its target when the target moves away from it.

I have added an option to clear the target or not. If not, the unit will be back into `Mission::Attack` so that it can start pursuing its target.
- `KeepTargetOnMove.NoMorePursuit` controls whether the unit will restart chasing the target for attack when it stops again, otherwise it will clear the target when it moves away.


Coronia (approved):

Coronia:
> The code looks fine but I can't say if the logic is without further testing, which should hopefully be conducted by multiple people, preferably those who were already using this logic in the first place.
> 
> I will do a proper review after such testing.

tested, the current behavior is for techno with `KeepTargetOnMove` and in Guard mission, if the target is leaving its range, then it'll keep chasing the target until another command is given

CrimRecya:
> The code looks fine but I can't say if the logic is without further testing, which should hopefully be conducted by multiple people, preferably those who were already using this logic in the first place.
> 
> I will do a proper review after such testing.

It's ready.

Coronia:
since it's still unresolved, maybe we could just add a tag to customize whether the unit will chase target once it's leaving its range, and merge it already. The default behavior could be like before (no chasing) in order to not break things unintentionally

CrimRecya:
> since it's still unresolved, maybe we could just add a tag to customize whether the unit will chase target once it's leaving its range, and merge it already. The default behavior could be like before (no chasing) in order to not break things unintentionally

In that case, I think it may not align with `KeepTargetOnMove.ExtraDistance`‘s design intention.

PR #1539, opened 2025-02-19
[Vanilla Fix] Fix amphibious harvesters can not automatically return to refineries with `WaterBound`

CrimRecya:
- Fix an issue that harvesters with amphibious movement zone can not automatically return to refineries with `WaterBound` on water surface.

Thx to @NetsuNegi .


Coronia:
there might be some path finding issue when miner is at the border of land or water, but that's not related to this pull request I think. Otherwise fine in testing

Coronia (approved):

PR #1549, opened 2025-02-25
Map Action 606、Map Action 38 extend

a851903106:
This is about the trigger moderation measure for AngerLevel, as the campaign may have AI players attacking another country that shouldn't be attacked.

Map Action 38 - adds a Param3 that adds additional AngerLevels.

Map Action 606 - clear all AngerLevels for a given country.

### DIFF
--- a/src/Commands/Commands.h
+++ b/src/Commands/Commands.h
@@ -10,7 +10,7 @@ template <typename T>
 void MakeCommand()
 {
 	T* command = GameCreate<T>();
-	CommandClass::Array->AddItem(command);
+	CommandClass::Array.AddItem(command);
 };
 
 #define CATEGORY_TEAM StringTable::LoadString(GameStrings::TXT_TEAM)
--- a/src/Commands/Dummy.cpp
+++ b/src/Commands/Dummy.cpp
@@ -25,5 +25,5 @@ const wchar_t* DummyCommandClass::GetUIDescription() const
 void DummyCommandClass::Execute(WWKey eInput) const
 {
 	Debug::Log("[Phobos] Dummy command runs.\n");
-	MessageListClass::Instance->PrintMessage(L"[Phobos] Dummy command runs.");
+	MessageListClass::Instance.PrintMessage(L"[Phobos] Dummy command runs.");
 }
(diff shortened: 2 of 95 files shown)


DeathFishAtEase:
I think the script number needs to be changed from 9935 to a suitable number and the corresponding `AI-Scripting-and-Mapping.md` explanation should be added. Maybe its type is suitable for classification into the 14000-14999 range?

---
我想脚本序号需要从 9935 更改为合适的序号并补充对应的 `AI-Scripting-and-Mapping.md` 说明，或许它的类型适合分类进 14000-14999 这个区域内？

a851903106:
> I think the script number needs to be changed from 9935 to a suitable number and the corresponding `AI-Scripting-and-Mapping.md` explanation should be added. Maybe it is suitable for the 506 or 511?
> 
> 我想脚本序号需要从 9935 更改为合适的序号并补充对应的 `AI-Scripting-and-Mapping.md` 说明，或许它适合 507 或 511？

Oh, and I apologize for not noticing this, as I saw Trigger Action 506 followed by Trigger Action 510, which leads me to believe that blindly filling them out in order would disrupt the original plan, which is why I chose the new serial number of 9935.

DeathFishAtEase:
@Metadorius 
Currently, Trigger Actions numbered `507-509` have been left empty for a long time, while the Script has used and reserved many large numbers. As a result, some people may also set a large number for the Action, including but not limited to the content in this PR (use `9935`). So, how should the sequence number of a new Action be arranged at present? I think we should use `511+` for now.

Coronia (changes requested):

DeathFishAtEase:
Indeed, the old PR that has not been merged can make way for the PR that is about to be merged.

FS-21:
Why not leave the map action 38 intact and make a new one for modifying the specified house anger level?
Is what I planned reusing the scripttype actions I sent here once they got merged reusing the code.

a851903106:
> Why not leave the map action 38 intact and make a new one for modifying the specified house anger level? Is what I planned reusing the scripttype actions I sent here once they got merged reusing the code.



> Why not leave the map action 38 intact and make a new one for modifying the specified house anger level? Is what I planned reusing the scripttype actions I sent here once they got merged reusing the code.

That's one way, but I'm not used to it.

Metadorius:
@DeathFishAtEase I think we need to use the same approach as with Script Actions: reserve big range for Phobos and smaller sub-ranges for categories.

PR #1550, opened 2025-02-26
[Minor] Perf tweak

Coronia:
some attempts to improve performance
1 - Phobos radiation (both vanilla and custom) will iterate cells for every RadSite when processing, which might be perf consuming. This pull request will seize the processing early on once the cell doesn't have `RadLevel` greater than 0. Need to test if it has other side effect tho
2 - replace a std::function to lambda in `DetonateOnAllMapObjects` logic
3 - added reserve for some vectors that has estimated max size, mostly for laser trails and attached effects
4 - use `emplace_back` to replace `push_back` for digits, CellStruct and CoordStruct
5 - store the variables that'll be used at least 3 times, for those functions that being called frequently
6 - replaced some nested loop with unordered indices + 1 loop
7 - added early jump out for target selectors

### DIFF
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -1284,7 +1284,7 @@ DrawTurretShadow=false  ; boolean
 
 In `artmd.ini`:
 ```ini
-[SOMEUNIT]      ; UnitType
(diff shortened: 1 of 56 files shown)


Coronia:
Might apply some of them as code optimization in the future, but no much perf improvement with it as of now

PR #1551, opened 2025-02-26
Fix incorrect implementation of tunnel loco speed

chaserli:
Fix incompatiblity with https://github.com/CnCNet/yrpp-spawner/pull/27

### DIFF
--- a/src/Misc/Hooks.BugFixes.cpp
+++ b/src/Misc/Hooks.BugFixes.cpp
@@ -923,12 +923,14 @@ DEFINE_HOOK(0x71ADE4, TemporalClass_Release_SlaveTargetFix, 0x5)
 // which means it didn't consider the actual speed of the unit. Now we check it and the units won't get stuck
 // even at high speeds - NetsuNegi
 
-DEFINE_HOOK(0x7295C5, TunnelLocomotionClass_ProcessDigging_SlowdownDistance, 0x9)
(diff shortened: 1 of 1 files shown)


Starkku:
Any chance of sharing with the class what's up here instead of being cryptic as always for no good reason?

chaserli:
> Any chance of sharing with the class what's up here instead of being cryptic as always for no good reason?

No clue so far. I got crash at the debut of S24 when the SC base was revealed. This just fixed that. 
Perhaps @shmocz @Belonit know. 
19.0 is in rdata and you should've altered its memory protection attributes in the first place.

Belonit:
> > Any chance of sharing with the class what's up here instead of being cryptic as always for no good reason?
> 
> No clue so far. I got crash at the debut of S24 when the SC base was revealed. This just fixed that. Perhaps @shmocz @Belonit know. 19.0 is in rdata and you should've altered its memory protection attributes in the first place.

I also have no idea what we are talking about

chaserli:
> > 19.0 is in rdata
> 
> I don't know what you are referring to

0x7F5B20 (19.0) is in rdata. Previously it was being modified constantly without being patched constantly.

Belonit:
We are talking about **TunnellocomotionClass::TunnelmovementSpeed (0x7F5B20)**

ZivDero (approved):

PR #1552, opened 2025-03-01
Fix Pathfinding Bugs by Decoupling SpeedType from Hardcoded Values

NetsuNegi:
`FreeUnit` uses the unit's own `SpeedType` to find the spawn location.
The bug where naval ships as StartUnit might spawn on land has been fixed.
When a building undeploy, it will normally use the target VehicleType's pathfinding method to decide whether it can move to the target cell.


CrimRecya (approved):
It seems unnecessary to rewrite the entire section, but there is no problem with implementation.

PR #1553, opened 2025-03-02
[Vanilla Enhancement] Enhance the laser, ebolt and rbeam drawing

CrimRecya:
- Fix the bug that laser, electric bolt and rad beam not support `FlakScatter=true` and `Inaccurate=true` projectiles.

---

- You can now add a random offset to visual effect's (`IsLaser=true`, `IsElectricBolt=true` or `IsRadBeam=true`) target location if set `VisualScatter` to true.

In `rulesmd.ini`:
```ini
[AudioVisual]
VisualScatter.Min=0.03  ; floating point value, distance in cells
VisualScatter.Max=0.13  ; floating point value, distance in cells

[SOMEWEAPON]            ; WeaponType
VisualScatter=false     ; boolean
```


Coronia:
maybe it's better to let VisualScatter be defined by individual projectiles? Like how BallisticScatter did. Think there'll be a need for different beams to have different visual

CrimRecya:
> maybe it's better to let VisualScatter be defined by individual projectiles? Like how BallisticScatter did. Think there'll be a need for different beams to have different visual

Is there really a need for this? What I think is that a switch can already solve most situations, because those who need this function usually use reshader, just need the visual effect of a laser bullet. If using the original laser, I think usually only the fix is needed.

Coronia (approved):

PR #1554, opened 2025-03-03
Modify AlphaImage respawn logic

NetsuNegi:
re-impl of [1281](https://github.com/Phobos-developers/Phobos/pull/1281#issuecomment-2693221648)


Coronia:
during our test, AlphaImage on buildings no longer consume large performance, but moving objects still have the same performance reduction as before. It's an optimization at least anyway.

CrimRecya (approved):

PR #1559, opened 2025-03-05
[Minor]RadSite Performance Improve

NetsuNegi:
Added CellExt to store RadSite pointers to avoid constantly traversing RadSiteClass: Array.


src/Ext/Cell/Body.h
```diff
@@ -0,0 +1,47 @@
...
+	using base_type = CellClass;
+
+	static constexpr DWORD Canary = 0xFDC49191;
+	static constexpr size_t ExtPointerOffset = 0x18;
+
+	class ExtData final : public Extension<CellClass>
+	{
+	public:
+		std::vector<RadSiteClass*> RadSites {};
+		std::vector<std::pair<RadSiteClass*, int>> RadLevels { };
```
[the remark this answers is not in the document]
NetsuNegi: > `PhobosMap` is technically any reason to not use that instead ?`std::vector<std::pair<>>`

how to traverse `PhobosMap`？

Coronia:
current test result: under environment with almost same amount of radiation and other conditions being as similar as possible, this pull request did increase performance a lot in all these conditions:
- no radiation on the map
- a bunch of units entering rad field
- radiation with RadHasInvoker/Owner
- radiation that can damage buildings

It's theoretically a bit laggier a the moment of radiation being created, but our test so far indicated that it's not worse than the current method
Hope others could give some more accurate test and check if all rad behaviors work properly

NetsuNegi:
Due to unknown reasons, the S/L has been broken.
I dont't know how to fix it

Metadorius:
Try uncommenting 0x144 offset, IIRC I had no trouble with it

Metadorius:
from @Otamaa on Discord:

> iam going different route for optimization 
> which using `RadSiteClass::Ai` to deliver damage instead
> `cellExt` rad vector ,is probably used for adding/creating new rad
> no more cursed radsite array inside techno every techno

Starkku (changes requested):
The radiation functionality appears to work (although there may be edge cases I did not catch in my tests) and the idea is sound, there is likely performance improvement to gain here.

However the implementation has some facets that likely need to be reworked. Main one is that `std::vector<std::pair<RadSiteClass*, int>>` is the cause of the save/load issues as whatever save/load functionality is implemented for vectors might not handle `std::pair` correctly. `PhobosMap` is effectively the same thing under the hood but it does not grant direct access to the underlying vectors or implement iteration functions, so switching to it would require some changes regardless.

There may be some alternate approaches to this like `CellExt` only tracking RadSites and RadSites themselves tracking levels per cell but I don't know what is the most feasible one right now.

NetsuNegi:
> The radiation functionality appears to work (although there may be edge cases I did not catch in my tests) and the idea is sound, there is likely performance improvement to gain here.
> 
> However the implementation has some facets that likely need to be reworked. Main one is that `std::vector<std::pair<RadSiteClass*, int>>` is the cause of the save/load issues as whatever save/load functionality is implemented for vectors might not handle `std::pair` correctly. `PhobosMap` is effectively the same thing under the hood but it does not grant direct access to the underlying vectors or implement iteration functions, so switching to it would require some changes regardless.
> 
> There may be some alternate approaches to this like `CellExt` only tracking RadSites and RadSites themselves tracking levels per cell but I don't know what is the most feasible one right now.

I'll try breaking it down into two std: vector

Starkku:
Otamaa's approach with all radiation logic being in RadSite and it keeping track which cells it exists on might be saner and cleaner in the end.

NetsuNegi:
I‘m using a struct instead of std::pair, and it has own Save() Load() Serialize()
hope it will be useful

Starkku:
I can confirm that S/L works now, not sure if any further tests with radiation functionality itself are required.

NetsuNegi:
> I can confirm that S/L works now, not sure if any further tests with radiation functionality itself are required.

ok

Coronia:
other logic for regular/custom radiation worked properly after the S/L fix

src/Ext/Bullet/Body.cpp
```diff
@@ -94,28 +95,25 @@ void BulletExt::ExtData::InterceptBullet(TechnoClass* pSource, WeaponTypeClass*
...
-			if (pRadExt->Type != pRadType)
-				return false;
+	const auto pThis = this->OwnerObject();
 
-			if (MapClass::Instance.TryGetCellAt(pSite->BaseCell) != MapClass::Instance.TryGetCellAt(Cell))
-				return false;
+	const auto pWeapon = pThis->GetWeaponType();
+	const auto pWeaponExt = WeaponTypeExt::ExtMap.Find(pWeapon);
+	const auto pRadType = pWeaponExt->RadType;
+	const auto pThisHouse = pThis->Owner ? pThis->Owner->Owner : this->FirerHouse;
```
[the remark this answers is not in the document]
NetsuNegi: > This can be placed just before `RadSiteExt::CreateInstance`.

what?

src/Ext/RadSite/Hooks.cpp
```diff
@@ -160,19 +165,14 @@ DEFINE_HOOK(0x43FB23, BuildingClass_AI_Radiation, 0x5)
...
-
 			if (pBuilding->IsAlive) // simple fix for previous issues
 			{
 				int damage = Game::F2I(radLevel * pType->GetLevelFactor());
 
 				if (maxDamageCount > 0)
 					damageCounts[pRadSite]++;
 
-				if (!pRadExt->ApplyRadiationDamage(pBuilding, damage, Game::F2I(orDistance)))
+				if (!pRadExt->ApplyRadiationDamage(pBuilding, damage, 0))
```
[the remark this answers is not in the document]
NetsuNegi: No, this parameter is required. The function does not define a default value for this parameter

CrimRecya (approved):

src/Ext/RadSite/Body.cpp
```diff
@@ -60,6 +62,10 @@ void RadSiteExt::CreateInstance(CellStruct location, int spread, int amount, Wea
 	pRadSite->SetSpread(spread);
 	pRadExt->SetRadLevel(amount);
 	pRadExt->CreateLight();
+
+	const auto pCell = MapClass::Instance.TryGetCellAt(location);
```
Coronia, line 66: should there be a sanity check here since it's using TryGetCellAt?

NetsuNegi: Actually, there have been checks before, but I still added one

Coronia (approved):

src/Ext/RadSite/Hooks.cpp
```diff
@@ -130,26 +129,32 @@ DEFINE_HOOK(0x43FB23, BuildingClass_AI_Radiation, 0x5)
...
 		CellStruct nCurrentCoord = buildingCoords + *pFoundation;
+		const auto pCell = MapClass::Instance.TryGetCellAt(nCurrentCoord);
+
+		if (!pCell)
+			continue;
+
+		const auto pCellExt = CellExt::ExtMap.Find(pCell);
 
-		for (auto const pRadSite : RadSiteClass::Array)
+		for (const auto& [pRadSite, radLevel] : pCellExt->RadLevels)
```
Metadorius, line 145: nice syntax, didn't know C++ has such 👍

src/Ext/RadSite/Hooks.cpp
```diff
@@ -333,3 +327,96 @@ DEFINE_HOOK(0x65BB67, RadSite_Deactivate, 0x6)
...
+	const auto it = std::find_if(radLevels.begin(), radLevels.end(), [pThis](CellExt::RadLevel const& item) { return item.Rad == pThis; });
+
+	if (it != radLevels.end())
+	{
+		const int amount = Game::F2I(static_cast<double>(max - distance) / max * pThis->RadLevel / pThis->LevelSteps);
+		it->Level -= amount;
+	}
+
+	return SkipGameCode;
+}
```
Metadorius, line 422: Those 3 hooks seem to have a lot of repeated code, could you make a function or share the hook code using `DEFINE_HOOK_AGAIN` and checking `R->Origin()` address?

Metadorius (changes requested):

src/Ext/RadSite/Hooks.cpp
NetsuNegi: it's structured bindings :)

Metadorius (approved):

PR #1561, opened 2025-03-06
[Minor] Fix art section reading bug

CrimRecya:
Fix the bug that `IsAnimDelayedBurst` and `ZShapePointMove.OnBuildup` can only be used when own section name is the same as image section name.


src/Ext/BuildingType/Body.cpp
```diff
@@ -166,7 +166,7 @@ void BuildingTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 	this->FactoryPlant_AllowTypes.Read(exINI, pSection, "FactoryPlant.AllowTypes");
 	this->FactoryPlant_DisallowTypes.Read(exINI, pSection, "FactoryPlant.DisallowTypes");
 
-	this->IsAnimDelayedBurst.Read(exArtINI, pSection, "IsAnimDelayedBurst");
+	this->IsAnimDelayedBurst.Read(exArtINI, pArtSection, "IsAnimDelayedBurst");
```
Coronia, line 169: should move Art tag to the same section, together with `ZShapePointMove.OnBuildup`

Coronia (changes requested):

Starkku (approved):

PR #1563, opened 2025-03-07
[Minor] `NoManualMove` for buildings

TaranDahl:
- Set this to `true` on a building with `UndeploysInto` set could prevent it from undeploying when setting the rally point.

### DIFF
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -1426,6 +1426,10 @@ In `rulesmd.ini`:
 NoManualMove=false  ; boolean
 ```
 
+```{note}
+Note that you can still undeploy the building by using a `BuildingUndeploy=true` warhead or by setting a rally point and selling it.
+```
+
 ### Promoted Spawns
 
 ![image](_static/images/promotedspawns-01.gif)
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -1418,6 +1418,7 @@ MultiMindControl.ReleaseVictim=false  ; boolean
 ### No Manual Move
 
 - You can now specify whether a TechnoType is unable to receive move command.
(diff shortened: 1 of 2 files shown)


Coronia:
ugh, how's the building supposed to move in this case? If it can still somehow undeploy with manual command, then the doc should suggest that

TaranDahl:
Note that you can still undeploy the building by using the undeploy warhead made by CrimRecya or by setting a rally point and selling it.

TaranDahl:
what is the topic to be discussed?

Coronia:
generally when I marked something as Need Discussion, it means I have doubt about either its usefulness or the impact on player interaction. For this pull request I have 2 concerns:
1 - the 'set rally point and sell to get the vehicle' thing is unintuitive and against most players' intuition. And if we simply make it Unsellable, the problem becomes how to let player move the building. BuildingUndeploy can do that _technically_ but we're discussing the _interaction_ aspect here
2 - there's a method to set rally point for movable building already: using Alt key to make it forced-move on a unit, then instead of moving it'll set the rally point. This is also kinda tricky but at least accepted by players, so we need to consider what else can this pull request achieve, or its usefulness will be reduced

TaranDahl:
I proposed this because I have other ways to detonate warheads on buildings to trigger undeploy without interacting with other units. I think the setting that one must click somewhere and then the building will be packed into a vehicle is very stupid, so I made a button for undeploying buildings. 
Currently, there is no such method in Phobos. If necessary, we can make a similar one in Phobos in the future, but I don't think this is the purpose that this pr needs to achieve. If you plan to reject this pull request for this reason, I will not object. You can directly tell me the conclusion.

Metadorius:
Something to consider is that this is a very small PR and one that just extends the existing Phobos logic a bit to also work as one would expect on a building, so I think even if it's fairly niche - I personally don't see a drawback in merging this.

CrimRecya (approved):

PR #1568, opened 2025-03-09
[Minor] Some Optimization

Coronia:
This is a subset of https://github.com/Phobos-developers/Phobos/pull/1550, which including the following tweaks:
- user reserve and emplace_back for some vectors
- early jump out for target selectors and rad damage
- reduce nested loop and replace std::function with lambda
- move/remove some unnecessary variables and checks w
- UpdateTypeData now also handles Gunner change

Average performance did improve a bit in my own test, but more testing and verifying might be needed


src/Ext/WarheadType/Detonate.cpp
```diff
@@ -147,7 +147,7 @@ void WarheadTypeExt::ExtData::Detonate(TechnoClass* pOwner, HouseClass* pHouse,
 
 void WarheadTypeExt::ExtData::DetonateOnOneUnit(HouseClass* pHouse, TechnoClass* pTarget, TechnoClass* pOwner, bool bulletWasIntercepted)
 {
-	if (!pTarget || pTarget->InLimbo || !pTarget->IsAlive || !pTarget->Health || pTarget->IsSinking || pTarget->BeingWarpedOut)
+	if (!pTarget || pTarget->InLimbo || !pTarget->IsAlive || pTarget->Health > 0 || pTarget->IsSinking || pTarget->BeingWarpedOut)
```
Starkku, line 150: I'll go over all of this at better opportunity but just commenting that `!pTarget->Health` and `pTarget->Health > 0` are not interchangeable, `!pTarget->Health` can be expanded to `pTarget->Health == 0`.

Coronia: that's intended in case some situation that strength is below 0, which also means the techno is 'dying'. I've checked other places with Health check in Phobos and they're mostly using this as the condition

Metadorius:
How's the progress?

Coronia:
currently still keep on experimenting everything I can think of to improve perf, but I doubt there's much room left. Would be appreciated if someone can review existing changes, or suggest other places that can be optimized

Coronia:
since there're too many minor changes stacking in this one, gonna gradually implement some changes which won't affect actual logic (cache variables early on, move positions of variables, vector reserve, etc) in other related PRs or standalone commits. Ideally the rest of this PR will only be those that'll truly impact game logic in the end

Starkku:
This honestly probably should be split into multiple PR's, commits or whatever cause atm it is a nightmare to keep up to date and/or review.

Coronia:
Many of the optimization from here are already implemented, the rest are either not that crucial or has controversial. Hence, this PR could be closed, and I'll make PRs for the rest bits if needed

PR #1569, opened 2025-03-10
[Vanilla Fix] Fix the incorrect interaction between cloaking and spawners / projs

TaranDahl:
- Fixed a bug where the owner could not gain experience from the spawner after the owner cloaks


CrimRecya (approved):

Coronia:
tested and worked

btw, when a techno enter a transport during its projectile/spawner is flying, it'll also lose experience due to similar reason. Maybe this could be fixed as well

TaranDahl:
Well, I checked the IDA. It does have a problem. I will try to deal with it.

TaranDahl:
> tested and worked-----------谷歌翻译-----------测试和工作
> 
> btw, when a techno enter a transport during its projectile/spawner is flying, it'll also lose experience due to similar reason. Maybe this could be fixed as well-----------谷歌翻译-----------顺便说一句，当技术在弹丸/产卵器飞行过程中进入运输时，由于类似原因，它也会失去经验。也许这也可以解决

It should work now. Please test again.

CrimRecya (approved):

PR #1570, opened 2025-03-10
Improve aircraft area guard behaviour

Starkku:
This PR should improve behaviour of aircraft when assigned area guard mission (`G` on keyboard or the guard button in advanced command bar) with `[General]` -> `ExtendedAircraftMissions=true`, particularly when landed on a dock. Before, aircraft would take off from the dock even if there was no ammo, no targets nearby etc. Now they will only take off if they have a target to engage.

Some notes:
- Repeatedly assigning area guard on docked aircraft that have partially loaded ammo will stall the reloading process even if the jet won't take off. The same was true before, just more obvious (and longer duration) with the aircraft actually taking off. This is no longer true for aircraft with no ammo at all, however.
- Area guard on aircraft is still not persistent (e.g doesn't expire without player command or some other explicity reset of the unit''s mission) like it is on other types. Implementing this, if it would even be desirable, would require completely different approach due to different missions being assigned to aircraft as part of the docking and reloading process and frequently interrupting the area guard mission.

### DIFF
--- a/src/Ext/Aircraft/Hooks.cpp
+++ b/src/Ext/Aircraft/Hooks.cpp
@@ -1,10 +1,10 @@
 #include <AircraftClass.h>
+#include <EventClass.h>
 #include <FlyLocomotionClass.h>
 
 #include <Ext/Aircraft/Body.h>
 #include <Ext/Techno/Body.h>
 #include <Ext/Anim/Body.h>
-#include <Ext/Techno/Body.h>
 #include <Ext/WeaponType/Body.h>
 #include <Utilities/Macro.h>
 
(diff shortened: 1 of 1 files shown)


CrimRecya (approved):
LGTM.

Coronia:
worked properly in testing

CrimRecya:
Working very well. When commanding the aircraft to `Mission::Area_Guard`, they won't make meaningless up and down twitches at the airport anymore.
In addition, there is a question that has been troubling me for a long time: when aircraft are returning to the airport, if they are given a movement command at the beginning of their landing, they will enter `Mission::Guard` instead of `Mission::Move`, which will cause them to abandon the command and start returning after moving a short distance. Do you know anything about this?

PR #1571, opened 2025-03-11
Optimize per-frame update logic

Starkku:
Splits functionality away from `TechnoClass_AI` / `TechnoExt::OnEarlyUpdate()` that does not need to be processed for all types of technos.

- Splits logic concerning `FootClass` only from `TechnoExt::UpdateTypeData()` to `TechnoExt::UpdateTypeData_Foot()` called from `FootClass_AI`. While `OpenTopped` is read for all technos, it does not work correctly for buildings so it was also moved.
- Moves tunnel enter / exit processing and relevant hooks entirely to appropriate update functions instead of general ones.
- `TechnoExt::DepletedAmmoActions()` is now called from `UnitClass_AI` (not a new hook - was repurposed/generalized from `KeepTargetOnMove` feature which was split into its own function too).
- `TechnoExt::UpdateGattlingRateDownReset()` is now called from `UnitClass_AI` and `BuildingClass_AI`.

Some notes:
- `TechnoExt::EatPassengers()` was kept in `TechnoExt::OnEarlyUpdate()` because it does work with passenger logic used on buildings such as Bio Reactor.
- Theoretically under current circumstances the entire  `TechnoExt::UpdateTypeData()` could've been moved to be processed for `FootClass` only as there are no building type changes in the game, but I suppose this is the future-proofed approach.

### DIFF
--- a/src/Ext/Building/Hooks.cpp
+++ b/src/Ext/Building/Hooks.cpp
@@ -21,9 +21,7 @@ DEFINE_HOOK(0x43FE69, BuildingClass_AI, 0xA)
 	auto const pExt = BuildingExt::ExtMap.Find(pThis);
 	pExt->DisplayIncomeString();
 	pExt->ApplyPoweredKillSpawns();
-
-	auto const pTechnoExt = TechnoExt::ExtMap.Find(pThis);
-	pTechnoExt->UpdateGattlingRateDownReset();
+	pExt->TechnoExtData->UpdateGattlingRateDownReset();
 
 	return 0;
 }
--- a/src/Ext/Building/Hooks.cpp
+++ b/src/Ext/Building/Hooks.cpp
@@ -13,24 +13,17 @@
 
 #pragma region Update
 
-//After TechnoClass_AI?
+// After TechnoClass_AI
 DEFINE_HOOK(0x43FE69, BuildingClass_AI, 0xA)
 {
 	GET(BuildingClass*, pThis, ESI);
 
-	// Do not search this up again in any functions called here because it is costly for performance - Starkku
-	auto pExt = BuildingExt::ExtMap.Find(pThis);
-
-	/*
-	// Set only if unset or type has changed - Not currently useful as building type does not change.
-	auto pType = pThis->Type;
-
-	if (!pExt->TypeExtData || pExt->TypeExtData->OwnerObject() != pType)
-		pExt->TypeExtData = BuildingTypeExt::ExtMap.Find(pType);
-	*/
-
+	auto const pExt = BuildingExt::ExtMap.Find(pThis);
 	pExt->DisplayIncomeString();
 	pExt->ApplyPoweredKillSpawns();
+
+	auto const pTechnoExt = TechnoExt::ExtMap.Find(pThis);
+	pTechnoExt->UpdateGattlingRateDownReset();
 
 	return 0;
 }
--- a/src/Ext/Techno/Body.Update.cpp
+++ b/src/Ext/Techno/Body.Update.cpp
@@ -17,14 +17,12 @@
 // It's not recommended to do anything more here it could have a better place for performance consideration
 void TechnoExt::ExtData::OnEarlyUpdate()
 {
-	auto pType = this->TypeExtData->OwnerObject();
+	auto pType = this->OwnerObject()->GetTechnoType();
 
 	// Set only if unset or type is changed
 	// Notice that Ares may handle type conversion in the same hook here, which is executed right before this one thankfully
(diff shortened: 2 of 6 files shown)


Metadorius:
> Moves laser trail processing to FootClass_AI.

I am not sure this should be done? I imagine a niche use case that could be done is a big turret of a building with a laser trails far from the center of rotation.

chaserli (changes requested):

Starkku:
> > Moves laser trail processing to FootClass_AI.
> 
> I am not sure this should be done? I imagine a niche use case that could be done is a big turret of a building with a laser trails far from the center of rotation.

The laser trailer update function had the following even before my changes.

```cpp
auto const pThis = generic_cast<FootClass*>(this->OwnerObject());

if (!pThis)
    return;
```

I am not certain about what `generic_cast` returns if the type doesn't match but if it comes back with `nullptr` then it already didn't work for buildings.

src/Ext/Techno/Body.Update.cpp
```diff
@@ -413,25 +412,16 @@ void TechnoExt::ExtData::ApplySpawnLimitRange()
...
 {
 	auto const pThis = this->OwnerObject();
-	auto const pOldTypeExt = this->TypeExtData;
-	auto const pOldType = this->TypeExtData->OwnerObject();
-
-	if (this->LaserTrails.size())
-		this->LaserTrails.clear();
-
+	auto const pOldTypeExt = TechnoTypeExt::ExtMap.Find(pOldType);
+	auto const pCurrentType = pThis->GetTechnoType();
```
CrimRecya, line 419: I don't understand here because `GetTechnoType()` should be more complex than directly using pointers.
Originally, we used the new type passed in here, and the old type can be obtained through `this->TypeExtData->OwnerObject()`, so at this point, it's like the work has already ended.
However, the modification here has become the repeated use of `GetTechnoType()` (Within and outside of this function), and use the `TechnoTypeExt::ExtMap.Find(this->TypeExtData->OwnerObject())` process.

src/Ext/Techno/Hooks.cpp
```diff
@@ -16,13 +16,41 @@
...
+
+// After TechnoClass_AI
+DEFINE_HOOK(0x4DA54E, FootClass_AI, 0x6)
+{
+	GET(FootClass*, pThis, ESI);
+
+	auto const pExt = TechnoExt::ExtMap.Find(pThis);
+
+	if (pExt->PreviousType)
+		pExt->UpdateTypeData_Foot(pExt->PreviousType);
```
CrimRecya, line 37: I think there's no need to pass this as a parameter?

CrimRecya (commented):
I have some small questions, all of which are listed below.
Also, I noticed the movement of the `KeepTargetOnMove` code, so I thought of another thing - could you take some time to review #1538 ?

src/Ext/Techno/Hooks.cpp
Starkku: There is because it has already been changed at this point.

src/Ext/Techno/Body.Update.cpp
Starkku: I have no clue what you're trying to say.

src/Ext/Building/Hooks.cpp
```diff
@@ -13,25 +13,18 @@
...
-	if (!pExt->TypeExtData || pExt->TypeExtData->OwnerObject() != pType)
-		pExt->TypeExtData = BuildingTypeExt::ExtMap.Find(pType);
-	*/
-
+	auto const pExt = BuildingExt::ExtMap.Find(pThis);
 	pExt->DisplayIncomeString();
 	pExt->ApplyPoweredKillSpawns();
 
+	auto const pTechnoExt = TechnoExt::ExtMap.Find(pThis);
+	pTechnoExt->UpdateGattlingRateDownReset();
```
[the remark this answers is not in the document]
Starkku: I forgot that exists.

Coronia:
nothing wrong happened during our test so far, and the performance did obviously improved

Coronia (commented):
I'm just being carping here, but maybe we can further split UpdateTypeData for unit/aircraft (MoveSound + transport) and infantry (MoveSound + Deployer fix). Similarly, EatPassengers could only be called by UnitClass_AI and BuildingClass_AI. Not sure if aircraft needs it as well

It should be fine if it remains unchanged too, guess there's no much perf consuming from these fields

mevitar:
Did few games against AI with this build, found no issues so far.

PR #1572, opened 2025-03-11
Reduce floating point precision

chaserli:
Do you really need that precision?

### DIFF
--- a/Phobos.vcxproj
+++ b/Phobos.vcxproj
@@ -372,6 +372,7 @@
       <DisableSpecificWarnings>4100;4201;4530;4731;4740;4458;4819;5103;5105</DisableSpecificWarnings>
       <EnableModules>true</EnableModules>
       <OpenMPSupport>true</OpenMPSupport>
+      <FloatingPointModel>Fast</FloatingPointModel>
     </ClCompile>
     <Link>
       <EnableCOMDATFolding>true</EnableCOMDATFolding>
@@ -417,6 +418,7 @@
       <EnableModules>true</EnableModules>
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
       <OpenMPSupport>true</OpenMPSupport>
+      <FloatingPointModel>Fast</FloatingPointModel>
     </ClCompile>
     <Link>
       <EnableCOMDATFolding>true</EnableCOMDATFolding>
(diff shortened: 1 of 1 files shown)


Metadorius (commented):
This just needs testing. No idea about consequences.

Coronia:
with nightly from this pull request we've experienced 6 crashes so far, while 4 of them are sudden quit that didn't generate anything other than debug.log. Uploaded these logs here, not sure if it's related to the floating point change

Version: MO 3.3.6 + edits for new Phobos logic testing
[Fatal Errors.zip](https://github.com/user-attachments/files/19284862/Fatal.Errors.zip)

chaserli:
> This just needs testing. No idea about consequences.

I am simply asking about which /fp option to use. As far as I know, there are not a great many FP operations in the current Phobos. However, if someone intends to rewrite some low level features, it could be an important factor. Although I've been using this option over the last 2 years, I'm not sure about its potential risks. Therefore, I am raising this question in advance to see if anyone has an answer. Before that, just hold this for a while. I compiled ddraw with avx and /fp:fast to take some advantage of simd operations, yet the effect is limited because of its narrow application scope.

> with nightly from this pull request we've experienced 6 crashes so far, while 4 of them are sudden quit that didn't generate anything other than debug.log. Uploaded these logs here, not sure if it's related to the floating point change

Alphaimage/blitter crashes should already be present before

mevitar:
> with nightly from this pull request we've experienced 6 crashes so far, while 4 of them are sudden quit that didn't generate anything other than debug.log. Uploaded these logs here, not sure if it's related to the floating point change

I did few games against AI and didn't encounter any crashes. Also tried MO and no crashes either (although i will still try to play the campaign more).
Perhaps those crashes are caused by ddraw.dll?

mevitar:
SCRN 20250413-152906-00737

Had this graphical problem when loading a MO campaign save. It fixed itself only when i loaded the game again. Not sure if it's related to ddraw used or because of this build.
That game also crashed to desktop when i was close to finishing the mission. Only syringle.log and debug.log are available,
[250413-crashtodesktop_b1572.zip](https://github.com/user-attachments/files/19725647/250413-crashtodesktop_b1572.zip)

PR #1573, opened 2025-03-13
[Minor] Fix on #1569

TaranDahl:


CrimRecya (approved):
The code is fine, just need testing.

Coronia (approved):
so far no crash occured in our testing, while experience can pass properly

for those who don't understand the context: we found a crash on aircrafts in some occasions (mostly when spawned in missions) after applying https://github.com/Phobos-developers/Phobos/pull/1569 and this pull request is supposed to fix this issue

PR #1580, opened 2025-03-19
[Dehardcode] Sinking improvement

TaranDahl:
### Sinking behavior dehardcode

- In vanilla, whether a ship sinks when it dies on the water is determined by multiple settings of hardcoding. The speed of the sinking is hardcoded to 5 Leptons per frame.
- Now you can determine whether a ship sinks with a dedicated flag `Sinkable`, and use `SinkSpeed` to customize the speed at which the ship sinks.

In `rulesmd.ini`:
```ini
[SOMEVEHICLE]           ; VehicleType
Sinkable=               ; bool
SinkSpeed=5             ; integer, lepton per frame
```


tyuah8:
I noticed that when a Giant Squid kills a target, that target will always sink even if it doesn't normally.  Does this PR address this issue?

TaranDahl:
> I noticed that when a Giant Squid kills a target, that target will always sink even if it doesn't normally. Does this PR address this issue?

The new submission has handled the relevant situation but has not been tested. If possible, please assist in testing.

src/Ext/Unit/Hooks.Sinking.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include <ScenarioClass.h>
+#include <TunnelLocomotionClass.h>
+
+#include <Ext/Techno/Body.h>
+#include <Ext/TechnoType/Body.h>
+#include <Ext/BuildingType/Body.h>
+#include <Ext/House/Body.h>
+#include <Ext/WarheadType/Body.h>
+#include <Ext/WeaponType/Body.h>
+#include <Utilities/EnumFunctions.h>
```
CrimRecya, line 12: I think it's no need to include so many header files.

Coronia:
tested and worked

Coronia:
It seems that units are still sinking with Sinkable=no and SinkableBySquid=no when destroyed by SquidGrab
Also, since the tag for vanilla Squid attack is `SquidGrab`, this tag should better be renamed to `Sinkable.SquidGrab`
image
image

Coronia (approved):
Sinkable.SquidGrab also worked in the newest nightly

PR #1587, opened 2025-03-26
Fixes & improvements to electric bolt handling

### LINKED ISSUE #1585: IsElectricBolt=true weapon with 2 burst cannont draw the first bolt correctly
### Description

When using a weapon with IsElectricBolt=true and Burst=2, the first electric bolt stops rendering immediately after the second bolt is drawn, making the first bolt almost invisible. This issue becomes more pronounced if BurstDelay0 is set to 0. It significantly affects the visual effects of the Tesla Cruiser.

### Phobos Version

Latest Nightly Build: compiled-dll-278e0a07fe29c532c1fae0cfaea25caf6acf3ee7       and also tried on Self-compiled develop branch (no additional code modifications).

### Conditions to reproduce

1. Phobos Latest Nightly Build: compiled-dll-278e0a07fe29c532c1fae0cfaea25caf6acf3ee7 or Self-compiled develop branch on March 25, 2025
2. mod: Mental Omega 3.3.6.

### INI code

```ini
Just vanilla Code of Tesla Cruiser in Mental Omega 3.3.6
```

### Steps to reproduce

1. Install Phobos in Mental Omega 3.3.6.
2. Launch Mental Omega 3.3.6.
3. Use the Tesla Cruiser to attack any target.
4. Observe the incorrect rendering of electric bolts between Bursts.

### Expected behaviour

Under normal circumstances, electric bolts from each Burst should render independently without interference. Attached is the expected behavior observed in version B46 for reference.

https://github.com/user-attachments/assets/e860a161-f619-4e7f-a58e-f2cd0633d11e

### Actual behaviour

The first electric bolt stops rendering immediately after the second bolt is drawn, making the first bolt almost invisible.

https://github.com/user-attachments/assets/50e37dc0-9070-4386-9af4-9ba1ae6c0e6f

### Additional context

_No response_

### Checklist

- [x] The issue happens on the **latest official** version of Phobos and wasn't fixed yet.
- [x] I agree to elaborate the details if requested and provide thorough testing if the bugfix is implemented.
- [x] I added a very descriptive title to this issue.
- [x] I have attached as much information as possible *(screenshots, gifs, videos, debug and exception logs, etc)*.

Starkku:
- Reverts the previous, partial or non-functional fixes for crash at `0x4C2C19`.
- Replaces the code for handling tracking of electric bolts and their owner technos to support multiple electric bolt instances per techno (vanilla tracks this with single pointer in UnitClass only). This tracking code also correctly works for units in limbo which should prevent the aforementioned crash from occuring at all.
- This tracking is used to determine whether or not the electric bolt updates its source coordinate based on FLH on every frame, allowing it to follow techno's rotation, movement etc. By default this is only enabled for vehicles just as before but can be overridden per weapon by explicitly setting value for `Bolt.FollowFLH`. It now correctly takes burst into accord thanks to changes mentioned in previous point.
 
Fixes #1585

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -256,6 +256,7 @@ This page lists all the individual contributions to the project by their author.
   - Custom exit cell for infantry factory
   - Vehicles keeping target on move command
   - `IsSonic` wave drawing crash fix
+  - Customizable electric bolt duration and electric bolt-related fixes
 - **Morton (MortonPL)**:
   - `XDrawOffset` for animations
   - Shield passthrough & absorption
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -194,6 +194,7 @@ This page describes all ingame logics that are fixed or improved in Phobos witho
 - Fixed issues caused by incorrect reference removal (f.ex. If the unit cloaks/enters transport, it cannot gain experience from previously launched spawners/C4/projectiles).
 - Fixed an issue that caused `IsSonic=true` wave drawing to crash the game if the wave traveled over a certain distance.
 - Buildings with foundation bigger than 1x1 can now recycle spawner correctly.
+- Electric bolts that are supposed to update their position based on units current firing coords (by default, those fired by vehicles) now do so correctly for more than one concurrent electric bolt.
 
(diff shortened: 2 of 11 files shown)


Coronia:
worked properly in testing

PR #1588, opened 2025-03-27
Fix Ares3.0p1 BuildingAnim issue

chaserli:


Coronia:
no visible lag during our test, and the profiler didn't show it as a lag source either

CrimRecya (approved):

PR #1592, opened 2025-03-30
[Vanilla Fix] Fixed an issue where `FireAngle` would not work properly under certain circumstances.

TaranDahl:
WW uses `SetDesired` on `BarrelFacing`, which causes the barrel image cache to be incorrectly initialized.


TaranDahl:
image
A buggy example. Normally, the barrel should be tilted up.

CrimRecya (approved):
LGTM. But do you know what 0x6F6DEA and 0x6F6DF1 are doing?

TaranDahl:
> LGTM. But do you know what 0x6F6DEA and 0x6F6DF1 are doing?

no ...

PR #1600, opened 2025-03-31
[Vanilla Fix] Fixed an issue that AI would select unreachable buildings as target

TaranDahl:


Coronia:
this one is hard to be tested, so more code review is appreciated

CrimRecya (approved):
looks fine.

TaranDahl:
> this one is hard to be tested, so more code review is appreciated-----------谷歌翻译-----------这很难测试，因此对更多的代码审查表示赞赏

I tested the battle bunker one and it works normally. Others are basically similar.

TaranDahl:
To test this, you can place a building on an isolated island and let the technos execute the corresponding script. 
In vanilla, technos will stand in place and continuously look at the target. For infantry, they will maintain a guard action and will not play idle actions. (For vehicles, there may be turret turning? Uncertain.) 
With this fix, technos will not select this building as a target. Infantry will play idle actions normally.

PR #1603, opened 2025-04-03
[Dehardcode] Prone speed customization

TaranDahl:

### DIFF
--- a/src/Ext/Techno/Hooks.cpp
+++ b/src/Ext/Techno/Hooks.cpp
@@ -715,3 +715,15 @@ DEFINE_HOOK(0x465D40, BuildingClass_Is1x1AndUndeployable_BuildingMassSelectable,
 
 #pragma endregion
 
+DEFINE_HOOK(0x521D94, InfantryClass_CurrentSpeed_ProneSpeed, 0x6)
+{
+	GET(InfantryClass*, pThis, ESI);
+	GET(int, currentSpeed, ECX);
+
+	auto pType = pThis->Type;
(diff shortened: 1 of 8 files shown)


src/Ext/Techno/Hooks.cpp
```diff
@@ -715,3 +715,15 @@ DEFINE_HOOK(0x465D40, BuildingClass_Is1x1AndUndeployable_BuildingMassSelectable,
...
 #pragma endregion
 
+DEFINE_HOOK(0x521D94, InfantryClass_CurrentSpeed_ProneSpeed, 0x6)
+{
+	GET(InfantryClass*, pThis, ESI);
+	GET(int, currentSpeed, ECX);
+
+	auto pType = pThis->Type;
+	auto pTypeExt = TechnoTypeExt::ExtMap.Find(pType);
+	auto multiplier = pTypeExt->ProneSpeed.Get(RulesExt::Global()->ProneSpeed.Get(pType->Crawls ? 0.67 : 1.5));
```
Coronia, line 725: should either split definition of speed when the infantry is crawling or not, or make ProneSpeed not affecting speed when it's not crawling

Coronia (changes requested):

Coronia:
considering the value is different when infantry has `Crawls` or not, suggest to seperate the tag in [General] into 2 tags that could refer to different default values in both cases

CrimRecya (approved):

Coronia (approved):

PR #1606, opened 2025-04-05
[Dehardcode]Customize damaged speed

NetsuNegi:
- In vanilla, units using drive/ship loco will has hardcoded speed multiplier when damaged. Now you can customize it.
- The max valuebale value is 1.0, you cannot make unit get faster on yellow condition by it.

In `rulesmd.ini`
```ini
[General]
DamagedSpeed=0.75        ; double

[SOMEUNIT]
DamagedSpeed=            ; double
```


TaranDahl:
You can return `R->Origin()+0x6`. Then you can use `DEFINE_HOOK_AGAIN`.

Coronia (approved):

PR #1607, opened 2025-04-06
[Dehardcode] RadarInvisible for non-enemy house

TaranDahl:
### RadarInvisible for non-enemy house

- In vanilla, `RadarInvisible` is ignored if the techno is allied with the current player. Now you can change this behavior.

In `rulesmd.ini`:
```ini
[SOMETECHNO]                         ; TechnoType
RadarInvisibleToHouse=               ; Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all), default to enemy if `RadarInvisible=true`, none otherwise
```


TaranDahl:
Need testing: Radar visibility when the alliance relationship is asymmetric.

Coronia:
suggest to change it into `RadarInvisibleToHouse` and use Phobos's AffectedHouse enum here, default to enemy

Coronia:
worked in testing, but the profiler show that it has a noticeable (not big tho) perf cost. Would like to hear others' words if this cost is acceptable
BE8474E8A42DEB2CD506881B3D4D086D

mevitar:
Don't think there will be many objects that people will use this on, it shouldn't be a problem (and I hope even less of a problem if it will be used on immobile objects).

Although i have to admit that i didn't expect _disabling_ drawing on minimap to be as costly as updating sensors.

Coronia (changes requested):
since there're some potential perf issues, suggest to rewrite it a bit to let it bail out earlier if the tag is not set

```c++
if (isInShrouded && !pTechno->Owner->IsControlledByCurrentPlayer())
	return Invisible;

auto pType = pTechno->GetTechnoType();

if (pType->RadarInvisible)
{
	auto pTypeExt = TechnoTypeExt::ExtMap.Find(pType);

	if (EnumFunctions::CanTargetHouse(pTypeExt->RadarInvisibleToHouse.Get(AffectedHouse::Enemies), pTechno->Owner, HouseClass::CurrentPlayer))
		return Invisible;
}

return GoOtherChecks;
```

src/Ext/Techno/Hooks.cpp
```diff
@@ -744,3 +744,35 @@ DEFINE_HOOK(0x4B3DF0, LocomotionClass_Process_DamagedSpeedMultiplier, 0x6)// Dri
...
+{
+	enum { Invisible = 0x655E66, GoOtherChecks = 0x655E19 };
+
+	GET_STACK(bool, isInShrouded, STACK_OFFSET(0x40, 0x4));
+	GET(TechnoClass*, pThis, EBP);
+
+	auto pTechno = abstract_cast<TechnoClass*>(pThis);
+
+	if (!pTechno)
+		return 0;
```
CrimRecya, line 760: Can it be omitted? In the original code, I saw that it directly used `Owner` and `GetTechnoType()` without any additional checks.

CrimRecya (approved):
Just a minor thing.

src/Ext/Techno/Hooks.cpp
TaranDahl: changed

Coronia (approved):

PR #1609, opened 2025-04-07
[Minor]Enhanced reveal & gap warhead

NetsuNegi:
- Warheads can now reveal an area or the entire map on impact.
  - Reveal only applies to the owner of the warhead.

In `rulesmd.ini`:
```ini
[SOMEWARHEAD] 
Reveal=0     ; integer - cell radius, negative values mean reveal the entire map
```

- Warheads can now shroud an area or the entire map on impact.
  - Shroud only applies to enemies of the warhead owner.

In `rulesmd.ini`:
```ini
[SOMEWARHEAD] 
CreateGap=0     ; integer - cell radius, negative values mean shroud the entire map
```

### DIFF
--- a/src/Ext/WarheadType/Detonate.cpp
+++ b/src/Ext/WarheadType/Detonate.cpp
@@ -17,6 +17,20 @@
 #include <Utilities/Helpers.Alex.h>
 #include <Utilities/EnumFunctions.h>
 
+#pragma region CreateGap Calls
+
+static void __stdcall Sub_4ADEE0(char a1, DWORD a2)
+{
+	JMP_STD(0x4ADEE0);
+}
+
+static void __stdcall Sub_4ADCD0(char a1, DWORD a2)
+{
+	JMP_STD(0x4ADCD0);
+}
+
+#pragma endregion
+
 void WarheadTypeExt::ExtData::Detonate(TechnoClass* pOwner, HouseClass* pHouse, BulletExt::ExtData* pBulletExt, CoordStruct coords)
 {
 	auto const pBullet = pBulletExt ? pBulletExt->OwnerObject() : nullptr;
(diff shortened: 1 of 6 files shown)


src/Ext/WarheadType/Detonate.cpp
```diff
@@ -31,22 +45,50 @@ void WarheadTypeExt::ExtData::Detonate(TechnoClass* pOwner, HouseClass* pHouse,
...
-					!pOtherHouse->IsObserver() &&         // Not Observer
-					!pOtherHouse->Defeated &&             // Not Defeated
-					pOtherHouse != pHouse &&              // Not pThisHouse
-					!pHouse->IsAlliedWith(pOtherHouse))   // Not Allied
+				if (pOtherHouse->IsControlledByHuman() &&     // Not AI
+					!pOtherHouse->IsObserver() &&		// Not Observer
+					!pOtherHouse->Defeated &&						// Not Defeated
+					pOtherHouse != pHouse &&							// Not pThisHouse
+					!pOtherHouse->SpySatActive &&					// No SpySat
+					!pOtherHouse->IsAlliedWith(pHouse))			// Not Allied
```
Coronia, line 82: it's better to use the same checking process when CreateGap is greater or smaller than 0

src/Ext/WarheadType/Detonate.cpp
```diff
@@ -31,22 +45,50 @@ void WarheadTypeExt::ExtData::Detonate(TechnoClass* pOwner, HouseClass* pHouse,
 
 	if (pHouse)
 	{
-		if (this->BigGap)
+		if (this->CreateGap > 0)
+		{
+			const auto pCurrent = HouseClass::CurrentPlayer;
+
+			if (pCurrent &&
```
Coronia, line 52: HouseClass::CurrentPlayer always exists so no need to do sanity check here

Coronia (commented):
since this one modify the existing tags, a migration script is needed as well

NetsuNegi:
> since this one modify the existing tags, a migration script is needed as well

see https://github.com/Phobos-developers/PhobosSupplementaries/pull/7

Coronia (approved):

PR #1615, opened 2025-04-08
[Vanilla Fix] Fix an issue that units on the slope tilted at an excessive angle

CrimRecya:


CrimRecya:
ced6f2abe9033639b158e539096c9eb5
->
3ce4920ecfa100ed0be48adf7df460e8

Coronia (approved):

Starkku:
what

How people (including yours truly) never noticed I will never know. LGTM ig.

src/Ext/TechnoType/Hooks.MatrixOp.cpp
```diff
@@ -703,3 +703,23 @@ Matrix3D* __fastcall BounceClass_ShadowMatrix(BounceClass* self, void*, Matrix3D
...
+DEFINE_HOOK_AGAIN(0x75564B, sub_754CB0_InitializeRampMatrix, 0x7)
+DEFINE_HOOK_AGAIN(0x7554CC, sub_754CB0_InitializeRampMatrix, 0x5)
+DEFINE_HOOK(0x755469, sub_754CB0_InitializeRampMatrix, 0x5)
+{
+	// All of them are float, but I don't understand why there would be problems if use float to pass in
+	GET(const int, phi, EBX);
+	R->EBX(R->EBP());
+	R->EBP(phi);
+	return 0;
+}
```
Metadorius, line 723: I checked the hook position, and I honestly didn't understand anything of it, can you explain in more details what is happening here? All I see is that EBX and EBP are interchangeably used in the code as arguments for the called functions.

Metadorius: From what I understand, the only thing that you effectively change is what register is read, swapping EBX and EBP when needed. Can't you just patch the needed bytes directly? Like replace mov EBX with mov EBP and vice versa.

PR #1617, opened 2025-04-09
[Vanilla Fix] 2 bugfixes

NetsuNegi:
- Fix the bug that `EnterBioReactorSound` and `LeaveBioReactorSound` on technotype does not used.
- Fix the bug that harvester dont stop unloading and cannot unload cargos anymore when lifting by `IsLocomotor=yes` warhead.


Coronia (approved):
worked properly in testing

TaranDahl:
I think `EnterGrinderSound` also has the same problem. If you are interested, you can fix it together.

NetsuNegi:
> I think `EnterGrinderSound` also has the same problem. If you are interested, you can fix it together.

soon

PR #1624, opened 2025-04-13
[Minor] Fix `JumpjetTilt` shadow drawing

CrimRecya:


Metadorius:
How does it look at the extreme angle like 90 degrees now?

CrimRecya:
> How does it look at the extreme angle like 90 degrees now?

Well, a line.

PR #1628, opened 2025-04-14
Map Action 606-608

a851903106:
This is the third time it's been pushed, and I don't know why every forced merge causes PR to shut down.

Function for three trigger actions 606-608:

- 606: Edit the hate-value that triggers houses against specified house. -1 for all houses.
- 607: Clear the hate-value that triggers houses against specified house. -1 for all houses
- 608: Set a forced enemy that is not affected by hate-value. -1 will remove forced enemies, -2 will force no enemies.

----------------------
这是第三次推送它了，我不知道为什么每次强制合并都会导致PR关闭。

功能为三个触发动作 606-608：

- 606 : 编辑触发所属方对指定所属方的仇恨值。-1为全部所属方。
- 607 : 清除触发所属方对指定所属方的仇恨值。-1为全部所属方。
- 608 : 设置一个不会被仇恨值影响的强制敌人。-1删除强制敌人，-2强制没有敌人。

### DIFF
--- a/src/Ext/House/Body.cpp
+++ b/src/Ext/House/Body.cpp
@@ -658,6 +658,7 @@ void HouseExt::ExtData::Serialize(T& Stm)
 		.Process(this->AIFireSaleDelayTimer)
 		.Process(this->SuspendedEMPulseSWs)
 		.Process(this->SuperExts)
+		.Process(this->ForceEnemyIndex)
 		;
 }
 
--- a/src/Ext/House/Body.cpp
+++ b/src/Ext/House/Body.cpp
@@ -604,9 +604,6 @@ void HouseExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 
 int HouseExt::ExtData::GetForceEnemyIndex()
 {
-	if (!this)
-		return -1;
-
 	auto const pHouse = this->OwnerObject();
 	if (!pHouse)
 		return -1;
@@ -614,11 +611,8 @@ int HouseExt::ExtData::GetForceEnemyIndex()
 	return this->ForceEnemyIndex;
 }
 
-void HouseExt::ExtData::SetForceEnemy(int EnemyIndex)
+void HouseExt::ExtData::SetForceEnemyIndex(int EnemyIndex)
 {
-	if (!this)
-		return;
-
 	if (EnemyIndex < 0 && EnemyIndex != -2)
 		this->ForceEnemyIndex = -1;
 	else
--- a/src/Ext/House/Body.cpp
+++ b/src/Ext/House/Body.cpp
@@ -602,6 +602,29 @@ void HouseExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 	}
 }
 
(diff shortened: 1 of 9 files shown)


a851903106:
I thought someone might come along who needed them, so I thought maybe it would be better to stay.

src/Ext/House/Body.cpp
```diff
@@ -602,6 +602,29 @@ void HouseExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 	}
 }
 
+int HouseExt::ExtData::GetForceEnemyIndex()
+{
+	if (!this)
```
Coronia, line 607: if `this` pointer is null, it would've already crashed when trying to access this function. Should move the sanity check to out of the function

src/Ext/TAction/Body.cpp
```diff
@@ -374,6 +382,133 @@ bool TActionExt::ToggleMCVRedeploy(TActionClass* pThis, HouseClass* pHouse, Obje
...
+			HouseClass::FindByIndex(pThis->Value) :
+			HouseClass::FindByCountryIndex(pThis->Value);
+
+		setValue(pTargetHouse);
+		pHouse->UpdateAngerNodes(0, nullptr);
+	}
+	else if (pThis->Value == -1)
+	{
+		for (auto pTargetHouse : HouseClass::Array)
+			setValue(pTargetHouse);
```
Coronia, line 434: for (auto pTargetHouse : HouseClass::Array)
{
			setValue(pTargetHouse);
}

src/Ext/TAction/Body.cpp
```diff
@@ -374,6 +382,133 @@ bool TActionExt::ToggleMCVRedeploy(TActionClass* pThis, HouseClass* pHouse, Obje
...
+		pHouse->UpdateAngerNodes(0, nullptr);
+	}
+
+	return true;
+}
+
+bool TActionExt::SetForceEnemy(TActionClass* pThis, HouseClass* pHouse, ObjectClass* pObject, TriggerClass* pTrigger, CellStruct const& location)
+{
+	auto const pHouseExt = HouseExt::ExtMap.Find(pHouse);
+	if (!pHouseExt)
```
Coronia, line 480: no need to do sanity check since `pHouse` is always valid

src/Ext/House/Body.cpp
```diff
@@ -602,6 +602,29 @@ void HouseExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
...
+		return -1;
+
+	auto const pHouse = this->OwnerObject();
+	if (!pHouse)
+		return -1;
+
+	return this->ForceEnemyIndex;
+}
+
+void HouseExt::ExtData::SetForceEnemy(int EnemyIndex)
```
Coronia, line 617: should probably rename this function to avoid misunderstanding of `SetForceEnemy` in TAction. Could be `SetForceEnemyIndex`

Coronia (approved):

PR #1630, opened 2025-04-16
MultiWeapon

a851903106:
Technology types are free to enable WeaponX and use them as usable weapons.

- TechnoType reads `WeaponX` as their weapon when `MultiWeapon=yes`, be careful not to forget `WeaponCount`.
 - `MultiWeapon.IsSecondary` can only be used for infantry and is responsible for determining which weapons should use `SecondaryFire` in the `Sequence`.
 - `MultiWeapon.SelectWeapon` determines how many weapons can be selected. maximum value is 4, minimum value is 1, defaults to 2.

In `rulesmd.ini`:
```ini
[SOMETECHNO]          ; TechnoType
MultiWeapon=          ; boolean
MultiWeapon.SelectWeapon=          ; integer

[SOMEINFANTRY]                 ; InfantryType
MultiWeapon.IsSecondary=       ; List of integers
```
---------
科技类型可以自由启用WeaponX并将它们作为可用武器使用。

- 当`MultiWeapon=yes`时科技类型会读取`WeaponX`作为它的武器，同时注意不要忘记写`WeaponCount`。
 - `MultiWeapon.IsSecondary`只对步兵有效，它会负责判断哪些武器开火时播放副武器动作帧。
 - `MultiWeapon.SelectWeapon`决定哪些武器可以被选择。最大值为4，最小值为1，默认2。

PR #1631, opened 2025-04-16
Forcing specific weapon by target type

Coronia:
This is a supplement for https://github.com/Phobos-developers/Phobos/pull/1630. With multiple weapons being set, you can use `ForceWeapon` to enable them when attacking different targets. This pull request is a basic one that should be enough for many cases

  - A series of tags can force specified weapons based on the target's type.
    - If `ForceWeapon.Defenses` is enabled, it'll be used if the target is a building with `IsBaseDefense=yes`. Otherwise it'll follow `ForceWeapon.Buildings`, if enabled.
    - `ForceWeapon.Infantry/Units/Aircraft` can be applied to both ground and air target if `ForceAAWeapon.Infantry/Units/Aircraft` is not set.
    - `ForceAAWeapon.Infantry/Units/Aircraft` do the same things but only for air target. Taking priority to `ForceWeapon.Infantry/Units/Aircraft`, which means that they can only be applied to ground target when they're both set.

In `rulesmd.ini`:
```ini
ForceWeapon.Buildings=-1                        ; integer, -1 to disable
ForceWeapon.Defenses=-1                         ; integer, -1 to disable
ForceWeapon.Infantry=-1                         ; integer, -1 to disable
ForceWeapon.Units=-1                            ; integer, -1 to disable
ForceWeapon.Aircraft=-1                         ; integer, -1 to disable
ForceAAWeapon.Infantry=-1                       ; integer, -1 to disable
ForceAAWeapon.Units=-1                          ; integer, -1 to disable
ForceAAWeapon.Aircraft=-1                       ; integer, -1 to disable
```

### DIFF
--- a/src/Ext/Techno/Hooks.Firing.cpp
+++ b/src/Ext/Techno/Hooks.Firing.cpp
@@ -122,6 +122,53 @@ DEFINE_HOOK(0x6F3428, TechnoClass_WhatWeaponShouldIUse_ForceWeapon, 0x6)
 			forceWeaponIndex = TechnoExt::ExtMap.Find(pThis)->ApplyForceWeaponInRange();
 			ForceWeaponInRangeTemp::SelectWeaponByRange = false;
 		}
+		else
+		{
+			switch (pTarget->WhatAmI())
+			{
+				case AbstractType::Building:
+				{
+					forceWeaponIndex = pTypeExt->ForceWeapon_Buildings;
+
+					if (pTypeExt->ForceWeapon_Defenses >= 0)
+					{
+						auto const pBuildingType = abstract_cast<BuildingTypeClass*>(pTarget->GetTechnoType());
+
+						if (pBuildingType->IsBaseDefense)
+							forceWeaponIndex = pTypeExt->ForceWeapon_Defenses;
+					}
+
+					break;
+				}
+				case AbstractType::Infantry:
+				{
+					if (pTypeExt->ForceAAWeapon_Infantry >= 0 && pTarget->IsInAir())
+						forceWeaponIndex = pTypeExt->ForceAAWeapon_Infantry;
+					else
+						forceWeaponIndex = pTypeExt->ForceWeapon_Infantry;
+
+					break;
+				}
+				case AbstractType::Unit:
+				{
+					if (pTypeExt->ForceAAWeapon_Units >= 0 && pTarget->IsInAir())
+						forceWeaponIndex = pTypeExt->ForceAAWeapon_Units;
+					else
+						forceWeaponIndex = pTypeExt->ForceWeapon_Units;
+
+					break;
+				}
+				case AbstractType::Aircraft:
+				{
+					if (pTypeExt->ForceAAWeapon_Aircraft >= 0 && pTarget->IsInAir())
+						forceWeaponIndex = pTypeExt->ForceAAWeapon_Aircraft;
+					else
+						forceWeaponIndex = pTypeExt->ForceWeapon_Aircraft;
+
+					break;
+				}
+			}
+		}
 
 		if (forceWeaponIndex >= 0)
 		{
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -449,7 +449,7 @@ This page lists all the individual contributions to the project by their author.
   - Fix `DefaultDisguise` showing wrong house colors for different players
   - Fire weapon when Warhead kills something
   - Promotion animation deglobalization
-  - Forcing specific weapon by range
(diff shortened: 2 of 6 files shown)


Coronia:
besides, if you're certain that weapons of a techno won't make use of the following things, you can use this ability to make it skip previous weapon index picking logic, to save some calculation cost
- CanTarget
- CanTargetHouse
- AttachEffect.RequiredTypes/Groups
- AttachEffect.DisallowedTypes/Groups

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -80,48 +80,91 @@ DEFINE_HOOK(0x6F3428, TechnoClass_WhatWeaponShouldIUse_ForceWeapon, 0x6)
...
+		{
+			switch (pTarget->WhatAmI())
+			{
+				case AbstractType::Building:
+				{
+					forceWeaponIndex = pTypeExt->ForceWeapon_Buildings;
+
+					if (pTypeExt->ForceWeapon_Defenses >= 0)
+					{
+						auto const pBuildingType = abstract_cast<BuildingTypeClass*>(pTargetTechno->GetTechnoType());
```
CrimRecya, line 131: `auto const pBuildingType = static_cast<BuildingClass*>(pTargetTechno)->Type;` or `auto const pBuildingType = static_cast<BuildingTypeClass*>(pTargetType);`

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -80,48 +80,91 @@ DEFINE_HOOK(0x6F3428, TechnoClass_WhatWeaponShouldIUse_ForceWeapon, 0x6)
...
+				}
+				case AbstractType::Unit:
+				{
+					if (pTypeExt->ForceAAWeapon_Units >= 0 && pTarget->IsInAir())
+						forceWeaponIndex = pTypeExt->ForceAAWeapon_Units;
+					else
+						forceWeaponIndex = pTypeExt->ForceWeapon_Units;
+
+					break;
+				}
```
CrimRecya, line 156: I think you can also add a `ForceWeapon.Naval`

CrimRecya (approved):

CrimRecya (approved):
I think after updating it, we can merge this.

---

By the way, my opinion on `ForceWeapon.Naval` is probably like this, what do you think?
```
case AbstractType::Unit:
{
	forceWeaponIndex = (this->ForceAAWeapon_Units >= 0 && pTarget->IsInAir())
		? this->ForceAAWeapon_Units
		: ((this->ForceWeapon_Naval >= 0 && pTargetType->Naval)
			? this->ForceWeapon_Naval
			: this->ForceWeapon_Units);

	break;
}
```

Coronia:
> I think after updating it, we can merge this.
> 
> By the way, my opinion on `ForceWeapon.Naval` is probably like this, what do you think?
> 
> ```
> case AbstractType::Unit:
> {
> 	forceWeaponIndex = (this->ForceAAWeapon_Units >= 0 && pTarget->IsInAir())
> 		? this->ForceAAWeapon_Units
> 		: ((this->ForceWeapon_Naval >= 0 && pTargetType->Naval)
> 			? this->ForceWeapon_Naval
> 			: this->ForceWeapon_Units);
> 
> 	break;
> }
> ```

`Naval=yes` can not only be applied to unit, but also other type of techno. If we're changing it to this way I'm afraid it might be somehow misleading

CrimRecya:
> `Naval=yes` can not only be applied to unit, but also other type of techno. If we're changing it to this way I'm afraid it might be somehow misleading

In this case, I think it's necessary to further distinguish between vehicles and buildings that `Naval=yes`.

FS-21:
> > I think after updating it, we can merge this.
> > By the way, my opinion on `ForceWeapon.Naval` is probably like this, what do you think?
> > ```
> > case AbstractType::Unit:
> > {
> > 	forceWeaponIndex = (this->ForceAAWeapon_Units >= 0 && pTarget->IsInAir())
> > 		? this->ForceAAWeapon_Units
> > 		: ((this->ForceWeapon_Naval >= 0 && pTargetType->Naval)
> > 			? this->ForceWeapon_Naval
> > 			: this->ForceWeapon_Units);
> > 
> > 	break;
> > }
> > ```
> 
> `Naval=yes` can not only be applied to unit, but also other type of techno. If we're changing it to this way I'm afraid it might be somehow misleading

Since when is allowed in the code style of this project the concatenated inlined if? (I dont know if is spelled in English the xx ? value1 : value2 )

PR #1640, opened 2025-04-21
[Minor] Fix the issue where infantry get stuck in their current cell during movement

NetsuNegi:


CrimRecya (approved):
Only need testing.
Additionally, I suggest adding some descriptions/comments of bugs between the code to avoid difficulties during later maintenance.

PR #1646, opened 2025-04-25
[Minor] Fixed the issue that the default attack interval of non `Strafe` aircraft was unexpectedly changed

CrimRecya:


Starkku:
Does not really need testing, this fixes an obvious bug where incorrect value is used for Strafing.Shots if it is not explicitly set. Merging.

PR #1653, opened 2025-05-05
Unit & infantry auto-conversion on ammo change

FS-21:
- Units/infantry can now be converted into another unit/infantry by ammo count.
- `Ammo.AutoConvertMinimumAmount` determines the minimal number of ammo at which a unit converts automatically.
- `Ammo.AutoConvertMaximumAmount` determines the maximum number of ammo at which a unit converts automatically.
- `Ammo.AutoConvertType` specify the new techno after the conversion. This unit must be of the same type of the original (vehicle -> vehicle or infantry -> infantry).
- Setting a negative number will disable ammo count check.

In `rulesmd.ini`:
```ini
[SOMETECHNO]                      ; InfantryType or VehicleType
Ammo.AutoConvertMinimumAmount=-1  ; integer
Ammo.AutoConvertMaximumAmount=-1  ; integer
Ammo.AutoConvertType=             ; InfantryType or VehicleType
```

- Tested in online games.
- This auto-conversion feature requires [Ares](https://github.com/Ares-Developers/Ares).
- Based / Inspired in Frione's work for "vehicle auto-deploy / deploy block on ammo change".


Coronia:
Think it's still better to batch these conditional conversion and AutoDeath in a same condition group, so it has the following benefits:
- use same condition calculation logic, so more consistent
- more conditions can be naturally supported by this system
see https://github.com/Phobos-developers/Phobos/pull/1346

TaranDahl:
Two concerns:
1. Why are the AircraftTypes excluded?
2. The behavior of checking during the techno update does not match the description in the document (on ammo change). I think you should check it when the ammo changes or clarify the actual checking timing in the document.

FS-21:
> Two concerns:
> 
> 1. Why are the AircraftTypes excluded?
> 2. The behavior of checking during the techno update does not match the description in the document (on ammo change). I think you should check it when the ammo changes or clarify the actual checking timing in the document.

1. [SOMETECHNO] ; InfantryType or VehicleType <--- do you mean this in the docs? I never tested what happens in AircraftTypes but since it uses the Ares conversion logic is the same case of any aircraft (again, I never tested it on this kind of unit), only infantry & vehicles that was the initial scope of the feature (the additional types are welcome if are supported without additional code changes).
2. Probably docs confussion but I don't see too much the difference. Modders shouldn't know that the check process is during an internal check done every frame in every techno and is more clear that it happens after the ammo changed the value. Do you have any better idea for changing this text for clarifying the possible confussion?

FS-21:
> Think it's still better to batch these conditional conversion and AutoDeath in a same condition group, so it has the following benefits:
> 
> * use same condition calculation logic, so more consistent
> * more conditions can be naturally supported by this system
> 
> see #1346

We can re-think the code after the merge into develop of #1346 but for now I need this feature for my mod :-D

TaranDahl:
> Probably docs confussion but I don't see too much the difference. Modders shouldn't know that the check process is during an internal check done every frame in every techno and is more clear that it happens after the ammo changed the value. Do you have any better idea for changing this text for clarifying the possible confussion?

There is not much difference indeed. I think you just need to additionally explain the actual check process in the document.

TaranDahl:
> [SOMETECHNO] ; InfantryType or VehicleType <--- do you mean this in the docs? I never tested what happens in AircraftTypes but since it uses the Ares conversion logic is the same case of any aircraft (again, I never tested it on this kind of unit), only infantry & vehicles that was the initial scope of the feature (the additional types are welcome if they are supported without additional code changes).

You don't need to worry about ConvertToType. If the conversion can not be processed, it will exit safely.

src/Ext/Techno/Body.Update.cpp
```diff
@@ -116,6 +118,44 @@ void TechnoExt::ExtData::DepletedAmmoActions()
...
+		ValueableVector<TechnoTypeClass*> convertFrom;
+		Nullable<TechnoTypeClass*> convertTo;
+		Nullable<AffectedHouse> convertAffectedHouses;
+
+		convertFrom.emplace_back(pThis->GetTechnoType());
+		convertTo = pTypeExt->Ammo_AutoConvertType;
+		convertAffectedHouses = AffectedHouse::Owner;
+		convertPair.emplace_back(convertFrom, convertTo, convertAffectedHouses);
+
+		TypeConvertGroup::Convert(pFoot, convertPair, pThis->Owner);
```
TaranDahl, line 155: No need for these, you can call TechnoExt::ConvertToType.

TaranDahl (approved):
LGTM. Only one small concern.

src/Ext/Techno/Body.Update.cpp
FS-21: Thanks!!! I didn't see this method existed so I had to figure out how this thing should work.
I already replaced the code & tested it :-)

src/Ext/Techno/Body.Update.cpp
```diff
@@ -116,6 +118,34 @@ void TechnoExt::ExtData::DepletedAmmoActions()
...
+	auto const pThis = abstract_cast<TechnoClass*>(this->OwnerObject());
+	if (!pThis || pThis->GetTechnoType()->Ammo <= 0)
+		return;
+
+	const auto pTypeExt = this->TypeExtData;
+	const bool skipMinimum = pTypeExt->Ammo_AutoConvertMinimumAmount < 0;
+	const bool skipMaximum = pTypeExt->Ammo_AutoConvertMaximumAmount < 0;
+
+	if (skipMinimum && skipMaximum)
+		return;
```
CrimRecya, line 132: - There's no need to check `pThis`, and it's already `TechnoClass*`, so there's also no need to convert it.
- In addition, considering the `TypeExtData` pointer, it can be checked first, and other temporarily unused ones can be put later.
- Also a minor code style thing: just choose one of `const auto` and `auto const`, do not mix them.
```
	const auto pTypeExt = this->TypeExtData;
	if (pTypeExt->OwnerObject()->Ammo <= 0)
		return;

	const bool skipMinimum = pTypeExt->Ammo_AutoConvertMinimumAmount < 0;
	const bool skipMaximum = pTypeExt->Ammo_AutoConvertMaximumAmount < 0;

	if (skipMinimum && skipMaximum)
		return;

	const auto pThis = this->OwnerObject();
```

src/Ext/Techno/Body.Update.cpp
```diff
@@ -116,6 +118,34 @@ void TechnoExt::ExtData::DepletedAmmoActions()
...
+
+	const bool moreThanMinimum = pThis->Ammo >= pTypeExt->Ammo_AutoConvertMinimumAmount;
+	const bool lessThanMaximum = pThis->Ammo <= pTypeExt->Ammo_AutoConvertMaximumAmount;
+
+	if ((skipMinimum || moreThanMinimum) && (skipMaximum || lessThanMaximum))
```
CrimRecya, line 137: - There's not a problem here, the variable names are very clear, and the compiler should help us put the judgment later, so that it will only check the condition when needed. You can also do this manually.
```
	if ((skipMinimum || pThis->Ammo >= pTypeExt->Ammo_AutoConvertMinimumAmount) // More than minimum
		&& (skipMaximum || pThis->Ammo <= pTypeExt->Ammo_AutoConvertMaximumAmount)) // Less than maximum
```

src/Ext/Techno/Body.Update.cpp
```diff
@@ -116,6 +118,34 @@ void TechnoExt::ExtData::DepletedAmmoActions()
...
+	{
+		const auto pFoot = abstract_cast<FootClass*>(pThis);
```
CrimRecya, line 139: - Recently, `abstract_cast` updated its usage. If you are certain that the pointer is not null, you can add a template parameter `true` after the converted type, which will skip checking for null pointer. For example, `pThis` is definitely not empty, so use:
```
const auto pFoot = abstract_cast<FootClass*, true>(pThis);
```
- In this way, it will only judge `pThis->AbstractFlags & AbstractFlags::Foot`.

CrimRecya (approved):

src/Ext/Techno/Body.Update.cpp
FS-21: Because I started the code with a copy/paste from a similar function, probably DepletedAmmoActions() :-D

src/Ext/Techno/Body.Update.cpp
FS-21: I have no idea what values can take pThis here, only I knew I had to write my code here. So by defaut I added the respective sanity check to prevent crashes.
If you are convinced that here pThis always has a value I'll apply your change.

FS-21: ah no, ignore my comment, I won't add this change because I have another feature that needs to know if isn't a FootClass object (a BuildingClass).

src/Ext/Techno/Body.Update.cpp
```diff
@@ -137,7 +137,7 @@ void TechnoExt::ExtData::AmmoAutoConvertActions()
 	if ((skipMinimum || pThis->Ammo >= pTypeExt->Ammo_AutoConvertMinimumAmount) // More than minimum
 	&& (skipMaximum || pThis->Ammo <= pTypeExt->Ammo_AutoConvertMaximumAmount)) // Less than maximum
 	{
-		const auto pFoot = abstract_cast<FootClass*>(pThis);
+		const auto pFoot = pThis && (pThis->AbstractFlags & AbstractFlags::Foot) ? static_cast<FootClass*>(pThis) : nullptr;
```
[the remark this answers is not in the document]
FS-21: But you are suggesting a code change or simply explaining the case for future situations? is for going direct & closing the pending conversation :-P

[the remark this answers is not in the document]
CrimRecya: You don't need to do sanity check on `pThis`. You have already used `pThis->Ammo` and are even still using `pThis->GetTechnoType()`, it cannot be empty.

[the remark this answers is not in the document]
CrimRecya: > if pThis is a FootClass object or something else for future things.

Sorry, I didn't understand what you meant. `pThis` is a `TechnoClass` pointer and the converted `pFoot` is a `FootClass` pointer. Unless it is not a `Foot` (is `Building`), they all point to the same entity.

DeathFishAtEase:
> **In Doc**:
> Aircrafts weren't tested but if Ares supported them then this feature should work on them too.

NeverUse

In `rulesmd.ini`:
```ini
[BEAG]
Ammo.AutoConvertType=ORCA
Ammo.AutoConvertMinimumAmount=0

[ORCA]
Ammo.AutoConvertType=BEAG
Ammo.AutoConvertMinimumAmount=1
Ammo.AutoConvertMaximumAmount=1
```

> [!NOTE]
> Since the Pull Request remained in the `Needs discussion` state before merging, this additional comment is for testing/proof of the feature's work status on AircraftTypes.

PR #1657, opened 2025-05-07
[Minor] Change `Strafing` default value to false if set `Trajectory`

CrimRecya:
Only change the default value.
If anyone finds it inappropriate, please provide feedback.


Starkku:
The previous behaviour is technically consistent with vanilla behaviour where only `ROT` (which defaults to 0 which allows for strafing) and `Inviso` and using something like `Arcing` alone *would* default to strafing behaviour. Not defaulting to strafing would probably make more sense so I do not object to this change, altho it will introduce a minor inconsistency between `Trajectory` and `Arcing` (but who even uses this on aircraft, not sure if it even works correctly).

Regardless of the outcome the change should be noted in migration guide in changelog.

PR #1658, opened 2025-05-08
[Minor] Make PassengerDeletion trigger UpdatePower for Bio Reactor

Coronia:
previously using `PassengerDeletion` on Bio Reactor can't change its power output until another recheck event happens, which should be fixed now. Also removed redundant checks in `LimboCreate` since `RecheckPower` already cover them

### DIFF
--- a/src/Ext/SWType/FireSuperWeapon.cpp
+++ b/src/Ext/SWType/FireSuperWeapon.cpp
@@ -76,10 +76,8 @@ inline void LimboCreate(BuildingTypeClass* pType, HouseClass* pOwner, int ID)
 		pBuilding->DiscoveredBy(pOwner);
 
 		pOwner->RegisterGain(pBuilding, false);
-		pOwner->UpdatePower();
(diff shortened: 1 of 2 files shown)


src/Misc/Hooks.BugFixes.cpp
```diff
@@ -1837,6 +1837,8 @@ DEFINE_HOOK(0x51A304, InfantryClass_UpdatePosition_EnterBioReactorSound, 0x6)
 	enum { SkipGameCode = 0x51A30A };
 
 	GET(BuildingClass*, pReactor, EDI);
+	GET(FootClass*, pFoot, ESI);
+	pFoot->Transporter = pReactor;
 	const int enterSound = pReactor->Type->EnterBioReactorSound;
```
Thrifinesma, line 1842: Since the Hooks now have more functionalities, I suggest changing theri names, for example, by removing 'Sound'.

NetsuNegi (approved):

PR #1661, opened 2025-05-09
[Vanilla Fix] Fix BalloonHover incorrectly considering ground factors when pathfinding

TaranDahl:
Fixed the bug where technos with `BalloonHover=yes` incorrectly considered ground factors when setting the destination and distributing moving commands. Use `[General]->BalloonHoverPathingFix=true` to enable this fix.


Metadorius:
What does this impact? Maybe the fix could be on by default?

NetsuNegi:
What did this fix

TaranDahl:
image
In vanilla, this is not allowed.
image
In vanilla, these commands will be distributed only on the bridge, because the clicked cell is on the bridge. This makes sense on the ground technos, but not for BalloonHovers. Also happens on cliff.

TaranDahl:
New commit:
27d25c11525cddf1ed6fe4f40ee8bfdb
Vanilla behavior👆
cb7b6a8178d67811ac1973331d79d676
New behavior👆
Pay attention to how BalloonHovers selects the destination when attacking the target.

Metadorius:
Looks to me like the fix could be on by default, and would just have to add migration notice to disable it if undesired.

Coronia:
image
These hooks didn't perform very well during performance test with profiler. But I guess there's no much options to optimize the code logic since it still needs to go through the hooks and check no matter if `BalloonHoverPathingFix` is enabled or not

Is it possible to move the hooks to another places that `BalloonHover` has already been checked?

TaranDahl:
> image These hooks didn't perform very well during performance test with profiler. But I guess there's no much options to optimize the code logic since it still needs to go through the hooks and check no matter if `BalloonHoverPathingFix` is enabled or not
> Is it possible to move the hooks to another places that `BalloonHover` has already been checked?-----------谷歌翻译-----------是否可以将钩子移至已经检查过BalloonHover的另一个地方？

It seems unlikely because vanilla doesn't check for BalloonHover in these functions.

TaranDahl:
I placed the IsInAir check at the end, which may slightly improve performance.

Metadorius:
using a naked hook may improve the situation

TaranDahl:
> using a naked hook may improve the situation-----------谷歌翻译-----------使用裸钩可能会改善情况

Dunno how to do it. Any example?

Metadorius:
> Dunno how to do it. Any example?

Search for DEFINE_NAKED_HOOK in code.

TaranDahl:
> > Dunno how to do it. Any example?
> 
> Search for DEFINE_NAKED_HOOK in code.-----------谷歌翻译-----------在代码中搜索Define_naked_hook。

Only one use case:

DEFINE_NAKED_HOOK(0x7CD8EA, _ExeTerminate)
{
	// Call WinMain
	SET_REG32(EAX, 0x6BB9A0);
	CALL(EAX);
	PUSH_REG(EAX);

	__asm {call Phobos::ExeTerminate};

	// Jump back
	POP_REG(EAX);
	SET_REG32(EBX, 0x7CD8EF);
	__asm {jmp ebx};
}

Should I write asm like that?

Metadorius:
> Should I write asm like that?

You're not really limited to assembly. You should just keep in mind that a naked hook is simply a jump to the code you write, no prolog, epilog or anything like that, so you should care to not mess up the stack, registers etc.

ZivDero:
If you want to look at naked hooks, you could look at Vinifera. It's *all* naked hooks there :P

TaranDahl (commented):
Kerb said this option should be enabled by default🤔

Metadorius:
@ZivDero maybe you could assist with a naked hook?

docs/Fixed-or-Improved-Logics.md
```diff
@@ -232,6 +232,7 @@ This page describes all ingame logics that are fixed or improved in Phobos witho
 - Infantry support `IsGattling=yes`.
 - Fixed the issue that the widespread damage caused by detonation on the bridge/ground cannot affect objects on the ground/bridge who are in the opposite case.
 - Fixed the bug that `DamageSelf` and `AllowDamageOnSelf` are ineffective on airforce.
+- Fixed the bug where technos with `BalloonHover=yes` incorrectly considered ground factors when setting the destination and distributing moving commands. If you don't want this anyway, use `[General]->BalloonHoverPathingFix=false` to disable this fix.
```
DeathFishAtEase: 1. `]_->_Ball...` (Maintain consistency) (`_` -> ` `)
2. Suggest writing the method to restore vanilla behavior to the `Whats-New.md/Migrating/From vanilla` section rather than here.

ZivDero (approved):
LGTM

PR #1662, opened 2025-05-10
[Minor] Update on Arcing.AllowElevationInaccuracy

Coronia:
The previous `Arcing.AllowElevationInaccuracy` fix will make the projectile fall and detonate at different positions, which looks awkward in many times
a9e84132490cddd7bd7a04df3833a55a

This fix is based on https://github.com/Phobos-developers/Phobos/pull/269 and https://github.com/Phobos-developers/Phobos/pull/270, which will adjust the projectile's trace to make them fall and detonate at the same elevated position, also enabled by setting `Arcing.AllowElevationInaccuracy=false`.

Also includes some bullet optimizations from https://github.com/Phobos-developers/Phobos/pull/1568

### DIFF
--- a/src/Ext/Bullet/Body.cpp
+++ b/src/Ext/Bullet/Body.cpp
@@ -334,13 +334,13 @@ void BulletExt::SimulatedFiringUnlimbo(BulletClass* pBullet, HouseClass* pHouse,
 		const auto targetCoords = pBullet->Target->GetCenterCoords();
 		const auto gravity = BulletTypeExt::GetAdjustedGravity(pType);
 		const auto distanceCoords = targetCoords - sourceCoords;
-		const auto horizontalDistance = Point2D{distanceCoords.X, distanceCoords.Y}.Magnitude();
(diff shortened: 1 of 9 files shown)


CrimRecya (changes requested):
There are two issues. Then just need testing.

src/Ext/Bullet/Body.cpp
```diff
@@ -390,6 +390,51 @@ void BulletExt::SimulatedFiringEffects(BulletClass* pBullet, HouseClass* pHouse,
...
+	bool elevationFix = !this->TypeExtData->Arcing_AllowElevationInaccuracy;
+
+	if (!inaccutate && !elevationFix)
+		return;
+	
+	auto theSourceCoords = pThis->SourceCoords;
+	auto theTargetCoords = pThis->TargetCoords;
+	const auto pWeapon = pThis->WeaponType;
+
+	if (inaccutate)
```
Metadorius, line 407: typo, should be inaccurate

PR #1665, opened 2025-05-12
[minor] Fix `OmniFire.TurnToTarget=true` not working when the unit is rearming

TaranDahl:


Coronia:
could we make it enable by `NoTurret.TrackTarget` so it's more consistent? Previously this tag can't work on `OmniFire=yes` weapon, and now that the fix is made, it'd be better to still let this tag control whether or not a unit should turn to its target when rearming

Coronia (approved):

PR #1669, opened 2025-05-12
[Minor] Weekly Regular Documentation Revisions 05.12-05.18 - develop

DeathFishAtEase:
For non-urgent documentation modifications, they should now be submitted through this dedicated Pull Request and will undergo weekly batch merging.

Documentation changes associated with other feature/bugfix pull requests should remain within their respective Pull Requests.

For example: If a Pull Request gets merged into the main branch without completing required documentation updates, the content should first be synchronized to this dedicated documentation branch. Subsequent corrections will then be made within this dedicated documentation Pull Request.

---
对于并不急需的文档修改现改为使用这个拉取请求进行提交，每周进行一次合并。

对于其他拉取请求中的文档修改则跟随其拉取请求。

例如某个拉取请求在合并前没有处理完毕文档问题，那么在它合并到主分支后，将内容更新到这个拉取请求所在的分支，然后在这个拉取请求所在的分支进行进一步更正


Starkku:
> @Starkku The description of `CanC4.AllowZeroDamage` seems problematic. I noticed your original description stated that
> 
>     * In YR, if `Damage <= 0`, it would be force-written to _1_.
> 
>     * Now, if `Damage < 0`, this rule is no longer applied.
>       
>       * When `Damage = 0`, if `CanC4.AllowZeroDamage = true`, this rule is also disabled.
> 
> 
> However, actual testing shows:
> 
> For `Damage = 0`, in YR, it inherently does no damage to BuildingType with `CanC4 = false`.
> 
> I now want to confirm: Is this flag working as intended?

The feature works correctly but the documentation is very unclear and should be rewritten.

`CanC4=false` on building makes building take atleast 1 point of damage **if** the raw damage is non-zero but is lowered to below 1 by `Verses` etc. `CanC4.AllowZeroDamage=true` disables this. Negative damage (that is, after `Verses` etc have been applied) also now bypasses this check entirely without having to enable anything.

DeathFishAtEase:
> The feature works correctly but the documentation is very unclear and should be rewritten.
> 
> `CanC4=false` on building makes building take atleast 1 point of damage **if** the raw damage is non-zero but is lowered to below 1 by `Verses` etc. `CanC4.AllowZeroDamage=true` disables this. Negative damage (that is, after `Verses` etc have been applied) also now bypasses this check entirely without having to enable anything.

I will update this content in the new week's document maintenance Pull Request.

PR #1670, opened 2025-05-16
Battle Points economy for super weapons

FS-21:
- This system displayes a new currency to be used (optionally) in Super Weapons.
- The new currency will be modified by the kills against other players and the system is enabled.
- If set `BattlePointsCollector` in any structure then the system will be enabled if the structure is built.
- If set `BattlePoints` in houses they will have the system enabled by default.
- If set [General] -> `BattlePoints` enables/disables the system for every player in the scenario, overriding the country preferences & `BattlePointsCollector` setting.
- If set [General] -> `BattlePoints.DefaultValue` and the enemy destroyed object doesn't have a value then this generic value will be used instead.
- If set [General] -> `BattlePoints.DefaultFriendlyValue` and the friendly destroyed object doesn't have a value then this generic value will be used instead. This tag doesn't work with the own objects of the player.
- If is set `BattlePoints.CanUseStandardPoints` and the destroyed object doesn't have the `BattlePoints` value then the points are obtained from the `Points` tag of the destroyed object. If `BattlePoints.DefaultValue` is present then this will be ignored.
- If set `BattlePoints` in the destroyed object the calculation is made with this value. Self-Kills done by the own player doesn't count in this system.
- If `BattlePoints.Amount` is set with a value different from `0` then the super weapon is influenced by this system. If the value is positive then this super weapon won't be launched until the affected player gathered the required ammount.

In `uimd.ini`:
```ini
[Sidebar]
BattlePointsSidebar.Label=<none>                ; CSF entry key `★ <VALUE>`, code U+2605
BattlePointsSidebar.Label.InvertPosition=false  ; bool, `<VALUE> ★`, code U+2605

[ToolTips]
BattlePoints.Label=<none>                       ; CSF entry key, default to `★: <VALUE>`, code U+2605
```

In `rulesmd.ini`:
```ini
[General]
BattlePoints=                            ; bool
BattlePoints.DefaultValue=               ; int
BattlePoints.DefaultFriendlyValue=       ; int

[SOMESIDE]                               ; Side
Sidebar.BattlePoints.Offset=0,0          ; X,Y, pixels relative to default
Sidebar.BattlePoints.Color=              ; integer - R,G,B
Sidebar.BattlePoints.Align=Left          ; Left, Right, Center/Centre

[SOMECOUNTRY]                            ; HouseType
BattlePoints=false                       ; bool
BattlePoints.CanUseStandardPoints=false  ; bool

[SOMEBUILDING]                           ; BuildingType
BattlePointsGenerator=                   ; bool

BattlePoints=                            ; int

[SOMESW]
BattlePoints.Amount=0                    ; int
```


FS-21:
battlepoints-demo-02

docs/New-or-Enhanced-Logics.md
```diff
@@ -967,6 +967,54 @@ In `rulesmd.ini`:
 AISuperWeaponDelay=  ; integer, game frames
 ```
 
+### Battle Points economy for super weapons
+- This system displayes a new currency to be used (optionally) in Super Weapons.
+- The new currency will be modified by the kills against other players and the system is enabled.
+- If set `BattlePointsGenerator` in any structure then the system will be enabled if the structure is built.
```
TaranDahl, line 973: `BattlePointsCollector` or `BattlePointsCounter` may describe this function more accurately. For the generation comes from kills rather than the generator.

FS-21: Since a structure enables the modification of that value then BattlePointsCollector maybe be the valid one.

FS-21:
Renamed tag from BattlePointsGenerator to BattlePointsCollector

TaranDahl:
It would be best if it could optionally display a FlyingString when obtaining BattlePoints, just like the bounty of Ares.

src/Ext/House/Body.cpp
```diff
@@ -1084,3 +1086,52 @@ bool HouseExt::ReachedBuildLimit(const HouseClass* pHouse, const TechnoTypeClass
...
+		{
+			enabledBattlePoints = true;
+			break;
+		}
+	}
+
+	enabledBattlePoints |= pOwnerTypeExt->BattlePoints;
+
+	if (RulesExt::Global()->BattlePoints.isset())
+		enabledBattlePoints = RulesExt::Global()->BattlePoints.Get();
```
TaranDahl, line 1115: The Nullable template should be used like this: 
enabledBattlePoints = RulesExt::Global()->BattlePoints.Get(enabledBattlePoints);

TaranDahl, line 1115: Also, these checks should be placed before looping through the vector. They are cheap, so you won't need to process the expensive ones if they are negative.

src/Ext/House/Body.cpp
```diff
@@ -1084,3 +1086,52 @@ bool HouseExt::ReachedBuildLimit(const HouseClass* pHouse, const TechnoTypeClass
...
+int HouseExt::ExtData::CalculateBattlePoints(TechnoClass* pTechno)
+{
+	if (!pTechno)
+		return 0;
+
+	const auto pThis = this->OwnerObject();
+	const auto pThisTypeExt = HouseTypeExt::ExtMap.Find(pThis->Type);
+	const auto pTechnoTypeExt = TechnoTypeExt::ExtMap.Find(pTechno->GetTechnoType());
+
+	int defaultValue = RulesExt::Global()->BattlePoints_DefaultValue.isset() ? RulesExt::Global()->BattlePoints_DefaultValue.Get() : 0;
```
TaranDahl, line 1129: same

src/Ext/House/Body.cpp
```diff
@@ -1084,3 +1086,52 @@ bool HouseExt::ReachedBuildLimit(const HouseClass* pHouse, const TechnoTypeClass
...
+	int defaultFriendlyValue = RulesExt::Global()->BattlePoints_DefaultFriendlyValue.isset() ? RulesExt::Global()->BattlePoints_DefaultFriendlyValue.Get() : 0;
```
TaranDahl, line 1130: same

src/Ext/House/Body.cpp
TaranDahl, line 1115: Also, I think it's unreasonable for a key in [General] to have the highest priority. They should usually have the lowest priority.

TaranDahl (changes requested):

FS-21:
> It would be best if it could optionally display a FlyingString when obtaining BattlePoints, just like the bounty of Ares.

I prefer keep the base foundation of the feature simple for a faster merge and in new PRs adding other features (or let other devs to develop them), like:
- Oil Derrick style for generating poings overtime.
- Points in crates
- Map Events/Actions for checking & manipulating values.
- Support for non-SW (structures & units)
- ...

src/Ext/House/Body.cpp
FS-21: No because you maybe want to hardcode easily the whole system enabled/disabled in 1 volley (for example a new game mode). If isn't set then other tags do the magic if is necessary.

TaranDahl: Yes

FS-21: At least the first one (enabledBattlePoints |= pOwnerTypeExt->BattlePoints;) because the last one forbids the use of the feature.

TaranDahl: Then I think `DisableBattlePoints` could be a better name.
And the logic would be:
if (RulesExt->DisableBattlePoints)
  return false;
if (pOwnerTypeExt->BattlePoints)
  return true;
if (HasAnyBattlePointsBuilding)
  return true;
return false;

FS-21: Ok I made some improvements so this message is outdated.

If the global setting is enabled the function returns the value.
If not but the House setting is enabled it returns the value and only if is false then checks the structures list looking for the tag.

TaranDahl: > Or the DisableBattlePoints is just for forcing the house to never use the logic (if the [General] tag isn't set) ?

Yeah.

TaranDahl: What I want to say is that, although there is no written rule, engine extension authors (as well as Westwood) tend to make global values to be the default.
Anyway it is not a big problem.

FS-21: But then without redesigning too much for the House's tag we could have a nullable bool instead of a valueable bool and do the same of the [General] global tag...

TaranDahl (approved):
LGTM.

ayylmaoRotE:
These 2 tags are not working:

BattlePointsCollector=true
BattlePointsGenerator=true

Is not working for me at the Techno level (building), can only get it to work at [General] or Side with BattlePoints=true, the documentation states otherwise that you can enable this system with a building instead - can you please confirm

FS-21:
I just tested now in my mod and it works.

With default settings, if you want a structure that enables the logic just put:
BattlePointsCollector=yes

AAAAAAAND, by default you get no points, unlike Ares bounty logic that forces you to put the value in all objects just put a value in:
[General]
BattlePoints.DefaultValue=               ; int

For example 100:
[General]
BattlePoints.DefaultValue=100

So the counter will appear and enable the count after new kills when the structure is built.
If is sold or destroyed the counter dissappears from the sidebar and resumed when built again.

If you want it always enabled in a specific house istead of waiting to build a structure then:
[SOMECOUNTRY]
BattlePoints=true

With my mod settings:
https://github.com/user-attachments/assets/f7148390-a727-41b6-9bfc-b07497f48740

Coronia:
Increase the rank of this PR to T3, since the essential of this is about adding a 'secondary resource' to the game, which need careful design for its functionality, interaction with various things and method of implementation first. We could indeed merge this PR first, but a rough structure of its future development needs to be decided I think. Hope @Metadorius and others could give some thought about it

Metadorius:
@FS-21 can you try and outline a base for new potential new resource system? perhaps introduce something like `CustomResourceType` which would have the icon definable and work from that?

Metadorius:
For now I propose to first make a mockup of how that would work in INI, and only after that implement it in code.

FS-21:
I already posted what is missing for a full resource support (ignoring the use of the resource for repairing things because is something that could hurt my head if I try to investigate).

First I would like to fix the bug regarding the WH thresholds posted few days ago before I could start thinking in how should be designed in INI

maikhoamKv:
If there is a new currency then why not set this currency to buy some special units?

FS-21:
> If there is a new currency then why not set this currency to buy some special units?

Because isn't supported yet and it was planned for Super Weapons not technos.

I was asked to re-make it into a generic system of new resources (like in warcraft or Age of Empires) but until I start that this is for now for Super Weapons.

FS-21:
I'll close this and create a new PR for the new multi-resource system requested/suggested,

PR #1671, opened 2025-05-18
[Minor] RadSite increase limit fix

NetsuNegi:


Coronia:
For those who don't know the context: Phobos' radiation damage was always wrong since the introduction of custom radiation. This pull request change it back to a similar way like vanilla, but might still have problem since there're fundamental difference between how vanilla and Phobos handling radiation. Hence, some more testing is appreciated, especially the case when multiple radiation stacking at a cell

Coronia (approved):

PR #1675, opened 2025-05-18
[Minor] Try optimize `TechnoClass_PointerGotInvalid_Airstrike`

CrimRecya:


Starkku (approved):
Appears to fix the most blatant performance problems with the old approach and I did not encounter crashes or other issues. Merging.

PR #1676, opened 2025-05-19
[Minor] Optimize script actions codes

CrimRecya:


Coronia (approved):
while I can't bother checking this line by line, the majority of this optimization looks correct and solid. Nothing strange occured during our test as well, while the profiler did suggest that it made script actions take less percentage of performance tax than it is

CrimRecya:
If there are no other issues, I will merge this tonight.

PR #1677, opened 2025-05-20
[Minor] User side laser trails switch

CrimRecya:
I'm not sure if CRC is needed, it's best to do a multiplayer online test.

In `ra2md.ini`:
```ini
[Phobos]
EnableLaserTrails=true           ; boolean
```

### DIFF
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -229,6 +229,12 @@ Due to performance concerns, unless any radiation type has `RadApplicationDelay.
 - Technos, Projectiles, and VoxelAnims can now have colorful trails of different transparency, thickness and color, which are drawn via laser drawing code.
 - Technos, Projectiles, and VoxelAnims can have multiple laser trails. For technos each trail can have custom laser trail type and FLH offset relative to turret and body.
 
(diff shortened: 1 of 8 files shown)


TaranDahl (approved):

Starkku:
User settings like this should be made with the possibility of them being toggleable in-game in mind, even though this is not currently. All of the existing user option toggles operate under this principle. This currently disables parsing of laser trails, not drawing.

CrimRecya:
> User settings like this should be made with the possibility of them being toggleable in-game in mind, even though this is not currently available. All of the existing user option toggles operate under this principle. This currently disables parsing of laser trails, not drawing.
> 
> If you only disable drawing it is a purely visual thing that does not affect game state so no additional changes to sync logic are needed.

Yes, but I just don't think anyone likes to turn on and off this in the game because those who really need to turn it off may not be able to afford the performance cost on their computer, so skipping from reading is the best performance solution.

Metadorius:
Doesn't Detail Level slider accomplish the same?

CrimRecya:
> Doesn't Detail Level slider accomplish the same?

People may prefer to turn it off rather than keep that kind of `Laser`.

PR #1678, opened 2025-05-24
Event 606: AttachEffect is attaching to a Techno

FS-21:
- Checks if an `AttachEffectType` is attaching to a techno. Doesn't work for [attached effects](New-or-Enhanced-Logics.md#attached-effects) that were attached prior to the trigger's enabling.
- To be elaborate, the event will be triggered during these occasions:
  - Self-owned effects: initial granted (triggered after `AttachEffect.InitialDelays` amount of frames), recreation (triggered after `AttachEffect.Delays` or `AttachEffect.RecreationDelays` amount of frames).
  - Effects from other sources: granted, refreshing when trying to apply the same type of attached effect to the techno.

In `mycampaign.map`:
```ini
[Events]
...
ID=EventCount,...,606,2,0,[AttachEffectType],...
...
```


Coronia (changes requested):
checking this in `AI()` function which is executed for every AE entity per frame is not a good idea, especially there's `HasAttachedEffects` call which needs to iterate all AEs of an object. Definitely not perf-friendly.

I would suggest to only check it once when the AE is attached to an object, aka `CreateAndAttach` function to avoid all the per frame calculation. But if you do want to check it per frame, then it'd better to only check the current entity when executing `AI()`, or adding a standalone function for this in `OnEarlyUpdate()` which only called `HasAttachedEffects` once.

FS-21:
We can try the simple instance approach and if I need that frequent AI check (for example for checking an attempt of a FS limpet drone using AE) always I can create & send a new PR for a new Map Event :-P

FS-21:
Discard what I said. I checked your suggestion of putting the event call in CreateAndAttach and works very well.

Coronia (approved):

PR #1682, opened 2025-05-27
New EVA voice after deploying a building

FS-21:
- You can now replace the current EVA voice when a specific building is placed/deployed.
- If any building is undeployed/sold/destroyed, the EVA voice will be evaluated again across all active player buildings defining `NewEVAVoice.Tag`.
- `NewEVAVoice.Tag` specifies the EVA voice name (defined under `[EVATypes]` in `evamd.ini` via Ares, or vanilla voices `Allied`, `Russian`, `Yuri`).
- In case of multiple buildings with different EVA voices, `NewEVAVoice.Priority` establishes a priority queue, where the building with the highest value is selected.
- `NewEVAVoice.RecheckOnDeath` controls whether to re-evaluate the active EVA voice upon building destruction or sale.
- `NewEVAVoice.InitialMessage` plays an EVA sound message to the player when a new EVA voice is activated.
- `NewEVAVoice.EndingMessage` plays an EVA sound message / sound effect with the outgoing EVA voice when it is deactivated/replaced before transitioning to the new voice.
- When no buildings with `NewEVAVoice` remain, the EVA voice automatically falls back hierarchically:
  1. `[HouseType] -> EVA.Tag` (Country-specific EVA voice)
  2. `[Side] -> EVA.Tag` (Side-specific EVA voice)
  3. Default vanilla voice according to side index (Allied, Russian, Yuri)

In `rulesmd.ini`:
```ini
[SOMESIDE]                        ; Side
EVA.Tag=                          ; EVA type name from [EVATypes] or Allied/Russian/Yuri

[SOMEBUILDING]                    ; BuildingType
NewEVAVoice.Tag=                  ; EVA type name from [EVATypes] or Allied/Russian/Yuri
NewEVAVoice.Priority=1            ; integer
NewEVAVoice.RecheckOnDeath=false  ; boolean
NewEVAVoice.InitialMessage=       ; EVA entry
NewEVAVoice.EndingMessage=        ; EVA entry
```


docs/User-Interface.md
```diff
@@ -27,6 +27,24 @@ IngameScore.WinTheme=  ; Soundtrack theme ID
...
+- `NewEVAVoice.Index` is the index of the new EVA voice. Ares is hightly recomended because these indexes are readed the new section `[EVATypes]` at `evamd.ini` created by Ares. Look at Ares documentation for more information.
+- In case of multiple structures with different EVA voices `NewEVAVoice.Priority` establish a priority queue, being the highest value the selected one.
+- `NewEVAVoice.RecheckOnDeath` re-checks a new EVA voice after the destruction/undeployment of of of these buildings.
+- `NewEVAVoice.InitialMessage` plays an EVA message to the player when a different EVA has been selected.
+
+In `rulesmd.ini`:
+```ini
+[SOMEBUILDING]                    ; BuildingType
+NewEVAVoice=                      ; boolean
+NewEVAVoice.Index=0               ; integer
```
Starkku: Get rid of the bool and use `Nullable` or negative default value index.

src/Ext/Rules/Body.cpp
```diff
@@ -313,6 +316,10 @@ void RulesExt::ExtData::LoadBeforeTypeData(RulesClass* pThis, CCINIClass* pINI)
 	}
 }
 
+namespace EVAIndexListTemp
+{
+	std::vector<std::string> EVAIndexList;
+}
```
Starkku: Is this namespace actually used anywhere?

Starkku (changes requested):
The EVA voice change logic should be generalized and put in a separate function from buildings, probably in `SideExt` or something. Firestorm has a map trigger for this purpose and I suppose someone might want to reimplement that here as well.

Some smaller things to take note of as well.

src/Ext/Rules/Body.cpp
FS-21: From initial tests, thanks for finding that garbage!

FS-21:
> The EVA voice change logic should be generalized and put in a separate function from buildings, probably in `SideExt` or something. Firestorm has a map trigger for this purpose and I suppose someone might want to reimplement that here as well.
> 
> Some smaller things to take note of as well.

Since Ares already has Map trigger actions for changing  these voices like in Firestorm I only focused in learning how to change it and apply this logic for buildings like in the video, not reimplementing again the EVA voices.

Let me see the SideExt you mentioned because I remember one of the first attempts was storing data in that place but things happened that made me move the code and I ended here.

Flactine:
Limbo Delivery buildings can't change EVA

FS-21:
> Limbo Delivery buildings can't change EVA

Yes, I never tried that so I added support now.

I updated this code and added support when a building is captured by engineers.

PR #1692, opened 2025-06-01
[Minor] Change overlays >255 to use NewINIFormat == 5

ZivDero:
Changes overlays with ID > 255 to be written in the same `[OverlayPack]` but as words, and with NewINIFormat == 5, to be in line with Vinifera and upcoming WAE changes so that this can actually be made use of.


Metadorius (changes requested):
docs where

ZivDero:
> docs where

Docs here

src/Misc/Hooks.Overlay.cpp
```diff
@@ -108,32 +131,24 @@ struct OverlayWriter
...
-		bytes[2] = ((nOverlay >> 16) & 0xFF);
-		bytes[3] = ((nOverlay >> 24) & 0xFF);
-		ByteWriters[0].Put(bytes[0]);
-		ByteWriters[1].Put(bytes[1]);
-		ByteWriters[2].Put(bytes[2]);
-		ByteWriters[3].Put(bytes[3]);
+		if (ScenarioClass::NewINIFormat >= 5)
+			ByteWriter.PutWord(static_cast<unsigned char>(nOverlay));
+		else
+			ByteWriter.PutByte(static_cast<unsigned short>(nOverlay));
```
Coronia: apparently PutWord is for unsigned short while PutByte is for unsigned char here, is it being opposite here? If not, then the compile warning should be fixed

Coronia (approved):

PR #1693, opened 2025-06-02
[Minor] Fix `ForceWeapon.InRange` not taking effect when attacking non-techno

CrimRecya:
- Also encapsulated the function for #1631 to use.
  > 顺便封装了函数，以便 #1631 使用。


Coronia (commented):
It’d better to add a tag to customize whether ForceWeapon.InRange should affect empty ground, cuz 1) other ForceWeapon doesn’t really work on empty ground, so this is an exception, and 2) there’re cases when modders only want to use it on units only

Coronia (approved):
add doc for `ForceWeapon.InRange.TechnoOnly` and we're good

PR #1698, opened 2025-06-04
EMPulseCannon.InaccurateRadius for projectiles in EMPulseCannon=yes superweapons and Ammo support

FS-21:
- Buildings with `EMPulseCannon=true` & `Ammo` now fire repeatedly until the weapon no longer has enough ammo.
- `EMPulseCannon.InaccurateRadius` defines an area around the target location where the projectile can land randomly.

In `rulesmd.ini`:
```ini
[PROJECTILE]                 ; Projectile
EMPulseCannon.InaccurateRadius=0 ; integer
```


Coronia:
Think it would be better to add an extra tag on the weapon or superweapon to control its maximum shot too, which could allow some more customization. E.g. an EMPulseCannon that can only fire once at a time, but has multiple charges after reloading more ammo

FS-21:
So instead of using the same weapon's ammo for consuming the structure's ammo and limiting the number of shots of that SW instance a new variable should be created to control the number of shots (as a counter) until the max number of shots is reached for that SW instance.
And then leave the structure's ammo for knowing how many instances this structure's SW can fire a EMPCannon?

Also, Ares has some tags with a similar name (shots or similar) so for avoiding confussion should be something like [SomeSW] -> EMPulseCannon.MaxShots or a better name.

Also, I thing a new label in the sidebar description tooltip should show the current available SW shots (the structure's ammo).

If I'm not wrong from your example I need to backport the SW method IsReady rewrite I the Battle points PR and include checks for EMPCannon ammo and maybe other small checks now I can't see yet.

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -1047,3 +1049,89 @@ DEFINE_HOOK(0x5223B3, InfantryClass_Approach_Target_DeployFireWeapon, 0x6)
...
+
+	if (radius > 0)
+	{
+		if (pExt->RandomEMPTarget == CellStruct::Empty)
+		{
+			// Calculate a new valid random target coordinate
+			do
+			{
+				pExt->RandomEMPTarget.X = (short)ScenarioClass::Instance->Random.RandomRanged(pHouse->EMPTarget.X - radius, pHouse->EMPTarget.X + radius);
+				pExt->RandomEMPTarget.Y = (short)ScenarioClass::Instance->Random.RandomRanged(pHouse->EMPTarget.Y - radius, pHouse->EMPTarget.Y + radius);
```
TaranDahl: I think it should be distributed in a circular shape rather than a square shape.
This clearly cannot be counted as any 'radius'.

docs/New-or-Enhanced-Logics.md
```diff
@@ -1129,12 +1129,17 @@ TabIndex=1  ; integer
   - Note that if you fire another `Type=EMPulse` superweapon with different weapon index that the same building is capable of launching before the first weapon was fired, the latter superweapon's settings will take precedence.
   - Additionally, due to technical limitations the targeting constraints will always default to primary weapon's `Range/MinimumRange` unless `SW.RangeMinimum` / `SW.RangeMaximum` are explicitly set.
 - Is is now also possible to have all other `Type=EMPulse` superweapons that can be fired by same buildings as current one be put on hold until first of the buildings currently set to fire goes off if the firing superweapon has `EMPulse.SuspendOthers=true`.
+- Structures with `Ammo` now fire repeatedly until the weapon no longer has enough ammo.
```
TaranDahl: I think it's unreasonable to bind the number of shots to Ammo. 
You should define the number of fire times on SW. As for whether the EMP cannon's firing should abide by Ammo or FireError, that should be another matter.

TaranDahl (changes requested):

src/Ext/Techno/Hooks.Firing.cpp
FS-21: changed

docs/New-or-Enhanced-Logics.md
FS-21: I changed the model and as Corina mentioned now it can be set a Burst for multiple shots. if the structure has low power in the middle of the burst it will resume when it gets back online. no ammo is used anymore

FS-21:
Uploaded the new tested version. I rejected the ammo idea, created a burst tag and it can pause & resume shots if the structure is offline/online.

fagonghaiwo:
boom when
[WallbusterP]
EMPulseCannon.InaccurateRadius=5
Eip:007564B4



my fault, I forget image

fagonghaiwo:
10952b902abfe5d3fb4932ac4e9f7e47

need doc

FS-21:
image
not related to this feature.

PR #1701, opened 2025-06-05
Customize the scatter caused by aircraft attack mission

TaranDahl:
### Customize the scatter caused by aircraft attack mission

- In vanilla, when an aircraft attacks, it forces the target's cell to trigger a scatter. Now you can disable this behavior by setting the following flag to `false`.

In `rulesmd.ini`:
```ini
[SOMEAIRCRAFT]            ; AircraftType
FiringForceScatter=true   ; boolean
```


TaranDahl:
> We had fixed the bug that aircraft will scatter objects in target cell which should not scatter. Do we still need this feature? Maybe it will be better if can specific for some aircraft instead of all.

Changed.

NetsuNegi (approved):

PR #1706, opened 2025-06-08
Electric trail for laser tails

NetsuNegi:
- It also can drawn as electric bolt if `IsElectricBolt=true`.
- 也可以设置 `IsElectricBolt=true` 来使其绘制成电弧。

[SOMETRAIL]                      ; LaserTrailType name
IsElectricBolt=false             ; boolean
IsAlternateColor=false          ; boolean
FadeDuration=                  ; integer, default 17 for electric, 64 for laser

### DIFF
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -228,6 +228,7 @@ Due to performance concerns, unless any radiation type has `RadApplicationDelay.
 
 - Technos, Projectiles, and VoxelAnims can now have colorful trails of different transparency, thickness and color, which are drawn via laser drawing code.
 - Technos, Projectiles, and VoxelAnims can have multiple laser trails. For technos each trail can have custom laser trail type and FLH offset relative to turret and body.
+- It also can drawn as electric bolt if `IsElectricBolt=true`.
 
 In `artmd.ini`:
 ```ini
@@ -237,8 +238,10 @@ In `artmd.ini`:
 [SOMETRAIL]                      ; LaserTrailType name
 IsHouseColor=false               ; boolean
 Color=255,0,0                    ; integer - R,G,B
-FadeDuration=64                  ; integer
 Thickness=4                      ; integer
+IsElectricBolt=false             ; boolean
+IsAlternateColor=false          ; boolean
+FadeDuration=                  ; integer, default 17 for electric, 64 for laser
 SegmentLength=128                ; integer, minimal length of each trail segment
 IgnoreVertical=false             ; boolean, whether the trail won't be drawn on vertical movement
 IsIntense=false                  ; boolean, whether the laser is "supported" (AKA prism forwarding)
--- a/src/New/Entity/LaserTrailClass.cpp
+++ b/src/New/Entity/LaserTrailClass.cpp
@@ -1,4 +1,5 @@
 #include "LaserTrailClass.h"
+#include <EBolt.h>
 
 #include <Utilities/TemplateDef.h>
 
@@ -17,15 +18,25 @@ bool LaserTrailClass::Update(CoordStruct location)
 	{
 		if (this->Visible && !this->Cloaked && (this->Type->IgnoreVertical ? (abs(location.X - this->LastLocation.Get().X) > 16 || abs(location.Y - this->LastLocation.Get().Y) > 16) : true))
 		{
-			// We spawn new laser segment if the distance is long enough, the game will do the rest - Kerbiter
-			LaserDrawClass* pLaser = GameCreate<LaserDrawClass>(
-				this->LastLocation.Get(), location,
-				this->CurrentColor, ColorStruct { 0, 0, 0 }, ColorStruct { 0, 0, 0 },
-				this->Type->FadeDuration.Get());
+			if (!this->Type->IsElectricBolt)
+			{
+				// We spawn new laser segment if the distance is long enough, the game will do the rest - Kerbiter
+				LaserDrawClass* pLaser = GameCreate<LaserDrawClass>(
+					this->LastLocation.Get(), location,
+					this->CurrentColor, ColorStruct { 0, 0, 0 }, ColorStruct { 0, 0, 0 },
+					this->Type->FadeDuration.Get(64));
 
-			pLaser->Thickness = this->Type->Thickness;
-			pLaser->IsHouseColor = true;
-			pLaser->IsSupported = this->Type->IsIntense;
+				pLaser->Thickness = this->Type->Thickness;
+				pLaser->IsHouseColor = true;
+				pLaser->IsSupported = this->Type->IsIntense;
+			}
+			else
+			{
+				const auto pBolt = GameCreate<EBolt>();
+				pBolt->Lifetime = 1 << (std::clamp(this->Type->FadeDuration.Get(17), 1, 31) - 1);
+				pBolt->AlternateColor = this->Type->IsAlternateColor;
+				pBolt->Fire(this->LastLocation, location, 0);
+			}
 
 			result = true;
 		}
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -403,8 +403,9 @@ This page lists all the individual contributions to the project by their author.
   - Original `Arcing` elevation inaccuracy fix
   - Fix the bug that uncontrolled scatter when elite techno attacked by aircraft or some unit try crush it
   - Exclusive SuperWeapon Sidebar
-  - Fix the bug that AlphaImage remained after unit entered tunnel.
+  - Fix the bug that AlphaImage remained after unit entered tunnel
   - Weapon target filtering by health percentage
+  - Electric trail for laser tails
 - **Apollo** - Translucent SHP drawing patches
 - **ststl**:
   - Customizable `ShowTimer` priority of superweapons
--- a/docs/Whats-New.md
+++ b/docs/Whats-New.md
@@ -388,6 +388,7 @@ New:
 - [Power plant damage factor](Fixed-or-Improved-Logics#power-plant-damage-factor) (by Otamaa and Ollerus)
 - Customize whether `Crater=yes` animation would destroy tiberium (by TaranDahl)
 - Weapon target filtering by health percentage (by NetsuNegi)
+- Electric trail for laser tails (by NetsuNegi)
 
 Vanilla fixes:
 - Fixed sidebar not updating queued unit numbers when adding or removing units when the production is on hold (by CrimRecya)
--- a/src/New/Type/LaserTrailTypeClass.cpp
+++ b/src/New/Type/LaserTrailTypeClass.cpp
@@ -17,9 +17,12 @@ void LaserTrailTypeClass::LoadFromINI(CCINIClass* pINI)
 
 	this->IsHouseColor.Read(exINI, section, "IsHouseColor");
 	this->Color.Read(exINI, section, "Color");
+	this->Thickness.Read(exINI, section, "Thickness");
+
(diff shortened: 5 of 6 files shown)


Metadorius:
I think we would need an enum to define a type instead of a bool, because RadBeam can also be used

Metadorius:
Is there also a way to apply EBolt customisations that Ares has?

NetsuNegi:
> Is there also a way to apply EBolt customisations that Ares has?

yes, but the color of bolt is define on weapon

NetsuNegi:
> Is there also a way to apply EBolt customisations that Ares has?

If you want define bolt color in here, maybe we need add a weapon, or takeover Ares's hook

NetsuNegi:
> I think we would need an enum to define a type instead of a bool, because RadBeam can also be used

like enum `LaserTrailDrawType` ?

Metadorius:
> If you want define bolt color in here, maybe we need add a weapon, or takeover Ares's hook

not sure but maybe if there isn't a lot of code could indeed take it over, also there are more customisations in Phobos already applied... begs for a `EBoltType` to be honest but I am not sure about backwards compatibility

CrimRecya (approved):

src/Ext/Bullet/Body.cpp
```diff
@@ -258,12 +259,33 @@ inline void BulletExt::SimulatedFiringElectricBolt(BulletClass* pBullet)
...
 		return;
 
-	const auto pEBolt = (AresFunctions::CreateAresEBolt ? AresFunctions::CreateAresEBolt(pWeapon) : GameCreate<EBolt>());
-	pEBolt->AlternateColor = pWeapon->IsAlternateColor;
-	auto& weaponStruct = WeaponTypeExt::BoltWeaponMap[pEBolt];
-	weaponStruct.Weapon = WeaponTypeExt::ExtMap.Find(pWeapon);
-	weaponStruct.BurstIndex = 0;
-	pEBolt->Fire(pBullet->SourceCoords, (pBullet->Type->Inviso ? pBullet->Location : pBullet->TargetCoords), 0);
+	const auto pBolt = GameCreate<EBolt>();
+	const auto pBoltExt = EBoltExt::ExtMap.Allocate(pBolt);
```
ZivDero: Why is the bolt extension allocated manually, and not by the bolt constructor? Seems unreliable

docs/New-or-Enhanced-Logics.md
```diff
@@ -228,17 +228,28 @@ Due to performance concerns, unless any radiation type has `RadApplicationDelay.
 
 - Technos, Projectiles, and VoxelAnims can now have colorful trails of different transparency, thickness and color, which are drawn via laser drawing code.
 - Technos, Projectiles, and VoxelAnims can have multiple laser trails. For technos each trail can have custom laser trail type and FLH offset relative to turret and body.
+- LaserTrail can now drawn as other behaviour like EBolt or RadBeam.
```
ZivDero: Does it make *sense* to mix laser trails, e bolts and rad beams? If so, explain why.
Unless they are the same class in vanilla (which at least for EBolts is not the case), I don't see why they should be intermixed in this way?

src/Ext/EBolt/Hooks.cpp
```diff
@@ -0,0 +1,241 @@
...
+#include <Ext/WeaponType/Body.h>
+#include <Utilities/Macro.h>
+#include <Utilities/AresHelper.h>
+#include <Helpers/Macro.h>
+
+int* EBoltExt::AresBoltColor1 = nullptr;
+int* EBoltExt::AresBoltColor2 = nullptr;
+int* EBoltExt::AresBoltColor3 = nullptr;
+
+namespace BoltTemp
```
ZivDero, line 10: This whole thing is hacks upon hacks upon hacks... Maybe just replace EBoltClass completely at this point? Just make NewEBoltClass and substitute it everywhere it's used

ZivDero: @Metadorius

src/Ext/EBolt/Hooks.cpp
```diff
@@ -0,0 +1,241 @@
...
+	GET_STACK(WeaponTypeClass*, pWeapon, STACK_OFFSET(0x30, 0x8));
+	GET_STACK(EBolt*, pBolt, STACK_OFFSET(0x30, -0x20));
+
+	if (const auto particle = WeaponTypeExt::ExtMap.Find(pWeapon)->Bolt_ParticleSystem.Get(RulesClass::Instance->DefaultSparkSystem))
+		GameCreate<ParticleSystemClass>(particle, pBolt->Point2, nullptr, nullptr, CoordStruct::Empty, nullptr);
+
+	return 0;
+}
+
+DWORD _cdecl EBoltExt::_EBolt_Draw_Colors(REGISTERS* R)
```
ZivDero, line 42: What is this _cdelc function and why does it exist?

src/Ext/EBolt/Hooks.cpp
```diff
@@ -0,0 +1,241 @@
...
+	{
+		boltColor1 = Drawing::RGB_To_Int(pExt->Color[0]);
+		boltColor2 = Drawing::RGB_To_Int(pExt->Color[1]);
+		boltColor3 = Drawing::RGB_To_Int(pExt->Color[2]);
+	}
+	else
+	{
+		GET_BASE(int, colorIdx, 0x20);
+		const COLORREF defaultAlternate = EBoltExt::GetDefaultColor_Int(FileSystem::PALETTE_PAL, colorIdx);
+		const COLORREF defaultWhite = EBoltExt::GetDefaultColor_Int(FileSystem::PALETTE_PAL, 15);
```
ZivDero: Why COLORREF? This ain't Win32

docs/New-or-Enhanced-Logics.md
Metadorius: it does at least make sense to have a common base, the principle of drawing of all of those is the same, it's just what specifically gets drawn as the graphic is different

src/Ext/EBolt/Hooks.cpp
NetsuNegi: it an Ares hook...

src/Ext/EBolt/Hooks.cpp
NetsuNegi: ...

NetsuNegi: this is Ares's global param, just use them then Ares will apply correct color

Metadorius: > this is Ares's global param, just use them then Ares will apply correct color

@ZivDero ~~has allergy to sharing space with other extensions and~~ is asking whether it's better to fully reimplement EBolt with all the Ares and Phobos additions at this point

NetsuNegi: I don't understand

NetsuNegi: Why do we have to increase the workload

NetsuNegi: If he insists on completely re implementing it, then I will close this PR

Metadorius: The concern is how cumbersome it may be to maintain. In my opinion, if it's alright to work with now -- there's no point to do premature refactoring if you don't feel like so.

NetsuNegi: I just don't think it's necessary, it's completely usable now, and Ares' custom colors have basically been taken over by Phobos

NetsuNegi: If necessary, I can also continue to take over the remaining 3 color hooks, but I don't think it's necessary to refactor the entire EBolt

src/New/Entity/LaserTrailClass.cpp
```diff
@@ -17,15 +18,54 @@ bool LaserTrailClass::Update(CoordStruct location)
...
 		if (this->Visible && !this->Cloaked && (this->Type->IgnoreVertical ? (abs(location.X - this->LastLocation.Get().X) > 16 || abs(location.Y - this->LastLocation.Get().Y) > 16) : true))
 		{
-			// We spawn new laser segment if the distance is long enough, the game will do the rest - Kerbiter
-			LaserDrawClass* pLaser = GameCreate<LaserDrawClass>(
-				this->LastLocation.Get(), location,
-				this->CurrentColor, ColorStruct { 0, 0, 0 }, ColorStruct { 0, 0, 0 },
-				this->Type->FadeDuration.Get());
+			if (this->Type->DrawType == LaserTrailDrawType::Laser)
+			{
+				// We spawn new laser segment if the distance is long enough, the game will do the rest - Kerbiter
```
CrimRecya: Minor issue: this comment should refer to line:19.

src/Ext/EBolt/Body.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+
+DEFINE_HOOK(0x4C2951, EBolt_DTOR, 0x5)
+{
+	GET(EBolt*, pItem, ESI);
+
+	EBoltExt::ExtMap.Remove(pItem);
+
+	return 0;
+}
```
CrimRecya, line 101: Not sure if 0x4C29FB also needs this?

src/Ext/Techno/Body.Update.cpp
```diff
@@ -525,14 +525,25 @@ void TechnoExt::ExtData::UpdateTypeData(TechnoTypeClass* pCurrentType)
...
-	this->LaserTrails.reserve(this->TypeExtData->LaserTrailData.size());
+		for (const auto& entry : this->TypeExtData->LaserTrailData)
+			this->LaserTrails.emplace_back(std::make_unique<LaserTrailClass>(entry.GetType(), pOwner, entry.FLH, entry.IsOnTurret));
 
-	for (auto const& entry : this->TypeExtData->LaserTrailData)
-	{
-		this->LaserTrails.emplace_back(entry.GetType(), pOwner, entry.FLH, entry.IsOnTurret);
+		for (const auto trail : addition)
+			this->LaserTrails.emplace_back(trail);
 	}
```
CrimRecya: I think this approach is unsafe and may lead to double deletion or memory leaks.. It's better to use `std::vector<std::unique_ptr<LaserTrailClass>> addition;`, `addition.push_back(std::move(trail));` and `this->LaserTrails.push_back(std::move(trail));` to transfer ownership.

src/Ext/EBolt/Hooks.cpp
```diff
@@ -1,78 +1,109 @@
...
 
 class EBoltFake final : public EBolt
 {
-	public:
-		void _SetOwner(TechnoClass* pTechno, int weaponIndex);
-		void _RemoveFromOwner();
+public:
+	void _SetOwner(TechnoClass* pTechno, int weaponIndex);
+	void _RemoveFromOwner();
 };
```
CrimRecya, line 107: I'm not sure if changing these to `EBoltExt`'s static function with `EBolt* pEBolt` would be better?

CrimRecya (changes requested):
In addition, regarding the original `LaserDrawClass` structure reference, it has now been modified to a pointer. In terms of variable naming, changing `trail` to `pTrail` should be better.

src/Ext/EBolt/Body.cpp
NetsuNegi: IDA shows it has no xrefs

src/Ext/EBolt/Hooks.cpp
NetsuNegi: I think it is not

PR #1707, opened 2025-06-08
[Vanilla Fix] Fix incorrect damage area logic

CrimRecya:
Fixed the issue that the widespread damage caused by detonation on the bridge/ground cannot affect objects on the ground/bridge who are in the opposite case.

---

- Perhaps WW wants to do some optimization, but obviously this is not reasonable.


Coronia:
since this is checked in every damage area calculation, the perf cost is large (which has also been proved by the profiler). Suggest to add a toggle for this, which should patch these hooks directly rather than inside them. See https://github.com/Phobos-developers/Phobos/commit/2ff45a0a4b120b7980104b1280de04d588dbd59b for reference

Coronia (approved):

PR #1710, opened 2025-06-08
Fix functions that return structs to return structs, not pointers

ZivDero:
Fixes functions that return CoordStruct, CellStruct or Matrix3D by value to actually do that, and dump the wrappers.
Needs to be tested, also `BuildingClass_GetFireCoords_Wrapper` is weird.

### DIFF
--- a/src/Ext/Building/Hooks.cpp
+++ b/src/Ext/Building/Hooks.cpp
@@ -172,11 +172,11 @@ DEFINE_HOOK(0x44CEEC, BuildingClass_Mission_Missile_EMPulseSelectWeapon, 0x6)
 	return SkipGameCode;
 }
 
-CoordStruct* __fastcall BuildingClass_GetFireCoords_Wrapper(BuildingClass* pThis, void* _, CoordStruct* pCrd, int weaponIndex)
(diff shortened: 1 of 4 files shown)


src/Ext/Building/Hooks.cpp
```diff
@@ -172,11 +172,11 @@ DEFINE_HOOK(0x44CEEC, BuildingClass_Mission_Missile_EMPulseSelectWeapon, 0x6)
 	return SkipGameCode;
 }
 
-CoordStruct* __fastcall BuildingClass_GetFireCoords_Wrapper(BuildingClass* pThis, void* _, CoordStruct* pCrd, int weaponIndex)
+CoordStruct* __fastcall BuildingClass_GetFireCoords_Wrapper(BuildingClass* pThis, CoordStruct* buffer, CoordStruct* pCrd, int weaponIndex)
```
Metadorius: *cough* it's a thiscall emulation

Metadorius (changes requested):

src/Ext/Building/Hooks.cpp
ZivDero: Yeah I know that, but what it actuall does is nonsense as far as I can see. See the Discord chat

Starkku:
Any unit/building trying to fire any weapon will crash the game
[except.txt](https://github.com/user-attachments/files/20660266/except.txt)
[debug.log](https://github.com/user-attachments/files/20660265/debug.log)

PR #1717, opened 2025-06-11
Ground line for select box

NetsuNegi:
- `Grounded` has been abandoned, you can now use `GroundShape` to specific a image which always draw on ground, it will only draw when techno is in air if set `Ground.AlwaysDraw=false`, this also affect on `GroundLine`.
- `Grounded` 已经被废弃了，你现在可以通过 `GroundShape` 指定一个始终绘制在地面的图像，若 `Ground.AlwaysDraw=false` ，那么它只会在单位位于空中时绘制，这也影响 `GroundLine` 的绘制。
- If `GroundLine=true` , the game will draw a line from techno's position to its vertical projection, `GroundLine.Dashed=true` means the projection line is a dashed line.
- 若 `GroundLine=true` ，那么游戏将会绘制一条从单位的位置到其垂直投影的线，`GroundLine.Dashed=true` 意味着投影线为虚线。

In `rulesmd.ini`:
```ini
[SOMESELECTBOXTYPE] ; Select box Type name
GroundShape=                            ; filename with .shp extension
GroundPalette=palette.pal               ; filename with .pal extension
GroundFrames=                           ; List of integer, default 1,1,1 for infantry, 0,0,0 for vehicle and aircraft
GroundOffset=0,0                        ; integers - horizontal, vertical
Ground.AlwaysDraw=true             ; boolean
GroundLine=false                        ; boolean
GroundLineColor=0,255,0                 ; R, G, B
GroundLineColor.ConditionYellow=        ; R, G, B
GroundLineColor.ConditionRed=           ; R, G, B
GroundLine.Dashed=false                 ; boolean
```
W@`N 6J G9@UM4MURW`LFIF

### DIFF
--- a/docs/User-Interface.md
+++ b/docs/User-Interface.md
@@ -712,35 +712,35 @@ Positional superweapon hotkeys are an experimental feature and currently the use
 In `uimd.ini`:
 ```ini
 [Sidebar]
-SuperWeaponSidebar=false              ; boolean
-SuperWeaponSidebar.Pyramid=true      ; boolean
-SuperWeaponSidebar.Interval=0         ; integer, pixels
-SuperWeaponSidebar.LeftOffset=0       ; integer, pixels
-SuperWeaponSidebar.CameoHeight=48     ; integer, pixels
-SuperWeaponSidebar.Max=0              ; integer
-SuperWeaponSidebar.MaxColumns=        ; integer
+SuperWeaponSidebar=false                    ; boolean
+SuperWeaponSidebar.Pyramid=true             ; boolean
+SuperWeaponSidebar.Interval=0               ; integer, pixels
+SuperWeaponSidebar.LeftOffset=0             ; integer, pixels
+SuperWeaponSidebar.CameoHeight=48           ; integer, pixels
+SuperWeaponSidebar.Max=0                    ; integer
+SuperWeaponSidebar.MaxColumns=              ; integer
 ```
 
 In `rulesmd.ini`:
 ```ini
 [GlobalControls]
-SuperWeaponSidebarKeysEnabled=false    ; boolean
+SuperWeaponSidebarKeysEnabled=false         ; boolean
 
 [AudioVisual]
-SuperWeaponSidebar.AllowByDefault=false   ; boolean
+SuperWeaponSidebar.AllowByDefault=false     ; boolean
 
-[SOMESIDE]
-SuperWeaponSidebar.OnPCX=             ; filename - including the .pcx extension
-SuperWeaponSidebar.OffPCX=            ; filename - including the .pcx extension
-SuperWeaponSidebar.TopPCX=            ; filename - including the .pcx extension
-SuperWeaponSidebar.CenterPCX=         ; filename - including the .pcx extension
-SuperWeaponSidebar.BottomPCX=         ; filename - including the .pcx extension
+[SOMESIDE]                                  ; Side
+SuperWeaponSidebar.OnPCX=                   ; filename - including the .pcx extension
+SuperWeaponSidebar.OffPCX=                  ; filename - including the .pcx extension
+SuperWeaponSidebar.TopPCX=                  ; filename - including the .pcx extension
+SuperWeaponSidebar.CenterPCX=               ; filename - including the .pcx extension
+SuperWeaponSidebar.BottomPCX=               ; filename - including the .pcx extension
 
-[SOMESW]
-SuperWeaponSidebar.Allow=             ; boolean
-SuperWeaponSidebar.PriorityHouses=    ; List of house types
-SuperWeaponSidebar.RequiredHouses=    ; List of house types
-SuperWeaponSidebar.Significance=0     ; integer
+[SOMESW]                                    ; SuperWeaponType
+SuperWeaponSidebar.Allow=                   ; boolean
+SuperWeaponSidebar.PriorityHouses=          ; List of house types
+SuperWeaponSidebar.RequiredHouses=          ; List of house types
+SuperWeaponSidebar.Significance=0           ; integer
 ```
 
 In `RA2MD.INI`:
--- a/src/Ext/Techno/Body.Visuals.cpp
+++ b/src/Ext/Techno/Body.Visuals.cpp
@@ -356,7 +356,7 @@ void TechnoExt::DrawSelectBox(TechnoClass* pThis, const Point2D* pLocation, cons
 		return;
 
 	const double healthPercentage = pThis->GetHealthPercentage();
-	const auto defaultFrame = whatAmI == InfantryClass::AbsID ? Vector3D<int> { 1, 1, 1 } : Vector3D<int> { 0,0,0 };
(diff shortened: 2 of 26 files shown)


Metadorius:
1. Maybe it's better to use a shp to draw ground line? Or somehow make it more tweakable. Maybe some would like a dotted line or something.
2. Any way to display GroundShape only when flying?
3. This would be a great candidate to add an image to docs.

NetsuNegi:
> 1. Maybe it's better to use a shp to draw ground line? Optionally perhaps. Or somehow make it more tweakable. Maybe some would like a dotted or thicker line or something.
> 2. Any way to display GroundShape only when flying?
> 3. This would be a great candidate to add an image to docs.

`GroundShape` is designed to draw a different image than `Shape` like a point, but I still can add a tag to prevent draw it when its height is 0.

Metadorius:
> I dont think shp is a good idea, that will make it harder to draw a line which length is changeable

No way to draw only a part of the SHP?
like imagine a 4px height SHP, and the line is 6px tall, then the SHP will be drawn 1.5 times starting from the top

NetsuNegi:
> > `GroundShape` is designed to draw a different image than `Shape`. For example, a point. But I still can add a tag to prevent draw it when its height is 0.

I guess you could use `InAir` check for this. My thought was so that GroundShape could indicate that the unit is grounded or flying and be only used for the airborne state. 

maybe check height greater than 0 also good?

NetsuNegi:
> > maybe check height greater than 0 also good?
> 
> I think it wouldn't be that useful. Hover units would also receive one then, and I don't think it would be beneficial in this case if the modder wants it only for airborne units.
> 
> IIRC `InAir()` determines whether AA weapons can fire on the unit too, so it could be useful for things like helicopters and planes taking off to know whether they are valid AA targets yet (when they take off). So I think it's only logical to tie those together.

alright

Metadorius:
Also is it intended that the shadow is drawn over it?

Metadorius:
> I dont have that plan

I meant that in your picture it looks like this: 

image

I am not sure if you wanted this

src/Ext/Techno/Body.Visuals.cpp
```diff
@@ -366,34 +361,49 @@ void TechnoExt::DrawSelectBox(TechnoClass* pThis, const Point2D* pLocation, cons
...
-			return;
+		if (pSelectBox->GroundLine)
+		{
+			Point2D projection { point.X, std::min(point.Y, pBounds->Height - 1) };
+			const int color = Drawing::RGB_To_Int(pSelectBox->GroundLineColor.Get(healthPercentage));
 
-		drawPoint = outClient;
+			if (pSelectBox->GroundLine_Dashed)
+			{
+				auto pattern = Make_Global<bool*>(0x84310C);
```
Metadorius: I think this has to be put into YRpp if it isn't already

NetsuNegi: Can you put it?
YRpp is a very big trouble for me :(

Speederovsky:
Seems like shape being covered by the shadows depends on DrawAboveTechno. 
What's up with this though? Why is the shape drawn off-center for jumpjet infantry?
offcenter

Metadorius:
I wonder how that dotted pattern is achieved too...

ZivDero:
> I wonder how that dotted pattern is achieved too...

There is a line draw functiont hat draws a line with a bool[] pattern.

Speederovsky:
Centering for Inf seems fixed.

docs/User-Interface.md
```diff
@@ -245,12 +245,16 @@ RealTimeTimers.Adaptive=false   ; boolean
 
 ### Select Box
 
+![placepreview](_static/images/selectbox.png)
```
Metadorius:
```suggestion
![selectbox](_static/images/selectbox.png)
```

src/Ext/Techno/Body.Visuals.cpp
```diff
@@ -366,34 +361,44 @@ void TechnoExt::DrawSelectBox(TechnoClass* pThis, const Point2D* pLocation, cons
...
-		const auto& [outClient, visible] = TacticalClass::Instance->CoordsToClient(coords);
+		if (visible && pGroundShape && (pSelectBox->GroundShape_AlwaysDraw || pThis->IsInAir()))
+		{
+			const Point2D drawPoint = point + pSelectBox->GroundOffset;
+			pSurface->DrawSHP(pPalette, pGroundShape, frame, &drawPoint, pBounds, flags, 0, 0, ZGradient::Ground, 1000, 0, nullptr, 0, 0, 0);
+		}
 
-		if (!visible)
-			return;
+		if (pSelectBox->GroundLine)
```
Metadorius, line 383: suggest to add `GroundLine.AlwaysDraw` too and maybe make the pattern customisable

src/New/Type/SelectBoxTypeClass.h
```diff
@@ -8,25 +8,34 @@ class SelectBoxTypeClass final : public Enumerable<SelectBoxTypeClass>
...
-	Valueable<bool> Grounded;
 	Valueable<Point2D> Offset;
+	Valueable<Point2D> GroundOffset;
 	TranslucencyLevel Translucency;
 	Valueable<AffectedHouse> VisibleToHouses;
 	Valueable<bool> VisibleToHouses_Observer;
 	Valueable<bool> DrawAboveTechno;
+	Valueable<bool> GroundShape_AlwaysDraw;
+	Valueable<bool> GroundLine;
+	Damageable<ColorStruct> GroundLineColor;
```
Metadorius, line 24: the fact that it's `Damageable` is not indicated in docs, maybe Shape and GroundShape should also be `Damageable` then too?

Metadorius:
@ZivDero could you consult @NetsuNegi regarding the pattern?

src/New/Type/SelectBoxTypeClass.h
NetsuNegi: no, I did

NetsuNegi:
> @ZivDero could you consult @NetsuNegi regarding the pattern?

what ?

src/New/Type/SelectBoxTypeClass.h
Metadorius: oh I missed that, pardon
what about the shapes being `Damageable`?

NetsuNegi: no need that just control by frame index

ZivDero:
> > > what ?
> > 
> > 
> > 
> > 1. he had some suggestion regarding usage of dashed line pattern
> > 2. it could be good to have it specified in INI, maybe as something like `GroundLine.DashedPattern=true,false,...`
> 
> I dont know what was that

The dashed line pattern is a static bool[16] that determines which pixels are drawn. There is no need to reference the one that exists in the game.
At a minimum, you can just define your own static bool[16]. Or do what Kerbiter suggested and istead of making it static, read the bool[16] from the ini so the user can customize the pattern (though I think that may potentially be a bit too much customization?)

Coronia (approved):

PR #1718, opened 2025-06-11
[Vanilla Enhancement] Customize limit when engineer repair a building

NetsuNegi:
- You can now set a maximum amount when engineer repair a building for either of them. 0 means the building will be repaired to full health.
- 现在你可以设置工程师修复建筑时的修复量限制了。0意味着建筑会被直接修满（不限制）。
  - Negative value means percentage. For example, `EngineerRepairAmount=-50` means you can only repair 50% of the building's health per Engineer.
  - 负数意味着百分比。例如， `EngineerRepairAmount=-50` 意味着每次工程师最多只可以修复50%的血量。
  - If both the building and the engineer has `EngineerRepairAmount` set, the actual repair amount will be the minimum of them.
  - 如果工程师与建筑都设置了 `EngineerRepairAmount` ，则实际的维修量将取更小者（按计算后的实际维修量对比）。
- `BuildingRepairedSound` can now be set individually for each building type.
- `BuildingRepairedSound` 现在可以被用来设置每个建筑物单独的修复音效。
- Fixed the bug that damaged particle dont disappear after building has repaired by engineer.
- 修复了工程师修复建筑后，建筑物的受损粒子不会立即消失的bug。

In `rulesmd.ini`
```ini
EngineerRepairAmount=0             ; integer
BuildingRepairedSound=             ; Sound entry, default to [AudioVisual] -> BuildingRepairedSound

[SOMEINFANTRY]                     ; InfantryType
EngineerRepairAmount=0             ; integer
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -412,6 +412,7 @@ This page lists all the individual contributions to the project by their author.
   - Weapon target filtering by health percentage
   - Fix the bug that `DamageSelf` and `AllowDamageOnSelf` are ineffective on airforce
   - Customize limit when engineer repair a building
+  - Fix the bug that damaged particle dont disappear after building has repaired by engineer
 - **Apollo** - Translucent SHP drawing patches
 - **ststl**:
   - Customizable `ShowTimer` priority of superweapons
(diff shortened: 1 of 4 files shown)


mevitar:
Shouldn't this be made customizable per-engineer?

Coronia:
> I think this should be made customizable per-engineer.

while I've made it work on Engineer now, it's still quite needed to keep it on building. Hence, when they're both set, the minimum of them will be used. E.g. to repair a building with 1000 strength, and `EngineerRepairAmount` on engi and building being 20% and 250, then it'll repair the building by 200 since 1000 * 20% < 250

PR #1721, opened 2025-06-12
Use `SkipCrushSlowdown=true` to avoid the bug related to `Accelerates=true` and `MovementZone=CrushAll`

TaranDahl:
- You can also disable the slowdown completely by using the flag `SkipCrushSlowdown`. This is not the same as `CrushSlowdownMultiplier=1.0`. Used to avoid a bug where the vehicle sometimes loses speed when `Accelerates=true` and `MovementZone=CrushAll` are set.


src/Ext/Unit/Hooks.Crushing.cpp
```diff
@@ -99,3 +99,10 @@ DEFINE_HOOK(0x6A108D, ShipLocomotionClass_WhileMoving_CrushTilt, 0xD)
 
 	return SkipGameCode;
 }
+
+DEFINE_HOOK_AGAIN(0x6A0809, SomeLocomotionClass_WhileMoving_SkipCrushSlowDown, 0x6) // Ship
+DEFINE_HOOK(0x4B1146, SomeLocomotionClass_WhileMoving_SkipCrushSlowDown, 0x6) // Drive
+{
+	GET(UnitClass*, pTechno, ECX);
```
CrimRecya: `GET(FootClass*, pLinked, ECX);`
If it's really an `UnitClass*`, there's no need to use `GetTechnoType()`.

PR #1722, opened 2025-06-12
[Vanilla Enhancement] Customizable spawn delay of `VoxelAnim`'s `TrailerAnim` and fix its incorrect position

CrimRecya:
- Fixed the issue of incorrect position of `TrailerAnim` in `VoxelAnim`.

---

- You can now customize the generation interval of VoxelAnim's trailer animation

In `rulesmd.ini`:
```ini
[SOMEVOXELANIM]       ; VoxelAnimType
Trailer.SpawnDelay=2  ; integer, game frames
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -48,6 +48,7 @@ This page lists all the individual contributions to the project by their author.
   - Ability to disable shadow for debris & meteor animations
   - Voxel light source position customization
   - `UseFixedVoxelLighting`
+  - Warhead activation target health thresholds
 - **Uranusian (Thrifinesma)**:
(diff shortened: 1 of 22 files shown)


Starkku (approved):
LGTM, although I am unfamiliar with the positioning bug and unsure what is actually being fixed here (if you don't mind satisfying my curiosity with an explanation, I see the game is doing bunch of float nonsense with the coord values).

The timer for trailer animation is technically a deviation from how game does it everywhere else but else but it is more accurate/proper approach rather than using `frame % delay` so I think it is fine.

PR #1723, opened 2025-06-13
Allows refineries to use multiple ActiveAnim simultaneously

TaranDahl:
### Allows refineries to use multiple ActiveAnim simultaneously

- In vanilla, the refinery uses different ActiveAnims depending on the storage. You can now make it use multiple ActiveAnims simultaneously like any other building.

In `artmd.ini`:
```ini
[SOMEBUILDING]                         ; BuildingType
Refinery.UseNormalActiveAnim=false     ; boolean
```

### DIFF
--- a/src/Ext/Building/Hooks.Refinery.cpp
+++ b/src/Ext/Building/Hooks.Refinery.cpp
@@ -57,3 +57,17 @@ DEFINE_HOOK(0x522E4F, InfantryClass_SlaveGiveMoney_CheckBalanceAfter, 0x6)
 
 	return 0;
 }
+
+DEFINE_HOOK(0x445FD6, BuildingClass_Place_RefineryActiveAnim, 0x6)
+{
+	GET(BuildingTypeClass*, pType, ESI);
+
+	return BuildingTypeExt::ExtMap.Find(pType)->Refinery_UseNormalActiveAnim ? 0x446183 : 0;
+}
+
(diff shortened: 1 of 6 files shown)


src/Ext/Building/Hooks.Refinery.cpp
```diff
@@ -57,3 +57,17 @@ DEFINE_HOOK(0x522E4F, InfantryClass_SlaveGiveMoney_CheckBalanceAfter, 0x6)
 
 	return 0;
 }
+
+DEFINE_HOOK(0x445FD6, BuildingClass_Place_RefineryActiveAnim, 0x6)
```
CrimRecya: How about hooking after the `Refinery` check is complete (for example 0x445FE4).

src/Ext/Building/Hooks.Refinery.cpp
```diff
@@ -57,3 +57,17 @@ DEFINE_HOOK(0x522E4F, InfantryClass_SlaveGiveMoney_CheckBalanceAfter, 0x6)
...
+{
+	GET(BuildingTypeClass*, pType, ESI);
+
+	return BuildingTypeExt::ExtMap.Find(pType)->Refinery_UseNormalActiveAnim ? 0x446183 : 0;
+}
+
+DEFINE_HOOK(0x450D9C, BuildingClass_UpdateAnimations_RefineryActiveAnim, 0x6)
```
CrimRecya: Same at 0x450DAA?

CrimRecya (approved):

PR #1728, opened 2025-06-17
[Minor] Damage.DealtByInvoker tweak & Damage.ApplyFIrepowerMult

Coronia:
previous `Damage.DealtByInvoker` doesn't really apply to anim warhead, this PR makes it inline with anim weapon. Notice that this might change the behavior of anim weapon and warhead a bit, but I think this should be the correct way. Any other feedbacks and reviews are welcome

`Damage.ApplyFirepowerMult` determines whether or not firepower modifiers from the animation's invoker are applied on the damage dealt from this animation, if exists

### DIFF
--- a/src/Ext/Anim/Hooks.cpp
+++ b/src/Ext/Anim/Hooks.cpp
@@ -51,9 +51,6 @@ DEFINE_HOOK(0x42453E, AnimClass_AI_Damage, 0x6)
 	const int damageMultiplier = isTerrain ? 5 : 1;
 	const double baseDamage = pThis->Type->Damage;
 
-	if (baseDamage < 1.0 && delay <= 0)
-		return SkipDamage;
-
 	int appliedDamage = 0;
 
 	if (pTypeExt->Damage_ApplyOncePerLoop) // If damage is to be applied only once per animation loop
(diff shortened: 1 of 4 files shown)


src/Ext/Anim/Hooks.cpp
```diff
@@ -42,70 +42,85 @@ DEFINE_HOOK(0x42453E, AnimClass_AI_Damage, 0x6)
...
+		if (pExt->InvokerHouse)
+			pOwner = pExt->InvokerHouse;
 
 		if (!pInvoker)
 		{
-			pInvoker = pThis->OwnerObject ? abstract_cast<TechnoClass*>(pThis->OwnerObject) : nullptr;
+			if (pThis->OwnerObject)
+				pInvoker = abstract_cast<TechnoClass*, true>(pThis->OwnerObject);
+			else if (pThis->IsBuildingAnim)
+				pInvoker = AnimExt::ExtMap.Find(pThis)->ParentBuilding;
```
CrimRecya: `pInvoker = pExt->ParentBuilding;`

src/Ext/Anim/Hooks.cpp
```diff
@@ -42,70 +42,85 @@ DEFINE_HOOK(0x42453E, AnimClass_AI_Damage, 0x6)
...
-	double damage = 0;
-	int appliedDamage = 0;
+	const bool isTerrain = pThis->OwnerObject && pThis->OwnerObject->WhatAmI() == AbstractType::Terrain;
+	const int damageMultiplier = isTerrain ? 5 : 1;
+	const double baseDamage = pThis->Type->Damage;
 
-	if (pThis->OwnerObject && pThis->OwnerObject->WhatAmI() == AbstractType::Terrain)
-		damageMultiplier = 5;
+	if (baseDamage < 1.0 && delay <= 0)
+		return SkipDamage;
```
CrimRecya: Why is the handling here different from the before?

CrimRecya (approved):

PR #1733, opened 2025-06-20
[Minor] Optimize Interceptor

Coronia:
split from https://github.com/Phobos-developers/Phobos/pull/1568. This one is a bit controversial so I'd pull this alone for test and review
- instead of selecting weapon for every bullet, used the fixed value from `Interceptor.Weapon` once and for all
- added a RearmTimer check for the interceptor process

### DIFF
--- a/src/Ext/Techno/Body.Update.cpp
+++ b/src/Ext/Techno/Body.Update.cpp
@@ -50,7 +50,7 @@ void TechnoExt::ExtData::ApplyInterceptor()
 	auto const pThis = this->OwnerObject();
 	auto const pTypeExt = this->TypeExtData;
 
-	if (pTypeExt->InterceptorType && !pThis->Target && !this->IsBurrowed)
+	if (pTypeExt->InterceptorType && !pThis->Target && pThis->RearmTimer.Expired() && !this->IsBurrowed)
 	{
 		BulletClass* pTargetBullet = nullptr;
 		const auto pInterceptorType = pTypeExt->InterceptorType.get();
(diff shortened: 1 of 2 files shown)


Starkku (changes requested):
You are misunderstanding how the interceptor code currently works. You have not taken into accord that all `TechnoExt::ExtData::ApplyInterceptor()` does is select an eligible `BulletClass` instance as the techno's target, it does not actually have the techno directly fire any weapons. It does not even execute most of the code if the techno currently has a target. As a result, removing weapon selector code makes it so that weapon actually fired against projectiles does not actually follow `InterceptorWeapon`.

Uncertain about the application of `RearmTimer` check as well. You are not technically checking for if the techno can fire here, you are supposed to be checking if it can target. If it already has a target, it bails out before this check anyway. A case could be made for separate targeting timer check here but the reality is that most projectiles are fast enough that not checking for it every frame can reduce accuracy dramatically. Normal target acquisition already suffers from similar issue with very fast moving units (such as AA vs. jets, potentially) due to `NormalTargetingDelay`/`GuardAreaTargetingDelay` being rather substantial.

Either rework how the logic works entirely or adjust/revert your fix to work with the current approach.

Coronia:
in this case I think the part that removing `SelectWeapon` from `ApplyInterceptor` could be kept, since it'll be forced to `Interceptor.Weapon` if the target is a bullet

other changes could be reverted for the sake of safety and accuracy

Starkku (approved):

PR #1737, opened 2025-06-23
[Vanilla Enhancement] Task subtitles display in the middle of the screen

CrimRecya:
- Now you can set `MessageApplyHoverState` to true，to make the upper left messages not disappear while mouse hovering over the top of display area.
- You can also let task subtitles (created by trigger 11) to display directly in the middle area of the screen instead of the upper left corner, with a semi transparent background, by setting `MessageDisplayInCenter` to true.
  - When the mouse hovers over the subtitle area (simply judged as a rectangle), its opacity will increase and it will not disappear during this period.

In `RA2MD.INI`:
```ini
[Phobos]
MessageApplyHoverState=false  ; boolean
MessageDisplayInCenter=false  ; boolean
```

---

I'm not sure if there is a better implementation, but it's the simplest implementation that comes to my mind.


Metadorius:
Can we have an example image added to docs and to PR?

TaranDahl:
I think you don't need to finish everything in one single pr🤔
Also, if you do so, it will result in a huge amount of review work, and as a result, no one will be willing to review😋

PR #1739, opened 2025-06-26
Fix the bug that `OpenToppedWarpDistance` is calculated incorrectly for building target

TaranDahl:
To verify the patch:
1. Set `OpenToppedWarpDistance` the same as CLEG's weapon range. Build a BFRT with no range bonus and a CLEG.
2. In vanilla, it will not warp big building targets (like a war factory) correctly on the edge distance.
3. With this fix, it can always warp the targets they fired at.


CrimRecya (approved):
LGTM.

PR #1740, opened 2025-06-27
[Minor] Optimize codestyle

Coronia:
Split from https://github.com/Phobos-developers/Phobos/pull/1535. Despite editing many files, it only contains 3 changes mostly:
- added const to variables that's not supposed to be changed, so the compiler might optimize it during building the dll
- cached some variables that would be used multiple time to reduce pointer call
- remove sanity check from `ExtMap.Find` and add `ExtMap.TryFind`


Coronia:
a new change that should be noticed here: the previous `ExtMap.Find` no longer has a nullptr sanity check now, while adding a new `ExtMap.TryFind` for that. Already changed all found instances that need to do sanity check too, and no crash occured in out test so far

pay attention that you should also use `TryFind` instead of `Find` when an object might not exist if this PR is merged. If you think this change is inappropriate, please also let me know

Coronia:
since this one doesn't have much issues in testing so far, gonna merge it recently. Any future optimization shall be coming with standalone commit or pull request

still, the new `TryFind` change might still break something accidentally, so this won't be merged into 0.4 as a whole, and please notice if there's any issue after it's merged. Meanwhile, if your PR is involving `Find` function that can be nullptr, you should also change it to `TryFind` once it's merged, especially for large change like Techno Attachment @Metadorius

PR #1745, opened 2025-06-29
[Minor] Fix number of issues with health threshold checks

### LINKED ISSUE #1743: WH target health thresholds broke ExtraWarheads on DeathWeapon
### Description

The commit '**Warhead activation target health thresholds** (https://github.com/Phobos-developers/Phobos/pull/1398)' made on 13/06 broke ExtraWarheads activating on DeathWeapon of a unit. I can only assume the new function interacts with ExtraWarheads differently than with the original Warhead and prevents its activation for a dying unit with 0 health. I was not able to make it activate.

### Phobos Version

Any dev nightly after 13/06

### Conditions to reproduce

_No response_

### INI code

```ini

```

### Steps to reproduce

1. Add an ExtraWarheads entry to whatever your _DeathWeapon=UnitDeathWeapon_ setup is.
2. Watch it fail.

### Expected behaviour

It should activate (effects, AnimList etc).

### Actual behaviour

It does not activate.

### Additional context

_No response_

### Checklist

- [x] The issue happens on the **latest official** version of Phobos and wasn't fixed yet.
- [x] I agree to elaborate the details if requested and provide thorough testing if the bugfix is implemented.
- [x] I added a very descriptive title to this issue.
- [x] I have attached as much information as possible *(screenshots, gifs, videos, debug and exception logs, etc)*.

Starkku:
• Warhead `AffectsBelow/AbovePercent` check is no longer bypassed in `BulletClass::Logics()` if bullet target and firer are same object.
• Removed incorrect `CanTarget.Max/MinHealth` check on firing techno itself in force weapon selection code.
• Warhead `AffectsAbovePercent` will now bypass the check if target has 0 health matching behaviour from before `AffectsAbove/BelowPercent` was introduced.
• Weapon `CanTarget.MinHealth`  will now by default bypass the check if target has 0 health left matching behaviour from before `CanTarget.Min/MaxHealth` was introduced.
• Added missing `Crit.AffectsAbovePercent` to be in line with other checks.
• Unified the health threshold checking code.

Fixes #1743

### DIFF
--- a/src/Ext/Techno/WeaponHelpers.cpp
+++ b/src/Ext/Techno/WeaponHelpers.cpp
@@ -39,11 +39,13 @@ int TechnoExt::PickWeaponIndex(TechnoClass* pThis, TechnoClass* pTargetTechno, A
 		if (pTargetCell && !EnumFunctions::IsCellEligible(pTargetCell, pSecondExt->CanTarget, true, true))
 			return weaponIndexOne;
 
+		if (!pSecondExt->IsHealthInThreshold(pTarget))
+			return weaponIndexOne;
+
 		if (pTargetTechno)
 		{
 			if (!EnumFunctions::IsTechnoEligible(pTargetTechno, pSecondExt->CanTarget)
 				|| !EnumFunctions::CanTargetHouse(pSecondExt->CanTargetHouses, pThis->Owner, pTargetTechno->Owner)
-				|| !pSecondExt->IsHealthRatioEligible(pTargetTechno)
 				|| !pSecondExt->HasRequiredAttachedEffects(pTargetTechno, pThis))
 			{
 				return weaponIndexOne;
@@ -68,11 +70,13 @@ int TechnoExt::PickWeaponIndex(TechnoClass* pThis, TechnoClass* pTargetTechno, A
 		if (pTargetCell && !EnumFunctions::IsCellEligible(pTargetCell, pFirstExt->CanTarget, true, true))
 			return weaponIndexTwo;
 
+		if (!pFirstExt->IsHealthInThreshold(pTarget))
+			return weaponIndexTwo;
+
 		if (pTargetTechno)
 		{
 			if (!EnumFunctions::IsTechnoEligible(pTargetTechno, pFirstExt->CanTarget)
 				|| !EnumFunctions::CanTargetHouse(pFirstExt->CanTargetHouses, pThis->Owner, pTargetTechno->Owner)
-				|| !pFirstExt->IsHealthRatioEligible(pTargetTechno)
 				|| !firstAllowedAE)
 			{
 				return weaponIndexTwo;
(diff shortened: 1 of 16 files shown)


src/Ext/Techno/WeaponHelpers.cpp
```diff
@@ -68,11 +70,13 @@ int TechnoExt::PickWeaponIndex(TechnoClass* pThis, TechnoClass* pTargetTechno, A
 		if (pTargetCell && !EnumFunctions::IsCellEligible(pTargetCell, pFirstExt->CanTarget, true, true))
 			return weaponIndexTwo;
 
+		if (!pFirstExt->IsHealthInThreshold(pTargetTechno))
```
CrimRecya: `pTargetTechno` -> `pTarget`

Metadorius:
> And regarding `IncludeZero`, I think negative numbers should be allowed? Then only need to provide a prompt in the document, without adding a new key.

Not sure on this, why?

Coronia:
I agree that 0 health is a rare edgy case that shouldn't even be considered in actual modding, so `IncludeZero` could be removed and just let it always be included or not, which should be written in the doc tho

src/Ext/WeaponType/Body.cpp
```diff
@@ -41,13 +41,17 @@ bool WeaponTypeExt::ExtData::HasRequiredAttachedEffects(TechnoClass* pTarget, Te
...
 }
 
-bool WeaponTypeExt::ExtData::IsHealthRatioEligible(TechnoClass* const pTarget) const
+bool WeaponTypeExt::ExtData::IsHealthInThreshold(AbstractClass* pTarget) const
 {
-	if (!pTarget)
+	if (auto const pObject = abstract_cast<ObjectClass*>(pTarget))
+		return GeneralUtils::IsHealthInThreshold(pObject, this->CanTarget_MinHealth, this->CanTarget_MaxHealth, this->CanTarget_MinHealth_IncludeZero);
+	else
 		return true;
```
CrimRecya: I have reviewed the discussion on DC, so in fact, `IncludeZero` is designed to handle the situation after the death of the target object. In this case, I think there are some issues.
For example, in `ExtraWarheads`, the target is reacquired every time it detonates. When the target dies, `AnnounceInvalidPointer` is called, resulting in an empty `Target`. In this case, the target obtained afterwards is empty and can no longer participate in subsequent checks. Therefore, some special handling like ` "return true;" -> "return this->CanTarget_MinHealth_IncludeZero;" ` is needed in this situation. (The external of the function may also require some processing, but I haven't looked closely yet)

Metadorius:
@fagonghaiwo can you elaborate what is being shown on the screenshot?

fagonghaiwo:
@Metadorius 
[SENT]
Interceptor=yes
Interceptor.Weapon=1
Interceptor.GuardRange=14
Interceptor.CanTargetHouses=all 

without WH changes
then they fire but cant intercept the projectile
it didnt occur in dev, only in this PR

Coronia:
The issue on bullet interception and terrain objects no longer appear with the newest build

Coronia (approved):

PR #1754, opened 2025-07-08
Smart default value of `DamageAlliesMultiplier` and `DamageOwnerMultiplier`

TaranDahl:
- Warheads are now able to define the extra damage multiplier for owner house, ally houses and enemy houses.
  - `DamageOwnerMultiplier.NotAffectsEnemies` and `DamageAlliesMultiplier.NotAffectsEnemies` is used as the default value if `AffectsEnemies=false` is set on the warhead.

### DIFF
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -2180,21 +2180,22 @@ SplashList.PickRandom=false  ; boolean
 ### Damage multipliers
 
 - Warheads are now able to define the extra damage multiplier for owner house, ally houses and enemy houses.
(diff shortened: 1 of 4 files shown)


TaranDahl:
What about `DamageMultiplierByHouse.IgnoreNotAffectsEnemies`
Maybe clearer, but ... too long.

Coronia:
how about this, instead of a boolean toggle, we can make it into floating point values that would be used when 1) `AffectsEnemies=false` and 2) these values are set, otherwise it'll still follow the original tags
```
DamageAlliesMultiplier.NotAffectsEnemies=
DamageOwnerMultiplier.NotAffectsEnemies=
```

CrimRecya:
Both the PR description and the doc have been forgotten.

CrimRecya (approved):

PR #1757, opened 2025-07-09
[Minor] Fix to InfantryType

a851903106:
Engineers can enter buildings normally when they don't need to be repaired (or you can force it by pressing Alt).
建筑不需要修理时，工程师可以正常进入建筑（也可以按 Alt 键强制进入）。

Player-controlled spies are not forced to perform other tasks while attacking buildings
玩家控制的间谍在攻击建筑时不会被迫执行其他任务。

Infantry in deployed state will not trigger `RateUp` on Gatling.
部署状态的步兵不会触发盖特逻辑的`RateUp`。

`MultiWeapon=yes` or `IsGattling=yes` allows infantry to select weapons normally when they are no longer in the deployed state.
`MultiWeapon=yes`或 `IsGattling=yes`允许步兵在不处于部署状态时正常选择武器。

If `BombDisarm=Yes` is not present for all weapon warheads, then the engineer will no longer use the appropriate mouse action.
如果所有武器弹头都不存在`BombDisarm=Yes`，那么工程师将不再使用相应的鼠标动作。

Fixed an unusual use of `DeployFireWeapon` for InfantryType.
修正了步兵类型使用DeployFireWeapon`的异常情况。

工程师修复


CrimRecya (commented):
When executing waypoint commands, you need to ignore the keys the player has pressed. You can see this processing at the beginning of the `MouseOverObject`.
Additionally, I suggest giving the return addresses some easily understandable names, which would be clearer.

CrimRecya (approved):

PR #1758, opened 2025-07-09
fixed cursor hardcoded to refresh at 60fps

FridgeRr:
Tested using latest cnc-ddraw with the renderer being direct3d9, minfps=120,maxfps=240 on a 240hz monitor, as well as cnc-ddraw onto dxvk.
singlecpu=false to be set in cnc-ddraw.ini is recommended if any initial problems are observed, as well as updating your cnc-ddraw.
Attached are the ddraw.ini settings I used. Note that refresh_rate=240 here, changing this to whatever your monitor uses may allow more compatibility. Note also that if you make a change to(mo install as example) mo/resources/cnc-ddraw.ini, make sure the change also applies to mo/ddraw.ini. For some reason I had to delete my ddraw.ini every time I made a change to my cnc-ddraw.ini.
[DDRAWSETTINGS.txt](https://github.com/user-attachments/files/21147261/DDRAWSETTINGS.txt)

also for some reason this commit has some git submodule jank


CrimRecya:
If there is no need to design a switch, I think only `DEFINE_PATCH(0x7B853C, 0x01);` is needed?

Metadorius:
So what's the status? was anyone able to test this?

Starkku:
I tested it but couldn't discern any difference but I have a very slow screen anyway.

TaranDahl (approved):
Doubt whether this really has any meaning. 
Anyway, it looks okay.

TaranDahl:
Will it make, for example, the cursor of Action::Move rotate faster? 🤣

Metadorius:
Not sure of the tested label, was the result actually visible? @FridgeRr says he can see the difference on high refresh rate screen.

FridgeRr:
> Not sure of the tested label, was the result actually visible? @FridgeRr says he can see the difference on high refresh rate screen.

here is a recording of it in action, recorded in 240fps. note that google drive preview will be in 60 fps, so please download the recording and use suitable software to verify, if you have only a 60hz monitor then .25x playback speed will work to see the illustrated effect. A couple of friends of mine with access to high refresh rate monitors have also confirmed the patch to be working. 
[test recording](https://drive.google.com/file/d/1EIymLwfb2oqkLlDmbMcc0AIy75WwjN7n/view?usp=sharing)

>Will it make, for example, the cursor of Action::Move rotate faster? 🤣

Actually i don't think it does, it seems that the cursor for actions such as Action::Move has a separate logic for determining the relative animation state of the cursor(animation states for cursors seem to be tied to real time?); for instance, the amount of rotation from t=0 state for Action::Move, that's independent of drawings.

Furthermore, part of the reason it's been difficult for people to recognise or 'see' the difference is because renderer wise, cncddraw requires a bit of setup(and maybe the latest version but im not sure in this) in order to actually have the game render in fps,refresh rate>60, as outlined in my initial post. It's apparently a bit finnicky to get cncddraw to actually do this on some systems I've noted, although i wouldn't know why. 
 Perhaps spreading some awareness in this aspect(perhaps some implementation to change these related options in the dta client [however it's unknown for me how much time/effort this would take to implement] would suffice for a user who doesn't want to meddle around editing ini files) in configuration would allow an easier time for testers/developers/users to verify the efficacy of this patch.
If any members would like to test this feature and is having trouble configuring or setting up cncddraw feel free to ask.

TaranDahl:
If you can confirm that it is working, then you can update the branch and get it ready for merging.

PR #1759, opened 2025-07-09
Fix vehicle with `AttackMove.StopWhenTargetAcquired` incorrectly lost megamission in some rare situation

TaranDahl:
Somehow, the vehicle called EnterIdleMode (probably in the functions of vehicle-related locomotors) in a move mission.
Then the megamission will be removed because the RefreshMegaMission function will only continue the megamission if the techno is in a guard mission.


CrimRecya (commented):
I think this should be considered a vanilla bug, and the default behavior can be directly changed. Also needs to be mentioned in the doc.

TaranDahl:
> Will only `AttackMove.StopWhenTargetAcquired=yes` units have such problems?

Yes, it won't happen if this flag is not set. That is the reason I suspect the locomotor.

CrimRecya (approved):

PR #1763, opened 2025-07-11
[Minor] Make `JumpjetTilt` units use `VxlCache` when tilting

CrimRecya:
Only optimized, as the title said.

### DIFF
--- a/src/Ext/TechnoType/Hooks.MatrixOp.cpp
+++ b/src/Ext/TechnoType/Hooks.MatrixOp.cpp
@@ -549,7 +549,6 @@ DEFINE_HOOK(0x73C47A, UnitClass_DrawAsVXL_Shadow, 0x5)
 		&& uTypeExt->JumpjetTilt
 		&& jjloco->State != JumpjetLocomotionClass::State::Grounded
 		&& jjloco->CurrentSpeed > 0.0
-		&& pThis->WhatAmI() == AbstractType::Unit
 		&& pThis->IsAlive
 		&& pThis->Health > 0
 		&& !pThis->IsAttackedByLocomotor)


src/Ext/TechnoType/Hooks.MatrixOp.cpp
```diff
@@ -200,31 +207,52 @@ Matrix3D* __stdcall JumpjetLocomotionClass_Draw_Matrix(ILocomotion* iloco, Matri
...
+			*pIndex = (((arfFace & 0x7Fu) << 7) + (arsFace & 0x7Fu)) << 7;
+		else if (onGround)
+			*pIndex <<= 6;
+
+		if (onGround)
+			*pIndex += slope_idx;
+
+		*pIndex <<= 5;
 		*pIndex |= curf.GetFacing<32>();
 	}
```
Metadorius: I think instead you should make it like this:

https://github.com/Phobos-developers/YRpp/blob/e6191d932ee27d26b06fbfd731988bfb70d5a861/Drawing.h#L303

src/Ext/TechnoType/Hooks.MatrixOp.cpp
```diff
@@ -200,31 +207,52 @@ Matrix3D* __stdcall JumpjetLocomotionClass_Draw_Matrix(ILocomotion* iloco, Matri
...
 }
 DEFINE_FUNCTION_JUMP(VTABLE, 0x7ECD8C, JumpjetLocomotionClass_Draw_Matrix);
 
+DEFINE_HOOK(0x73B748, UnitClass_DrawVXL_ResetKeyForTurretUse, 0x7)
+{
+	REF_STACK(int, vxlIndexKey, STACK_OFFSET(0x1C4, -0x1B0));
+
+	// Main body drawing completed, then enable accurate drawing of turrets and barrels
+	if (vxlIndexKey != -1 && (vxlIndexKey & 0x7FFE0000)) // Flags used by JumpjetTilt units
+		vxlIndexKey = -1;
```
Metadorius: Same here.

Metadorius (changes requested):
Change is good but I think we gotta have more refinement on how it's implemented. Maybe something like this?

```cpp

struct JumpjetTiltVoxelIndexKey
{
	// bitwise definitions here
};

struct PhobosVoxelIndexKey
{
	union
	{
		VoxelIndexKey Base;
		union
		{
			JumpjetTiltVoxelIndexKey JumpjetTiltVoxel;
			// add other definitions here as needed
		} CustomIndexKey;
	};

	// add funcs here if needed
};

static_assert(sizeof(PhobosVoxelIndexKey) == sizeof(VoxelIndexKey), "PhobosVoxelIndexKey size mismatch");
```

src/Ext/TechnoType/Hooks.MatrixOp.cpp
CrimRecya: The content provided here is not quite accurate.

Metadorius: hmmm, what do you mean?

CrimRecya: Like these.
```cpp
	struct MainKey
	{
		uint32_t bodyFrame  : 5; // 移动帧号（0-4）
		uint32_t bodyFace   : 5; // 车体朝向（5-9）
		uint32_t slopeIndex : 6; // 斜坡索引（10-15）
		uint32_t isSpawnAlt : 1; // SpawnAlt（16）
		uint32_t reserved   : 15;// 空保留位（17-31）
	}
	Body;

	struct TurretKey
	{
		uint32_t turretFace : 5; // 炮塔朝向（0-4）
		uint32_t bodyFace   : 5; // 车体朝向（5-9），如果 TurretOffset=0 ，则此段归零，因为从中心点开始画不需要有偏移
		uint32_t slopeIndex : 6; // 斜坡索引（10-15）
		uint32_t turretFrame: 8; // 炮塔帧号（16-23）
		uint32_t turretNum  : 8; // 炮塔编号（24-31）
	}
	Turret;
```

ZivDero: Would be nice if you provided references for where this structure comes from (e. g., see 0xFFFFFF, face index << 5, etc.)

Coronia (approved):

src/Ext/TechnoType/Hooks.MatrixOp.cpp
```diff
@@ -193,50 +266,63 @@ Matrix3D* __stdcall JumpjetLocomotionClass_Draw_Matrix(ILocomotion* iloco, Matri
...
-				arsFace = arsDir.GetFacing<128>();
+				arsFace = (arsDir.GetFacing<128>() + 96u) & 0x7Fu;
 
 				if (arsFace)
 					ret->RotateX(static_cast<float>(arsDir.GetRadian<128>()));
 			}
 
 			const auto arfDir = DirStruct(arf);
-			arfFace = arfDir.GetFacing<128>();
+			arfFace = (arfDir.GetFacing<128>() + 96u) & 0x7Fu;
```
Metadorius: not sure what are those additions?

src/Ext/TechnoType/Hooks.MatrixOp.cpp
```diff
@@ -103,6 +103,78 @@ DEFINE_HOOK(0x73CCE1, UnitClass_DrawSHP_TurretOffest, 0x6)
...
 #pragma region draw_matrix
 
+struct BodyVoxelIndexKey
+{
+	unsigned bodyFrame : 5;
+	unsigned bodyFace : 5;
+	unsigned slopeIndex : 6;
+	unsigned isSpawnAlt : 1;
+	unsigned reserved : 15;
+};
```
Metadorius, line 113: From my understanding, this is the vanilla definition? Could you instead amend the YRpp ones?

CrimRecya: Because `bitfield_17` is `private`.

src/Ext/TechnoType/Hooks.MatrixOp.cpp
CrimRecya: Because when changing the radian to DirStruct, it will rotate 90 degrees. To ensure that 0 is still 0, it needs to be rotated back, otherwise it will cause the `key` to be incorrect.
For example, in the previous nightly, a 90 degree forward tilt would result in using a horizontal cache.

Metadorius: I see. I think this has to be put in a comment, and I think this is probably doable with existing functions?

src/Ext/TechnoType/Hooks.MatrixOp.cpp
Metadorius: Couldn't you define a function that needs private access on the base voxel keys instead? Also you mentioned current one is incorrect, so improvements to YRpp are always welcome.

src/Ext/TechnoType/Hooks.MatrixOp.cpp
CrimRecya: You mean `offset`, right?

Metadorius: I just meant that what you are doing manually probably has a corresponding function, don't know for sure what is the function specifically. We could ask @secsome about this.

CrimRecya: I found it. Just fill in the `offset` parameter.

And can I also change `Value` in `VoxelIndexKey` to `public`?
The main problem is that if I don't do this, I have to rewrite the original functions.

Metadorius: Yeah, sure. We're a RE project anyways and we're pretty invasive, so we don't strictly follow encapsulation principle.

PR #1772, opened 2025-07-14
Improve attached animation position customization

Starkku:
Replaces `UseCenterCoordsIfAttached` with new enum key with additional options for position customization. Will add rule to migration tool if/after this PR is merged.

--------------

### Attached animation position customization

- You can now customize position of attached animations via different values for `AttachedAnimPosition`.
    - `default`: Animation shows up at the parent object's logical position (for buildings, this is the top-leftmost cell / cell #0).
    - `center`: Animation shows up at the parent object's visual center.
    - `ground`: Animation shows up at the parent object's visual center but forced to ground level. Note that the animations is still considered attached to the object and is likely to be drawn above objects that the object itself would cover unless other settings are used to compensate.

In `artmd.ini`:
```ini
[SOMEANIM]                   ; AnimationType
AttachedAnimPosition=object  ; Attached animation position enumeration (default|center|ground)
```

### DIFF
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -358,12 +358,15 @@ Damage.ApplyFirepowerMult=false ; boolean
 
 ### Attached animation position customization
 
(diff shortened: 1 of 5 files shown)


Coronia (approved):
reviewed and tested to work

PR #1773, opened 2025-07-15
[Minor] Remove NONE_STR shield

Coronia:
It seems that adding a NONE_STR shield is redundant, but not sure if removing it will result in any problem. Made a PR for testing first


Coronia:
tested and no issue other than https://discord.com/channels/338055131666972683/773636942775582720/1394713122983514155, which is not a problem introduced by this PR

PR #1774, opened 2025-07-15
KeepRange.EarlyStopFrame

TaranDahl:
- `KeepRange` controls how long the distance to maintain when the techno's ROF timer is ticking. What is actually read is its absolute value. If it is a positive value, it will be stayed outside this distance, just like it has a special `MinimumRange` after firing. If it is a negative value, it will be kept as close as possible to this distance, just like it has a special `Range` after firing. In addition, if the effective range section is too small, it will be considered unable to fire. It is best to have an effective range of 1.0, and 2.0 is best for Infantry.
    - `KeepRange.AllowAI` controls whether this function is effective for computer.
    - `KeepRange.AllowPlayer` controls whether this function is effective for human.
    - The function won't take effect if the techno's rearm time left is shorter than `KeepRange.EarlyStopFrame`.

### DIFF
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -2542,7 +2542,7 @@ In `rulesmd.ini`:
 KeepRange=0                       ; floating point value
 KeepRange.AllowAI=false           ; boolean
 KeepRange.AllowPlayer=false       ; boolean
-KeepRange.MaxRearmTimerAllowed=0  ; boolean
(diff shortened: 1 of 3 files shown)


TaranDahl:
@Coronia Is this flag name OK?

Coronia:
Honestly both testers and me didn't quite understand this feature, especially what it's supposed to achieve in actual practice. Also RearmTimer is a name that mostly for programmers, while modders would rather prefer names like ROF. Before we think of a better naming, it'd better to have some example where this feature can be used

PR #1776, opened 2025-07-17
Customize the chained damage of the wall

TaranDahl:
### Customize the chained damage of the wall

- In vanilla, when the wall is damaged, it will deal 200 damage to the walls in the 4 nearby cells. This makes connected walls more vulnerable to damage compared to single walls. 
- Now you can customize that damage by using the following flag.

In `rulesmd.ini`:
```ini
[CombatDamage]
AdjacentWallDamage=200  ; integer
```


Coronia:
could it be made into a value tag that decide how much damage should be dealt to adjacent wall instead of the fixed 200? Could be named as `AdjacentWallDamage` perhaps, and use 0 to omit the damage

Coronia:
c3d2231be09dc3188b8b943dc6571152
0 damage is not working in our testing, and it seemed that other damage value also didn't make a difference in the rate of wall destruction

TaranDahl:
@Coronia I guess your testing method might be incorrect. To clearly see the difference, you need a unit with a CellSpread less than 0.01. This is because the range within which projectiles damage the walls is the RangeFinding range, which is equal to CellSpread + 0.99.

TaranDahl:
Strange, the testing on my side seems to be working properly?

TaranDahl:
@Coronia did you notice that the flag should be written in the section `CombatDamage`?

CrimRecya (approved):

PR #1790, opened 2025-07-29
Support YRpp spawner's MP saves

Metadorius:
Also see https://github.com/CnCNet/yrpp-spawner/pull/48. You should test this PR with any recent YRpp spawner version with MP saves, should work.

Scenarios to test:
- quicksave
- save game trigger action

Conditions:
- online with YRpp spawner (should work)
- online with old cncnet5.dll / old YRpp spawner without support for MP saves

### DIFF
--- a/src/Utilities/SpawnerHelper.cpp
+++ b/src/Utilities/SpawnerHelper.cpp
@@ -1,7 +1,9 @@
 #include "SpawnerHelper.h"
+
 #include <Utilities/Debug.h>
+#include <Utilities/Macro.h>
 
 bool SpawnerHelper::IsSaveGameEventHooked()
 {
-	return SaveGameHookStart == 0xE9;
+	return SaveGameHookStart == LJMP_OPCODE;
 }


src/Commands/QuickSave.cpp
```diff
@@ -49,6 +51,12 @@ void QuickSaveCommandClass::Execute(WWKey eInput) const
 		Phobos::CustomGameSaveDescription += L" - ";
 		Phobos::CustomGameSaveDescription += GeneralUtils::LoadStringUnlessMissing("TXT_QUICKSAVE_SUFFIX", L"Quicksaved");
 	}
+	else if (SpawnerHelper::SaveGameEventHooked())
+	{
+		// Relinquish handling of the save game to spawner
+		EventClass event { HouseClass::CurrentPlayer->ArrayIndex, EventType::SaveGame };
+		EventClass::AddEvent(event);
```
ZivDero: One line (construct within the AddEvent call) perhaps?

ZivDero (approved):

Metadorius:
Took over the XNA client's handling of savegames. This should fix the "saving too fast" issue.

Please verify all previously faulting scenarios, and in addition to that also verify if the saves in skirmish/campaign are alright still, also verify that saves are alright with old `cncnet5.dll`.

src/Utilities/SpawnerHelper.cpp
```diff
@@ -0,0 +1,7 @@
+#include "SpawnerHelper.h"
+#include <Utilities/Debug.h>
+
+bool SpawnerHelper::IsSaveGameEventHooked()
+{
+	return SaveGameHookStart == 0xE9;
```
Belonit: Perhaps it is worth using the LJMP_OPCODE macro here instead of 0xE9

Belonit (approved):

src/Utilities/SpawnerHelper.cpp
Metadorius: agree, didn't realize such exists

PR #1807, opened 2025-08-12
[Minor] Units of the same type do not attack the same target

Fly-Star-him:
When a weapon has `OnlyAttacker=yes`, it prevents other units using that weapon from attacking the same target.
- 当武器存在`OnlyAttacker=yes`时，它会阻止使用该武器的其他单位攻击同一目标。


src/Ext/Techno/Body.Update.cpp
```diff
@@ -2044,3 +2057,48 @@ void TechnoExt::ExtData::UpdateTintValues()
 		calculateTint(Drawing::RGB_To_Int(pShieldType->Tint_Color), static_cast<int>(pShieldType->Tint_Intensity * 1000), pShieldType->Tint_VisibleToHouses);
 	}
 }
+
+void TechnoExt::ExtData::AddFirer(WeaponTypeClass* const Weapon, TechnoClass* const Attacker)
+{
+	if (Attacker->InLimbo)
```
TaranDahl: Why shouldn't the limbo Firestarter be recorded?

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -310,6 +310,11 @@ DEFINE_HOOK(0x6FC339, TechnoClass_CanFire, 0x6)
 
 	if (pTargetTechno)
 	{
+		const auto pTargetExt = TechnoExt::ExtMap.Find(pTargetTechno);
+
+		if (pWeaponExt->OnlyAttacker.Get() && !pTargetExt->ContainFirer(pWeapon, pThis))
```
TaranDahl, line 359: No need `.Get()` I think?

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -537,6 +542,24 @@ DEFINE_HOOK(0x6FDDC0, TechnoClass_FireAt_BeforeTruelyFire, 0x6)
...
 
+DEFINE_HOOK(0x6FDE0E, TechnoClass_FireAt_OnlyAttacker, 0x6)
+{
+	GET(TechnoClass* const, pThis, ESI);
+	GET(WeaponTypeClass* const, pWeapon, EBX);
+	GET_BASE(AbstractClass* const, pTarget, 0x8);
+
+	const auto pWeaponExt = WeaponTypeExt::ExtMap.Find(pWeapon);
+
+	if (pWeaponExt->OnlyAttacker.Get() && pTarget == pThis->Target
```
TaranDahl, line 727: Same

TaranDahl (changes requested):
I think it would be best to have the `OnlyAttacker.GroupAs` and `OnlyAttacker.Duration` functions.

PR #1808, opened 2025-08-12
[Minor] Speed=0 extension

Fly-Star-him:
When `Speed=0` or the TechnoTypes cell cannot move due to `MovementRestrictedTo`, vehicles cannot attack targets beyond the weapon's range. `Area Guard` and `Hunt` missions will also become ineffective.
- `Speed=0`或`MovementRestrictedTo`导致单元格无法移动时载具无法攻击超出武器射程的目标。`区域守卫（AreaGuard）`和`猎杀（Hunt）`任务也会失效。

- When the vehicle loses its target, you can customize whether to align the turret direction with the vehicle body.
  - When `Speed=0` or TechnoTypes cells cannot move due to `MovementRestrictedTo`, the default value is no; in other cases, it is yes.
- 当载具失去目标时可以自定义炮塔朝向和本体对齐。
  - `Speed=0`或者`MovementRestrictedTo`导致单元格无法移动时默认值为no，其他情况为yes。

In `rulesmd.ini`:
```ini
[SOMEVEHICLE]       ; VehicleType
TurretResponse=     ; boolean
```

### DIFF
--- a/src/Ext/Unit/Hooks.DisallowMoving.cpp
+++ b/src/Ext/Unit/Hooks.DisallowMoving.cpp
@@ -72,7 +72,7 @@ DEFINE_HOOK(0x736B60, UnitClass_Rotation_AI_DisallowMoving, 0x6)
 {
 	GET(UnitClass*, pThis, ESI);
 
-	return !TechnoTypeExt::ExtMap.Find(pThis->Type)->TurretResponse.Get(pThis->Type->Speed != 0) ? 0x736AFB : 0;
+	return !TechnoTypeExt::ExtMap.Find(pThis->Type)->TurretResponse.Get(!TechnoExt::CannotMove(pThis)) ? 0x736AFB : 0;
 }
 
 DEFINE_HOOK(0x73891D, UnitClass_Active_Click_With_DisallowMoving, 0x6)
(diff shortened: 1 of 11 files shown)


src/Ext/Unit/Hooks.DisallowMoving.cpp
```diff
@@ -72,12 +97,77 @@ DEFINE_HOOK(0x736B60, UnitClass_Rotation_AI_DisallowMoving, 0x6)
...
+	GET(UnitClass*, pThis, ESI);
+	GET(FireError, result, EAX);
+
+	if (result == FireError::RANGE && CannotMove(pThis))
+		R->EAX(FireError::ILLEGAL);
+
+	return 0;
+}
+
+DEFINE_HOOK(0x736E34, UnitClass_UpdateFiring_DisallowMoving, 0x6)
```
TaranDahl: Perhaps hooking `UnitClass::ApproachTarget` (0x7414E0) would be more appropriate?
Because `FootClass::ApproachTarget`(0x4D56FC) is doing something similar.

Fly-Star-him: > Perhaps hooking `UnitClass::ApproachTarget` (0x7414E0) would be more appropriate? Because `FootClass::ApproachTarget`(0x4D56FC) is doing something similar.

well, that's true.

PR #1625, opened 2025-04-13
[Vanilla Enhancement] Restore turret recoil effect

CrimRecya:
- Now you can use `TurretRecoil` to control units’ turret/barrel recoil effect when firing.
  - `TurretTravel` and `BarrelTravel` control the maximum recoil distance.
  - `TurretRecoil.Suppress` can prevent the weapon from producing this effect when firing.

In `rulesmd.ini`:
```ini
[SOMEVEHICLE]             ; VehicleType
TurretRecoil=no           ; boolean
TurretTravel=2            ; integer, pixels
TurretCompressFrames=1    ; integer, game frames
TurretHoldFrames=1        ; integer, game frames
TurretRecoverFrames=1     ; integer, game frames
BarrelTravel=2            ; integer, pixels
BarrelCompressFrames=1    ; integer, game frames
BarrelHoldFrames=1        ; integer, game frames
BarrelRecoverFrames=1     ; integer, game frames

[SOMEWEAPON]              ; WeaponType
TurretRecoil.Suppress=no  ; boolean
```

barrel

---

And:
  - Fix an issue that `FireAngle` was not taken into account when drawing barrel in `TurretShadow`.

PR #1809, opened 2025-08-13
[Minor] Fix BarrelAnimData being loaded incorrectly

saralmira:
When a unit's Section is present in the map file but lacks defined "BarrelTravel"/"BarrelCompressFrames"/"BarrelHoldFrames"/"BarrelRecoverFrames", the BarrelAnimData is incorrectly overwritten. This causes units such as [GTGCAN] to fail to render correctly after firing.

### DIFF
--- a/src/Misc/Hooks.BugFixes.cpp
+++ b/src/Misc/Hooks.BugFixes.cpp
@@ -2478,3 +2478,21 @@ DEFINE_HOOK(0x6FC8F5, TechnoClass_CanFire_SkipROF, 0x6)
 }
 
 #pragma endregion
+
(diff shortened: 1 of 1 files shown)


Coronia (approved):

saralmira:
这是来自QQ邮箱的假期自动回复邮件。你好，我最近正在休假中，无法亲自回复你的邮件。我将在假期结束后，尽快给你回复。

CrimRecya:
I think [the fix in1625](https://github.com/Phobos-developers/Phobos/pull/1625/files#diff-5e24fcccc37dd8fb07ae7b1fb3a409dfd21b68c5a0aa8782f79a4007d2afda62R1350) has already solved this problem. You can copy it over as a separate fix.

PR #1825, opened 2025-08-18
Improve/optimize arbitrary weapon/warhead detonation code

Starkku:
- Reuse a special BulletClass instance for these detonations instead of always creating a new one.
- Fix a bug that causes the detonation coords to snap to the logical target - this is not always desirable and should be handled separately where needed by making sure the detonation coords are same as target coords.


Coronia (approved):
LGTM, and worked in single/multiplayer testing

PR #1831, opened 2025-08-25
IsSimpleDeployer rework & improvements

Starkku:
## Refactoring / code changes
- Rewrote `IsSimpleDeployer` logic, all deploy logic save for the 'aftermath' aka resetting unit to guard mission (and handling Ares' type conversion) is now handled on Phobos' side. Any Ares code besides the aforementioned type conversion is disabled if Phobos is active.
- Moved `IsSimpleDeployer` code from `Hooks.Unload.cpp` to `Hooks.IsSimpleDeployer.cpp` as there is now much more of it (over 400 lines).
- Some Phobos hooks related to the matter have been consolidated e.g deploy and undeploy-related things are now handled in one hook each, with common code being in helper functions.
- Replaced use of TechnoClass Animation StageClass timer with a new bespoke one and simplified the deploy animation timer logic to allow easy use with all types of animations (normal and reversed) as well as animations using non-standard `Start`/`End` setups without any extra calculations. It also eliminates hypothetical conflict scenario where the TechnoClass animation timer is used for something else while also deploying (to my knowledge this does not currently occur, however).

## Migration
- `IsSimpleDeployer` units now obey deploying facing constraint even without deploying animation. To disable this, set `DeployDir` (defaults to `[AudioVisual]` -> `DeployDir` to -1.
- `DeployingAnim.AllowAnyDirection` has been superceded by `DeployDir`. Use value of -1 to re-enable the no facing restriction.

## New Features
- `DeployingAnims` can be used to set a list of directional deploy anims to use instead of only one animation that `DeployAnim` supports.
- `DeployDir` can now be used to customize facing constraint for deploying irregardless of presence of deploy animation. Units will automatically begin deploying once they are facing the correct way if constrained.

## Bugfixes
- `IsSimpleDeployer` `BalloonHover=true` units with `DeployToLand=false` are no longer forced to land when hovering.
- If `DeployingAnim` with `Shadow=true` is played for unit currently in air its shadow will now be drawn on ground.
- `DeployingAnim` now supports both `Normalized=true` and `Reverse=true`. Keep in mind `Reverse` uses `LoopEnd` for frame amount instead of `End` even without `LoopCount` > 1.
- `DeployingAnim` using unit drawer now also tint accordingly with the unit.

## Documentation
-------------------------------------------------
### IsSimpleDeployer facing and animation customization

- In vanilla game only units with `DeployingAnim` were constrained to a specific deploy facing and it was not customizable per unit. `DeployDir` can be set to override this per unit (defaults to `[AudioVisual]` -> `DeployDir`), including using value of -1 to disable the facing restriction.
- Multiple new options for deploy animations:
  - `DeployingAnims` can be used instead of `DeployingAnim` (if both are set, `DeployingAnims` takes precedence) to define a list of direction-specific deploy animations to play. Largest power of 2 the number of listed animations falls to is used as number of directions/animations. Less than 8 animations listed results in only first listed one being used.
  - `DeployingAnim.KeepUnitVisible` determines if the unit is **not** hidden while the animation is playing.
  - `DeployingAnim.ReverseForUndeploy` controls whether or not the animation is played in reverse for undeploying.
  - `DeployingAnim.UseUnitDrawer` controls whether or not the animation is displayed in the unit's palette and team colours or regular animation palette, including a potential custom palette.

In `rulesmd.ini`:
```ini
[SOMEVEHICLE]                          ; VehicleType
DeployDir=                             ; Facing type (integers from 0-7 or -1)
DeployingAnims=                        ; List of AnimationTypes
DeployingAnim.KeepUnitVisible=false    ; boolean
DeployingAnim.ReverseForUndeploy=true  ; boolean
DeployingAnim.UseUnitDrawer=true       ; boolean
```


yawnallday:
Can the IsSimpleDeployer and passenger Unload  functions be separated?while NoManualUnload=yes

Starkku:
> Can the IsSimpleDeployer and passenger Unload functions be separated?while NoManualUnload=yes

Probably doable but I think conditional changes on deploy cursor and/or deploy functionality itself is something that should be explored outside this PR.

For any people who may be testing, waypoint not working correctly on jumpjet units with `IsSimpleDeployer=true` anymore is a known issue pending further investigation.

src/Ext/Unit/Hooks.SimpleDeployer.cpp
```diff
@@ -0,0 +1,415 @@
...
+	{
+		auto const pType = pUnit->Type;
+		auto const pTypeExt = TechnoTypeExt::ExtMap.Find(pType);
+		auto pAnimType = pUnit->Type->DeployingAnim;
+
+		if (pTypeExt->DeployingAnims.size() > 0)
+			pAnimType = GeneralUtils::GetItemForDirection<AnimTypeClass*>(pTypeExt->DeployingAnims, pUnit->PrimaryFacing.Current());
+
+		auto const pAnim = GameCreate<AnimClass>(pAnimType, pUnit->Location, 0, 1, 0x600, 0,
+			!isDeploying ? pTypeExt->DeployingAnim_ReverseForUndeploy : false);
```
CrimRecya: `!isDeploying && pTypeExt->DeployingAnim_ReverseForUndeploy`

src/Ext/Unit/Hooks.SimpleDeployer.cpp
```diff
@@ -0,0 +1,415 @@
...
+
+	if (pThis->Deploying || pThis->Undeploying)
+	{
+		const auto pTypeExt = TechnoTypeExt::ExtMap.Find(pThis->Type);
+
+		if (pTypeExt->DeployingAnim_KeepUnitVisible || (pThis->Deploying && !pThis->DeployAnim))
+			return KeepUnitVisible;
+	}
+
+	return 0;
```
CrimRecya: Here `return 0;` will repeatedly check `pThis->Deploying || pThis->Undeploying`

src/Ext/Unit/Hooks.SimpleDeployer.cpp
```diff
@@ -0,0 +1,415 @@
...
+}
+
+// Disable deploy cursor if Ares type conversion on deploy is available and the new type is not allowed to move to the cell.
+DEFINE_HOOK(0x7010C1, TechnoClass_CanShowDeployCursor_IsSimpleDeployer, 0x5)
+{
+	enum { DoNotAllowDeploy = 0x700DCE };
+
+	GET(UnitClass*, pThis, ESI);
+
+	if (pThis->Type->IsSimpleDeployer && AresFunctions::ConvertTypeTo)
```
CrimRecya, line 413: I'm not sure if it's necessary to check `AresFunctions::ConvertTypeTo` here? Because we also has our own `ConvertToType`, did I miss something?

CrimRecya (approved):
LGTM, only a minor issue and a small question for me.
Additionally, I think that `return condition ? addr1 : addr2;` may be better?

src/Ext/Unit/Hooks.SimpleDeployer.cpp
CrimRecya: I think I understand this question now. Just that `Convert.Deploy` is not a reimplement.

Starkku:
Turns out the broken deploy during waypoint mode pathing was a bug introduced in #1759. A fix for it has been pushed in develop branch and this PR has been rebased on it.

PR #1832, opened 2025-08-26
Update the super weapon to play music.

ahasasjeb:
```ini
[SOMESW]              ; SuperWeaponType
Music.Theme=          ; Soundtrack theme ID from thememd.ini (such as GodsendOne)
Music.Duration=0      ; integer, game frames; 0 or below means do not auto-stop,with the game speed set to 4, 15 frames equal 1 second.
Music.AffectedHouses= ; owner|allies|enemies|all (default all)
```

- `Music.Theme` selects the soundtrack theme by its ID defined in `thememd.ini` (such as `GodsendOne`).
- `Music.Duration` sets how long to keep playing, in game frames. 0 or below means no auto-stop.
- If a different theme is already playing, it will be replaced when the superweapon fires.
- When the timer completes, the theme is stopped only if the currently playing theme still equals the configured `Music.Theme`; if music was changed during the countdown, it will not be altered.
- `Music.AffectedHouses` determines which houses will hear and be affected by the superweapon music on their client: `owner`, `allies`, `enemies`, or `all` (default). Playback and auto-stop are applied only for those houses.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -642,3 +642,4 @@ This page lists all the individual contributions to the project by their author.
 - **Damfoos** - extensive and thorough testing
 - **Dmitry Volkov** - extensive and thorough testing
 - **Rise of the East community** - extensive playtesting of in-dev features
+- **ahasasjeb** - Add music to super weapons
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -1226,12 +1226,14 @@ In `rulesmd.ini`:
 [SOMESW]              ; SuperWeaponType
 Music.Theme=          ; Soundtrack theme ID from thememd.ini (such as GodsendOne)
 Music.Duration=0      ; integer, game frames; 0 or below means do not auto-stop
+Music.AffectedHouses= ; owner|allies|enemies|all (default all)
 ```
 
 - `Music.Theme` selects the soundtrack theme by its ID defined in `thememd.ini` (such as `GodsendOne`).
 - `Music.Duration` sets how long to keep playing, in game frames. 0 or below means no auto-stop.
 - If a different theme is already playing, it will be replaced when the superweapon fires.
 - When the timer completes, the theme is stopped only if the currently playing theme still equals the configured `Music.Theme`; if music was changed during the countdown, it will not be altered.
+- `Music.AffectedHouses` determines which houses will hear and be affected by the superweapon music on their client: `owner` (firer only), `allies`, `enemies`, or `all` (default). Playback and auto-stop are applied only for those houses.
 
 ## Technos
 
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -1231,7 +1231,7 @@ Music.Duration=0      ; integer, game frames; 0 or below means do not auto-stop
 - `Music.Theme` selects the soundtrack theme by its ID defined in `thememd.ini` (such as `GodsendOne`).
 - `Music.Duration` sets how long to keep playing, in game frames. 0 or below means no auto-stop.
 - If a different theme is already playing, it will be replaced when the superweapon fires.
-- When the timer completes, the current theme is stopped.
+- When the timer completes, the theme is stopped only if the currently playing theme still equals the configured `Music.Theme`; if music was changed during the countdown, it will not be altered.
 
 ## Technos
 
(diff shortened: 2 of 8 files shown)


ahasasjeb:
> Perhaps a flag should be added to control which houses will be affected.

ok

ahasasjeb:
Add:
- `Music.AffectedHouses` determines which houses will hear and be affected by the superweapon music on their client: `owner` (firer only), `allies`, `enemies`, or `all` (default). Playback and auto-stop are applied only for those houses.

ahasasjeb:
> where is dll?

Go build it yourself at https://github.com/ahasasjeb/Phobos/tree/develop

TaranDahl (changes requested):
Should add docs in `CREDITS.md`

ahasasjeb:
> Should add docs in `CREDITS.md`

done

TaranDahl (approved):

Coronia:
I'm not sure if `Music.Duration` is a reasonable design. It can easily be interrupted by Esc screen, which will make the theme always play regardless of duration unless getting back to the game, and you can also switch music there. But on the other hand, it adds another per frame check which needs to loop check all houses' owned SWs which is bad for performance. Personally I would like to remove this tag and the related logic, but I'd like to hear if others have such demand

Coronia:
If we're keeping this feature then it'd better be done in a way like AutoDeath and TransportReloader, which is making a global vector to store the superweapon entity and only iterate it instead of looping all houses + all SWs. That would require a proper InvalidatePointer modification tho

Another idea is to only store a timer in HouseExt, which will be set once such SW is fired, and checking if it's completed in the same hook for music reset. This could at least decrease the complexity of checking every SW. It'll change the behavior a bit tho: if the original superweapon is gone (by building destruction or such), the music will still be reset once the time meets

fagonghaiwo:
maybe Music.AffectedHouses could use (owner/self|allies/ally|team|enemies/enemy|all) , just like other features whitch affect houses. At least, team is different from ally.

Coronia:
I'm not sure if a list is needed, given there's only one music which can be played at once, and making different SWs (and more sources in the future perhaps) temporarily change music in the same time would just make it very messy. Hence, my suggestion would be simpler:
- add a timer for HouseExt, which will be set to tick whenever a music switch with duration happens
- if the timer is ended, it'll trigger the music switchback
- if another music event occurs, it'll reset timer to the new duration, or stop it if it doesn't have a duration

in this case it's easier to understand and implement, less messy and perf costy, and can better interact with other (potential) music playing methods, making a whole dynamic music system as @Metadorius proposed
- trigger action 20 Play Theme, which could have a param to decide its duration
- play music when receive damage and killing enemies like RA3 did

and to achieve this system, it would be better to extract the music display part to a standalone function, so it can be called by SW and different sources for reusability

Flactine:
Hi, I’ve been testing the **SuperWeapon music switching** feature and noticed a potential issue related to non-looping themes and `Music.Duration=0`.

#### Observed behavior

* When a superweapon triggers music with:

  * `Music.Duration=0`
  * the selected theme is **not looping** (`repeat=no`),
* After the theme finishes playing, **no background music resumes automatically**.
* The game remains silent until the player manually changes music via the options menu or another event triggers music.

#### Expected / reference behavior

In vanilla RA2/YR, this situation behaves differently:

* When music is triggered via **Map Trigger Action 20 (Play Music)** with a non-looping theme (`repeat=no`),
* After the triggered theme finishes, the game **automatically resumes normal background music**.

This makes the Action 20 behavior feel more seamless compared to the current superweapon-triggered music logic.

So, could the logic be improved or an INI option be added so that the music triggered by the super weapon can resume playing normal music after it ends?

---

最近我在我的MOD里尝试了此功能，并且发现了一个情况；

#### 表现

* 如果超武触发音乐时：

  * `Music.Duration=0`
  * 这个音乐不是**循环音乐** (没有`repeat=yes`),
* 这首音乐播放结束后， 游戏就不会再播放其它音乐了，除非玩家去设置里手动重新开始音乐。

如果超武触发的音乐不是循环音乐（没有repeat=yes），并且超武触发音乐的参数`Music.Duration=0`，那么这首音乐播放完成之后，游戏就不会再播放任何音乐了，除非去设置里手动开启音乐。

#### 预期/参考行为

在原版里，使用地图触发来播放音乐的话：

* 如果使用触发20来播放一首同样不是**循环音乐** (没有`repeat=yes`)的音乐
* 当这首音乐播放结束后，游戏会自动重新开始播放正常的背景音乐。

这种播放方式似乎更适合在游戏中的表现。

所以，能否改进一下播放结束时的逻辑，或是增加INI选项，来让该音乐结束后，游戏能重新开始播放正常的BGM？

PR #1833, opened 2025-08-29
fix:Pause AI building production and spawner aircraft under EMP (respect ImmuneToEMP)

ahasasjeb:
Added functionality to freeze AI-controlled factory production when under EMP.

### DIFF
--- a/src/Ext/Techno/Body.Update.cpp
+++ b/src/Ext/Techno/Body.Update.cpp
@@ -10,6 +10,8 @@
 #include <Kamikaze.h>
 #include <JumpjetLocomotionClass.h>
 #include <FlyLocomotionClass.h>
+#include <BuildingClass.h>
+#include <FactoryClass.h>
 
 #include <Ext/Anim/Body.h>
 #include <Ext/Bullet/Body.h>
@@ -18,8 +20,6 @@
 #include <Ext/Scenario/Body.h>
 #include <Utilities/EnumFunctions.h>
 #include <Utilities/AresFunctions.h>
-#include <BuildingClass.h>
-#include <FactoryClass.h>
 
 
 // TechnoClass_AI_0x6F9E50
@@ -1629,8 +1629,9 @@ void TechnoExt::ExtData::UpdateTemporal()
 void TechnoExt::ExtData::UpdateRearmInEMPState()
 {
 	const auto pThis = this->OwnerObject();
+	const bool underEMP = pThis->IsUnderEMP();
 
-	if (!pThis->IsUnderEMP() && !pThis->Deactivated)
+	if (!underEMP && !pThis->Deactivated)
 		return;
 
 	const auto pTypeExt = this->TypeExtData;
@@ -1642,17 +1643,18 @@ void TechnoExt::ExtData::UpdateRearmInEMPState()
 		pThis->ReloadTimer.StartTime++;
 
 	// Freeze AI-controlled factory production while building is EMPed
-	if (pThis->IsUnderEMP() && pThis->WhatAmI() == AbstractType::Building)
+	if (underEMP)
 	{
-		auto const pBuilding = static_cast<BuildingClass*>(pThis);
-
-		if (pBuilding->Owner && !pBuilding->Owner->IsControlledByHuman())
+		if (auto const pBuilding = abstract_cast<BuildingClass*, true>(pThis))
 		{
-			if (auto const pFactory = pBuilding->Factory)
+			if (pBuilding->Owner && !pBuilding->Owner->IsControlledByHuman())
 			{
-				if (pFactory->Object && pFactory->Production.Rate > 0)
+				if (auto const pFactory = pBuilding->Factory)
 				{
-					pFactory->Production.Timer.StartTime++;
+					if (pFactory->Object && pFactory->Production.Rate > 0)
+					{
+						pFactory->Production.Timer.StartTime++;
+					}
 				}
 			}
 		}
--- a/src/Ext/Techno/Body.Update.cpp
+++ b/src/Ext/Techno/Body.Update.cpp
@@ -18,6 +18,8 @@
 #include <Ext/Scenario/Body.h>
 #include <Utilities/EnumFunctions.h>
 #include <Utilities/AresFunctions.h>
+#include <BuildingClass.h>
+#include <FactoryClass.h>
 
 
 // TechnoClass_AI_0x6F9E50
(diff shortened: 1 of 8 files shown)


TaranDahl (changes requested):
More documentation needs to be added. Could look into other PRs.

src/Ext/Techno/Body.Update.cpp
```diff
@@ -1638,6 +1640,23 @@ void TechnoExt::ExtData::UpdateRearmInEMPState()
 
 	if (pThis->ReloadTimer.InProgress() && pTypeExt->NoReload_UnderEMP.Get(RulesExt::Global()->NoReload_UnderEMP))
 		pThis->ReloadTimer.StartTime++;
+
+	// Freeze AI-controlled factory production while building is EMPed
+	if (pThis->IsUnderEMP() && pThis->WhatAmI() == AbstractType::Building)
+	{
+		auto const pBuilding = static_cast<BuildingClass*>(pThis);
```
CrimRecya: No need to repeatedly check `IsUnderEMP()`, and you can use `abstract_cast` like `if (auto const pBuilding = abstract_cast<BuildingClass*, true>(pThis))`.

src/Ext/Techno/Body.Update.cpp
```diff
@@ -18,6 +18,8 @@
 #include <Ext/Scenario/Body.h>
 #include <Utilities/EnumFunctions.h>
 #include <Utilities/AresFunctions.h>
+#include <BuildingClass.h>
+#include <FactoryClass.h>
```
CrimRecya: These 2 lines can be placed above of line 13.

src/Ext/Techno/Hooks.Misc.cpp
```diff
@@ -155,6 +160,15 @@ DEFINE_HOOK(0x6B78D3, SpawnManagerClass_Update_Spawns, 0x6)
 	auto const pOwner = pThis->Owner;
 	auto const pTypeExt = TechnoExt::ExtMap.Find(pOwner)->TypeExtData;
 
+	if (auto const pBuilding = abstract_cast<BuildingClass*, true>(pOwner))
+	{
+		if (pBuilding->Type->Powered && !pBuilding->IsPowerOnline())
```
TaranDahl: It would be best to have a toggle. IMO this should not be considered as a fix.

src/Ext/Techno/Hooks.Misc.cpp
```diff
@@ -155,6 +160,15 @@ DEFINE_HOOK(0x6B78D3, SpawnManagerClass_Update_Spawns, 0x6)
...
+			return 0;
+	}
+
+	if (const auto pOwnerType = pOwner->GetTechnoType(); pOwnerType && pOwnerType->SpawnRegenRate <= 0)
```
TaranDahl, line 164: BTW, what's the vanilla behavior of `SpawnRegenRate <= 0`? Is it buggy?

src/Ext/Techno/Hooks.Misc.cpp
ahasasjeb: No toggle needed. EMP behavior is already controlled by ImmuneToEMP—if not immune, the building should be paralyzed, including spawner logic. Letting an EMP’d building keep launching spawns is inconsistent and unintuitive.

src/Ext/Techno/Hooks.Misc.cpp
ahasasjeb: Vanilla SpawnRegenRate <= 0 = instant spawning (timer expires immediately). This check prevents spam-spawning in Phobos's custom spawn queue feature where <= 0 should mean "disabled".

src/Ext/Techno/Hooks.Misc.cpp
TaranDahl: In vanilla (or Ares), will the FootClass affected by the EMP pause the spawner logic?

ahasasjeb: Vanilla and Ares likely did not account for buildings acting as spawners. As a result, a powered building with ImmuneToEMP=no can still launch spawns while under EMP, which is inconsistent with RA2 semantics. EMP implies a power blackout and a paralyzed structure, so a carrier‑style spawner should not be doing anything, least of all launching aircraft.

TaranDahl: 我是说，我们可以参考FootClass被EMP时其子机逻辑是否会暂停。如果FootClass会暂停那么我们可以将这个作为一个修复，如果不会那最好留一个开关。

src/Ext/Techno/Body.Update.cpp
```diff
@@ -1638,6 +1641,24 @@ void TechnoExt::ExtData::UpdateRearmInEMPState()
 
 	if (pThis->ReloadTimer.InProgress() && pTypeExt->NoReload_UnderEMP.Get(RulesExt::Global()->NoReload_UnderEMP))
 		pThis->ReloadTimer.StartTime++;
+
+	// Freeze AI-controlled factory production while building is EMPed
+	if (underEMP)
```
CrimRecya: Is there any reasons that need to exclude `Deactivated`?

src/Ext/Techno/Hooks.Misc.cpp
```diff
@@ -155,6 +160,15 @@ DEFINE_HOOK(0x6B78D3, SpawnManagerClass_Update_Spawns, 0x6)
...
+		return 0;
+
```
CrimRecya: Two questions:
- I saw in your doc that you want to prevent the respawn of the spawnee in EMP state, but is this necessary? Will the building still launch them in EMP state?
- And, you used `return 0;`, it will not affect any of the vanilla logic, but it will cause issues with the `Spawns.Queue` logic later on.

src/Ext/Techno/Hooks.Misc.cpp
ahasasjeb: done

src/Ext/Techno/Hooks.Misc.cpp
CrimRecya: Please take a clear look at the problem first and do not mark it as resolved at will

ahasasjeb: Re-review changes, please

TaranDahl: I also have the same concern that your code may not work as you expect because it `return 0`. Have you tested it?

ahasasjeb: Code works fine

CrimRecya: In EMP state, `SpawnManagerClass` will not make any updates, which means these are unnecessary.
Even if this has any effect, `return 0` means continuing to execute the game's vanilla instructions, which will break `Spawns.Queue`'s logic.

What I want to express is that the code here actually has no effect. This approach is no different from the vanilla, as `SpawnManagerClass` would never work under EMP.
If you want to pause the timer, you can do it like the production. Although this may not be the most efficient approach, it is feasible.

PR #1841, opened 2025-09-03
Further adjustments to health threshold checks

Starkku:
@Metadorius 

- Makes lower bound check of health threshold checks (weapon, warhead, crit) exclusive e.g equal to it does not satisfy the condition. This allows for creating distinct ranges for weapon/warhead eligibility e.g no overlapping.
- Ignore health threshold checks for objects with zero health. They are not relevant consideration for current health checks (other checks filter out unalive objects), if need to consider them in future appears, conditions can be added then and there.

Migration notice is currently absent due to uncertainty about where to place it.

### DIFF
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -2111,7 +2111,7 @@ Conventional.IgnoreUnits=false  ; boolean
 
 ### Customizable Warhead trigger conditions
 
-- It is now possible to make warheads only trigger when target's (TechnoTypes only) HP is above/below or equal to certain percentage. Both conditions need to evaluate to true in order for the warhead to trigger.
+- It is now possible to make warheads only trigger when target's (TechnoTypes only) HP is above and/or below/equal to certain percentage. Both conditions need to evaluate to true in order for the warhead to trigger.
 - If set to `false`, `AffectsNeutral` makes the warhead can't damage or affect target that belongs to neutral house.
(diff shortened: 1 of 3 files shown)


docs/Fixed-or-Improved-Logics.md
```diff
@@ -2111,7 +2111,7 @@ Conventional.IgnoreUnits=false  ; boolean
 
 ### Customizable Warhead trigger conditions
 
-- It is now possible to make warheads only trigger when target's (TechnoTypes only) HP is above/below or equal to certain percentage. Both conditions need to evaluate to true in order for the warhead to trigger.
+- It is now possible to make warheads only trigger when target's (TechnoTypes only) HP is above and/or below/equal to certain percentage. Both conditions need to evaluate to true in order for the warhead to trigger.
```
Metadorius: I think the whole phrase would need a redo to be clear at this point

Metadorius: I also think a note mentioning that health 0 is exempt should be done

src/Ext/Techno/Body.cpp
```diff
@@ -727,8 +727,11 @@ bool TechnoExt::HandleDelayedFireWithPauseSequence(TechnoClass* pThis, int weapo
 
 bool TechnoExt::IsHealthInThreshold(TechnoClass* pObject, double min, double max)
 {
+	if (!pObject->Health)
+		return true;
```
Metadorius: In case of 0/0 health I think this is valid, but for health 0 objects with strength >= I think this has to check for whether the min HP is equal to 0

Coronia (approved):

PR #1847, opened 2025-09-08
[Vanilla Fix] Fix an issue that the currently hovered planning node not update up-to-date

CrimRecya:
- Fix an issue that the currently hovered planning node not update up-to-date, such as using hotkeys to select technos
  > 修复高等路径学


CrimRecya:
> Can you please wrote it on human readable way , this will make thing hard to review because not everyone understand ASM stuffs .

Changed.

PR #1848, opened 2025-09-09
[Minor] Fix create building map trigger action

Starkku:
Fixes an issue with map trigger action 125 causing buildups to play incorrectly sometimes. A separate issue remains where any building that is mid-buildup will interrupt buildup if it changes owner but this is not exclusive to buildings created through the trigger action.

Also re-enabled use fo the third parameter but this time it turns buildups off even if otherwise available if set to != 0.


fagonghaiwo:
it works
bandicam 2025-09-09 21-54-06-732

Metadorius:
Since the bug has been there for a long time, it shouldn't be tagged with minor

Coronia (approved):

PR #1853, opened 2025-09-11
Allow using <Player @ X> as trigger owner in skirmish/MP

Starkku:
Skirmish / multiplayer maps can now use <Player @ X> as trigger owner. If the player is not present in game, the trigger is destroyed and never sprung.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -280,7 +280,7 @@ This page lists all the individual contributions to the project by their author.
   - Build area customizations
   - `Scorch` / `Flamer` fire animation customization
   - EM Pulse cannon logic improvements
-  - `<Player @ X>` as owner for pre-placed objects
+  - `<Player @ X>` as owner for pre-placed objects and triggers
   - Custom exit cell for infantry factory
   - Vehicles keeping target on move command
   - `IsSonic` wave drawing crash fix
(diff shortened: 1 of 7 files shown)


TaranDahl (approved):
I think WW must got shit in their mind to come up with the design of setting the trigger owner as HouseTypeClass instead of HouseClass. 
Is there any way to solve these problems at the root?

Starkku:
Shoving HouseClass pointer in what game expects to be HouseTypeClass and fixing the use cases on game's side I suppose. Problem remains any other extensions etc. that might not be aware of this change so I do not consider this approach safe. Other option which isn't really fixing the 'root cause' is using extension class instead of a map but it is basically just a different way of doing what has been done here.

Starkku:
Map editors/files also treat the field for trigger types as HouseTypeClass ID. Changing this would brick a lot of map file/editor compatibility. Not sure if that's worth the ability to allow matching triggers to specific HouseClass instances, given that <Player @ X> covers skirmish/MP already and singleplayer maps generally speaking have one HouseClass per HouseTypeClass to begin with.

TaranDahl:
I'm just asking randomly. It's not something to be solved here.

Starkku:
Yeah I am just explaining (or trying to) why the current approach (or alternatively the extension class one but that just depends on preference I suppose) is the sane one for now atleast.

TaranDahl:
Going to merge this in 24h since there is nobody willing to test it.

Starkku:
It's been in public versions of Project Phantom and has worked there, if that's worth anything.

TaranDahl:
That's convincing enough for me.

PR #1854, opened 2025-09-11
Attached animation draw offset customizations

Starkku:
### Draw offset customization

- `YDrawOffset.ApplyBracketHeight` makes Y axis position follow it's owner object's selection bracket height (for buildings, this is based on `Height` and `Foundation`, for others it is influenced by `PixelSelectionBracketDelta`) if it is attached to one.
  - By default this will only apply if the bracket position is negative e.g it is moved upwards from the object center. If `YDrawOffset.InvertBracketShift` is set to true, the opposite is true and negative shift is ignored.
  - The bracket-based shift can be further adjusted with offset from `YDrawOffset.BracketAdjust`, overridden by `YDrawOffset.BracketAdjust.Buildings` for buildings only.
 
In `artmd.ini`:
```ini
[SOMEANIM]                            ; AnimationType
YDrawOffset.ApplyBracketHeight=false  ; boolean
YDrawOffset.InvertBracketShift=false  ; boolean
YDrawOffset.BracketAdjust=0,0         ; X,Y, pixels relative to default
YDrawOffset.BracketAdjust.Buildings=  ; X,Y, pixels relative to default
```

### AttachEffect animation-specific

 - `Animation.DrawOffsetN` (where N is integer starting from 0) can be used to define draw offset rules for the animation. These are parsed starting from index 0 until offset with value 0,0 is encountered.
   - `Animation.DrawOffsetN.RequiredTypes` contains list other AttachEffectTypes that need to be attached on the same techno as the current one for the draw offset rule to apply.

In `rulesmd.ini`:
```ini
Animation.DrawOffsetN=0,0                          ; X,Y, pixels relative to default
Animation.DrawOffsetN.RequiredTypes=               ; List of AttachEffectTypes
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -308,6 +308,7 @@ This page lists all the individual contributions to the project by their author.
   - Tank Bunker improvements
   - `ProjectileRange` weapon range modifiers interaction fix
   - Berzerk duration stacking behaviour customization
+  - Attached animation draw offset customizations
 - **Morton (MortonPL)**:
   - `XDrawOffset` for animations
   - Shield passthrough & absorption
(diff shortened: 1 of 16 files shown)


TaranDahl (approved):

TaranDahl:
Going to merge this in 24h since there is no other person willing to test it.

PR #1859, opened 2025-09-15
Attack and damage technos under ground 

TaranDahl:
### Attack technos underground

- Now, you can enable projectiles to attack technos underground.
  - To actually damage the technos, you need [AffectsUnderground](#damage-technos-underground).

In `rulesmd.ini`:
```ini
[SOMEPROJECTILE]      ; Projectile
AU=false              ; boolean
```

### Damage technos underground

- Now you can make the warhead damage technos underground!
  - To allow weapons to target underground technos, you need [AU](#attack-technos-underground).
- Notice that if the projectile detonates underground, its animation effect may look strange.
  - You can use `[SOMEWARHEAD] -> PlayAnimUnderground=false` to prevent the warhead animation from playing when the projectile detonates underground.
  - You can also use `[SOMEWARHEAD] -> PlayAnimAboveSurface=true` to make the warhead animation play on the ground directly above when the projectile detonates underground.

In `rulesmd.ini`:
```ini
[SOMEWARHEAD]                         ; WarheadType
AffectsUnderground=false              ; boolean
PlayAnimUnderground=true              ; boolean
PlayAnimAboveSurface=false            ; boolean,
```

### DIFF
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -2295,7 +2295,7 @@ In `rulesmd.ini`:
 [SOMEWARHEAD]                         ; WarheadType
 AffectsUnderground=false              ; boolean
 PlayAnimUnderground=true              ; boolean
-PlayAnimAboveSurface=false            ; boolean,
+PlayAnimAboveSurface=false            ; boolean
 ```
 
 ### Detonate Warhead on all objects on map
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -747,6 +747,17 @@ OnlyUseLandSequences=false  ; boolean
 
 ## Projectiles
 
+### Attack technos underground
+
+- Now, you can enable projectiles to attack technos underground.
+  - To actually damage the technos, you need [AffectsUnderground](#damage-technos-underground).
+
+In `rulesmd.ini`:
+```ini
+[SOMEPROJECTILE]      ; Projectile
+AU=false              ; boolean
+```
+
 ### Parabombs
 
 - Restored feature from Red Alert 1 (also partially implemented in Ares but undocumented, if used together Phobos' version takes priority) that allows projectiles to be parachuted down to ground if fired by an aerial unit.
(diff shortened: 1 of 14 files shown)


Coronia:
According to our test, only `Inviso` projectiles can effectively penetrate ground and hit underground units. Others will all explode when impact the ground, while still able to damage underground units if `CellSpread` is large enough.

This is a hardcoded thing so it's fine to keep it as it is, but it should be noted in the doc.

docs/New-or-Enhanced-Logics.md
```diff
@@ -747,6 +747,21 @@ OnlyUseLandSequences=false  ; boolean
...
+  - To actually damage the technos, you need [AffectsUnderground](#damage-technos-underground).
+
+In `rulesmd.ini`:
+```ini
+[SOMEPROJECTILE]      ; Projectile
+AU=false              ; boolean
+```
+
+```{note}
+Only vanilla projectiles with `Inviso=yes` set or [Phobos projectiles](#projectile-trajectories) with `SubjectToGround=yes` enabled can go beneath the ground. Otherwise, the projectile will be forced to detonate upon hitting the ground.
```
CrimRecya: Besides `Inviso`, maybe only `Straight` with `Trajectory.Straight.SubjectToGround=false` and `Bombard` with `Trajectory.Bombard.SubjectToGround=false`  can hit.

src/Ext/Techno/Body.cpp
```diff
@@ -987,6 +989,12 @@ DEFINE_HOOK(0x6F4500, TechnoClass_DTOR, 0x5)
 {
 	GET(TechnoClass*, pItem, ECX);
 
+	if (TechnoExt::ExtMap.Find(pItem)->UndergroundTracked)
+		ScenarioExt::Global()->UndergroundTracker.Remove(pItem);
+
+	if (TechnoExt::ExtMap.Find(pItem)->SpecialTracked)
+		ScenarioExt::Global()->SpecialTracker.Remove(pItem);
+
```
CrimRecya: You can place these in `TechnoExt::ExtData::~ExtData()`
```
	if (this->UndergroundTracked)
		ScenarioExt::Global()->UndergroundTracker.Remove(pThis);

	if (this->SpecialTracked)
		ScenarioExt::Global()->SpecialTracker.Remove(pThis);
```

docs/New-or-Enhanced-Logics.md
TaranDahl: I thought you were planning to add `SubjectToGround` for other types of trajectories?

CrimRecya: Not very accurate. Even after the update, `SubjectToGround` still only affects the selection of attacking area. Just in `Straight` and `Bombard`, because their judgment criteria are simple, allowing them to penetrate the ground will not cause any problems.

Issue #1851, opened 2025-09-10
AI removes from existence the offensive structure with UndeploysInto when AI player is forced to undeploy


FS-21:
### Description

I found the culprit of why deployed offensive units into structures, controlled by AI dissappears after finishing the undeploy animation:
<img width="553" height="52" alt="Image" src="https://github.com/user-attachments/assets/499e3ae9-88b7-45ae-ac9d-6ed75218a458" />

Started with this commit from the **build-34**: e359a87ba7b0f1ab428f7bd02c58775a2e551be4

the AI undeployment "from structure into vehicle" was ruined when an enemy forced the deployed vehicle/structure to undeploy (for example when the deployer's weapon has a minimum range and the enemy is inside the minimum range).
The deployer after finishing the undeploy animation is simply "banished" from existence.

### Phobos Version

Any version since build-34 (to be exact since commit e359a87ba7b0f1ab428f7bd02c58775a2e551be4 )

### Conditions to reproduce

Wait for the AI to deploy any affected unit and attack the deployed unit forcing the AI to undeploy into vehicle again.

### INI code

```ini
Any offensive vehicle-To-Building that uses the Tiberian Sun logic and with "UndeploysInto" tag for the transformation from structure:

[ARTYDEP]
UIName=name:ARTYDEP
Name=Deployed Avd. Artillery
TechLevel=11
Strength=250
Points=28
BattlePoints=14
FireAngle=25
Cost=1400
Soylent=700
Bounty.Value=140
Power=0
Armor=deployed_light_armor
Sight=9
Explosion=TSTWLT070,TSS_BANG48,TSS_BRNL58,TSS_CLSN58,TSS_TUMU60
MaxDebris=4
DebrisAnims=TSDBRIS1LG,TSDBRIS2LG,TSDBRIS3LG,TSDBRIS4LG,TSDBRIS5LG,TSDBRIS6LG,TSDBRIS7LG,TSDBRIS8LG,TSDBRIS9LG,TSDBRS10LG,TSDBRIS1SM,TSDBRIS2SM,TSDBRIS3SM,TSDBRIS4SM,TSDBRIS5SM,TSDBRIS6SM,TSDBRIS7SM,TSDBRIS8SM,TSDBRIS9SM,TSDBRS10SM
Prerequisite=NODWEAP,TSNARADR
UndeploysInto=ART2
Owner=NodCountry,NodCountry2
BaseNormal=no
VoiceSelect=TSHVRSelect
[...]

PR #1868, opened 2025-09-24
Fixed AI bug when an UndeploysInto building undeploys

FS-21:
The AI bug never created the expected unit when is undeployed.

Tested in Map action 60 & forcing the AI to undeploy (for example when the affected deployed has a minimum range in the weapon and the enemy is inside that range).

Bugfix of the reported issue:
https://github.com/Phobos-developers/Phobos/issues/1851

Since this is only fixing a very small coding mistake no documentation of any kind should be required.


TaranDahl (changes requested):
Add another flag

PR #1879, opened 2025-09-28
Implement open-topped building support and fix building self-attack issue

Metadorius:
- Fixed `OpenTopped` to work with buildings.
- Fixed buildings giving an order to attack self when unloading (manifested with opentopped buildings and garrisons leaving buildings sometimes).

TODO: need to move UpdateTypeData_Foot thing related to OpenTopped to generic UpdateTypeData


Metadorius:
> There are many buildings using the occupant logic. It would be very troublesome to modify them one by one to OpenTopped.
> It would be best to have a global switch, such as `ReplaceOccupiableAsOpenTopped`, to batch replace it on `LoadFromINI`.

it is not a drop-in replacement, currently the engine lacks logics for that

PR #1582, opened 2025-03-20
[General Purpose] New trajectory system

CrimRecya:
- Universalized many of the original functions (#1293 , #1294 , #1295 , #1374 , #1404 , #1481 ).
- Fixed the issue where the trajectory projectile had to wait for one frame before detonating.

---

### Projectile trajectories

- Projectiles can now have customizable trajectories.
  - `Trajectory` should not be combined with original game's projectile trajectory logics (`Arcing`, `ROT`, `Vertical` or `Inviso`). Attempt to do so will result in the other logics being disabled and a warning being written to log file. It also has linkage functions with `Inaccurate`, `BallisticScatter.Min`, `BallisticScatter.Max`, `Gravity`, `SubjectToGround`.
  - `Trajectory.Speed` defines the speed of the projectile, which unlike `Speed` used by `ROT` > 0 projectiles is defined on projectile not weapon.
    - In `Trajectory=Straight`, it refers to the whole distance speed of the projectile.
    - In `Trajectory=Bombard`, it refers to the initial speed of the projectile.
    - In `Trajectory=Missile`, it refers to the final speed of the projectile. `Trajectory.Speed` will be fixed at 192 by setting `Trajectory.Missile.UniqueCurve=true`.
    - In `Trajectory=Engrave`, it refers to the horizontal engrave speed of the projectile and it cannot exceed 128. Recommend set as about 40.
    - In `Trajectory=Parabola`, it refers to the horizontal velocity of the projectile and is only used for modes 0, 3, or 5.
    - In `Trajectory=Tracing`, it refers to the moving speed of the projectile.
  - `Trajectory.Duration` controls the duration the projectile can exist, and at the end of the time, the projectile will detonate. If it is a non positive number, there will be no timing.
    - In `Trajectory=Engrave`, if it is a non positive number, automatically use `Trajectory.Engrave.SourceCoord` and `Trajectory.Engrave.TargetCoord` to calculate the process duration. At this point, `Trajectory.Engrave.TargetCoord` can be regarded as the endpoint coordinates of the cutting line segment.
    - In `Trajectory=Tracing`, if set to zero, use weapon's `ROF`-10 as the duration. At least 1 frame. If it is negative, do not time it.
  - `Trajectory.TolerantTime` controls how long the projectile will detonate after losing the target. If it is 0, it will detonate directly when switching targets.
[...]

PR #1886, opened 2025-10-04
[Review Only] Trajectory base class refactor

CrimRecya:
Split from #1582 . For review only.

### DIFF
--- a/src/Ext/Bullet/Body.cpp
+++ b/src/Ext/Bullet/Body.cpp
@@ -78,12 +78,9 @@ bool BulletExt::ExtData::CheckOnEarlyUpdate()
 void BulletExt::ExtData::CheckOnPreDetonate()
 {
 	const auto pBullet = this->OwnerObject();
-	const auto pBulletTypeExt = this->TypeExtData;
 
-	// To skip all extra effects, no damage, no anims...
-	pBullet->Health = 0;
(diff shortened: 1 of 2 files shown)


src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -284,7 +265,7 @@ DEFINE_HOOK(0x5218F3, InfantryClass_WhatWeaponShouldIUse_DeployFireWeapon, 0x6)
 #pragma region TechnoClass_GetFireError
 DEFINE_HOOK(0x6FC339, TechnoClass_CanFire, 0x6)
 {
-	enum { CannotFire = 0x6FCB7E };
+	enum { CannotFire = 0x6FCB7E, TemporarilyCannotFire = 0x6FCD0E };
```
secsome, line 268: Did I missing something? where does the `TemporarilyCannotFire` being used?

CrimRecya: Sorry for not organizing it cleanly enough. It was used [here](https://github.com/Phobos-developers/Phobos/pull/1890/commits/f26a28af301276e945f33578321d2ac92484c41a#diff-a64b29573d90ea4e690de090d75e4ac5877a3b28aa38439cd5d38497c0703a78R346).

PR #1888, opened 2025-10-04
[Review Only] Trajectory new types

CrimRecya:
Split from #1582 . For review only. (The latest commit)


src/Ext/Bullet/Trajectories/PhobosVirtualTrajectory.cpp
```diff
@@ -0,0 +1,213 @@
...
+}
+
+bool VirtualTrajectoryType::Save(PhobosStreamWriter& Stm) const
+{
+	this->PhobosTrajectoryType::Save(Stm);
+	const_cast<VirtualTrajectoryType*>(this)->Serialize(Stm);
+	return true;
+}
+/*
+void VirtualTrajectoryType::Read(CCINIClass* const pINI, const char* pSection) // Read separately
```
secsome, line 30: Would it be better to read it here, and the child types call VirtualTrajectoryType::Read before they read their data?

src/Ext/Bullet/Trajectories/VirtualTrajectories/TracingTrajectory.cpp
```diff
@@ -0,0 +1,333 @@
+#include "TracingTrajectory.h"
+
+#include <Ext/Bullet/Body.h>
+#include <Ext/Techno/Body.h>
+
+namespace detail
+{
+	template <>
+	inline bool read<TraceTargetMode>(TraceTargetMode& value, INI_EX& parser, const char* pSection, const char* pKey)
```
secsome, line 9: Should the INI parsing stuffs be put together or separate like this one

src/Ext/Bullet/Trajectories/VirtualTrajectories/TracingTrajectory.cpp
```diff
@@ -0,0 +1,333 @@
...
+		{
+			const double radius = BulletExt::Get2DDistance(offset);
+
+			// Individual or entirety
+			if (!pType->StableRotation || !pBulletExt->TrajectoryGroup)
+			{
+				const auto distanceCoords = pBullet->Location - destination;
+
+				// Rotate around the center only when the distance is less than 1.2 times the radius
+				if ((radius * 1.2) > BulletExt::Get2DDistance(distanceCoords))
```
secsome, line 275: Any idea about the constant 1.2 comes from?

src/Ext/Bullet/Trajectories/VirtualTrajectories/TracingTrajectory.cpp
```diff
@@ -0,0 +1,333 @@
...
+	const auto pBulletExt = BulletExt::ExtMap.Find(pBullet);
+
+	// Calculate survival time
+	if (const auto pWeapon = pBullet->WeaponType)
+		pBulletExt->LifeDurationTimer.Start((pWeapon->ROF > 10) ? pWeapon->ROF - 10 : 1);
+	else
+		pBulletExt->LifeDurationTimer.Start(120);
+}
+
+bool TracingTrajectory::ChangeVelocity()
```
secsome, line 170: Personally I do not think the `tracing` one should be a trajectory, it looks more like a spawner, which should become a independent system? putting it in trajectory seems not reasonable enough?

secsome (commented):
After reading the two specific VirtualTrajectory I'm doubting the `VirtualTrajectories`. The VirtualTrajectories doesn't seem to be trajectories but some specific systems which simply borrow the trajectory generic process routine? I can somewhat understand this line of thinking, but shouldn't such logic be separated into a class like spawners for handling? Not sure would thich one is more reasonable.

Also the other parts of this PR looks ok, but now I think the main problem is whether the `VirtualTrajectories` exists or not. 

*Don''t mind the other parts I mentioned below, just ignore them cause they are not the main issue now*

CrimRecya:
> The `engrave` one, however, seems have nothing to do with the trajectory (aka the trail of the bullet while flying)?

`Engrave` draws a "straight line" based on the target, which can rotate or move with the relative position, or just be a fixed straight line ([Demo U](https://github.com/Phobos-developers/Phobos/pull/1582/files#diff-f6fbe0428af7072c2f1acbdc2334cb8bff3ec98771a81a52652be067f0c43900)).

secsome:
I think its something kinda like burst but much more customizable, not thinking its a trajectory but a standalone bullet attribute for me

Metadorius:
> I think its something kinda like burst but much more customizable, not thinking its a trajectory but a standalone bullet attribute for me

I agree. Something like `BurstTargetFLH`, and add interpolation for FLH (we have the tools in #1304, interpolation mode there, also @Starkku pls fix/merge).

secsome:
I'd paste some of my opinions here:

```
My thought is that
Maybe the fire location, the pattern it flies, what it done when explode should be separated
into three different parts at least
Instead of putting them into trajectory all together
These parts should all belonging to the bullet
controlling how it work
Bullet is what Westwood call them in codes
And Weapon, Warhead is a component of Bullet
aka Projectiles
Now trajectory is a part of bullet system, what Im thinking about  is making the Firing and the Exploding into parts of the bullet system too 
https://github.com/Phobos-developers/YRpp/blob/phobos-dev/BulletClass.h#L128-L155
As you can see most of the logic are components of the Bullet(Projectile) in original WW code too
```

TaranDahl:
>> Because Engrave's damage is actually achieved by multiple small-cellspread warheads detonating. If not restricted, then certain units (such as buildings) may receive multiple damages from a single shot.

> Yes, and why is this a problem? This is exactly how I expect it to work and it's logical. Imagine a laser firing in 1 shot vs continuously for a period of time. It's a feature, not an issue.

1. The Colossus in Starcraft 2, which was mentioned by CrimRecya, is not working like that.
2. Even if they are all targets of the FootClass, they may still receive variable damage due to random positions. I hope to make the damage of the weapon relatively constant.

PR #1889, opened 2025-10-04
[Review Only] Trajectory retargeting

CrimRecya:
Split from #1582 . For review only. (The latest commit)


secsome (approved):
Most things are simply rename and splits, LGTM

PR #1900, opened 2025-10-08
Restored ScritType action 24 - Play speech with index...

FS-21:
RA2/YR doesn't have implemented the Tiberian Sun action "24" (Play Speech). It just marks the action ended and then jumps to the next action from the script list.
- Given a speech index, the game will play it. Just like action 24 from Tiberian Sun.

In `aimd.ini`:
```ini
x=24,n
```

Already tested in my mod and very simple action.


TaranDahl (approved):
LGTM.

FS-21:
I don't mind renaming this but I only use the naming commonly used in FinalSun, FinalAlert2, WAE and all knowledge pages like ModEnc.

PR #1901, opened 2025-10-08
New 190xx ScriptType actions that are masks of existing actions that used indexes as arguments

FS-21:
- These actions are "masks" of existing actions:
These actions translate IDs into indices while preserving the original action values. The reason for using these masks is that some ScriptType actions rely on fixed indices rather than ID labels (basically because all arguments are readed as integers). When these lists change, there's a high probability of breaking the original index of the pointed element.
So for example if is readed:
x=19046,GAPOWR,LeastThreat
It will be translated into:
x=46,0
The modder won't have to calculate the index every time because is calculated the first time the scenario is readed.

- Yes, arguments now support **reading** strings (if the action supports it that currently are these cases).

#### `19017` Change Script

- Similar to action 17, but uses a string ID from the `[ScriptTypes]` list instead of an index.

In `aimd.ini`:
```ini
x=19017,ID        ; text
```

#### `19018` Change TeamType

- Similar to action 18, but uses a string ID from the `[TeamTypes]` list instead of an index.

In `aimd.ini`:
```ini
x=19018,ID        ; text
```

#### `19020` Change House

- Similar to action 20, but uses a string ID from the `[Countries]` list instead of an index.

In `aimd.ini`:
```ini
x=19020,ID        ; text
```

#### `19024` Play Speech

- Similar to action 24 (when the respective PR gets merged), but uses a string ID from the `[DialogList]` list instead of an index.

In `aimd.ini`:
```ini
x=19024,ID        ; text
```

#### `19025` Play Sound

- Similar to action 25, but uses a string ID from the `[SoundList]` list instead of an index.

In `aimd.ini`:
```ini
x=19025,ID        ; text
```

#### `19027` Play Theme

- Similar to action 27, but uses a string ID from the `[Themes]` list instead of an index.

In `aimd.ini`:
```ini
x=19027,ID        ; text
```

#### `19051` Play Animation

- Similar to action 51, but uses a string ID from the `[Animations]` list instead of an index.

In `aimd.ini`:
```ini
x=19051,ID        ; text
```

#### `19046` Attack Enemy Structure

- Similar to action 46, but uses a string ID from the `[BuildingTypes]` list instead of an index.
- The last parameter is the BwP. Check below the possible strings.

In `aimd.ini`:
```ini
x=19046,ID,BwP    ; text
```

| *BwP values*  | *Description*                                             |
|--------------:|:----------------------------------------------------------|
| LeastThreat   | Index of the instance of the building with least threat   |
| HighestThreat | Index of the instance of the building with highest threat |
| Nearest       | Index of the instance of the building which is nearest    |
| Farthest      | Index of the instance of the building which is farthest   |

```{note}
More BwP information in https://modenc.renegadeprojects.com/ScriptTypes/ScriptActions
For this action to work in multiplayer - you need to use a version of [Script Actions information at Modenc](https://modenc.renegadeprojects.com/ScriptTypes/ScriptActions).
```

#### `19047` Move To Enemy Structure

- Similar to action 47, but uses a string ID from the `[BuildingTypes]` list instead of an index.
- The last parameter is the BwP. Check the table of the action `19046`.

In `aimd.ini`:
```ini
x=19047,ID,BwP    ; text
```

#### `19056` Chronoshift TaskForce To Structure

- Similar to action 56, but uses a string ID from the `[BuildingTypes]` list instead of an index.
- The last parameter is the BwP. Check the table of the action `19046`.

In `aimd.ini`:
```ini
x=19056,ID,BwP    ; text
```

#### `19058` Move To Friendly Structure

- Similar to action 58, but uses a string ID from the `[BuildingTypes]` list instead of an index.
- The last parameter is the BwP. Check the table of the action `19046`.

In `aimd.ini`:
```ini
x=19058,ID,BwP    ; text
```


src/Ext/Script/Hooks.cpp
```diff
@@ -103,3 +112,147 @@ DEFINE_HOOK(0x6F01B0, TMission_ChronoShiftToTarget_SuperWeapons, 0x6)
...
+		node->Argument = argument;
+		R->ECX(node);
+
+		return SkipCode;
+	}
+
+	long val = strtol(Phobos::readBuffer, &endptr, 10);
+
+	if (*endptr == '\0'
+		&& val >= std::numeric_limits<int>::min()
```
secsome: Meaningless check where long is exactly int32_t so it will always satisfy the boundary condition here.

https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:2,endLineNumber:10,positionColumn:2,positionLineNumber:10,selectionStartColumn:2,selectionStartLineNumber:10,startColumn:2,startLineNumber:10),source:'//+Type+your+code+here,+or+load+an+example.%0A%23include+%3Climits%3E%0A%0Abool+check(long+num)+%7B%0A++++if+(num+%3E%3D+std::numeric_limits%3Cint%3E::min()+%0A+++++%26%26+num+%3C%3D+std::numeric_limits%3Cint%3E::max())+%7B%0A++++++++return+true%3B%0A++++%7D%0A++++return+false%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:49.47631463318796,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:vcpp_v19_latest_x86,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-O2',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86+msvc+v19.latest+(Editor+%231)',t:'0')),k:37.87161724029135,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86+msvc+v19.latest',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86+msvc+v19.latest+(Compiler+%231)',t:'0')),k:12.652068126520678,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4

src/Ext/Script/Hooks.cpp
```diff
@@ -1,10 +1,19 @@
 #include "Body.h"
 
 #include <MapClass.h>
+#include <ThemeClass.h>
 
 #include <Ext/House/Body.h>
 #include <Helpers/Macro.h>
 
+enum class BuildingWithProperty : unsigned int
```
secsome, line 8: Declare the enum here with bitshifts or in hex form may be better?
```cpp
LeastThreat = 0 << 16,
HIghestThreat = 1 << 16,
Nearest = 2 << 16,
Farthest = 3 << 16
```
or 
```cpp
LeastThreat = 0x00000,
HIghestThreat = 0x10000,
Nearest = 0x20000,
Farthest = 0x30000
```

I prefer the bitshift one

src/Ext/Script/Hooks.cpp
```diff
@@ -103,3 +112,147 @@ DEFINE_HOOK(0x6F01B0, TMission_ChronoShiftToTarget_SuperWeapons, 0x6)
...
+			Information from https://modenc.renegadeprojects.com/ScriptTypes/ScriptActions
+			Computed Value                           Description
+			-------------------------------------		-------------------------------------------------------
+			0 (Hex 0x0) + Building Index          -> Index of the instance of the building with least threat
+			65536 (Hex 0x10000) + Building Index  -> Index of the instance of the building with highest threat
+			131072 (Hex 0x20000) + Building Index -> Index of the instance of the building which is nearest
+			196608 (Hex 0x30000) + Building Index -> Index of the instance of the building which is farthest
+			*/
+
+			//strcpy_s(id, strtok_s(textArgument, ",", &context));
```
secsome: Those useless comments can be removed imo

secsome (changes requested):
Can be merged after testing and doing those cleanups.

src/Ext/Script/Hooks.cpp
FS-21: Seems more easy/readable the hex one but since these values won't be used directly I don't mind what values to be used here and both ways are the same so why not the first one too.

src/Ext/Script/Hooks.cpp
FS-21: Including the explanation or only the old code?

FS-21:
I think all comments were cleaned as requested.

PR #1904, opened 2025-10-09
MultiWeapon Auto-Target Selection Fix

Fly-Star-him:
Fixed an issue where the game would only use `Weapon1` and `Weapon2` for auto-targeting even when `MultiWeapon=yes` was set.
修复`MultiWeapon`启用后仍然只使用`Weapon1`、`Weapon2`进行自动索敌的问题。

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -9,4 +9,4 @@ community_bridge: # Replace with a single Community Bridge project-name e.g., cl
 liberapay: # Replace with a single Liberapay username
 issuehunt: # Replace with a single IssueHunt username
 otechie: # Replace with a single Otechie username
-custom: [github.com/Phobos-developers/Phobos#developers]
+custom: [github.com/Phobos-developers/Phobos?tab=readme-ov-file#credits]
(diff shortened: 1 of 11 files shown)


src/Ext/TechnoType/Body.cpp
```diff
@@ -362,6 +362,69 @@ void TechnoTypeExt::ExtData::ParseVoiceWeaponAttacks(INI_EX& exINI, const char*
...
+			pEliteWeapon = pWeapon;
+
+		if (pWeapon)
+		{
+			if (const auto pBulletType = pWeapon->Projectile)
+			{
+				if (pBulletType->AA)
+					this->ThreatTypes.X |= ThreatType::Air;
+
+				if (pBulletType->AG && !BulletTypeExt::ExtMap.Find(pBulletType)->AAOnly)
```
TaranDahl: Should deal with https://github.com/Phobos-developers/Phobos/blob/ade8820390593e5fb3cec45c5462c190378aa5dd/src/Ext/Techno/Hooks.cpp#L1346

TaranDahl: You may call that hooked function instead. If you persist in rewriting it, you should also remove the AAOnly hook.

Fly-Star-him: Since it has already been processed, that's naturally better.

PR #1908, opened 2025-10-11
[Minor] Auto Target Fix - Prioritize checking if the target is dead

Fly-Star-him:
Units that have already died enter auto-target checking, which can cause some issues, so try to fix them.
已经死亡的单位会进入自动索敌检查，它们可能会造成一些问题，因此尝试修复。

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -394,6 +394,7 @@ This page lists all the individual contributions to the project by their author.
   - Health bar permanently displayed
   - Unlimbo Detonate warhead
   - Fast access structure
+  - The target of death is excluded from automatic target selection
 - **NetsuNegi**:
   - Forbidding parallel AI queues by type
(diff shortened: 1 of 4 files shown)


CrimRecya (changes requested):
I think this is an invalid fix. All places where this function is called have already used the `pTarget` pointer to call virtual functions. So if this issue cannot be solved from the root, then it cannot be placed here either. You should make these sanity check before calling `IsAlliedWith` which has called `GetOwningHouse`.
Moreover, I think that this fix will affect the performance of the enemy searching and will not occur frequently. A switch should be set to disabled this fix by applying patch.

CrimRecya (commented):
It looks fine now, but I still hope it can be turned off by `Patch::Apply_RAW` like `DisableRadDamageOnBuildings`, as in most cases these hooks are not needed.

CrimRecya (approved):

Issue #1912, opened 2025-10-14
FireUp on shp vehicles breaks the vanilla animation behavior


mevitar:
### Description

In regular YR, shp vehicles that have a fire animation will be forced restart the animation with each shot, even if the sequence is longer than its weapon's ROF. This means a high rate of fire weapon will work as expected even if its owned by an shp vehicle that is animating, and the animation can be made to look smoothly.

It used to work like that with Phobos too, until FireUp for shp vehicles got introduced, and it completely broke high ROF weapons on shp vehicles. Now the weapon is forced to pause until the vehicle finishes playing its entire sequence, which makes giving a rapid fire weapon not only pointless, but impossible.

The old behavior should have been kept as a default behavior. Instead, it got removed completely.

### Phobos Version

https://github.com/Phobos-developers/Phobos/actions/runs/15516858646/job/43684937391

### Conditions to reproduce

Give an shp vehicle with a long firing sequence a weapon with ROF=5 or something like that.
Before FireUp=, the weapon fires properly and the sequence is reset with every shot.
Since adding FireUp=, the weapon will wait with every shot until the sequence finishes playing.

### INI code

```
[ARapidFireWeapon]
ROF=5
```

### Steps to reproduce

1.
2.
3.
...

### Expected behaviour

Low ROF on the weapon should turn it into a rapid fire weapon.

### Actual behaviour

The shp unit will fire one and then wait for the entire sequence to finish.

### Checklist

- [x] The issue happens on the **latest official** version of Phobos and wasn't fixed yet.
- [x] I agree to elaborate the details if requested and provide thorough testing if the bugfix is implemented.
- [x] I added a very descriptive title to this issue.
- [x] I used the GitHub search and read the issue list to find a similar issue and didn't find it.
- [x] I have attached as much information as possible *(screenshots, gifs, videos, debug and exception logs, etc)*.

mevitar:
I rechecked with the newest build, and it seems i did a mistake. All of the above is true _if FireUp is defined on the unit_.

If FireUp _is not defined_, then the weapon _will_ be fired properly. But the entire sequence will also keep playing, unlike before.

This means a unit cannot be given an animation with a smooth transition between firing and idle states, because it will look terribly (it will be firing some shots while its animation indicates it's not firing).
And in order for it to have a proper transition and not look terribly, it needs to fire one shot per entire sequence, which prohibits it from using rapid fire weapons.

PR #1913, opened 2025-10-15
[Minor] SHP Vehicle Firing Animation Fix When FireUp is Undefined

Fly-Star-him:
Attempt to resolve the issue mentioned in [#Issues 1912](https://github.com/Phobos-developers/Phobos/issues/1912).


mevitar:
The looping sequence problem is fixed in this build, and it resets now with each shot if FireUp= is not defined.
However, the sequence is still being broken when the unit stops firing, and that doesn't change even if i add FireUp.ResetInRetarget=no (although still with no FireUp= defined)..

To demonstrate, this is how it looked like before FireUp support for shp vehicles:
fireupproblem1

And this is how it looks now:
fireupproblem2

I hope the difference is easily noticeable here. It looks properly when firing, but the sequence doesn't finish like it used to when it stops.

Coronia (approved):

PR #1916, opened 2025-10-18
Dehardcode the `ZAdjust` of warhead anim

TaranDahl:
### Dehardcode the `ZAdjust` of warhead anim

- In vanilla, the animations generated by `AnimList` have a hard-coded `ZAdjust=-15`. Now you can customize it in the following ways.
  - If these flags is set to 0, the `ZAdjust` defined by the anim will be used.

In `rulesmd.ini`:
```ini
[AudioVisual]
WarheadAnimZAdjust=-15           ; Integer

[SOMEWARHEAD]                    ; WarheadType
AnimZAdjust=                     ; Integer, default to [AudioVisual] -> WarheadAnimZAdjust
```


docs/Fixed-or-Improved-Logics.md
```diff
@@ -2164,6 +2164,20 @@ Parasite.CullingTarget=infantry   ; List of Affected Target Enumeration (none|ai
 Parasite.GrappleAnim=             ; animation
 ```
 
+### Dehardcode the `ZAdjust` of warhead anim
+
+- In vanilla, the animations generated by `AnimList` have a hard-coded `ZAdjust=-15`. Now you can customize it in the following ways.
+  - If these flags is set to 0, the `ZAdjust` defined by the anim will be used.
```
Coronia: If these flags are set to 0

Coronia (commented):
correct me if I'm wrong, but according to how the code is currently implemented, wouldn't its ZAdjust also becomes 0 when setting the customized value to 0?

docs/Fixed-or-Improved-Logics.md
TaranDahl: That is processed inside the CTOR of AnimClass. That arg is named ForceZAdjust. If you set it to 0 and the CTOR will use Type->ZAdjust.

Coronia (approved):

PR #1926, opened 2025-10-26
[Vanilla Fix]Fixed the bug that vehicle fall on infantries will make all cell content has been removed

NetsuNegi:
- Fixed the bug that vehicle fall on infantries will make all cell content has been removed.
- 修复了载具落到步兵上时会导致格子内全部物体被移除的bug。

### DIFF
--- a/src/Misc/Hooks.BugFixes.cpp
+++ b/src/Misc/Hooks.BugFixes.cpp
@@ -2799,7 +2799,7 @@ DEFINE_HOOK(0x54CC9C, JumpjetLocomotionClass_ProcessCrashing_DropFix, 0x5)
 		{
 			const auto pTerrain = static_cast<TerrainClass*>(*object);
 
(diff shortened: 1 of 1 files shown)


src/Misc/Hooks.BugFixes.cpp
```diff
@@ -2744,3 +2744,63 @@ DEFINE_HOOK(0x44242A, BuildingClass_ReceiveDamage_SetLATime, 0x8)
...
+			pInfantry->Limbo();
+			pInfantry->UnInit();
+			continue;
+		}
+
+		if (whatAmObject == TerrainClass::AbsID)
+		{
+			const auto pTerrain = static_cast<TerrainClass*>(*object);
+
+			if (pTerrain->Type->SpawnsTiberium || pTerrain->Type->Immune)
```
CrimRecya: I think checking `Immune` is enough.

PR #1927, opened 2025-10-26
Main

obsidian14:


Metadorius (changes requested):
1. Your license entry is duplicated many times.
3. You added Phobos license to the external libs' source code too.
4. Use SPDX license identifiers if you're gonna do this. https://spdx.dev/learn/handling-license-info/

PR #1934, opened 2025-10-27
Fixed an issue where some effects pointing to a unit were not properly cleared when the unit changed its owner

TaranDahl:
some effects:
CLEG
CARRIER
LastTarget set by Override_Mission (such as retaliation)


src/Misc/Hooks.BugFixes.cpp
```diff
@@ -2813,3 +2813,52 @@ DEFINE_HOOK(0x54CC9C, JumpjetLocomotionClass_ProcessCrashing_DropFix, 0x5)
 
 	return fallOnSomething ? SkipGameCode2 : SkipGameCode;
 }
+
+#pragma region ClearTargetOnOwnerChanged
+
+DEFINE_HOOK(0x701681, TechnoClass_SetOwningHouse_ClearManagerTarget, 0x6)
```
CrimRecya: I'm not sure, how about placing this hook here at 70D580?

TaranDahl: Moved to the start of that func.

Coronia (approved):

PR #1939, opened 2025-11-02
[Minor] Independent SHP Vehicle Turret Files

Fly-Star-him:
SHP turret vehicles support the use of `*tur.shp` files.
SHP炮塔车辆支持使用`*tur.shp`文件。


DeathFishAtEase:
I believe that even without any additional flags, it should be documented in detail as an *Improvement*. For instance, for Shape vehicles that use `*tur.shp`, what sequence should be used for the shp files that act as the body, what are the advantages of the new approach, how to transition from the old usage to the new usage, and other such details, similar to how it's done for [Stationary vehicles](https://phobos.readthedocs.io/en/latest/Fixed-or-Improved-Logics.html#stationary-vehicles).
> 我想它即便没有任何新增语句也应当作为一个功能改进在文档中详细说明，例如对与使用了 `*tur.shp` 的 Shape 载具其作为车体的 shp 文件应当使用什么样的序列、新的方式有哪些优势、旧的使用方式如何转为新的使用方式等内容，就像 [静止的载具](https://phobos.readthedocs.io/zh-cn/latest/Fixed-or-Improved-Logics.html#stationary-vehicles) 那样。

TaranDahl (approved):

Rinimabi:
look like work (DRONTUR.shp)
```
[DRON]
Primary=120mmx
Turret=yes
```

DeathFishAtEase:
> look like work (DRONTUR.shp)

The effect of this asset display is not very obvious. If you want to test it, it might be more convenient to use TS's `mmch.shp` to split it into Shape files for the *body* and *turret* parts, and you can also check the differences between the two file organization methods.

PR #1953, opened 2025-11-13
Fix scattered infantry units on game start

11EJDE11:
Fixes an issue where grouped starting infantry would scatter randomly instead of spawning together at game start.

The hooks `ScenarioClass_ScanPlaceUnit_CheckMovement` and `ScenarioClass_ScanPlaceUnit_CheckMovement2` were introduced to [fix a bug where naval ships could spawn on land](https://github.com/Phobos-developers/Phobos/pull/1552). It used:

```cpp
IsClearToMove(SpeedType, false, false, -1, MovementZone, -1, 1)
```
The first false parameter (ignoreInfantry = false) causes the function to check for infantry occupancy. During starting infantry placement, when multiple infantry are placed in a group, IsClearToMove() would return false for cells that already have infantry from the same group. Since infantry can share cells it causes scattering.

Before:
before

After:
after


NetsuNegi:
> @NetsuNegi since you were the original author of that PR, can you check whether this introduces any issues regarding your original improvement?

seems like fine

PR #1959, opened 2025-11-16
Allow deploy controlled MCV

NetsuNegi:
- In vanilla, you cannot deploy a controlled vehicle to `ConstructionYard=true` building, the reverse is also disabled. Now you can customize it.

In `rulesmd.ini`:
```ini
[General]
AllowDeployControlledMCV=false   ; boolean
```

### DIFF
--- a/src/Ext/Unit/Hooks.DeploysInto.cpp
+++ b/src/Ext/Unit/Hooks.DeploysInto.cpp
@@ -8,24 +8,12 @@
 
 #pragma region AllowDeployControlledMCV
 
-DEFINE_HOOK(0x700ED0, UnitClass_CanDeploySlashUnload_MindControlled, 0x6)
+DEFINE_HOOK_AGAIN(0x443770, TechnoClass_AllowDeployControlledMCV, 0x6)// BuildingClass::CellClickedAction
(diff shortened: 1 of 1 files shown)


DeathFishAtEase:
NeverUse
*Encountered a very strange problem in PR testing*
Suspect that there is a conflict with other functions that change the behavior of `ConstructionYard`.
(`0x449CC1`?)

src/Ext/Unit/Hooks.DeploysInto.cpp
```diff
@@ -6,6 +6,30 @@
 #include <Ext/WarheadType/Body.h>
 #include <Ext/Building/Body.h>
 
+#pragma region AllowDeployControlledMCV
+
+DEFINE_HOOK(0x700ED0, UnitClass_CanDeploySlashUnload_MindControlled, 0x6)
+{
+	return RulesExt::Global()->AllowDeployControlledMCV ? 0x700EDE : 0;
```
TaranDahl: Could use DEFINE_HOOK_AGAIN and the following code:
return RulesExt::Global()->AllowDeployControlledMCV ? R->Origin()+0xE : 0;

TaranDahl (approved):

PR #1969, opened 2025-11-22
Allow the attached animations to be retained and hidden when the unit disappears (f.ex. cloak, enters a vehicle)

TaranDahl:
- When a unit disappears (e.g., cloaks, enters a vehicle), the animations attached to it will be hidden instead of removed. Enable this feature through `[General] -> KeepAnimOnLimbo=true`.

### DIFF
--- a/src/Ext/Anim/Hooks.cpp
+++ b/src/Ext/Anim/Hooks.cpp
@@ -550,3 +550,25 @@ DEFINE_HOOK(0x4250E1, AnimClass_Middle_CraterDestroyTiberium, 0x6)
 	return AnimTypeExt::ExtMap.Find(pType)->Crater_DestroyTiberium.Get(RulesExt::Global()->AnimCraterDestroyTiberium) ? 0 : SkipDestroyTiberium;
 }
 
+#pragma region KeepAnimOnLimbo
+
+DEFINE_HOOK(0x422C70, AnimClass_DrawIfVisible_DontDrawIfOwnerInLimbo, 0x6)
+{
+	if (!RulesExt::Global()->KeepAnimOnLimbo)
+		return 0;
+
+	GET(AnimClass*, pThis, ECX);
+	R->EAX(pThis->LoopDelay || pThis->OwnerObject && (pThis->OwnerObject->InLimbo || pThis->OwnerObject->VisualCharacter(true, HouseClass::CurrentPlayer) == VisualType::Hidden));
+	return R->Origin() + 0x6;
+}
+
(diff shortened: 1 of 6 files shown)


Starkku:
There is already per-animation control for this (`DetachOnCloak`) for cloak only. Should extend it and maybe add a global default instead of adding something completely new.

src/Ext/Anim/Hooks.cpp
```diff
@@ -550,3 +550,25 @@ DEFINE_HOOK(0x4250E1, AnimClass_Middle_CraterDestroyTiberium, 0x6)
...
+}
+
+DEFINE_HOOK(0x425174, AnimClass_PointerExpired_KeepAnimOnLimbo, 0x6)
+{
+	if (!RulesExt::Global()->KeepAnimOnLimbo)
+		return 0;
+
+	GET_STACK(bool, bRemoved, STACK_OFFSET(0xC, 0x8));
+	return bRemoved ? 0 : 0x4251A3;
+}
```
Starkku, line 572: Also I am fairly certain that what this is doing is blanket preventing deletion of the animation if its parent object is being detached for any reason whatsoever, which seems like it might cause unwanted side effects somewhere.

TaranDahl: Yeah it is.
Actually, I can't understand why WW would remove all the animations attached to it just because the unit disappeared. I think it should actually be hidden rather than removed.

Starkku: My main concern is that it may skip cases where the removal is warranted e.g destruction as well. But I assume the `bRemoved` check *might* take care of that if it is being used correctly.

TaranDahl: What do you think are the reasons for removing the animation when the unit disappears? What are the use cases?

Starkku: If by 'disappear' you mean limbo then there's probably no good reason to delete it vs. hiding since most cases for limboing are temporary anyway. As long as it is deleted if the parent object disappears permanently it should be fine. Overlap with `DetachOnCloak` should be resolved still, though.

TaranDahl: > My main concern is that it may skip cases where the removal is warranted e.g destruction as well. But I assume the bRemoved check might take care of that if it is being used correctly.

I'm actually also worried whether it might cause some problems. But so far, I haven't encountered any during my tests.

TaranDahl:
TODO: deal with `DetachOnCloak`

Issue #1970, opened 2025-11-22
Random stably reproducible CMIN insta-desync with CnC-DDraw only for some users


Metadorius:
### Description

Commit 9c8bd9ecf9a81b4da294875cf5e1f842471fae49 causes desyncs *only for a part of users*. User Shero on CnCNet YR Discord had the desyncs when using CnC-DDraw instantly when CMIN got spawned from a ref, or in a few seconds. This pattern can be reproduced in a pretty stable fashion, but no staff members were able to reproduce it.

Commit was found by bisecting builds then blind pick of the change to be reverted on top of 0.4.

### Phobos Version

0.4

### Conditions to reproduce

CnCNet YR 9.0.x
(spawner version .12, Ares 3.0p1, public Syringe with patches for multithreading and whitelist, Phobos v0.4)

### INI code

```ini

```

### Steps to reproduce

self-explanatory

### Expected behaviour

no desync

### Actual behaviour

desync

### Checklist

- [x] The issue happens on the **latest official** version of Phobos and wasn't fixed yet.
- [x] I agree to elaborate the details if requested and provide thorough testing if the bugfix is implemented.
- [x] I added a very descriptive title to this issue.
- [x] I used the GitHub search and read the issue list to find a similar issue and didn't find it.
- [x] I have attached as much information as possible *(screenshots, gifs, videos, debug and exception logs, etc)*.

PR #1973, opened 2025-11-23
Replace game's CRT function call in floating point number INI parser

Starkku:
Attempts to address one of the issues presented in #1970 by working around the changes introduced in 9c8bd9ecf9a81b4da294875cf5e1f842471fae49 by replacing calling game's CRT functions. `sscanf(_s)` produces results with different precision from game's `sscanf`. Using `strtod` and casting the result double into float appears to be more accurate but whether or not these results are consistent across different systems I do not know.

This merely addresses the floating point number parsing in Phobos (which also replaces game's floating point number INI parser, see hook `INIClass_ReadDouble_Overwrite` in `Misc\Hooks.INIInheritance.cpp`) and does not in any way involve any potential root causes for why calling game's CRT functions is problematic in the first place.

### DIFF
--- a/src/Utilities/Parser.h
+++ b/src/Utilities/Parser.h
@@ -263,8 +263,10 @@ inline bool Parser<double>::TryParse(const char* pValue, OutType* outValue)
 	errno = 0;
 	char* end;
 
-	// Nov 23, 2025 - Starkku: strtod() + cast result to float produces results
-	// more similar to game's CRT functions than using sscanf_s.
+	// Nov 23, 2025 - Starkku, Kerbiter: strtod() + cast result to float produces
+	// results more similar to game's CRT functions than using sscanf_s. For some
+	// reason CnC-DDraw in DirectX mode causes game's CRT function usage here to change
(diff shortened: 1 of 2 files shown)


src/Utilities/Parser.h
```diff
@@ -261,35 +260,45 @@ inline bool Parser<int>::TryParse(const char* pValue, OutType* outValue)
 template<>
 inline bool Parser<double>::TryParse(const char* pValue, OutType* outValue)
 {
+	errno = 0;
+	char* end;
 
-	// Game doesn't use double precision when parsing, using double here would create inconsistency.
-	float buffer = 0.0;
+	// Nov 23, 2025 - Starkku: strtod() + cast result to float produces results
+	// more similar to game's CRT functions than using sscanf_s.
```
Metadorius:
```suggestion
	// Nov 23, 2025 - Starkku, Kerbiter: strtod() + cast result to float produces
	// results more similar to game's CRT functions than using sscanf_s. For some
	// reason CnC-DDraw in DirectX mode causes game's CRT function usage here to change
	// for some players, causing stably reproducible desyncs when CMIN appears, f.ex.
```

PR #1976, opened 2025-11-25
Add Expected Object

obsidian14:
# Expected object

I add a Expected.h under the src/Utilities directory. The header contains a implementation of [`std::expected`](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0323r3.pdf). It is written in C++20.

## Note

Calling the `Error()` method on an expected value, or using the `*` or `->`  operators on an unexpected value, is undefined behaviour. In this implementation it causes an unreachable error. The implementation con be overridden by defining the macro `TERMITE_ASSERT(boolean_condition)` before including Expected.h.

Phobos disabled exception handling by default. Considering that the `Value()` method throws an exception when be called on an unexpected value, the implementation deleted the method. Users should use the `*` or `->` operators to access the  expected value.

### DIFF
--- a/Phobos.vcxproj
+++ b/Phobos.vcxproj
@@ -327,6 +327,7 @@
     <ClInclude Include="src\Utilities\TemplateDef.h" />
     <ClInclude Include="src\Utilities\AresHelper.h" />
     <ClInclude Include="src\Utilities\AresFunctions.h" />
+    <ClInclude Include="src\Utilities\Expected.h"/>
     <ClInclude Include="lib\nameof\nameof.h" />
     <ClInclude Include="YRpp\GameTextManager.h" />
   </ItemGroup>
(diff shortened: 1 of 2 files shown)


obsidian14:
fixes: The implementation can be overridden by defining the macro TERMITE_ASSERT(boolean_condition) before including Expected.h.

TaranDahl:
Topic to discuss: Does Phobos really need this?

secsome:
I can't see a point to implement this manually now. Neither we do need expect struct for current code base nor this is expected to be implemented ourselves cause this template has alrewdy been implemented in msvc STL already in C++23 which we will migrate to when it's necessary.

TaranDahl:
> Topic to discuss: Does Phobos really need this?

My word: no.

PR #1978, opened 2025-11-28
Typeselect consider ifv mode

NetsuNegi:
In vanilla game, when using type selection command on IFVs, all of them will be selected regardless of their current modes, which is allowed to customize now.
- `WeaponGroupAsN` determines which group the IFV is in when enabling `WeaponN`, where N stands for 1-based weapon mode index. IFVs in the same group will be selected together during type a selection, while not included those in different groups.
- `TypeSelectUseIFVMode` determines whether all IFV modes will be considered as its own group by default during a type selection.
  - If it's set to true, `WeaponGroupAsN` will be default to N for each `WeaponN`, which makes each of them become a standalone type during a type selection.
  - If it's set to false, `WeaponGroupAsN` will be default to 0 for all weapons, which makes type selection on IFVs work the same as before.

In `rulesmd.ini`:
```ini
[General]
TypeSelectUseIFVMode=false   ; boolean

[SOMEVEHICLE]                ; VehicleType
WeaponGroupAsN=              ; string, default to N if [General] -> TypeSelectUseIFVMode=true, and 0 if false
```

### DIFF
--- a/src/Misc/Selection.cpp
+++ b/src/Misc/Selection.cpp
@@ -145,7 +145,7 @@ public:
 				char* gunnerID = pTypeExt->WeaponGroupAs[pTechno->CurrentWeaponNumber];
 
 				if (!GeneralUtils::IsValidString(gunnerID))
-					sprintf_s(gunnerID, 0x20, "%d", pTechno->CurrentWeaponNumber + 1);
+					sprintf_s(gunnerID, 0x20, "%d", RulesExt::Global()->TypeSelectUseIFVMode ? pTechno->CurrentWeaponNumber + 1 : 0);
 
 				if (std::ranges::none_of(ExtSelection::IFVGroups, [gunnerID](const char* pID) { return !_stricmp(pID, gunnerID); }))
 					break;
@@ -193,12 +193,9 @@ DEFINE_FUNCTION_JUMP(CALL, 0x4ABCEB, ExtSelection::Tactical_MakeFilteredSelectio
 // Replace vanilla function. For in case another module tries to call the vanilla function at offset
 DEFINE_FUNCTION_JUMP(LJMP, 0x6D9FF0, ExtSelection::Tactical_MakeFilteredSelection)
 
-DEFINE_HOOK(0x73298D, TypeSelectCommand_Execute_UseIFVMode, 0x5)
+DEFINE_HOOK(0x73298D, TypeSelectExecute_UseIFVMode, 0x5)
 {
-	if (!RulesExt::Global()->TypeSelectUseIFVMode)
-		return 0;
-
-	ExtSelection::IFVGroups.clear();
+	const bool useIFVMode = RulesExt::Global()->TypeSelectUseIFVMode;
 
 	for (const auto pObject : ObjectClass::CurrentObjects)
 	{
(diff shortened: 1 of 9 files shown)


src/Misc/Selection.cpp
```diff
@@ -146,8 +168,28 @@ class ExtSelection
 	}
 };
 
+DEFINE_FUNCTION_JUMP(LJMP, 0x732C30, ExtSelection::TypeSelectExecute_Filter)
```
TaranDahl: Should deal with the existing flag `BuildingTypeSelectable`.

TaranDahl (changes requested):

src/Misc/Selection.cpp
NetsuNegi: I dont think so, it is hotkey execute's sub function

TaranDahl: IIRC it has a hook in that func. So if you are going to replace that func then that hook will be lost.

NetsuNegi: this contains that Ares hook‘s code

TaranDahl: I mean [BuildingTypeSelectable](https://phobos.readthedocs.io/en/latest/User-Interface.html#type-select-for-buildings).

src/Misc/Selection.cpp
```diff
@@ -146,8 +185,73 @@ class ExtSelection
...
+DEFINE_HOOK(0x732C06, TypeSelectExecute_Clear, 0x6)
+{
+	ExtSelection::IFVGroups.clear();
+	return 0;
+}
+
+#pragma region BuildingTypeSelectable
+
+DEFINE_HOOK_AGAIN(0x732B28, TypeSelectExecute_SetContext, 0x6)
+DEFINE_HOOK(0x732A85, TypeSelectExecute_SetContext, 0x7)
```
TaranDahl, line 234: You can set this variable at the beginning of the IDMatches function you rewrite and delete these two hooks.

TaranDahl (approved):
Other things LGTM.

DmitryVolkov666:
The game freezes in multiplayer, and there's nothing that can be done about it. The dump file indicates that this is the cause:

WARNING: Stack unwind information not available. Following frames may be wrong.
0019d144 721d8956     00a8e9c0 00647689 00000000 Phobos!TypeSelectExecute_UseIFVMode+0xc07
00000000 00000000     00000000 00000000 00000000 Ares!WaveClass_Update_Beam+0xa16

(im not use TypeSelectUseIFVMode=yes)

dump file: https://drive.google.com/drive/folders/1OFy1FQdRusUKPilEHVAXEWMLWaDGMMnC?usp=sharing

Coronia:
@DmitryVolkov666 did the crash happen right after using a type select?

also looking at the code it seems to be a issue that's not only restricted to multiplayer. Could you see if it also happens in singleplayer scenario?

and check your IFV settings to see if the weapon count is mismatched

DmitryVolkov666:
@Coronia, Hi
Right after selecting an IFV weapon? Possibly. The crash behaves in an extremely strange way. There have even been cases where the game crashed when there were no units using IFV logic at all, although I'm not 100% sure about that.

Even if there is a mismatch in the number of weapons, it shouldn't necessarily explain the crash, since this configuration didn't cause crashes before.

I haven't been able to reproduce the crash in single-player so far, and I'm not even sure whether it's possible to trigger it there.

PR #1979, opened 2025-11-28
Allow techno type considered as other type when AI recruit techno for teams

NetsuNegi:
### Allow techno type considered as other type when recruiting techno for teams

- It is now possible to make techno type considered as other type when recruiting techno for teams, both for AI team recruitment and `Create Team` action.
  - Only affect techno that's presented on the map. Cannot make AI produce this type of techno if it doesn't have any.

In `rulesmd.ini`:
```ini
[SOMETECHNO]                      ; TechnoType
TeamMember.ConsideredAs=        ; list of technotypes
```

### DIFF
--- a/src/Ext/Team/Hooks.cpp
+++ b/src/Ext/Team/Hooks.cpp
@@ -1,5 +1,6 @@
 #include "Body.h"
 #include <Utilities/AresHelper.h>
+#include <Ext/TechnoType/Body.h>
 
 // Bugfix: TAction 7,80,107.
 DEFINE_HOOK(0x65DF67, TeamTypeClass_CreateMembers_LoadOntoTransport, 0x6)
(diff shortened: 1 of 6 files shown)


src/Ext/Team/Hooks.cpp
```diff
@@ -53,3 +54,95 @@ DEFINE_HOOK(0x65DF67, TeamTypeClass_CreateMembers_LoadOntoTransport, 0x6)
...
+
+		++idx;
+	}
+	while (pTaskForce->CountEntries > idx);
+
+	R->Stack(STACK_OFFSET(0x14, 0x8), idx);
+	return SkipGameCode;
+}
+
+DEFINE_HOOK(0x6EAD73, TeamClass_Sub_6EAA90_Consideration, 0x7)
```
TaranDahl: `TryRecruitMemberByIdx` may be a name.

TaranDahl (approved):

ayylmaoRotE:
Can confirm it works from my tests. Different scenarios tested:
2. TechnoType converted to Technotype (no Team)
3. TechnoType converted to TechnoType (with Team)
4. AircraftType to other classes e.g. TechnoType
5. TechnoType GiftBox'd into another TechnoType class (e.g. AircraftType - not Phobos native, but works, as expected)
6. All of the above work with Technos converted during, before, or after being recruited

PR #1981, opened 2025-12-01
Document organization for the release v0.4.0.1 branch

DeathFishAtEase:
To-do:
1. Properly move Crate improvements into the correct categories
  - [x] Split this paragraph into two paragraphs: global flags and VehicleType flag
  - [x] Add a global flags category and organize the content
    - we currently have many features that are only global flags scattered in various places
  - [x] Convert the global switches that are off by default in `## Bugfixes and miscellaneous` into a feature paragraph.

### DIFF
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -383,6 +383,7 @@ In vanilla, IvanBomb images display and the bombs detonate at the top-leftmost c
 
 In `rulesmd.ini`:
 ```ini
+[CombatDamage]
 IvanBombAttachToCenter=false  ; boolean
 ```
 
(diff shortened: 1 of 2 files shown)


DeathFishAtEase:
It was originally planned to list global features that are exclusively related to specific Types, such as [Voxel light source customization](https://phobos.readthedocs.io/en/latest/Fixed-or-Improved-Logics.html#voxel-light-source-customization) and [Customizing effect of level lighting on air units](https://phobos.readthedocs.io/en/latest/Fixed-or-Improved-Logics.html#customizing-effect-of-level-lighting-on-air-units), under the corresponding Types categories. But later, features like [Veinholes & Weeds](https://phobos.readthedocs.io/en/latest/Fixed-or-Improved-Logics.html#veinholes-weeds) were found to be too difficult to separate, so they were just summarized according to the locations where INI flags are written. After all, the end users are modders.

PR #1982, opened 2025-12-01
Don't close Syringe immediately

11EJDE11:
Keep Syringe open until the game exits. Helps stop the client calling GameProcessExited when the "attach debugger" message appears.

Ported from Vinifera - [originally done ](https://github.com/Vinifera-Developers/Vinifera/pull/1548/commits/bbc37f5a6e6665189e37166d94197657da25c504)by ZivDero & secsome.


ZivDero:
> @ZivDero can you review this one?

It looks like a direct copy of what I did. There is a side-effect that if you close the game via task manager/taskkill it won't kill syringe, so you need to close that manually, though otherwise it works alright.

ZivDero (approved):

ZivDero (changes requested):
Actually, I guess per Phobos policty, need docs? 🤣

Coronia:
> Actually, I guess per Phobos policty, need docs? 🤣

added

PR #1983, opened 2025-12-03
RA1-Style Multi-Turret and Multi-Barrel

TaranDahl:
### RA1-Style Multi-Turret and Multi-Barrel

image
*Shogun Battleship in Project Rush: Conquer*

- Vehicles can now draw and fire the same turret at multiple positions (RA1-style multi-turret) and draw multiple barrels for each turret, only supports regular VXL vehicles.
  - `ExtraTurretCount` controls how many extra turret positions a unit has. `ExtraTurretOffsetX` controls the offset of the X-th turret position relative to the center of the unit. X is 0-based.
  - `ExtraBarrelCount` controls how many extra barrels each turret draws. `BarrelOffset` controls the offset distance of the main barrel relative to the center of the turret in the left-right direction (positive values shift left). `ExtraBarrelOffsetX` controls the offset distance of the X-th extra barrel of each turret relative to the center of the turret in the left-right direction (positive values shift left). X is 0-based.
  - `BarrelOverTurret` controls whether the barrel layer is always above the turret layer. Otherwise, the layer relationship will be judged according to the direction of the turret.
  - When a unit fires, it will first use the original turret to fire `BurstPerTurret` times, then switch to the 0th additional turret position to fire `BurstPerTurret` times, then switch to the 1st, and so on.
- The firing FLH is also calculated relative to the firing turret's position. If the unit's turret or barrel has recoil effects set, the turret/barrel index will be automatically calculated based on `Burst`.
- The firing position is essentially determined by the current `Burst`. If you want the rear turrets to fire, you need weapons with high `Burst`. For example, a battleship with 2 additional turret positions (3 turrets total) and `BurstPerTurret=3` would need at least `Burst=9` to allow all turrets to fire completely once. The same applies to barrels on each turret.

In `artmd.ini`:
```ini
[SOMEVEHICLE]               ; VehicleType
BarrelOverTurret=           ; boolean
BarrelOffset=0              ; integer
ExtraBarrelCount=0          ; integer
ExtraBarrelOffsetX=0        ; integer
ExtraTurretCount=0          ; integer
ExtraTurretOffsetX=0,0,0    ; FLH coordinates
BurstPerTurret=0            ; integer
```

```{note}
This is essentially designed to restore multi-turrets in RA1, not to help you make a Zantos. Therefore, features like different turrets using different weapons/targeting separately/calculating cooldown separately/having separate health are not within the scope of this function. Please use techno attachment for such requirements.
```


TaranDahl:
Thanks for testing from @DreamingDays.

CrimRecya:
Since I rewrote the drawing part, having me review this PR would undermine the purpose of a code review. It would be better to have someone else provide the assessment.

docs/New-or-Enhanced-Logics.md
```diff
@@ -2424,6 +2167,35 @@ JumpjetTilt.SidewaysRotationFactor=1.0  ; floating point value
 JumpjetTilt.SidewaysSpeedFactor=1.0     ; floating point value
 ```
 
+#### Multiple Barrels and Turrets
+
+![image](_static/images/multiturret-01.png)
```
DeathFishAtEase: It seems that no GIF file has been added.

docs/New-or-Enhanced-Logics.md
```diff
@@ -2424,6 +2167,35 @@ JumpjetTilt.SidewaysRotationFactor=1.0  ; floating point value
 JumpjetTilt.SidewaysSpeedFactor=1.0     ; floating point value
 ```
 
+#### Multiple Barrels and Turrets
```
DeathFishAtEase:
```suggestion
### Multiple Barrels and Turrets
```

docs/Whats-New.md
```diff
@@ -504,102 +459,16 @@ HideShakeEffects=false           ; boolean
...
-- [Allow infantry to perform type conversion when deploying and undeploying](New-or-Enhanced-Logics.md#allow-infantry-to-perform-type-conversion-when-deploying-and-undeploying) (by Noble_Fish)
-- Add `ClampToScreen` tag for `BannerType` (defaults to `true`) to control whether banner position is clamped to the visible area (by Chang_zhi)
-- [Customizable Berzerk mission](Fixed-or-Improved-Logics.md#enhanced-berzerk-behavior) (by TaranDahl)
-- [Tank Bunker foundation and state update delay improvements](Fixed-or-Improved-Logics.md#tank-bunker-improvements) (by Starkku)
-- [Custom cruise missiles](New-or-Enhanced-Logics.md#custom-cruise-missiles) (by Noble_Fish)
-
-#### Vanilla fixes:
+- CellSpread in cylinder shape (by TaranDahl)
+- CellSpread damage check if victim is in air or on floor (by TaranDahl)
+- Multiple Barrels and Turrets (by TaranDahl & CrimRecya)
```
DeathFishAtEase:
```suggestion
- [Multiple Barrels and Turrets](New-or-Enhanced-Logics.html#multiple-barrels-and-turrets) (by TaranDahl & CrimRecya)
```

DeathFishAtEase (commented):
Does it additionally need a control statement similar to `TurretXShadow`?

~~What is the use case for Barrel being independent of the Turret count setting?~~

~~Does only controlling the number of turrets and allowing users to disable the drawing of the X-th turret so that only the barrel is drawn help make the syntax more standardized (where X is fixed to correspond to one tur and one barl, and Y corresponds to one hidden tur and one barl with the same center coordinates as X; rather than X corresponding to turretX and barrelY, and YZ corresponding to turretW and barrelZ)?~~

TaranDahl:
> Does it additionally need a control statement similar to TurretXShadow?

I don't think so. What is the use case?

> What is the use case for Barrel being independent of the Turret count setting?
> 
> Does only controlling the number of turrets and allowing users to disable the drawing of the X-th turret so that only the barrel is drawn help make the syntax more standardized (where X is fixed to correspond to one tur and one barl, and Y corresponds to one hidden tur and one barl with the same center coordinates as X; rather than X corresponding to turretX and barrelY, and Z corresponding to turretW and barrelZ)?

@CrimRecya did the barrel part, so it would be better if he could answer this himself.
IMO no extensions are necessary, as this feature is intended to recreate the multi-turret design of battleships from RA1, and its implementation is also modeled after that very design. Any other requirements should be directed to TA.

DeathFishAtEase:
~~I surveyed some users' opinions: the most frequent application scenario currently is that someone expressed the wish that among multiple turrets on a battleship, only one large gun should draw a shadow, while the other small turrets should not draw shadows.~~

I think you are referring to the Cruiser with 2 turrets in RA? Anyway, I misunderstood earlier; that is to say, now each turret will use `ExtraBarrelCount` barrels, and it is impossible to have 3 barrels for the main gun and 2 barrels for the secondary gun? And the barrel offset applies to the 3 barrels of all turrets?

Since it is recreating the behavior of RA, there is nothing unreasonable about this; and it is no longer reasonable that only some of the identical turrets cannot draw shadows, so ignore it.

> [!Tip]
> Additionally, for Voxel art authors, according to the VXL coordinate system, X is left-right, Y is up-down, and Z is front-back.

DeathFishAtEase:
Now this name makes me mistakenly think that it allows art authors to freely design the respective styles and modes of multiple turrets and barrels of a vehicle. I suggest changing the feature name to 'Allow drawing a vehicle's turrets and barrels at multiple positions', 'RA1-style multi-turret', or other similar expressions that can clearly avoid this misunderstanding.

src/Ext/Techno/Body.Internal.cpp
```diff
@@ -168,6 +187,57 @@ void TechnoExt::ExtData::InitializeAttachEffects()
...
+{
+	const auto pTypeExt = this->TypeExtData;
+	const auto pType = pTypeExt->OwnerObject();
+
+	if (!pType->TurretRecoil)
+		return;
+
+	if (pTypeExt->ExtraTurretCount)
+	{
+		if (static_cast<int>(this->ExtraTurretRecoil.size()) < pTypeExt->ExtraTurretCount)
```
DeathFishAtEase: Would such judgment cause that after a vehicle with 9 turrets transforms into a vehicle with 2 turrets, the elements storing the remaining 7 turrets' information are retained and occupy memory as redundant data and increase save file storage?

src/Ext/Techno/Body.Internal.cpp
```diff
@@ -168,6 +187,57 @@ void TechnoExt::ExtData::InitializeAttachEffects()
...
+			data.State = RecoilData::RecoilState::Inactive;
+			data.TravelFramesLeft = 0;
+		}
+	}
+
+	if (pTypeExt->ExtraTurretCount || pTypeExt->ExtraBarrelCount)
+	{
+		const auto dataCount = (pTypeExt->ExtraBarrelCount + 1) * (pTypeExt->ExtraTurretCount + 1) - 1;
+
+		if (static_cast<int>(this->ExtraBarrelRecoil.size()) < dataCount)
```
DeathFishAtEase: same

src/Ext/Techno/Body.Internal.cpp
TaranDahl: I reckon this is consistent with the current handling of vanilla recoil. Should we change all of them here?

TaranDahl: I don't think this matters much. I can modify it if you decide to, anyway.

DeathFishAtEase (approved):
LGTM, expecting different opinions to be given in time before merging if any.

src/Ext/Techno/Body.Internal.cpp
TaranDahl: I asked the agent, and it says that the resize won't take much performance. Since every recoil data needs to be updated per frame, I changed to resize on every Initiating.

PR #1984, opened 2025-12-03
Improve our existing tag name

DeathFishAtEase:
> [!Important]
> Since this involves historical usage and naming habits, we are opening this PR to collect opinions and discuss possible improvements. The goal is to establish clearer standards for similar cases in the future.

**Take the problem that this PR initially hoped to fix as an example**:
- In the documentation, `Crit.AffectsAbovePercent` is used, although it is more standard compared to `Crit.AffectAbovePercent` (~~&nbsp;s&nbsp;~~), but what is actually checked is `this->Crit_AffectAbovePercent.Read(exINI, pSection, "Crit.AffectAbovePercent");`
However, there is also a Debug Log that uses Affect ***s***:
  
  ```cpp
	  if (this->Crit_AffectsAbovePercent > this->Crit_AffectsBelowPercent)
		  Debug::Log("[Developer warning][%s] Crit.AffectsAbovePercent is bigger than Crit.AffectsBelowPercent, crit will never activate!\n", pSection);
  ```

- And, similarly, the warhead logic [Customizable Warhead trigger conditions](https://phobos.readthedocs.io/en/latest/Fixed-or-Improved-Logics.html#customizable-warhead-trigger-conditions) uses the following flags:
  
  ```ini
  ```
  
  ---

- Unifying the use of Affect ***s*** has the following advantages:
  1. It is better to maintain the same naming rules as similar logic, and this helps avoid spelling errors in INI code;
  2. Using `Affects` is more *standard* in expression compared to `Affect`;

- But if we do this, we need to consider the issue of user INI flag Migration.
  1. Its paired flag `Crit.AffectBelowPercent` was added as early as DevBuild#27, meaning that since its addition, it has gone through two *major* versions, 0.3 and 0.4, while `Crit.AffectsAbovePercent` or the one it should actually use, `Crit.AffectBelowPercent`, are both new content added in DevBuild#48.
  2. There is a previous handling scheme: when `BelowPercent` was [first introduced](https://github.com/Phobos-developers/Phobos/commit/80622e1d9646dd1002a91e46ebe1b490a6196cd6#diff-ed2191a6b64a75680ba92822821030ac5afce9042d09ab6d940a838328568f93R618), the documentation also used Affect ***s***, but this was [corrected as a spelling error](https://github.com/Phobos-developers/Phobos/commit/5aaf5c36f402178ea207fa37c1dfc1f57106c891) after 9 months.

---

*The above example issue has been temporarily fixed based on the source code operation method as the standard.*

But indeed, we have always lacked specifications for handling such issues—whether naming or minor content INI tag changes such as typos. We need to clarify the handling methods for these types of problems.

> [!NOTE]
> Currently, users have expressed two viewpoints: some believe that since it has already been added, it should not be changed; others believe that as a non-standard and inconsistent flag name, it should be unified, especially since we have [MigrationUtility](https://github.com/Phobos-developers/PhobosSupplementaries/tree/develop/MigrationUtility), which makes flag migration not troublesome.

In summary, for issues similar to it, please provide your opinions, such as

- Listing other existing similar issues.
- Specifying how migration documents should be written.
- Defining the standards we need to follow.

After sufficient discussion, we can pre-process all similar issues and descriptions of changes through this PR.

---
Currently existing tag replacement plan:

```diff
# 1. Standardization and Misspelling - Affect → Affect**s**
- Crit.AffectBelowPercent
+ Crit.AffectsBelowPercent

# 2. Standardization and Misspelling - Affect → Affect**s**
- Crit.AffectAbovePercent
+ Crit.AffectsAbovePercent

# 3.
# Standardization and Misspelling - Affect → Affect**s**
# Keep consistent with Ares num flags - AffectsHouses → AffectsHouse
- DetonateOnAllMapObjects.AffectHouses
+ DetonateOnAllMapObjects.AffectsHouse

# 4. Keep consistent with Ares num flags - AffectTargets → AffectsTarget
- DetonateOnAllMapObjects.AffectTargets
+ DetonateOnAllMapObjects.AffectsTarget

# 5. Keep consistent with Ares num flags - AffectTargets → AffectsTarget
- AffectTargets
+ AffectsTarget

# 6. AutoFire's misleading issue
- AutoFire
- AutoFire.TargetSelf
+ AutoTargetOwnPosition
+ AutoTargetOwnPosition.Self

# 7.
# Standardization and Misspelling - Affected → Affect**s**
# Keep consistent with Ares num flags - AffectsHouses → AffectsHouse
- Convert(N).AffectedHouses
+ Convert(N).AffectsHouse

# 8. Standardization and Misspelling - Affected → AffectsHouse
- LimboKill.Affected
+ LimboKill.AffectsHouse

# 9. Crit.Affects misleading issue
- Crit.Affects
+ Crit.AffectsTarget

# 10. Keep consistent with Ares num flags - AffectsHouses → AffectsHouse
- Crit.AffectsHouses
+ Crit.AffectsHouse

# 11. Keep consistent with Ares num flags - AffectsHouses → AffectsHouse
- RevengeWeapon.AffectsHouses
+ RevengeWeapon.AffectsHouse

# 12. Keep consistent with Ares num flags - AffectsHouses → AffectsHouse
- ReflectDamage.AffectsHouses
+ ReflectDamage.AffectsHouse

# 13. KillWeapon.Affects misleading issue
- KillWeapon.Affects
+ KillWeapon.AffectsTarget

# 14. Keep consistent with Ares num flags - AffectsHouses → AffectsHouse
- KillWeapon.AffectsHouses
+ KillWeapon.AffectsHouse

# 15. KillWeapon.OnFirer.Affects misleading issue
- KillWeapon.OnFirer.Affects
+ KillWeapon.OnFirer.AffectsTarget

# 16. Keep consistent with Ares num flags - AffectsHouses → AffectsHouse
- KillWeapon.OnFirer.AffectsHouses
+ KillWeapon.OnFirer.AffectsHouse
```


Coronia:
there's also a similar case elsewhere: while other tags are all `AffectsHouses`, it's `AffectHouses` for the case of `DetonateOnAllMapObjects`

mevitar:
Since you're changing it, AffectsHouses= would still be inconsistent with Ares, as the tags there are singular, not plural (like SW.AffectsHouse= or SW.RequiresHouse=, or even SW.AffectsTarget= as opposed to DetonateOnAllMapObjects.AffectTargets= ).
Out of all Ares tags i think only AffectsEnemies= is plural, but that is also a boolean, not an enum (and it stays consistend with YR's AffectsAllies= ).

And don't know if anyone noticed it, but Phobos' AttachEffects also has AffectTargets=, so that too needs to be changed.

Starkku:
`AutoFire` should realistically be something like `AutoTargetSelf` and `AutoFire.TargetSelf` could be `AutoTargetSelf.Snap` or whatever. What it does functionality-wise is skip passive target selection checks altogether and always pick its own cell as target, or if the latter key is set to true, the firer itself. It does not strictly speaking concern firing weapons itself at all.

DeathFishAtEase:
> `AutoFire` should realistically be something like `AutoTargetSelf` and `AutoFire.TargetSelf` could be `AutoTargetSelf.Snap` or whatever.

Uncertain whether `AutoTargetSelf.Snap` will be misunderstood as immediately detonating on oneself while ignoring the ballistic flight process. How about the flag name example in [issuecomment-3617388253](https://github.com/Phobos-developers/Phobos/pull/1984#issuecomment-3617388253)?

> What it does functionality-wise is skip passive target selection checks altogether and always pick its own cell as target

You mentioned here "skip passive target selection checks altogether", but currently `AutoFire` is still blocked by `CanPassiveAquire=false`. Is this a design intention or an oversight?

Starkku:
> You mentioned here "skip passive target selection checks altogether", but currently `AutoFire` is still blocked by `CanPassiveAquire=false`. Is this a design intention or an oversight?

Right, I was not very clear. It skips all target-related checks but not whether or not the firer can passively acquire targets. That is still technically intentional per the way it currently works altho potentially misleading.

NetsuNegi (approved):

Coronia (approved):

Metadorius:
@mevitar why exactly should we use singular for enums, apart from consistency? I think it is misleading to name it like this when the value accepts multiple entries.

mevitar:
> @mevitar why exactly should we use singular for enums, apart from consistency? I think it is misleading to name it like this when the value accepts multiple entries.

Because it is confusing and i had situations that i needed to recheck the spelling because i typed a Phobos tag like the Ares tag, and an Ares tag like the Phobos tag.

Metadorius:
> Because it is confusing and i had situations that i needed to recheck the spelling because i typed a Phobos tag like the Ares tag, and an Ares tag like the Phobos tag.

now my question is, which of those are more confusing, the inconsistency or being illogical? I am on the fence about it

DeathFishAtEase:
I have also hesitated about this, but after some thought, I think for many users, the more important point is the reusability of word segments. They will simply double-click the content in a document or another configuration file (due to the blocking of double-click selection by the dot, a single double-click usually does not select the entire flag but rather these subclass setting entries after the function flag word) and then copy it into the configuration file they are currently editing. When there are different options such as `AffectsHouse`, `AffectHouses`, `AffectsHouses`, `AffectedHouse`, etc. at the same time, they will have to spend extra effort to distinguish which is which.

Even for users and developers who often manually spell INI flags and variable names instead of copying them, the problem caused by this naming inconsistency still exists. For example, when `Crit.AffectsBelowPercent` first appeared in the documentation, the engine actually used `Affect`, and it was only later that correction steps were taken. The entire cycle spanned multiple versions from DevBuild#27 to DevBuild#33, so that in the 0.3.0.1 version released during this period, [the documentation](https://phobos.readthedocs.io/en/v0.3.0.1/New-or-Enhanced-Logics.html#chance-based-extra-damage-or-warhead-detonation-critical-hits) still had the description text as `Affects` while the ini-block had `Affect`.

---
In fact, #1380 has already changed many singular form INI flag values to accept multiple entries; and describing what values an INI flag allows mainly relies on the comment content immediately following the flag.

- In Ares, it is 
  > `[Type]►Key=` (value type)
- In Phobos, it is
  ```ini
  [SOMETYPE]        ; type name (with additional requirements)
  Key=default value ; value type
  ```

When seeking a balance between two requirements, such comments provide a guarantee for filling in values of the correct type under non-standard syntax.

mevitar:
> now my question is, which of those are more confusing, the inconsistency or being illogical? I am on the fence about it

I don't think people that care about tags being logical even mod this game. If they did, they quit the moment they saw Westwood's own tag misspellings (like CanPassiveAquire=).
Consistency seems far more important, not overthinking about whether tag naming makes sense.

Metadorius:
Aight, should be good then.

PR #1985, opened 2025-12-04
[MINOR] Fix MPDebug timer displaying when debug's visibility is off

11EJDE11:
In vanilla, the MP debug number at top-left shows even when debug stats display is toggled off. This change hides the number when `DrawMPDebugStats` is false which makes it consistent with the rest of the debug stats panel.

I'm not sure what the number represents - event timing perhaps.

image


Coronia (approved):

Metadorius:
should have added docs, especially since I said about it above

@Coronia please do that on develop then

Metadorius:
@Coronia reminder

Coronia:
It's been added in a recent commit already

PR #1986, opened 2025-12-04
[Minor] Cumulative.SimpleStack

Coronia:
add an option to allow Cumulative attached effect simply stack on one entity, instead of all being its own entities.

- If `Cumulative.SimpleStack` also set to true, latter applied effects will be counted as a part of the first one, instead of its own entity. This is better for performance but will make the effect always refresh during a reapplication.


TaranDahl (approved):

Coronia:
didn't work well in actual testing, and a bit useless since cumulative AE is not a great source of perf cost that should be handled specifically

PR #1989, opened 2025-12-05
Map Event 1000 - Force the check of events in sequential order

FS-21:
- By default, the game evaluates all map triggers in parallel. Adding this map event forces short-circuit evaluation as soon as any subsequent event returns `false`.
- This only affects evaluation from this control event (inclusive) to the last event in the list.
- All events placed before this control event still work in a non-short-circuiting manner.
- Sequential events will not begin evaluation until all preceding events have successfully completed.

In `mycampaign.map`:
```ini
[Events]
...
ID=EventCount,...,1000,0,0,0,...
...
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -152,6 +152,7 @@ This page lists all the individual contributions to the project by their author.
   - Warhead activation target health thresholds enhancements
   - Event 606: AttachEffect is attaching to a Techno
   - Linked superweapons
+  - Force the check of events in sequential order
 - **Starkku**:
   - Misc. minor bugfixes & improvements
   - AI script actions:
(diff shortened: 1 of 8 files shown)


FS-21:
So if you have:

TRIGGER_ID=4,E1,E2,E3,E4
E1, E2, E3 & E4 are executed in parallel as always happened. Is the same as:
TRIGGER_ID=5,E1,E2,E3,E4,1000

TRIGGER_ID=5,1000,E1,E2,E3,E4
E1, E2, E3 & E4 are executed in sequential order.

TRIGGER_ID=5,E1,1000,E2,E3,E4
E1 is executed in parallel as always happened.
E2, E3 & E4 are executed in sequential order. Only start when all the parallel events are completed.

TRIGGER_ID=5,E1,E2,1000,E3,E4
E1 & E2 are executed in parallel as always happened.
E3 & E4 are executed in sequential order. Only start when all the parallel events are completed.

TRIGGER_ID=5,E1,E2,1000,E4
E1, E2, E3 are executed in parallel as always happened.
E4 is executed in sequential order. Only start when all the parallel events are completed.

FS-21:
Until I find the cause of a rare crash this branch will stay in quarantine.

FS-21:
In theory the rare crash has been fixed. I also improved it and added a switch between sequential (1000) or parallel (1001, vanilla behaviour) modes.

I'll keep it as draft until I have time for more stress tests.

PR #1992, opened 2025-12-07
AttachmentUnit improvement - Selection and Flash

DeathFishAtEase:
Fixed the abnormal [SelectionFlash](https://phobos.readthedocs.io/en/latest/User-Interface.html#flashing-technos-on-selecting) issue caused by synchronization states and added a flag to pass the selection target to the Parent Unit when a Child Unit is selected.

- When `PassSelection=false`, the Child Unit will be selected as usual, but unlike before, it can correctly reference SelectionFlash instead of being terminated after 1 frame.
- When `PassSelection=true`, if a Child Unit is selected, it will search upward for the nearest Parent Unit within an AttachmentType where `PassSelection=false` (or the final caller).
  - This will cause all Child Units along the tree path to flash due to the Parent Unit being selected.

Before (in [the current latest version compiled three days ago](https://github.com/Phobos-developers/Phobos/pull/352#issuecomment-3613072028)):
Before_5ecdf4e023c926f621732d1998a82cceade1fca9

Now :
Now

In `rulesmd.ini`:
```ini
[AttachmentTypes]
+=PassSelect_False
+=PassSelect_True

[PassSelect_False]
PassSelection=false

[APOC]
Attachment0.Type=PassSelect_False
Attachment0.TechnoType=SREF
Attachment0.FLH=-512,0,208

Attachment1.Type=PassSelect_True
Attachment1.TechnoType=DLPH
Attachment1.FLH=512,0,208

[HTNK]
Attachment0.Type=PassSelect_False
Attachment0.TechnoType=DLPH
Attachment0.FLH=-512,256,208

Attachment1.Type=PassSelect_True
Attachment1.TechnoType=DLPH
Attachment1.FLH=-512,-256,208

[TNKD]
Attachment0.Type=PassSelect_True
Attachment0.TechnoType=APOC
Attachment0.FLH=256,0,208

[SREF]
Attachment0.Type=PassSelect_True
Attachment0.TechnoType=HTNK
Attachment0.FLH=256,0,208
```

The line with a red cross mark indicates that the selection status of a Child Unit is not passed to the Parent Unit when the Child Unit is selected, due to `PassSelection=false`.
image
*Selected State Transmission Relationship Diagram*

### DIFF
--- a/src/Ext/Techno/Hooks.TechnoAttachment.cpp
+++ b/src/Ext/Techno/Hooks.TechnoAttachment.cpp
@@ -470,7 +470,7 @@ bool __fastcall TechnoClass_Select(TechnoClass* pThis)
 	const auto pAttachment = pExt->ParentAttachment;
 	return pAttachment && pAttachment->GetType()->PassSelection
 		? pAttachment->Parent->Select()
-		: reinterpret_cast<bool(__thiscall*)(TechnoClass*)>(0x6FBFA0)(pThis);
+		: pThis->TechnoClass::Select();
 }
 DEFINE_FUNCTION_JUMP(VTABLE, 0x7F5DBC, TechnoClass_Select) // UnitClass
 DEFINE_FUNCTION_JUMP(VTABLE, 0x7EB1A4, TechnoClass_Select) // InfantryClass
@@ -487,13 +487,13 @@ void __fastcall TechnoClass_Flash(TechnoClass* pThis, void*, int duration)
 			pAttachment->Child->Flash(duration);
 	}
 
-	return reinterpret_cast<void(__thiscall*)(TechnoClass*, int)>(0x6F9DD0)(pThis, duration);// TechnoClass::Flash
+	return pThis->TechnoClass::Flash(duration);
 }
 DEFINE_FUNCTION_JUMP(VTABLE, 0x7F5DB8, TechnoClass_Flash) // UnitClass
 DEFINE_FUNCTION_JUMP(VTABLE, 0x7EB1A0, TechnoClass_Flash) // InfantryClass
 DEFINE_FUNCTION_JUMP(VTABLE, 0x7E23EC, TechnoClass_Flash) // AircraftClass
 
-void __fastcall BuildingClass_Flash(TechnoClass* pThis, void*, int duration)
+void __fastcall BuildingClass_Flash(BuildingClass* pThis, void*, int duration)
 {
 	const auto pExt = TechnoExt::ExtMap.Find(pThis);
 
@@ -503,7 +503,7 @@ void __fastcall BuildingClass_Flash(TechnoClass* pThis, void*, int duration)
 			pAttachment->Child->Flash(duration);
 	}
 
-	return reinterpret_cast<void(__thiscall*)(TechnoClass*, int)>(0x456E00)(pThis, duration);// BuildingClass::Flash
+	return pThis->BuildingClass::Flash(duration);
 }
 DEFINE_FUNCTION_JUMP(VTABLE, 0x7E4004, BuildingClass_Flash) // BuildingClass
 


src/Misc/Hooks.AI.cpp
```diff
@@ -5,7 +5,24 @@
...
+			if (childFlashRemaining > 0)
+			{
+				attachment->Child->Flash(childFlashRemaining);
+			}
+		}
+		else
+		{
+			attachment->AI();
+		}
+	}
```
Metadorius: I think this should be done by amending AttachmentClass::AI instead

src/Ext/Techno/Hooks.Misc.cpp
```diff
@@ -722,15 +722,88 @@ DEFINE_HOOK(0x5F46AE, ObjectClass_Select, 0x7)
...
+			for (int i = 0; i < ObjectClass::CurrentObjects.Count; i++)
+			{
+				if (ObjectClass::CurrentObjects[i] == pOriginal)
+				{
+					int lastIndex = ObjectClass::CurrentObjects.Count - 1;
+					ObjectClass::CurrentObjects[i] = ObjectClass::CurrentObjects[lastIndex];
+					ObjectClass::CurrentObjects.Count--;
+					break;
+				}
+			}
```
Metadorius: Hmm, can't we take over the game's code before anything is actually selected and redirect it to as if we were clicking or selecting only the parent unit? Then we won't need that extra handling, supposedly.

src/Ext/Techno/Hooks.Misc.cpp
```diff
@@ -722,15 +722,88 @@ DEFINE_HOOK(0x5F46AE, ObjectClass_Select, 0x7)
...
+				{
+					parentAlreadySelected = true;
+					break;
+				}
+			}
+
+			if (!parentAlreadySelected)
+			{
+				ObjectClass::CurrentObjects.AddItem(pFinalTarget);
+			}
```
Metadorius: same here

Metadorius (commented):
looking good, couple of thoughts though

also do all the children flash because of InheritStateEffects? maybe that should be excluded from state effects if it's an UI flash and be controlled by other tags, like InheritCommands to signify they will also be commanded when you give order to the parent unit?

src/Misc/Hooks.AI.cpp
DeathFishAtEase: I originally planned to do that, but later found that directly modifying `AttachmentClass::AI` was somewhat difficult, so I went up and found the call point of `AttachmentClass::AI` to operate on it. If you have any good methods, feel free to edit it.

src/Ext/Techno/Hooks.Misc.cpp
DeathFishAtEase: I made some attempts, but ultimately failed. Although it is not elegant, currently it indeed is such that clicking Child will directly redirect to Parent, but I think the query here cannot be ignored, although if possible, it would indeed be better to handle it earlier.

Metadorius: Perhaps @TaranDahl or @NetsuNegi could help with this and the AI modification?

Metadorius:
> Indeed, that was passed by `InheritStateEffects`. I'm not entirely sure what your original code intended to do—for instance, whether synchronizing the flash state at that moment was intentional, since if the goal was to prevent individual flashing when directly selecting a child, it would still take effect for one frame.

Well, the whole point of all those customisation tags is to allow people to make different attachment types depending on what they need. `InheritStateEffects` is meant so that the child gets chaosed when parent gets chaosed, IC'd with parent etc. Flash on the other hand is an UI thing that is not necessarily related. So a separate tag would make sense IMO.

src/Ext/Techno/Hooks.Misc.cpp
TaranDahl: if (pExt->ParentAttachment && PassSelection)
{
  pExt->ParentAttachment->Parent->Select();
  return RETN;
}

Isn't this okay?

NetsuNegi: I'll replace the function in vtable

src/Ext/Techno/Hooks.TechnoAttachment.cpp
```diff
@@ -462,6 +462,53 @@ DEFINE_FUNCTION_JUMP(VTABLE, 0x7EB0D4, TechnoClass_IsSurfaced);
...
 
+#pragma region Select & Flash
+
+bool __fastcall TechnoClass_Select(TechnoClass* pThis)
+{
+	const auto pExt = TechnoExt::ExtMap.Find(pThis);
+	const auto pAttachment = pExt->ParentAttachment;
+	return pAttachment && pAttachment->GetType()->PassSelection
+		? pAttachment->Parent->Select()
+		: reinterpret_cast<bool(__thiscall*)(TechnoClass*)>(0x6FBFA0)(pThis);
```
NetsuNegi: will update YRpp later

src/Ext/Techno/Hooks.TechnoAttachment.cpp
```diff
@@ -462,6 +462,53 @@ DEFINE_FUNCTION_JUMP(VTABLE, 0x7EB0D4, TechnoClass_IsSurfaced);
...
+{
+	const auto pExt = TechnoExt::ExtMap.Find(pThis);
+
+	for (const auto& pAttachment : pExt->ChildAttachments)
+	{
+		if (pAttachment->GetType()->InheritStateEffects && pAttachment->Child)
+			pAttachment->Child->Flash(duration);
+	}
+
+	return reinterpret_cast<void(__thiscall*)(TechnoClass*, int)>(0x6F9DD0)(pThis, duration);// TechnoClass::Flash
```
NetsuNegi: same this

src/Ext/Techno/Hooks.TechnoAttachment.cpp
```diff
@@ -462,6 +462,53 @@ DEFINE_FUNCTION_JUMP(VTABLE, 0x7EB0D4, TechnoClass_IsSurfaced);
...
+{
+	const auto pExt = TechnoExt::ExtMap.Find(pThis);
+
+	for (const auto& pAttachment : pExt->ChildAttachments)
+	{
+		if (pAttachment->GetType()->InheritStateEffects && pAttachment->Child)
+			pAttachment->Child->Flash(duration);
+	}
+
+	return reinterpret_cast<void(__thiscall*)(TechnoClass*, int)>(0x456E00)(pThis, duration);// BuildingClass::Flash
```
NetsuNegi: this also

PR #1949, opened 2025-11-08
[Highly Customized] Distribution click action mode

CrimRecya:
Reopen #1453.

Previously mentioned that the remaining items in TODO can be considered in the future, so I will not make any changes before merging this.

PR #1993, opened 2025-12-08
Distribution click action mode

TaranDahl:
Reopen #1453 again.

### `[ ]` Switch No-Move Command

- Now you can change the click action by using `AllowSwitchNoMoveCommand` hotkey. If the behavior to be executed by the current techno is different from the behavior displayed by the mouse, and the behavior to be executed will make the techno move near the target, the behavior will be replaced with area guard. Regardless of whether or not switch hotkey is used, default behavior can be changed through `DefaultApplyNoMoveCommand`.
- For localization add `TXT_SWITCH_NOMOVE` and `TXT_SWITCH_NOMOVE_DESC` into your `.csf` file.

In `rulesmd.ini`:
```ini
[GlobalControls]
AllowSwitchNoMoveKeyEnabled=false                   ; boolean
DefaultApplyNoMoveCommand=true                      ; boolean
```

### `[ ]` Distribution Mode

- You can now distribute commands across similar targets in a specified range when holding down a hotkey if `AllowDistributionCommand` is enabled. This behavior is like using the selected objects one by one to click on each target within the spread range.
  - The targets within the spread range will be allocated equally to the selected technos. Only when the behavior to be performed by the current techno is the same as that displayed by the mouse will it be allocated (e.g. a healer will only be allocated friendly targets it can heal, while combat units will only be allocated the targets they would attack). Neutral targets are only eligible if the clicked target itself is neutral. Otherwise, the techno will fall back to the vanilla behavior of clicking the original target (the move fallback, or area guard if the switch no-move command is active). This will display a range ring.
  - `AllowDistributionFilterKeyEnabled` allows you to set target filter by hotkey, which default to `DefaultDistributionFilterMode`.
    - When the filter is `None`, it is the default behavior of the game. If the range is not zero at this time, a green ring will be displayed. You can adjust the filter mode to:
      - `Like` - only targets with the same armor type (Completely identical `Armor`) will be selected among the targets allocated in the range. At this time, a blue ring will be displayed.
      - `Type` - only targets of the same type (like infantries, vehicles or buildings) will be selected among the targets allocated in the range. At this time, a yellow ring will be displayed.
      - `Name` - only targets of the same name (or with the same `GroupAs`) will be selected among the targets allocated in the range. At this time, a red ring will be displayed.
  - `DefaultDistributionSpreadRange` controls the initial spread range, which is a number that's corresponding to the amount of cell radius * 512.
- `AllowDistributionSpreadKeyEnabled` allows you to cycle through preset spread ranges by hotkey. There're 4 tiers of range that can be selected by this hotkey which are identical to 0, 4, 8 and 16 cells.
  - When the range is 0, it is the original default behavior of the game.
- `AllowDistributionUseClick` controls whether distribution mode is activated by clicking on a target. When set to false, distribution mode only works via press-and-drag, and the range ring will only be shown while dragging.
- You can also adjust spread range by using the mouse wheel while holding down the specific hotkey if `AllowDistributionSpreadScroll` set to true. This allows a more precise control of spread range that each step will increase/decrease it by `DistributionSpreadScrollStep`, with 20 cells as its maximum value.
- `AllowDistributionSpreadDrag` allows you to adjust the spread range by pressing and dragging the mouse while holding down the specific hotkey. The drag distance from the starting point determines the spread range. This can naturally co-exist with `AllowDistributionSpreadScroll`.
- `AllowDistributionCommandOnOwner`, `AllowDistributionCommandOnAllies`, `AllowDistributionCommandOnEnemies` & `AllowDistributionCommandOnNeutral` allow the distribution command to work on owner, allies, enemies or neutral target. If picking a target that's not eligible, it'll fallback to vanilla command.
- It's possible to add a button for distribution mode in the bottom bar by adding `DistributionMode` in the `ButtonList` of `AdvancedCommandBar` and `MultiplayerAdvancedCommandBar`.
  - The positions of each button are hardcoded, so it'll only decide whether enable this button or not. Distribute Mode button is now always listed after all the vanilla ones.
  - The asset of these buttons should be added in `sidec0x.mix` files which correspond to different sides, with the name `button12.shp`.
- The distribution mode hotkey works like the vanilla waypoint mode hotkey: holding it down activates the mode while releasing it deactivates it, a sound is played and a message is shown when the mode is turned on/off, and the bottom bar button (if added) toggles the mode as well. The sounds played when the mode is turned on/off can be customized with `StartDistributionModeSound` and `EndDistributionModeSound`, which default to the planning mode sounds.
- A sound is also played each time a distribution command is issued, either by clicking a target or on drag release. It can be customized with `AddDistributionModeCommandSound`.
- For localization add `TXT_SWITCH_NOMOVE`, `TXT_DISTR_SPREAD`, `TXT_DISTR_FILTER`, `TXT_DISTR_HOLDDOWN`, `TXT_SWITCH_NOMOVE_DESC`, `TXT_DISTR_SPREAD_DESC`, `TXT_DISTR_FILTER_DESC`, `TXT_DISTR_HOLDDOWN_DESC`, `MSG:DistributionModeOn`, `MSG:DistributionModeOff`, `TIP:DistributionMode` into your `.csf` file.

In `rulesmd.ini`:
```ini
[GlobalControls]
AllowDistributionCommand=false                      ; boolean
AllowDistributionCommandOnOwner=true                ; boolean
AllowDistributionCommandOnAllies=true               ; boolean
AllowDistributionCommandOnEnemies=true              ; boolean
AllowDistributionCommandOnNeutral=true              ; boolean
AllowDistributionSpreadKeyEnabled=true              ; boolean
AllowDistributionFilterKeyEnabled=false             ; boolean
DefaultDistributionSpreadRange=2048                 ; integer between 0 and 5120
DefaultDistributionFilterMode=2                     ; integer, 0 - None , 1 - Like , 2 - Type , 3 - Name
DistributionSpreadScrollStep=256                    ; integer, minimum 16

[AudioVisual]
StartDistributionModeSound=                         ; sound entry, defaults to `StartPlanningModeSound`
EndDistributionModeSound=                           ; sound entry, defaults to `EndPlanningModeSound`
AddDistributionModeCommandSound=                    ; sound entry
```

In `ra2md.ini`:
```ini
[Phobos]
AllowDistributionSpreadScroll=true                  ; boolean
AllowDistributionSpreadDrag=true                    ; boolean
AllowDistributionUseClick=true                      ; boolean
```

In `uimd.ini`:
```ini
[AdvancedCommandBar]
ButtonList=[Button1],DistributionMode,[ButtonX]     ; List of button entry

[MultiplayerAdvancedCommandBar]
ButtonList=[Button1],DistributionMode,[ButtonX]     ; List of button entry
```


TaranDahl:
TODO:
(Summarized from https://github.com/Phobos-developers/Phobos/pull/1949#pullrequestreview-3473595716)

- [x] Create independent file for `AdvancedCommandBarButtonClass`.
- [x] Allow registering new `AdvancedCommandBarButtonClass` like `MakeCommand`.
- [x] Rewrite the spread range code to allow exact range.
- [x] Press-and-drag mode

@Metadorius Please confirm the above summary, or supplement/correct the incorrect parts.

TaranDahl:
> there was also a comment about implementing press-and-drag mode (and not sure if the "same type" (infantry/buildings/vehicles/etc) is needed

I think there is not enough labor force to add more features.
I think the existing work is sufficient in terms of completion. We should just perfect the existing features and then merge them. 
As for more features, they should be another work, and can be implemented later, by anyone who wants them.

Metadorius:
> I think there is not enough labor force to add more features. I think the existing work is sufficient in terms of completion. We should just perfect the existing features and then merge them. As for more features, they should be another work, and can be implemented later, by anyone who wants them. Breaking it into two PR can also help avoid the problem of having an excessive amount of code piled up in one PR, which may lead to no one willing to review it.

This isn't a feature though? It is just a somewhat small improvement that brings it in line with how modern games do it.

- `0x6D2280` (exists in YRpp) for transforming screen coords into world coords (though Z is 0, maybe would want to account for that)
- `0x4AC4CC` is where the drag-selection band is set and the mouse hold is done
- `0x4ABCEB` is where the drag-selection release is handled
- `0xD90` in `TacticalClass` is `Rect` that contains X, Y, Width, Height

From that you could calculate 2 points, get world coords via function above, use this info to draw a corresponding circle and set the mode to such. I am not sure what is complex here. If needed I can send my decompile for this.

TaranDahl:
@Metadorius Is there any difference between `ShapeButtonClass` and `AdvancedCommandBarButtonClass`?

Metadorius:
@TaranDahl `ShapeButtonClass` is a vanilla engine class, `AdvancedCommandBarButtonClass` is something I propose to invent (perhaps a descendant from `ShapeButtonClass`) that stores the extra things that are currently stored as static arrays, for example.

TaranDahl:
> @TaranDahl `ShapeButtonClass` is a vanilla engine class, `AdvancedCommandBarButtonClass` is something I propose to invent (perhaps a descendant from `ShapeButtonClass`) that stores the extra things that are currently stored as static arrays, for example.

Judging from the current code, there seems to be no need for it to inherit from `ShapeButtonClass`. After all, the game originally stores those attributes in a static array.
image

Metadorius:
> After all, the game originally stores those attributes in a static array.


Which is a bad pattern. You have class fields and methods for that.

TaranDahl:
Yeah I will make a new class to arrange the new buttons. But for the vanilla buttons, maybe we should just let them be?

Metadorius:
> Yeah I will make a new class to arrange the new buttons. But for the vanilla buttons, maybe we should just let them be?

Yeah I didn't mean we should be squeezing vanilla static array shitcode into proper classes necessarily, should be good.

TaranDahl:
If we want to implement dragging, how should we handle the selection range?
Right now, it's several specific values.

src/Commands/DistributionMode.cpp
```diff
@@ -0,0 +1,457 @@
...
+
+	if (SessionClass::Instance.MultiplayerObserver)
+		return;
+
+	VocClass::PlayGlobal(RulesExt::Global()->EndDistributionModeSound, 0x2000, 1.0);
+
+	if (!DistributionModeHoldDownCommandClass::OffMessageShowed)
+	{
+		DistributionModeHoldDownCommandClass::OffMessageShowed = true;
+		MessageListClass::Instance.PrintMessage(GeneralUtils::LoadStringUnlessMissing("MSG:DistributionModeOff", L"Distribution mode unabled."), RulesClass::Instance->MessageDelay, HouseClass::CurrentPlayer->ColorSchemeIndex, true);
```
Metadorius:
```suggestion
		MessageListClass::Instance.PrintMessage(GeneralUtils::LoadStringUnlessMissing("MSG:DistributionModeOff", L"Distribution mode disabled."), RulesClass::Instance->MessageDelay, HouseClass::CurrentPlayer->ColorSchemeIndex, true);
```

src/Commands/DistributionMode.cpp
```diff
@@ -0,0 +1,457 @@
...
+	{
+		const auto pCell = MapClass::Instance.GetCellAt(DisplayClass::Instance.CurrentFoundation_CenterCell);
+		const auto color = (filterMode > 1)
+			? ((filterMode == 3) ? ColorStruct { 255, 0, 0 } : ColorStruct { 200, 200, 0 })
+			: ((filterMode == 1) ? ColorStruct { 0, 100, 255 } : ColorStruct { 0, 255, 50 });
+		Game::DrawRadialIndicator(false, true, pCell->GetCoords(), color, static_cast<float>(spreadMode ? (2 << spreadMode) : 0.5), false, true);
+	}
+
+	return 0;
+}
```
Metadorius, line 529: I think since this is so complex and spans across than a single "execute" method -- the actual distribution mode handling should not be a part of the command declaration/definition but elsewhere, and command classes should just call what they need in execute method. Same for hooks.

src/Commands/DistributionMode.cpp
```diff
@@ -0,0 +1,457 @@
...
+				&& (HouseClass::CurrentPlayer->IsAlliedWith(pTechno->Owner)
+					? Phobos::Config::AllowDistributionCommand_AffectsAllies
+					: Phobos::Config::AllowDistributionCommand_AffectsEnemies))
+			{
+				VocClass::PlayGlobal(RulesExt::Global()->AddDistributionModeCommandSound, 0x2000, 1.0);
+				const bool targetIsNeutral = pTechno->Owner->IsNeutral();
+				const auto pType = pTechno->GetTechnoType();
+				const int range = (2 << spreadMode);
+				const auto center = pTechno->GetCoords();
+				const auto pItems = Helpers::Alex::getCellSpreadItems(center, range);
```
Metadorius: You said that the code would need significant changes previously to accomodate arbitrary precise range, but I don't see what is significant here?

src/Commands/DistributionMode.cpp
```diff
@@ -0,0 +1,457 @@
...
+	Unsorted::MoveFeedback = false;
+}
+
+void __fastcall DistributionModeHoldDownCommandClass::AreaGuardAction(TechnoClass* pTechno)
+{
+	pTechno->ClickedMission(Mission::Area_Guard, reinterpret_cast<ObjectClass*>(pTechno->GetCellAgain()), nullptr, nullptr);
+	Unsorted::MoveFeedback = false;
+}
+
+DEFINE_HOOK(0x4AE7B3, DisplayClass_ActiveClickWith_Iterate, 0x0)
```
Metadorius: I think this big hook should be split out into functions/methods called from a hook.

Metadorius (changes requested):
partial review

src/Commands/DistributionMode.cpp
TaranDahl: Once we used shortcut keys because there were only a few available values. If there are too many available values, then the shortcut keys will become impractical and we need to delete them. There probably won't be much actual work, but there will be changes to the existing functions.

Metadorius: I think there's a misunderstanding. I think shortcut keys *could* be left in with pre-set ranges, and dragging will mean an arbitrary range is selected. Both could exist in parallel.

Although if I am being honest I personally don't think there's a need in anything else other than drag-distribution. This is how it works in many modern RTS/RTT games with advanced controls and there's no customisation, I never heard it being a problem.

src/Commands/DistributionMode.cpp
TaranDahl: Changed

src/Commands/DistributionMode.cpp
TaranDahl: Changed

src/Commands/DistributionMode.cpp
TaranDahl: Changed

Coronia:
maybe the default `DistributionSpreadScrollStep` of 16 is too small. Now it takes long time to scroll it to a desirable value. Maybe make it 256 which is 1 cell per scroll, making it more intuitive and easy to use while keeping the preciseness

also the upper limit of `DistributionSpreadScrollStep` seems to be not needed, as it was 0-4-8-16 range before which are all greater than 512

as for the drag, I think instead of a choice between it and scroll, it'd better just make a toggle in rulesmd.ini to enable or disable it, like `AllowDistributionCommand.SpreadModeScroll` did, since these 2 patterns can naturally co-exist without conflict

TaranDahl:
> maybe the default `DistributionSpreadScrollStep` of 16 is too small. Now it takes long time to scroll it to a desirable value. Maybe make it 256 which is 1 cell per scroll, making it more intuitive and easy to use while keeping the preciseness
> 
> also the upper limit of `DistributionSpreadScrollStep` seems to be not needed, as it was 0-4-8-16 range before which are all greater than 512
> 
> as for the drag, I think instead of a choice between it and scroll, it'd better just make a toggle in rulesmd.ini to enable or disable it, like `AllowDistributionCommand.SpreadModeScroll` did, since these 2 patterns can naturally co-exist without conflict

Changed.

docs/User-Interface.md
```diff
@@ -585,6 +585,66 @@ For this command to work in multiplayer - you need to use a version of [YRpp spa
 - Deselect 1 or 5 object(s) from current selected objects.
 - For localization add `TXT_DESELECT`, `TXT_DESELECT_DESC`, `TXT_DESELECT5` and `TXT_DESELECT5_DESC` into your `.csf` file.
 
+### `[ ]` Distribution Mode Spread / Filter / Enable
+
+- Now you can change the click action by using `AllowSwitchNoMoveCommand` hotkey. If the behavior to be executed by the current techno is different from the behavior displayed by the mouse, and the behavior to be executed will make the techno move near the target, the behavior will be replaced with area guard. Regardless of whether or not switch hotkey is used, default behavior can be changed through `DefaultApplyNoMoveCommand`.
```
Metadorius, line 622: 1. I don't think this has to be in Distribution Mode docs
2. it's not clear what this does, is it a toggle for "fallback" behavior if different from the mouse one and allows disabling it? if so - maybe `UnavailableActionFallsBackToMove=action` will be more intuitive? or `UnavailableActionFallBack=none/action` (defaults to move), with customisability per type (not sure what handles actions, MissionType?) maybe useful?

docs/User-Interface.md
```diff
@@ -585,6 +585,66 @@ For this command to work in multiplayer - you need to use a version of [YRpp spa
...
+- Now you can also change the click action when hold down the specific hotkey if enabled `AllowDistributionCommand`. The new behavior is like using the selected objects one by one to click on each target within the spread range.
```
Metadorius:
```suggestion
- You can now distribute commands across similar targets in a specified range when holding down a hotkey if `AllowDistributionCommand` is enabled. This behavior is like using the selected objects one by one to click on each target within the spread range.
```

docs/User-Interface.md
```diff
@@ -585,6 +585,66 @@ For this command to work in multiplayer - you need to use a version of [YRpp spa
...
+  - The targets within the spread range will be allocated equally to the selected technos. Only when the behavior to be performed by the current techno is the same as that displayed by the mouse will it be allocated. Otherwise, it will return to the original default behavior of the game (it will not be effective for technos in the air). This will display a range ring.
```
Metadorius, line 624: > Otherwise, it will return to the original default behavior of the game

do you mean the move fallback? if so - you should mention that and/or link it

> (it will not be effective for technos in the air)

hmm?

docs/User-Interface.md
```diff
@@ -585,6 +585,66 @@ For this command to work in multiplayer - you need to use a version of [YRpp spa
...
+  - `DefaultDistributionSpreadRange` controls the initial spread range, which is a number that's corresponding to the amount of cell radius * 512.
```
Metadorius, line 625: we should probably move the modes description earlier
```suggestion
  - `DefaultDistributionSpreadRange` controls the initial spread range for the scrollwheel mode, which is a number that's corresponding to the amount of cell radius * 512.
```

docs/User-Interface.md
```diff
@@ -585,6 +585,66 @@ For this command to work in multiplayer - you need to use a version of [YRpp spa
...
+- For localization add `TXT_SWITCH_NOMOVE`, `TXT_DISTR_SPREAD`, `TXT_DISTR_FILTER`, `TXT_DISTR_HOLDDOWN`, `TXT_SWITCH_NOMOVE_DESC`, `TXT_DISTR_SPREAD_DESC`, `TXT_DISTR_FILTER_DESC`, `TXT_DISTR_HOLDDOWN_DESC`, `MSG:DistributionModeOn`, `MSG:DistributionModeOff`, `TIP:DistributionMode` into your `.csf` file.
+
+In `rulesmd.ini`:
+```ini
+[GlobalControls]
+AllowSwitchNoMoveCommand=false                      ; boolean
+AllowDistributionCommand=false                      ; boolean
+AllowDistributionCommand.SpreadModeHotKey=true       ; boolean
+AllowDistributionCommand.SpreadModeScroll=true      ; boolean
+AllowDistributionCommand.SpreadModeDrag=true        ; boolean
```
Metadorius, line 649: I don't think it should be controlled from `rules.ini`. Control schemes are not up to mod author.

The mod author should be able to enable or disable the feature (because it significantly alters the gameplay), but the players must be able to choose their preferred implementation themselves via `RA2MD.ini`.

docs/User-Interface.md
```diff
@@ -585,6 +585,66 @@ For this command to work in multiplayer - you need to use a version of [YRpp spa
...
+- `AllowDistributionCommand.FilterMode` allows you to set target filter by hotkey, which default to `DefaultDistributionFilterMode`.
+  - When the filter is `None`, it is the default behavior of the game. If the range is not zero at this time, a green ring will be displayed. You can adjust the filter mode to:
+    - `Like` - only targets with the same armor type (Completely identical `Armor`) will be selected among the targets allocated in the range. At this time, a blue ring will be displayed.
+    - `Type` - only targets of the same type (like infantries, vehicles or buildings) will be selected among the targets allocated in the range. At this time, a yellow ring will be displayed.
+    - `Name` - only targets of the same name (or with the same `GroupAs`) will be selected among the targets allocated in the range. At this time, a red ring will be displayed.
+- `AllowDistributionCommand.AffectsAllies`, `AllowDistributionCommand.AffectsEnemies` & `AllowDistributionCommand.AffectsNeutral` allow the distribution command to work on allies (including owner), enemies or neutral target. If picking a target that's not eligible, it'll fallback to vanilla command.
```
Metadorius, line 636: I don't think this is needed? It should derive this info from context. If I select a healer unit and distribute - it should heal any similar unit, with similar in this context being a friendly eligible target. Same for attacking, if it's hostile then attack it.

docs/User-Interface.md
```diff
@@ -585,6 +585,66 @@ For this command to work in multiplayer - you need to use a version of [YRpp spa
...
+- `AllowDistributionCommand.SpreadModeHotKey` allows you to cycle through preset spread ranges by hotkey. There're 4 tiers of range that can be selected by this hotkey which are identical to 0, 4, 8 and 16 cells.
+  - When the range is 0, it is the original default behavior of the game.
+- `AllowDistributionCommand.UseClick` controls whether distribution mode is activated by clicking on a target. When set to false, distribution mode only works via press-and-drag, and the range ring will only be shown while dragging.
+- You can also adjust spread range by using the mouse wheel while holding down the specific hotkey if `AllowDistributionCommand.SpreadModeScroll` set to true. This allows a more precise control of spread range that each step will increase/decrease it by `DistributionSpreadScrollStep`, with 20 cells as its maximum value.
+- `AllowDistributionCommand.SpreadModeDrag` allows you to adjust the spread range by pressing and dragging the mouse while holding down the specific hotkey. The drag distance from the starting point determines the spread range. This can naturally co-exist with `SpreadModeScroll`.
```
Metadorius, line 630: Does it issue the command on drag release? if so - it should be mentioned in documentation.

docs/User-Interface.md
```diff
@@ -585,6 +585,66 @@ For this command to work in multiplayer - you need to use a version of [YRpp spa
...
+- `AllowDistributionCommand.AffectsAllies`, `AllowDistributionCommand.AffectsEnemies` & `AllowDistributionCommand.AffectsNeutral` allow the distribution command to work on allies (including owner), enemies or neutral target. If picking a target that's not eligible, it'll fallback to vanilla command.
+- It's possible to add a button for distribution mode in the bottom bar by adding `DistributionMode` in the `ButtonList` of `AdvancedCommandBar` and `MultiplayerAdvancedCommandBar`.
+  - The positions of each button are hardcoded, so it'll only decide whether enable this button or not. Distribute Mode button is now always listed after all the vanilla ones.
+  - The asset of these buttons should be added in `sidec0x.mix` files which correspond to different sides, with the name `button12.shp`.
+- For localization add `TXT_SWITCH_NOMOVE`, `TXT_DISTR_SPREAD`, `TXT_DISTR_FILTER`, `TXT_DISTR_HOLDDOWN`, `TXT_SWITCH_NOMOVE_DESC`, `TXT_DISTR_SPREAD_DESC`, `TXT_DISTR_FILTER_DESC`, `TXT_DISTR_HOLDDOWN_DESC`, `MSG:DistributionModeOn`, `MSG:DistributionModeOff`, `TIP:DistributionMode` into your `.csf` file.
```
Metadorius, line 640: Does this work exactly like waypoint mode hotkey? sound and a message when on/off, pressing on the bottom bar enables it on/off, and holding hotkey will also activate it (releasing deactivates).

I think this should be mentioned.

docs/User-Interface.md
```diff
@@ -585,6 +585,66 @@ For this command to work in multiplayer - you need to use a version of [YRpp spa
...
+AllowDistributionCommand.SpreadModeDrag=true        ; boolean
+AllowDistributionCommand.UseClick=true              ; boolean
+AllowDistributionCommand.FilterMode=true            ; boolean
+AllowDistributionCommand.AffectsAllies=true         ; boolean
+AllowDistributionCommand.AffectsEnemies=true        ; boolean
+AllowDistributionCommand.AffectsNeutral=true        ; boolean
+
+[AudioVisual]
+StartDistributionModeSound=                         ; sound entry
+EndDistributionModeSound=                           ; sound entry
```
Metadorius, line 658: let's default this to the same sound as waypoint mode

docs/User-Interface.md
```diff
@@ -585,6 +585,66 @@ For this command to work in multiplayer - you need to use a version of [YRpp spa
...
+AddDistributionModeCommandSound=                    ; sound entry
```
Metadorius, line 659: what's this about

docs/Whats-New.md
```diff
@@ -991,6 +990,7 @@ HideShakeEffects=false           ; boolean
 - `<Player @ X>` can now be used as owner for pre-placed objects on skirmish and multiplayer maps (by Starkku)
 - Allow customizing charge turret delays per burst on a weapon (by Starkku)
 - Draw visual effects for airburst weapons (by CrimRecya)
+- Distribution click action mode (by CrimRecya)
```
Metadorius:
```suggestion
- Distribution click action mode (by CrimRecya & TaranDahl)
```

docs/User-Interface.md
```diff
@@ -585,6 +585,66 @@ For this command to work in multiplayer - you need to use a version of [YRpp spa
...
+  - When the range is 0, it is the original default behavior of the game.
+- `AllowDistributionCommand.UseClick` controls whether distribution mode is activated by clicking on a target. When set to false, distribution mode only works via press-and-drag, and the range ring will only be shown while dragging.
+- You can also adjust spread range by using the mouse wheel while holding down the specific hotkey if `AllowDistributionCommand.SpreadModeScroll` set to true. This allows a more precise control of spread range that each step will increase/decrease it by `DistributionSpreadScrollStep`, with 20 cells as its maximum value.
+- `AllowDistributionCommand.SpreadModeDrag` allows you to adjust the spread range by pressing and dragging the mouse while holding down the specific hotkey. The drag distance from the starting point determines the spread range. This can naturally co-exist with `SpreadModeScroll`.
+- `AllowDistributionCommand.FilterMode` allows you to set target filter by hotkey, which default to `DefaultDistributionFilterMode`.
```
Metadorius, line 631: I am not sure how this tag works? at least from the documentation

docs/User-Interface.md
```diff
@@ -585,6 +585,66 @@ For this command to work in multiplayer - you need to use a version of [YRpp spa
 - Deselect 1 or 5 object(s) from current selected objects.
 - For localization add `TXT_DESELECT`, `TXT_DESELECT_DESC`, `TXT_DESELECT5` and `TXT_DESELECT5_DESC` into your `.csf` file.
 
+### `[ ]` Distribution Mode Spread / Filter / Enable
```
Metadorius, line 620: I wonder if we could assign it by default to something like C? or maybe (crazy idea) holding X (and making specifically the release of the same hotkey in a short time scattering)?

Metadorius (commented):
I did a doc pass, sorry, haven't got to the rest yet. treat it both as a doc review (if what I asked for is true but docs are not reflecting it yet) and design review (if what I asked about is not true yet)

docs/User-Interface.md
TaranDahl: Too complicated to write such an algorithm

TaranDahl: Moreover, I'm convinced that even if you develop such an algorithm, some people will still be dissatisfied.

docs/User-Interface.md
TaranDahl: Set it to false, and it will function exactly as you described earlier. It is fully tailored for you.

docs/User-Interface.md
TaranDahl: I believe the current architecture does not allow the same key to be used for different functions.

docs/User-Interface.md
Metadorius: I don't see how it is complicated. You already have filtering on weapon/warhead level. Filter out those whose cursor would show the same action and be done with it. no need to overthink it.

Even then, if you think this won't be enough, those flags are the wrong customisation. You're just cutting away the feature instead of placing customisations in decisions where it's not obvious what would be better.

docs/User-Interface.md
Metadorius: so this is a toggle between togglable and not togglable mode? I am not sure why this needs to be customised at `rulesmd.ini` level though?

TaranDahl:
@Metadorius How about the rest of your review?

docs/User-Interface.md
Metadorius: waypoint mode somehow works like this, so this should too

Metadorius:
> How about the rest of your review?

Have you addressed the design changes I requested? without that it would be a moot point to get deeper into code

TaranDahl:
> > How about the rest of your review?
> 
> Have you addressed the design changes I requested? without that it would be a moot point to get deeper into code

So there are no further issues with the design, right?

Metadorius:
> So there are no further issues with the design, right?

you left some of my comments/suggestion unaddressed though, either by changes or replies, that's what I was pointing at

TaranDahl:
> > So there are no further issues with the design, right?
> 
> you left some of my comments/suggestion unaddressed though, either by changes or replies, that's what I was pointing at

Yeah I know it. I was looking forward to some other reviews, to deal with them together.

docs/User-Interface.md
TaranDahl: > waypoint mode somehow works like this, so this should toowaypoint mode somehow works like this, so this should too

I don't quite know what you're referring to.
By default, the hotkey for waypoint mode is Z, and the Z key is exclusively used by waypoint mode.
It is not feasible under the current hotkey architecture for different CommandClass to share the same hotkey and distinguish between short presses and long presses only. Also I don't think it is worthwhile to significantly expand the existing hotkey architecture for this at the moment; it would make more sense to devote that effort to the development of OpenYR.

docs/User-Interface.md
TaranDahl: > 1. I don't think this has to be in Distribution Mode docs

wdym? A function without docs?

> 2\. it's not clear what this does, is it a toggle for "fallback" behavior if different from the mouse one and allows disabling it? if so - maybe  will be more intuitive? or  (defaults to move), with customisability per type (not sure what handles actions, MissionType?) maybe useful?`UnavailableActionFallsBackToMove=action``UnavailableActionFallBack=none/action`

The current fallback behavior is not simply changing the Action, but a specially customized behavior targeting the AreaGuard unit itself instead of the target.

docs/User-Interface.md
TaranDahl: Are you asking about the `UseClick` or `FilterMode`?

docs/User-Interface.md
TaranDahl: It depends on the `UseClick`.

docs/User-Interface.md
TaranDahl: It applies to all modes except PressAndDrag. I don't think it needs special explanation, because it should be intuitive that PressAndDrag does not take this into account.

docs/User-Interface.md
Metadorius: oh I confused the response thread, nevermind. just ensure it works like waypoint mode hotkey, that's all

docs/User-Interface.md
Metadorius: > wdym? A function without docs?

no, just a separate section. it's not really relevant to distribution mode and can be used without it, right? it's just a complementary function

Metadorius: > The current fallback behavior is not simply changing the Action, but a specially customized behavior targeting the AreaGuard unit itself instead of the target.

then the tag name is still not clear IMO

docs/User-Interface.md
Metadorius: elaborate

TaranDahl: `UseClick=true`: PressAndDrag only changes the spread.
`UseClick=false`: PressAndDrag issues the command on drag release, and other spread changing mode is ignored.

docs/User-Interface.md
TaranDahl: Changed.

docs/User-Interface.md
```diff
@@ -624,6 +624,63 @@ For this command to work in multiplayer - you need to use a version of [YRpp spa
...
+
+In `ra2md.ini`:
+```ini
+[Phobos]
+DefaultApplyNoMoveCommand=true                      ; boolean
+AllowDistributionSpreadHotkey=true                 ; boolean
+AllowDistributionSpreadScroll=true                 ; boolean
+AllowDistributionSpreadDrag=true                    ; boolean
+AllowDistributionUseClick=true                      ; boolean
+AllowDistributionFilterHotkey=true                 ; boolean
```
Coronia, line 669: I think filter mode shouldn't be made toggleable by player tho, as it did affect how the target is picking up. Also this rule is unintuitive for normal players to memorize. Should still stay in rules or at least add a way to make it control by modders.

Other spread mode toggles are purely for control and is fine to be customized I guess

docs/User-Interface.md
TaranDahl: Changed.

docs/User-Interface.md
TaranDahl: Changed.

docs/User-Interface.md
TaranDahl: Docs added.

docs/User-Interface.md
TaranDahl: Yeah it should work like that.

docs/User-Interface.md
TaranDahl: Changed.

docs/User-Interface.md
TaranDahl: Any candidate?

docs/User-Interface.md
TaranDahl: Changed.

docs/User-Interface.md
TaranDahl: Changed.

TaranDahl:
@Metadorius Please check again when you are free.

Coronia:
The house filter is still needed for modders to decide targets from which houses could distribution command be applied to

Also you get me wrong, I'm just saying that whether filter mode could be controlled by players should be decided by the modders, since it's affecting how the target will be selecting which can be balance-breaking. The hotkey itself could still be kept if modders allow that, which is why it's been put in rulesmd.ini instead of ra2md.ini and default to false

And add some handling for Armor filter mode to make it take shield armor into account

Coronia:
@TaranDahl @Metadorius after re-reviewing the current design I think only `AllowDistributionSpreadScroll`, `AllowDistributionSpreadDrag` and `AllowDistributionUseClick` should be made toggleable by players through client. `AllowDistributionSpreadHotkey` is meaningless since players can already decide whether or not to set the hotkey. So it should either be removed or put back to rules for global control

As for `DefaultDistributionSpreadRange`, `DistributionSpreadScrollStep` and `DefaultDistributionFilterMode`, I think it's hard to make it controlled by players through client either since it's neither supporting a drag box setting nor a manually enter setting. Besides, normal players can't understand that '256 means 1 cell' and what are those filter mode means, so it's also not feasible to make them as settings. Since `DefaultDistributionSpreadRange` and `DefaultDistributionFilterMode` are default values that can be changed by other method later on, it's also a bit meaningless to customize them

Last but not least, I'm still going back to the topic of 'whether a hotkey needs toggle'. Now I believe that every hotkey needs a toggle even if it's purely QoL change, because it'll still require the modders to manually set them up in client or it'll go wrong ingame, which can't be automatically completed whenever a new hotkey is added. I don't think it's a wise idea to let modders having to redefine client ini again when something new is added. If some hotkey is really about QoL, then we can make its toggle default to true, but it's still needed

PR #1994, opened 2025-12-08
[Vanilla Fix] Fix the issue that units will goto farest location if target is closer than `MinimumRange`

NetsuNegi:
- In the vanilla game, when the distance from unit to target is less than the `MinimumRange`, he still prioritizes searching for the farthest position from the target. In this situation, he will prioritize searching for positions that are closer in distance.
- 在原版游戏中，当单位到目标的距离小于MinimumRange时，他仍然会优先搜索距离目标最远的位置。现在这种情况下他会优先搜索距离更近的位置了.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -464,6 +464,7 @@ This page lists all the individual contributions to the project by their author.
   - Customize Ares's radar jam logic
   - Customize if cloning need power
   - Customize type selection for IFV
+  - Fix the issue that units will goto farest location if target is closer than `MinimumRange`
 - **Apollo** - Translucent SHP drawing patches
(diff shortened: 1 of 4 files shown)


Coronia (approved):

TaranDahl:
> Why was this merged so quickly? This change should be made optional though a tag IMO.

Personally, I think this is indeed a bug fix, so there is no need for a toggle.

> Also the documentation does not mention this at all.

Which document do you mean? Docs are already added to What's new, CREDITS and manual.

mevitar:
I meant Fixed / Improved Logics, i didn't see it there when i mentioned it, but i can see it now.

PR #1998, opened 2025-12-10
Allow other vehicles to attempt to move above the target like `BalloonHover=yes`

TaranDahl:
### Allow other vehicles to attempt to move above the target like ZEP

- In vanilla, vehicles with `BalloonHover=yes` and a weapon with projectile with `Vertical=yes` will attempt to move above the target.
- Now you can make other vehicles do the same without `BalloonHover=yes` or `Vertical=yes`.

In `rulesmd.ini`:
```ini
[SOMEWEAPON]                        ; WeaponType
GoAboveTarget=false               ; boolean
```


Coronia (changes requested):
This add another SelectWeapon call which is debateable for performance. IDA suggests that there're already some existing calls of it, could we somehow reuse that instead of reimplementing it again?

Image

TaranDahl:
Closing this because a similar function is implemented in https://github.com/Phobos-developers/Phobos/pull/2088.

PR #2008, opened 2025-12-17
Add veterancy-based target filtering for weapons and warheads

Flactine:
This PR introduces **veterancy-based filtering** for both weapon targeting and warhead effect application.

---

## 1. Weapon target filtering by target veterancy

Weapons can now restrict which targets they are allowed to attack based on the target’s veterancy level.
This affects both weapon firing eligibility and weapon selection logic (except for special cases where weapon selection is fixed by the engine).

### New INI options:

```ini
[SOMEWEAPON]             ; WeaponType
CanTargetVeterancy=all     ; List of Affected Veterancy Enumeration (none|rookie|veteran|elite|all)
```

Targets whose veterancy is not included in the specified enumeration will not be considered valid targets for the weapon.

---

## 2. Warhead effect filtering by target veterancy

Warheads can now conditionally apply their effects based on the target’s veterancy level.
Unlike weapon filtering, this logic does **not** affect whether a weapon can fire, but instead determines whether the warhead’s effects will be applied after detonation.

### New INI options:

```ini
AffectsVeterancy=all        ; List of Affected Veterancy Enumeration (none|rookie|veteran|elite|all)
```

* The warhead will only affect targets whose veterancy is explicitly listed in AffectsVeterancy.

---

---

---

本次更新新增了**基于目标等级（Veterancy）的筛选逻辑**，分别作用于**武器攻击判定**和**弹头效果生效判定**。

---

## 1. 基于目标等级的武器筛选

武器现在可以根据目标单位的经验等级，决定是否允许对其进行攻击。
该逻辑不仅影响武器能否开火，也会影响武器选择流程（部分引擎强制固定选择的特殊情况除外）。

### 新增 INI 选项（WeaponType）：

```ini
[SOMEWEAPON]             ; 武器类型
CanTargetVeterancy=all     ; List of Affected Veterancy Enumeration (none|rookie|veteran|elite|all)
```

当目标的等级未被列入指定枚举时，武器将不会将其视为可攻击目标。

---

## 2. 基于目标等级的弹头效果筛选

弹头现在可以根据目标单位的经验等级，决定爆炸后是否对其产生效果。
与武器筛选不同，此逻辑**不影响武器是否开火**，而是作用于爆炸结算阶段。

### 新增 INI 选项：

```ini
[SOMEWARHEAD]               ; 弹头类型
AffectsVeterancy=all        ; List of Affected Veterancy Enumeration (none|rookie|veteran|elite|all)
```

* 只有等级被列入指定枚举的目标才会受到弹头效果影响

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -656,7 +656,9 @@ This page lists all the individual contributions to the project by their author.
 - **solar-III (凤九歌)**
   - Target scanning delay customization (documentation)
   - Skip target scanning function calling for unarmed technos (documentation)
-- **Flactine** - add target filtering options to attacheffect system
+- **Flactine** 
+  - add target filtering options to attacheffect system
+  - add veterancy-based target filtering for weapons and warheads
 - **tyuah8**:
   - Drive/Jumpjet/Ship/Teleport locomotor did not power on when it is un-piggybacked bugfix
   - Destroyed unit leaves sensors bugfix
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -657,8 +657,8 @@ This page lists all the individual contributions to the project by their author.
   - Target scanning delay customization (documentation)
   - Skip target scanning function calling for unarmed technos (documentation)
 - **Flactine** 
-  - add target filtering options to attacheffect system
-  - add veterancy-based target filtering for weapons and warheads
+  - Add target filtering options to attacheffect system
+  - Add veterancy-based target filtering for weapons and warheads
 - **tyuah8**:
   - Drive/Jumpjet/Ship/Teleport locomotor did not power on when it is un-piggybacked bugfix
   - Destroyed unit leaves sensors bugfix
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -2326,8 +2326,8 @@ In `rulesmd.ini`:
 [SOMEWARHEAD]               ; WarheadType
 AffectsBelowPercent=1.0     ; floating point value, percents or absolute
 AffectsAbovePercent=0.0     ; floating point value, percents or absolute
-AffectsBelowVeterancy=2.0     ; floating point value, percents or absolute, default to [General] -> VeteranCap
-AffectsAboveVeterancy=0.0     ; floating point value, percents or absolute
+AffectsBelowVeterancy=2.0     ; double, default to [General] -> VeteranCap
+AffectsAboveVeterancy=0.0     ; double
 AffectsNeutral=true         ; boolean
 EffectsRequireVerses=false  ; boolean
 ```
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -2326,8 +2326,8 @@ In `rulesmd.ini`:
 [SOMEWARHEAD]               ; WarheadType
 AffectsBelowPercent=1.0     ; floating point value, percents or absolute
 AffectsAbovePercent=0.0     ; floating point value, percents or absolute
-AffectsBelowVeterancy=2.0     ; double, default to [General] -> VeteranCap
-AffectsAboveVeterancy=0.0     ; double
+AffectsBelowVeterancy=     ; floating point value, percents or absolute, default to [General] -> VeteranCap
+AffectsAboveVeterancy=0.0     ; floating point value, percents or absolute
 AffectsNeutral=true         ; boolean
 EffectsRequireVerses=false  ; boolean
 ```
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -2326,6 +2326,8 @@ In `rulesmd.ini`:
 [SOMEWARHEAD]               ; WarheadType
 AffectsBelowPercent=1.0     ; floating point value, percents or absolute
 AffectsAbovePercent=0.0     ; floating point value, percents or absolute
+AffectsBelowVeterancy=2.0     ; floating point value, percents or absolute, default to [General] -> VeteranCap
+AffectsAboveVeterancy=0.0     ; floating point value, percents or absolute
 AffectsNeutral=true         ; boolean
 EffectsRequireVerses=false  ; boolean
 ```
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -2326,8 +2326,8 @@ In `rulesmd.ini`:
 [SOMEWARHEAD]               ; WarheadType
 AffectsBelowPercent=1.0     ; floating point value, percents or absolute
 AffectsAbovePercent=0.0     ; floating point value, percents or absolute
-AffectsBelowVeterancy=     ; floating point value, percents or absolute, default to [General] -> VeteranCap
-AffectsAboveVeterancy=0.0     ; floating point value, percents or absolute
+AffectsBelowVeterancy=      ; floating point value, percents or absolute, default to [General] -> VeteranCap
+AffectsAboveVeterancy=0.0   ; floating point value, percents or absolute
 AffectsNeutral=true         ; boolean
 EffectsRequireVerses=false  ; boolean
 ```
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -2312,7 +2312,7 @@ Conventional.IgnoreUnits=false  ; boolean
 ### Customizable Warhead trigger conditions
 
 - `AffectsBelowPercent` and `AffectsAbovePercent` can be used to set the health percentage thresholds that target needs to be below/equal and/or above of for the Warhead to detonate. If target has zero health left this check is bypassed.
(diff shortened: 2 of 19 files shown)


docs/Whats-New.md
```diff
@@ -465,6 +465,10 @@ New:
 - [Customize if cloning need power](Fixed-or-Improved-Logics.md#customize-if-cloning-need-power) (by NetsuNegi)
 - [Added Target Filtering Options to AttachEffect System](New-or-Enhanced-Logics.md#attached-effects) (by Flactine)
 - [Customize type selection for IFV](Fixed-or-Improved-Logics.md#customize-type-selection-for-ifv) (by NetsuNegi)
+- CellSpread in cylinder shape (by TaranDahl)
+- CellSpread damage check if victim is in air or on floor (by TaranDahl)
```
TaranDahl: Wrong docs.

src/Ext/WeaponType/Body.h
```diff
@@ -114,6 +116,8 @@ class WeaponTypeExt
 			, CanTargetHouses { AffectedHouse::All }
 			, CanTarget_MaxHealth { 1.0 }
 			, CanTarget_MinHealth { 0.0 }
+			, CanTarget_MaxVeterancy { 2.0f }
```
TaranDahl: Perhaps you should use Nullable and use `RulesClass::Instance->VeteranCap` as the default value?

TaranDahl (changes requested):

src/Ext/WarheadType/Body.h
```diff
@@ -192,6 +192,8 @@ class WarheadTypeExt
 
 		Valueable<double> AffectsBelowPercent;
 		Valueable<double> AffectsAbovePercent;
+		Nullable<float> AffectsBelowVeterancy;
+		Valueable<float> AffectsAboveVeterancy;
```
TaranDahl: Why not use double?

src/Ext/WeaponType/Body.cpp
```diff
@@ -157,6 +164,7 @@ void WeaponTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 	// handle SkipWeaponPicking
 	if (this->CanTarget != AffectedTarget::All || this->CanTargetHouses != AffectedHouse::All
 		|| this->CanTarget_MaxHealth < 1.0 || this->CanTarget_MinHealth > 0.0
+		|| this->CanTarget_MaxVeterancy.Get(static_cast<float>(RulesClass::Instance->VeteranCap)) < static_cast<float>(RulesClass::Instance->VeteranCap) || this->CanTarget_MinVeterancy > 0.0f
```
TaranDahl: You may test this in-game. I'm not sure if RulesClass::Instance has been loaded at this time.

Flactine: It is loaded. Debug information can display "RulesClass::Instance->VeteranCap" at that time.

docs/Fixed-or-Improved-Logics.md
```diff
@@ -2320,6 +2326,8 @@ In `rulesmd.ini`:
 [SOMEWARHEAD]               ; WarheadType
 AffectsBelowPercent=1.0     ; floating point value, percents or absolute
 AffectsAbovePercent=0.0     ; floating point value, percents or absolute
+AffectsBelowVeterancy=2.0     ; floating point value, percents or absolute, default to [General] -> VeteranCap
```
TaranDahl: You forgot the docs.

docs/New-or-Enhanced-Logics.md
```diff
@@ -2856,6 +2862,8 @@ CanTarget=all            ; List of Affected Target Enumeration (none|land|water|
 CanTargetHouses=all      ; List of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
 CanTarget.MaxHealth=1.0  ; floating point value, percents or absolute
 CanTarget.MinHealth=0.0  ; floating point value, percents or absolute
+CanTarget.MaxVeterancy=2.0  ; floating point value, percents or absolute, default to [General] -> VeteranCap
```
TaranDahl: Same

TaranDahl (approved):

docs/Whats-New.md
```diff
@@ -465,6 +465,8 @@ New:
 - [Customize if cloning need power](Fixed-or-Improved-Logics.md#customize-if-cloning-need-power) (by NetsuNegi)
 - [Added Target Filtering Options to AttachEffect System](New-or-Enhanced-Logics.md#attached-effects) (by Flactine)
 - [Customize type selection for IFV](Fixed-or-Improved-Logics.md#customize-type-selection-for-ifv) (by NetsuNegi)
+- [Weapon target filtering by target veterancy](New-or-Enhanced-Logics.md#weapon-targeting-filter) (by Flactine)
+- [Warhead effect filtering by target veterancy](Fixed-or-Improved-Logics.md#customizable-warhead-trigger-conditions) (by Flactine)
```
Coronia: these 2 lines should be put at the end of this section

Coronia (approved):
maybe it's better to put veterancy in an enum, so it can be represented in a clearer way than having to remember what 0.0, 1.0 and 2.0 stand for, and could be reused in the future

CREDITS.md
```diff
@@ -656,7 +656,9 @@ This page lists all the individual contributions to the project by their author.
 - **solar-III (凤九歌)**
   - Target scanning delay customization (documentation)
   - Skip target scanning function calling for unarmed technos (documentation)
-- **Flactine** - add target filtering options to attacheffect system
+- **Flactine** 
+  - add target filtering options to attacheffect system
+  - add veterancy-based target filtering for weapons and warheads
```
ZivDero: Capitalization please.

src/Ext/Techno/Body.cpp
```diff
@@ -743,6 +743,16 @@ bool TechnoExt::IsHealthInThreshold(TechnoClass* pObject, double min, double max
...
 }
 
+bool TechnoExt::IsVeterancyInThreshold(TechnoClass* pTechno, double min, double max)
+{
+	double veterancy = 0.0;
+
+	if (pTechno->GetTechnoType()->Trainable)
+		veterancy = pTechno->Veterancy.Veterancy;
+
+	return veterancy >= min && veterancy <= max;
```
Metadorius: One of the limits should be inclusive, other exclusive. Otherwise imagine if you make one warhead effect that works on 0-1, another that works on 1-2. With your implementation it's impossible to create non-overlapping ranges of effect in a good fashion. See how `IsHealthInThreshold` handles this.

I think the lower bound should be inclusive, and the upper bound should be exclusive. The reasoning is that the game itself uses it like this. [0, 1) is Rookie, [1, 2) is Veteran, and [2, +inf) is Elite. This means


```suggestion
	return veterancy >= min && veterancy < max;
```

should be used.

However, the same issue as with `IsHealthInThreshold` arises when you can't then cover all ranges in a nice way, hence why we made 0 HP follow a slightly different set of rules. Accepting ideas on how to handle this similarly.

ZivDero (changes requested):
It looks solid overall, but I can't help shake the feeling that the "can target" and "can affect" stuff is getting a bit spaghetti. Could combine it all into single "can target" and "can affect" functions (with subfunctions for the various logics). Though this is not the fault of this specific PR.

src/Ext/WeaponType/Body.cpp
```diff
@@ -46,6 +46,11 @@ bool WeaponTypeExt::ExtData::IsHealthInThreshold(TechnoClass* pTarget) const
 	return TechnoExt::IsHealthInThreshold(pTarget, this->CanTarget_MinHealth, this->CanTarget_MaxHealth);
 }
 
+bool WeaponTypeExt::ExtData::IsVeterancyInThreshold(TechnoClass* pTarget) const
+{
+	return TechnoExt::IsVeterancyInThreshold(pTarget, this->CanTarget_MinVeterancy, this->CanTarget_MaxVeterancy.Get(RulesClass::Instance->VeteranCap));
```
Metadorius: > `this->CanTarget_MaxVeterancy.Get(RulesClass::Instance->VeteranCap)`

@ZivDero or whoever, defaultable template pleeeeeeeeeeease 🥺

Metadorius (changes requested):

src/Ext/WeaponType/Body.cpp
Flactine: Is it OK to change into "RulesClass::Instance->VeteranCap + 1.0"?

Metadorius: maybe... not sure about the usability

@Starkku any ideas?

TaranDahl:
> maybe it's better to put veterancy in an enum, so it can be represented in a clearer way than having to remember what 0.0, 1.0 and 2.0 stand for, and could be reused in the future

@Flactine You can make that flag read both the enum (Rookie, Veteran, Elite) and float types simultaneously.
After that, I think this PR can be merged? @Coronia @ZivDero @Metadorius

TaranDahl (approved):

TaranDahl:
@Coronia @Metadorius Going to merge this in 24h if no one has further instructions.

PR #2010, opened 2025-12-19
Fixed the issue where non-repairer units needed sensors to attack cloaked friendly units

TaranDahl:

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -643,6 +643,7 @@ This page lists all the individual contributions to the project by their author.
   - Dehardcode the `ZAdjust` of warhead anim
   - Fix an issue where some effects pointing to a unit were not properly cleared when the unit changed its owner
   - Fix an issue where the vanilla script ignores jumpjets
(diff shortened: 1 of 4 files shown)


Metadorius:
If I recall correctly, @chaserli had a PR in spawner related to that, can you check it out?

TaranDahl:
> If I recall correctly, @chaserli had a PR in spawner related to that, can you check it out?

You mean this one?
https://github.com/CnCNet/yrpp-spawner/pull/20

Metadorius:
yes, exactly

TaranDahl:
I think that one is not related to this one. This one is about `CanFire`, while that one is about `DecideAction`.

Fryone:
There is also an issue with cloaked techno's that is not owned by player, but owner house is under player's control and techno is invisible, selectable but not usable.

Coronia (approved):

TaranDahl:
> There is also an issue with cloaked techno's that is not owned by player, but owner house is under player's control and techno is invisible, selectable but not usable.

You can open an issue.

PR #2011, opened 2025-12-21
[Vanilla Fix] Fix the bug that ai will try to product aircraft even the airport has no free dock for it

NetsuNegi:


Coronia (approved):

wenrui1245:
e4a24ee78deefd8ab81d737f1e3fa320
这个问题的确有所改善，日志中也不再输出Computer is abandoning production of SomeAircraft 的字样，不过从观察者视角来看，工厂上的图标却还在，这是正常现象吗？

This issue has indeed improved, and the log no longer outputs 'Computer is abandoning production of SomeAircraft.' However, from the observer's perspective, the icon on the factory is still there. Is this normal?
(My English is not good, so I use translation software.)

PR #2014, opened 2025-12-22
Apply per-target warhead effects on `ReceiveDamage`

TaranDahl:
### Toggle per-target warhead effects apply timing

- Now you can set the following flag to `false` to apply the **Phobos** warhead effects that take effect on each target when taking damage, rather than when the projectiles detonate.
  - This will allow such effects to be applied through damage without projectiles, including but not limited to damage from particles, vanilla radiation, and Ares' `GenericWarhead` superweapon.
  - This will also cause all effects that can completely prevent damage to also prevent these warhead effects, including but not limited to `DamageSelf`, `DamageAirThreshold`, `AffectsAllies`, `AffectsAir`.
  - If you use a warhead with CellSpread to damage a building multiple times, then these effects will be applied multiple times. If you don't want this to happen, use [`MergeBuildingDamage`](#allow-merging-aoe-damage-to-buildings-into-one).
  - The affected effects include:
    - [Remove mind-control](#break-mind-control-on-impact)
    - [Type convertion](#convert-technotype-on-impact)
    - [`BuildingSell` & `BuildingUndeploy`](#sell-or-undeploy-building-on-impact)
    - [`RemoveDisguise`](#remove-disguise-on-impact)
    - [`ReverseEngineer`](#reverse-engineer-warhead)
    - [Modify shield](#shields)
    - [Modify attach-effects](#attached-effects)
    - [Critical hits](#chance-based-extra-damage-or-warhead-detonation--critical-hits)
      - Due to technical reasons, `Crit.SuppressWhenIntercepted=false` and `Crit.ApplyChancePerTarget=true` will forced to be used.

In `rulesmd.ini`:
```ini
[CombatDamage]                        ; WarheadType
ApplyPerTargetEffectsOnDetonate=true  ; boolean

[SOMEWARHEAD]                         ; WarheadType
ApplyPerTargetEffectsOnDetonate=      ; boolean, default to [CombatDamage] -> ApplyPerTargetEffectsOnDetonate
```

```{note}
- Ares' warhead effects, such as EMP or IronCurtain warhead, will not be affected. 
- Ares' warhead effect controllers, such as `EffectsRequireDamage`, only affect Ares' effects. So they have nothing to do with this.
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -656,6 +656,7 @@ This page lists all the individual contributions to the project by their author.
   - CellSpread in cylinder shape
   - CellSpread damage check if victim is in air or on floor
   - Fix an issue where non-repairer units needed sensors to attack cloaked friendly units
+  - Toggle per-target warhead effects apply timing
 - **solar-III (凤九歌)**
   - Target scanning delay customization (documentation)
   - Skip target scanning function calling for unarmed technos (documentation)
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -2632,6 +2632,7 @@ SpawnsCrate(N).Weight=1  ; integer
 
 - Now you can set the following flag to `false` to apply the Phobos warhead effects that take effect on each target when taking damage, rather than when the projectiles detonate.
   - This will allow such effects to be applied through damage without projectiles, including but not limited to damage from particles, vanilla radiation, and Ares' `GenericWarhead` superweapon.
+  - This will also cause all effects that can completely prevent damage to also prevent these warhead effects, including but not limited to `DamageSelf`, `DamageAirThreshold`, `AffectsAllies`, `AffectsAir`.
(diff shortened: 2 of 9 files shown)


Starkku:
Code and documentation are fine.

I assume with `ApplyPerTargetEffectsOnDetonate=false` the Warhead effects require non-zero damage to be forwarded to receive damage function. Ares added a key called `EffectsRequireDamage` to remove this requirement for its damage-adjacent Warhead effects, does that still function in this case too? Something maybe worth mentioning in doc if so.

TaranDahl:
> Code and documentation are fine.
> 
> I assume with `ApplyPerTargetEffectsOnDetonate=false` the Warhead effects require non-zero damage to be forwarded to receive damage function. Ares added a key called `EffectsRequireDamage` to remove this requirement for its damage-adjacent Warhead effects, does that still function in this case too? Something maybe worth mentioning in doc if so.

I don't think it should have an impact. Because this change only affects the Phobos-related effects, and Ares effects still use its own area search.

Starkku (approved):

PR #2020, opened 2025-12-28
[Ares Fix] 2 Bugfixes about spawn survivors

NetsuNegi:
- Fixed the issue that technos cannot spawn survivors due to non-probabilistic reasons when the tech type was destroyed.
- 修复了科技类型会因为非概率原因而无法生成出幸存者的问题.
- Fixed the bug that vehicle survivor can spawn on wrong position when transport has been destroyed.
- 修复了载具幸存者会被生成在错误的位置的bug.

### DIFF
--- a/src/Ext/Techno/Body.cpp
+++ b/src/Ext/Techno/Body.cpp
@@ -876,6 +876,103 @@ bool TechnoExt::SimpleDeployerAllowedToDeploy(UnitClass* pThis, bool defaultValu
 	return true;
 }
 
+bool TechnoExt::EjectRandomly(FootClass* pEjectee, const CoordStruct& coords, int distance, bool select)
+{
+	std::vector<CoordStruct> usableCoords;
+
+	for (int direction = 0; direction < 8; ++direction)
+	{
+		const CellStruct tmpCoords = Unsorted::AdjacentCell[direction];
+		CoordStruct ejectCoords { coords.X + tmpCoords.X * distance, coords.Y + tmpCoords.Y * distance, coords.Z };
+		const auto pCell = MapClass::Instance.TryGetCellAt(ejectCoords);
+
+		if (!pCell)
+			continue;
+
+		const auto occupied = pEjectee->IsCellOccupied(pCell, FacingType::None, -1, nullptr, true);
+
+		if (occupied != Move::OK && occupied != Move::MovingBlock)
+			continue;
+
+		if (pEjectee->WhatAmI() == InfantryClass::AbsID)
+		{
+			ejectCoords = pCell->FindInfantrySubposition(ejectCoords, false, false, false);
+			ejectCoords.Z = coords.Z;
+		}
+		else
+		{
+			ejectCoords = CellClass::Cell2Coord(pCell->MapCoords, coords.Z);
+		}
+
+		usableCoords.emplace_back(ejectCoords);
+	}
+
+	const int count = static_cast<int>(usableCoords.size());
+
+	if (!count)
+		return false;
+
+	return TechnoExt::EjectSurvivor(pEjectee, usableCoords[ScenarioClass::Instance->Random(0, count - 1)], select);
+}
+
+bool TechnoExt::EjectSurvivor(FootClass* pSurvivor, CoordStruct coords, bool select)
+{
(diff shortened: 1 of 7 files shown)


ltyylt:
It's already implemented in https://github.com/Phobos-developers/Phobos/blob/develop/src/Utilities/AresAddressInit.cpp#L30

NetsuNegi:
> It's already implemented in https://github.com/Phobos-developers/Phobos/blob/develop/src/Utilities/AresAddressInit.cpp#L30

alright, but why disable it

TaranDahl (approved):

TaranDahl:
I simply gave it a try, but it seems to be ineffective for the first bug. 🤔

chaserli:
> > It's already implemented in https://github.com/Phobos-developers/Phobos/blob/develop/src/Utilities/AresAddressInit.cpp#L30
> 
> alright, but why disable it

There were issues that couldn't confirm if they were actually the repercussion of this or something else. If you're also sure   it's safe, toggle the tweak before releasing a dev build.

src/Ext/Techno/Body.cpp
```diff
@@ -876,6 +876,103 @@ bool TechnoExt::SimpleDeployerAllowedToDeploy(UnitClass* pThis, bool defaultValu
...
+		const CellStruct tmpCoords = Unsorted::AdjacentCell[direction];
+		CoordStruct ejectCoords { coords.X + tmpCoords.X * distance, coords.Y + tmpCoords.Y * distance, coords.Z };
+		const auto pCell = MapClass::Instance.TryGetCellAt(ejectCoords);
+
+		if (!pCell)
+			continue;
+
+		const auto occupied = pEjectee->IsCellOccupied(pCell, FacingType::None, -1, nullptr, true);
+
+		if (occupied != Move::OK && occupied != Move::MovingBlock)
```
CrimRecya, line 894: I don't quite understand why `Move::MovingBlock` is allowed. When another unit is entering this position, I think directly generating units on it will cause overlap.

src/Ext/Techno/Body.cpp
```diff
@@ -876,6 +876,103 @@ bool TechnoExt::SimpleDeployerAllowedToDeploy(UnitClass* pThis, bool defaultValu
...
+	pSurvivor->Transporter = nullptr;
+	pSurvivor->LastMapCoords = pCell->MapCoords;
+
+	if (chuted)
+	{
+		const bool scat = pSurvivor->OnBridge;
+		const auto occupation = scat ? pCell->AltOccupationFlags : pCell->OccupationFlags;
+
+		if (occupation & 0x1C)
+			pCell->ScatterContent(CoordStruct::Empty, true, true, scat);
```
CrimRecya, line 959: As an aside: I'm not sure why only the occupation of infantry was assessed here although the same judgment is also made in Ares.

CrimRecya (commented):
Looks good overall. I just have two small questions.

momokama:
the problem of "viehcle cannot spawn survivors" due to "suicide weapon" is unfixed yet.

src/Ext/Techno/Body.cpp
NetsuNegi: it it just copy from vanilla and ares's code

TaranDahl:
@Coronia How did you verify that it worked? I just did the same test as @momokama, and the result shows that it isn't working.

Coronia:
we did several rounds of transport destruction test with survivability setting to 100%, and so far all passengers are alive

as for the Suicide weapon thing, I don't consider this as a bug or something that should be solved in this PR, it's another issue if anything

TaranDahl:
> as for the Suicide weapon thing, I don't consider this as a bug or something that should be solved in this PR, it's another issue if anything

I think that's also a problem that should be addressed here, because 

> "Fixed the issue that technos cannot spawn survivors due to non-probabilistic reasons when the tech type was destroyed."

and this issue was caused by Ares using the wrong parameters, it is a non-probabilistic reason.
I will directly submit it to the dev branch to solve it.

TaranDahl:
@chaserli Excuse me, why hasn't the fix in https://github.com/Phobos-developers/Phobos/blob/develop/src/Utilities/AresAddressInit.cpp#L30 been enabled? Is there any special reason?

PR #2022, opened 2025-12-29
Fixed the issue where units recruited by a team with `AreTeamMembersRecruitable=false` cannot be recruited even if they have been liberated by that team

TaranDahl:


Coronia:
This is only for AI recruitable I think? As these team members can clearly be recruited again by Create Team actions

TaranDahl:
> This is only for AI recruitable I think? As these team members can clearly be recruited again by Create Team actions

I don't quite understand TAction.
As I understand it, what TAction Create Team (ActionKind == 4) does is create an instance of TeamClass (6DEB57). This should have nothing to do with the recruit.

Coronia (approved):

PR #2029, opened 2026-01-06
Updateable firing anim

TaranDahl:
### Updateable firing anim

- In vanilla, firing anims is attached to the firer, but it won't update its type and location to fit the firer's facing. This is now customizable by the following flags.

In `rulesmd.ini`:
```ini
[AudioVisual]
FiringAnim.Update=false   ; boolean

[SOMEWEAPON]              ; WeaponType
Anim.Update=              ; boolean, default to [AudioVisual] -> FiringAnim.Update
```

```{note}
This effect will cause problem when used together with `[AnimType] -> Next`. `Next` modifies the Anim type over time, while this function changes it back, resulting in the Anim being unable to end.
```

### DIFF
--- a/src/Ext/Anim/Body.cpp
+++ b/src/Ext/Anim/Body.cpp
@@ -112,7 +112,8 @@ void AnimExt::ExtData::UpdateAsFiringAnim()
 			pNewType = pWeapon->Anim.GetItemOrDefault(0);
 		}
 
-		pThis->Type = pNewType;
+		if (pNewType)
+			pThis->Type = pNewType;
 
 		auto burstIdx = pOwner->CurrentBurstIndex;
 		pOwner->CurrentBurstIndex = this->FromBurstIdx;
--- a/src/Ext/Anim/Body.cpp
+++ b/src/Ext/Anim/Body.cpp
@@ -65,29 +65,6 @@ void AnimExt::ExtData::DeleteAttachedSystem()
 	}
 }
 
-inline unsigned int TranslateFixedPoint(size_t bitsFrom, size_t bitsTo, unsigned int value, unsigned int offset = 0)
-{
-	const size_t MaskIn = ((1u << bitsFrom) - 1);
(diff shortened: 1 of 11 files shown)


src/Ext/Anim/Body.cpp
```diff
@@ -65,6 +65,64 @@ void AnimExt::ExtData::DeleteAttachedSystem()
...
+		if (highest >= 3)
+		{
+			auto offset = 1u << (highest - 3);
+			auto index = TranslateFixedPoint(16, highest, static_cast<WORD>((pOwner)->GetRealFacing().GetValue<16>()), offset);
+			pNewType = pWeapon->Anim.GetItemOrDefault(index);
+		}
+		else
+		{
+			pNewType = pWeapon->Anim.GetItemOrDefault(0);
+		}
```
Starkku: Maybe use `GeneralUtils::GetItemForDirection()` if possible.

TaranDahl: Changed.

src/Utilities/GeneralUtils.h
```diff
@@ -90,4 +90,37 @@ class GeneralUtils
 
 		return items[0];
 	}
+
+	// Returns item from vector based on given direction and number of items in the vector, f.ex directional animations.
+	// Vector is expected to have 2^n items where n >= 3 and n <= 16 for the logic to work correctly, other cases return first item.
+	// Do not pass an empty vector, size/indices are not sanity checked here.
+	template<typename T>
+	static T GetItemForDirection(DynamicVectorClass<T> const& items, DirStruct const& direction)
```
Otamaa: probably shoud use `Iterator<T>`  instead so it compatible with other vectors , instead of making duplicated codes.

TaranDahl: Changed.

Otamaa:
LGTM

PR #2035, opened 2026-01-08
[Vanilla Bug] Fixed the bug that if object has been removed from LogicClass in Update(), next object will be skip

NetsuNegi:


TaranDahl (approved):

TaranDahl:
After simply thinking about it, I think there seem to be no exceptions.

PR #2038, opened 2026-01-11
[Phobos Fix] Fix the bug that weapon cannot used to intercept on gound bullet if it's projectile has `AG=no`

NetsuNegi:


src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -481,6 +481,19 @@ DEFINE_HOOK(0x6FC749, TechnoClass_GetFireError_AntiUnderground, 0x5)
 	return GoOtherChecks;
 }
 
+DEFINE_HOOK(0x6FC7EB, TechnoClass_GetFireError_InterceptBullet, 0x7)
+{
+	enum { IgnoreAG = 0x6FC815, ContinueCheck = 0x6FC7F2 };
+
+	GET(AbstractClass*, pTarget, EBX);
+
+	if (pTarget->WhatAmI() == AbstractType::Bullet)
```
TaranDahl, line 490: I don't think there is any need for checking. Just return IgnoreAG.

TaranDahl (approved):

Rinimabi:
seems allright...


``` ini
[ArmorTypes]
Projectile=none

[InterceptorDrone]
Primary=InterceptorLaser
Interceptor.GuardRange=12
Interceptor=yes

[InterceptorLaser]
Projectile=InterceptorPro
Warhead=InterceptorLaserWH

[InterceptorPro]
Inviso=yes
Image=none
AA=yes
AG=no

[InterceptorLaserWH]
Verses=0%,0%,0%,0%,0%,0%,0%,0%,0%,0%,0%
Versus.Projectile=100%

[AAHeatSeeker]
Interceptable=yes
Strength=1
Armor=Projectile
```

src/Ext/Techno/Hooks.Firing.cpp
NetsuNegi: > I don't think there is any need for checking. Just return IgnoreAG.

why?

TaranDahl: I said the wrong thing. I mean it should be to replace this check with `pTarget->WhatAmI() != AbstractType::Cell`.

NetsuNegi: > I said the wrong thing. I mean it should be to replace this check with `pTarget->WhatAmI() != AbstractType::Cell`.

Actually, `pTarget` could also be of `TerrainClass`, and I'm not sure if it should be excluded, as this is something that vanilla can do

PR #2045, opened 2026-01-18
Add low priority deploy filtering

Starkku:
### Low priority for deploy

- You can now set lower priority for TechnoType deploying which means it will be excluded from deploy command if selected together with other units. This will not affect the cursor action which requires no other objects to be selected in the first place.

In `rulesmd.ini`:
```ini
LowDeployPriority=false  ; boolean
```

In `RA2MD.INI`:
```ini
[Phobos]
PriorityDeployFiltering=true  ; boolean
```

### DIFF
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -1811,6 +1811,24 @@ AutoDeath.TechnosExist.Houses=owner            ; Affected House Enumeration (non
 Please notice that if the object is a unit which carries passengers, they will not be released even with the `kill` option **if you are not using Ares 3.0+**.
 ```
 
+### Low priority for deploy
+
+- You can now set lower priority for TechnoType deploying which means it will be excluded from deploy command if selected together with other units. This will not affect the cursor action which requires no other objects to be selected in the first place.
+
+In `rulesmd.ini`:
+```ini
+[SOMETECHNO]             ; TechnoType
+LowDeployPriority=false  ; boolean
+```
+
+- This behavior is designed to be toggleable by users. For now you can only do that externally via client or manually.
+
+In `RA2MD.INI`:
+```ini
+[Phobos]
+PriorityDeployFiltering=true  ; boolean
+```
+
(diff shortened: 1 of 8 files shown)


docs/New-or-Enhanced-Logics.md
```diff
@@ -1811,6 +1811,24 @@ AutoDeath.TechnosExist.Houses=owner            ; Affected House Enumeration (non
 Please notice that if the object is a unit which carries passengers, they will not be released even with the `kill` option **if you are not using Ares 3.0+**.
 ```
 
+### Low priority for deploy
+
+- You can now set lower priority for TechnoType deploying which means it will be excluded from deploy command if selected together with other units. This will not affect the cursor action which requires no other objects to be selected in the first place.
```
TaranDahl, line 1816: I think a similar handling method to the low selection priority should be used.
That is, if units with normal priority and low priority are selected simultaneously, only the units with normal priority will be deployed. If only units with low priority are selected, then they will be deployed.

TaranDahl (changes requested):

docs/New-or-Enhanced-Logics.md
Metadorius: Isn't that what is done?

Starkku: I suppose I can adjust it to work like that. Current behaviour takes a shortcut and disables it for low priority no matter what is selected with it.

Metadorius: > Thats how it works?

Wait, no, I just examined the code. You simply skip deploy at all if more than 1 object is selected. That's not the same. Imagine a group of Desolators only selected, could they be deployed? Or a group of desolators and (say, for MO) eradicators. I think it just has to check whether there are NOT low-priority technos in the mix too, and only then exclude.

Starkku:
Addressed the issue with `LowDeployPriority=true` technos not being able to deploy together.

TaranDahl (approved):

PR #2046, opened 2026-01-18
Audio logging improvements

Starkku:
- Patch out sound buffer size log calls due to them polluting the debug log disproportionately while being of questionable use.
- Output developer warnings for missing audio samples f.ex `VocClass [VO_FuryMove] has missing sample 'vo_regu_mov07'`.

### DIFF
--- a/src/Utilities/Debug.cpp
+++ b/src/Utilities/Debug.cpp
@@ -233,6 +233,10 @@ void Console::PatchLog(DWORD dwAddr, void* fakeFunc, DWORD* pdwRealFunc)
 	VirtualProtect((LPVOID)dwAddr, 5, dwOldFlag, NULL);
 }
 
+// Patch out sound buffer size etc. logging calls.
+DEFINE_JUMP(LJMP, 0x40A55D, 0x40A562);
+DEFINE_JUMP(LJMP, 0x40A5BC, 0x40A5C1);
+
 DEFINE_HOOK(0x7504C9, VocClass_ReadINI_LogMissingSamples, 0x5)
 {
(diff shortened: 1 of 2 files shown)


src/Utilities/Debug.cpp
```diff
@@ -230,3 +232,34 @@ void Console::PatchLog(DWORD dwAddr, void* fakeFunc, DWORD* pdwRealFunc)
 
 	VirtualProtect((LPVOID)dwAddr, 5, dwOldFlag, NULL);
 }
+
+// Patch out sound buffer size etc. logging calls.
+DEFINE_JUMP(LJMP, 0x40A55D, 0x40A562);
+DEFINE_JUMP(LJMP, 0x40A5BC, 0x40A5C1);
```
TaranDahl, line 238: My IDA says this func is never used 🤔

TaranDahl (approved):

src/Utilities/Debug.cpp
Starkku: If you mean `0x4068E0` it is emptied out in the release build of the game, Ares makes it actually print out to the log.

If you mean `0x40A340` then something's wrong on your end.

TaranDahl: It's really strange. Can you tell me who is calling `0x40A340`?

Starkku: `0x402767`. It is some kind of virtual I think.

TaranDahl: The virtual table that references it has been destroyed by my IDA. It's very strange. Thinking it should be able to automatically identify the virtual table.

PR #2048, opened 2026-01-18
Enable playing ingame movie in non-campaign modes

TaranDahl:


TaranDahl:
Needs testing:

- [x] TAction PlayIngameMovie in skirmish
- [x] TAction PlayIngameMovieAndPause in skirmish
- [x] The two TAction in MP

TaranDahl:
image

Metadorius (approved):
The patch looks correct. You need to test in multiplayer though.

Also just to be sure (maybe some other place calls it) - patch 0x5BF2B9 and also see if there are any other similar checks by xrefs to movie statics

Metadorius:
I also give up the bounty.

TaranDahl:
Tested by @Flactine 
https://www.phoboscn.top/t/topic/364/2

TaranDahl:
> Also just to be sure (maybe some other place calls it) - patch 0x5BF2B9 and also see if there are any other similar checks by xrefs to movie statics

Oh, I forgot this. I'm looking into it right now.
My IDA says that the function at 0x5BF2B9 is never called.
I checked the movie array references; there are no similar checks, I think.

Metadorius:
> My IDA says that the function at 0x5BF2B9 is never called.

Same, but some extension may use it. Just to be safe

PR #2050, opened 2026-01-18
[Vanilla Enhancement] Allow draw SuperWeapon timer as percentage

NetsuNegi:
- Superweapon cd timer can now draw as percentage.
- 超级武器的CD倒计时现在可以以百分比形式显示了.

In `rulesmd.ini`:
```ini
[AudioVisual]
SuperWeaponTimer.Percentage=false  ; boolean

[SOMESW]              ; SuperWeaponType, with ShowTimer=yes
ShowTimer.Percentage=     ; boolean
```

### DIFF
--- a/docs/User-Interface.md
+++ b/docs/User-Interface.md
@@ -397,8 +397,11 @@ ShowTimer.Priority=0  ; integer
 
 In `rulesmd.ini`:
 ```ini
+[AudioVisual]
+SuperWeaponTimer.Percentage=false  ; boolean
+
 [SOMESW]              ; SuperWeaponType, with ShowTimer=yes
-ShowTimer.Percentage=false  ; boolean
+ShowTimer.Percentage=     ; boolean
 ```
 
 ### Task subtitles display in the middle of the screen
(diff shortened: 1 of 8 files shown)


docs/User-Interface.md
```diff
@@ -391,6 +391,16 @@ In `rulesmd.ini`:
...
 ```
 
+### Allow draw SuperWeapon timer as percentage
+
+- Superweapon cd timer can now draw as percentage.
+
+In `rulesmd.ini`:
+```ini
+[SOMESW]              ; SuperWeaponType, with ShowTimer=yes
+ShowTimer.Percentage=false  ; boolean
```
TaranDahl: This should be allowed to be defined in General or RA2MD.
Also, I doubt whether it is really necessary to define it per type.

TaranDahl (changes requested):

docs/User-Interface.md
TaranDahl: RA2MD would be better, I think.

Metadorius: there is a Chemical Missile superweapon which @zivdero restored, for which one should always use percentage I think. Correct me if I am wrong.

TaranDahl (approved):

PR #2061, opened 2026-01-25
Technos with Walk locomotor spawn wake like ship

TaranDahl:
- Also, you can now custom whether the techno makes wake when moving.
  - Walk locomotor is able to make wake like ship now.
  - Use `MakesWake` to custom this per type.

In `rulesmd.ini`:
```ini
[AudioVisual]
WalkLocomotorMakesWake=false   ; boolean
HoverLocomotorMakesWake=true   ; boolean
DriveLocomotorMakesWake=true   ; boolean
ShipLocomotorMakesWake=true    ; boolean

[SOMETECHNO]         ; TechnoType
MakesWake=           ; boolean, default to the global value that matches the techno's current locomotor

### CODE REVIEW

File: src/Ext/Techno/Hooks.Misc.cpp [hunk 1]
```diff
@@ -378,6 +378,25 @@ DEFINE_HOOK(0x737F05, UnitClass_ReceiveDamage_SinkingWake, 0x6)
...
+DEFINE_HOOK(0x75AC93, WalkLocomotionClass_Process_Wake, 0x6)
+{
+	if (!RulesExt::Global()->WalkLocomotorMakesWake)
+		return 0;
+
+	GET(ILocomotion* const, pThis, ESI);
+
+	const auto pLinkedTo = static_cast<LocomotionClass*>(pThis)->LinkedTo;
+
+	if (pThis->Is_Moving_Now() && !(Unsorted::CurrentFrame % 10) && !pLinkedTo->OnBridge && pLinkedTo->GetCell()->LandType == LandType::Water)
```
Coronia: I'm not sure, but should it be `Is_Really_Moving_Now`?

Coronia (commented):
not a suggestion but just some thoughts: would it be better to define a global param for locomotors that can trigger wake anim? Something like:
`UseWakeAnim=drive,hover,walk`

while also make a tag to decide whether wake should be displayed for individual techno. Now you can set `Wake` to a dummy anim for that but it's still different from disabling it

TaranDahl:
@Coronia No new enumeration was established because I think only these three loco are meaningful here.

TaranDahl:
image

NetsuNegi (approved):
LGTM

TaranDahl:
Bounty transferred.

PR #2062, opened 2026-01-26
[Minor] Update TechnoType's loading

Coronia:
put many tags into the existing `LoadFromINIByWhatAmI` function to make it won't read a tag that's incapable to use. If there're more tags that could be put in, or shouldn't be put in certain types, please point it out

### DIFF
--- a/src/Ext/TechnoType/Body.cpp
+++ b/src/Ext/TechnoType/Body.cpp
@@ -730,17 +730,6 @@ void TechnoTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 	this->LowSelectionPriority.Read(exINI, pSection, "LowSelectionPriority");
 	this->LowDeployPriority.Read(exINI, pSection, "LowDeployPriority");
 
-	if (pThis->Gunner)
-	{
-		this->WeaponGroupAs.resize(pThis->WeaponCount);
-
-		for (int idx = 0; idx < pThis->WeaponCount; ++idx)
-		{
-			_snprintf_s(tempBuffer, sizeof(tempBuffer), "WeaponGroupAs%d", idx + 1);
-			this->WeaponGroupAs[idx].Read(pINI, pSection, tempBuffer);
-		}
-	}
-
 	this->RadarJamHouses.Read(exINI, pSection, "RadarJamHouses");
 	this->RadarJamDelay.Read(exINI, pSection, "RadarJamDelay");
 	this->RadarJamAffect.Read(exINI, pSection, "RadarJamAffect");
@@ -758,8 +747,6 @@ void TechnoTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 	this->Spawner_UseTurretFacing.Read(exINI, pSection, "Spawner.UseTurretFacing");
 
 	this->Harvester_Counted.Read(exINI, pSection, "Harvester.Counted");
-	if (!this->Harvester_Counted.isset() && pThis->Enslaves)
-		this->Harvester_Counted = true;
 
 	this->Promote_IncludeSpawns.Read(exINI, pSection, "Promote.IncludeSpawns");
 	this->ImmuneToCrit.Read(exINI, pSection, "ImmuneToCrit");
@@ -769,7 +756,6 @@ void TechnoTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 	if (this->InitialStrength.isset())
 		this->InitialStrength = Math::clamp(this->InitialStrength, 1, pThis->Strength);
 
-	this->ReloadInTransport.Read(exINI, pSection, "ReloadInTransport");
 	this->ForbidParallelAIQueues.Read(exINI, pSection, "ForbidParallelAIQueues");
 	this->IgnoreForBaseCenter.Read(exINI, pSection, "IgnoreForBaseCenter");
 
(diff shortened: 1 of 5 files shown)


src/Ext/TechnoType/Body.cpp
```diff
@@ -1394,9 +1219,291 @@ void TechnoTypeExt::ExtData::LoadFromINIByWhatAmI(INI_EX& exINI, const char* pSe
...
+		this->HarvesterScanAfterUnload.Read(exINI, pSection, "HarvesterScanAfterUnload");
+		this->OreGathering_Anims.Read(exINI, pSection, "OreGathering.Anims");
+		this->OreGathering_Tiberiums.Read(exINI, pSection, "OreGathering.Tiberiums");
+		this->OreGathering_FramesPerDir.Read(exINI, pSection, "OreGathering.FramesPerDir");
+
+		this->Passengers_SyncOwner.Read(exINI, pSection, "Passengers.SyncOwner");
+		this->Passengers_SyncOwner_RevertOnExit.Read(exINI, pSection, "Passengers.SyncOwner.RevertOnExit");
+		this->Explodes_KillPassengers.Read(exINI, pSection, "Explodes.KillPassengers");
+		this->Ammo_Shared.Read(exINI, pSection, "Ammo.Shared");
+		this->Ammo_Shared_Group.Read(exINI, pSection, "Ammo.Shared.Group");
```
TaranDahl, line 1240: This should work for both infantry and unit.

TaranDahl (commented):
I'm not sure if it's really necessary. I think the performance of `LoadFromINI` is irrelevant.
Some functions may be adapted to other types in the future. For example, Kerbiter is developing the `OpenTopped` for buildings. This will cause inconvenience for these future developments.
In addition, for those flags that support multiple types, such as `NoQueueUpToEnter`, devs still need to write read operations in more places.

Coronia:
The main point for this is to make it less coupled than performance increment. As of now these tags are a bit clustering and it would be better to keep their job clear

the ideal solution for this might be extending Building/Infantry/Unit/AircraftTypeClass and making them inherit TechnoTypeExt, but I guess that's too much of a revamp to do at this point

TaranDahl:
@Metadorius Come to think of it, why didn't we make Ext inherit like a base class in the beginning?

Metadorius:
Because when we started, we only had Ares architecture and had not much idea of anything else.

And yes, I do think that making a proper hierarchy is something that needs to be done.

ExtData needs to inherit each other, and Ext should inherit the class it's extending I think (without new virtuals or fields).

src/Ext/TechnoType/Body.cpp
```diff
@@ -1358,34 +1134,82 @@ void TechnoTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
...
+
+		this->ParseBurstFLHs(exArtINI, pArtSection, this->DeployedWeaponBurstFLHs, this->EliteDeployedWeaponBurstFLHs, "Deployed");
+		this->ParseBurstFLHs(exArtINI, pArtSection, this->CrouchedWeaponBurstFLHs, this->EliteCrouchedWeaponBurstFLHs, "Prone");
+		this->PronePrimaryFireFLH.Read(exArtINI, pArtSection, "PronePrimaryFireFLH");
+		this->ProneSecondaryFireFLH.Read(exArtINI, pArtSection, "ProneSecondaryFireFLH");
+		this->DeployedPrimaryFireFLH.Read(exArtINI, pArtSection, "DeployedPrimaryFireFLH");
+		this->DeployedSecondaryFireFLH.Read(exArtINI, pArtSection, "DeployedSecondaryFireFLH");
+
+		if (!this->Harvester_Counted.isset() && pThis->Enslaves)
+			this->Harvester_Counted = true;
```
TaranDahl, line 1190: Should be under building and unit I think

src/Ext/TechnoType/Body.cpp
```diff
@@ -1395,9 +1219,291 @@ void TechnoTypeExt::ExtData::LoadFromINIByWhatAmI(INI_EX& exINI, const char* pSe
...
+		this->DeployingAnims.Read(exINI, pSection, "DeployingAnims");
+		this->DeployingAnim_KeepUnitVisible.Read(exINI, pSection, "DeployingAnim.KeepUnitVisible");
+		this->DeployingAnim_ReverseForUndeploy.Read(exINI, pSection, "DeployingAnim.ReverseForUndeploy");
+		this->DeployingAnim_UseUnitDrawer.Read(exINI, pSection, "DeployingAnim.UseUnitDrawer");
+
+		this->Wake.Read(exINI, pSection, "Wake");
+		this->Wake_Grapple.Read(exINI, pSection, "Wake.Grapple");
+		this->Wake_Sinking.Read(exINI, pSection, "Wake.Sinking");
+
+		exINI.ReadSpeed(pSection, "SubterraneanSpeed", &this->SubterraneanSpeed);
```
TaranDahl, line 1288: I don't think this is a type-restricted thing, at most it should be restricted to footclass.

TaranDahl (commented):
> The main point for this is to make it less coupled than performance increment.

Most of the tags work independently. I don't think there is any coupling between them. Are there any examples?

> As of now these tags are a bit clustering and it would be better to keep their job clear

It might make sense, but I'm not sure if it makes enough sense to justify the additional effort of cleaning up. 
I think it's sufficient to place those flags that act on a single type under the corresponding RTTI. Also, we should consider whether it's valuable to expand to multiple types in the future.

Coronia:
no longer needed now that there're standalone Infantry/UnitType Ext

PR #2065, opened 2026-01-27
[Ares Fix] Fix the bug that building with `Explodes=yes` use Ares's rubble logic will cause it's owner cannot defeat normally

NetsuNegi:


TaranDahl:
Is there anyone who can briefly explain why this bug occurred?
I've looked at the code of Ares 0A. What the rubble logic does is generate a new building and place it in its original position.
What makes the building get stuck in the memory?

NetsuNegi:
> Is there anyone who can briefly explain why this bug occurred? I've looked at the code of Ares 0A. What the rubble logic does is generate a new building and place it in its original position. What makes the building get stuck in the memory?



> Is there anyone who can briefly explain why this bug occurred? I've looked at the code of Ares 0A. What the rubble logic does is generate a new building and place it in its original position. What makes the building get stuck in the memory?

I don't know the exact reason, but this is indeed effective, but I can't be sure if there are any side effects

TaranDahl (approved):
As the tester says that it works, I think it is good enough to be merged.
However, since the reason is not clearly known, the situation should be stated in the doc as well as the code comment.

PR #2066, opened 2026-01-27
Return warhead

Coronia:
- For each techno affected by this warhead, a warhead owned by the target techno will be detonated at the owner of the original warhead.
  - `ReturnWarhead` determines the warhead that'll 'return' to the original owner from the target.
  - `ReturnWarhead.Damage` determines the damage dealt by the return warhead.
  - `ReturnWarhead.Chance` determines chance for a return warhead to occur. By default this is checked once when the original warhead is detonated and every target that is susceptible to critical hits will be affected. If `ReturnWarhead.ApplyChancePerTarget` is set, then whether or not the chance roll is successful is determined individually for each target.
  - `ReturnWarhead.FullDetonation` controls whether or not the return warhead is detonated fully on the original owner (as part of a dummy weapon) or simply deals area damage and applies Phobos' Warhead effects.
  - `ReturnWarhead.AffectsTarget` can be used to customize types of targets that original warhead can trigger a return warhead. Critical hits cannot affect empty cells or cells containing only TerrainTypes, overlays etc.
  - `ReturnWarhead.AffectsHouse` can be used to customize houses that original warhead can trigger a return warhead.

In `rulesmd.ini`:
```ini
ReturnWarhead=                              ; WarheadType
ReturnWarhead.Damage=0                      ; integer
ReturnWarhead.Chance=0.0                    ; floating point value, percents or absolute (0.0-1.0)
ReturnWarhead.ApplyChancePerTarget=false    ; boolean
ReturnWarhead.FullDetonation=true           ; boolean
ReturnWarhead.AffectsTarget=all             ; List of Affected Target Enumeration (none|land|water|infantry|units|buildings|all)
ReturnWarhead.AffectsHouse=all              ; List of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
```


TaranDahl:
If you're thinking about something like "life steal", then I think there may need to be more ways to determine damage, for example, determining damage based on a certain percentage of the actual damage.
Also, I think this thing may need to be defined in the AEs of both the firer and the victim.

Coronia:
I would call this a bridge for an opposite direction of interaction between the attacker and the victim. I've thought about what you said as well, but this should be part of another feature that make use of this bridge. No need to bloat this one.

Actually, the key reason I made this is to establish a way to let victim attach AEs and such to attacker. It will be helpful for a series of interaction between AE's invoker and carrier, which I'm working on too. This can also work as a semi 'life steal' which provides healing based on hit targets like you've said. It's just a start and maybe rough as of now indeed, but again I don't think we need to bump them all in a PR.

TaranDahl (approved):

Metadorius:
> If you're thinking about something like "life steal", then I think there may need to be more ways to determine damage, for example, determining damage based on a certain percentage of the actual damage.

...like Transact system 💀

Coronia:
This feature should also make it easier to do a transaction system, as we only need to do it in one direction (from target to attacker), and let the other direction make use of this ReturnWarhead. In this way we can save some workload and reduce tag clustering

PR #2067, opened 2026-01-27
`AllowBerzerkOnAllies`

TaranDahl:
### Berzerk on allies

- In vanilla, `Psychedelic` warheads is hardcoded to ignore allies' target. Now you can turn this off.

In `rulesmd.ini`:
```ini
[CombatDamage]
AllowBerzerkOnAllies=false  ; boolean
```

```{note}
No per-warhead setting because `AffectsAllies` etc. is respected.
```


src/Ext/Techno/Hooks.ReceiveDamage.cpp
```diff
@@ -427,3 +428,8 @@ DEFINE_HOOK(0x5F5480, ObjectClass_ReceiveDamage_FlashDuration, 0x6)
 
 	return SkipGameCode;
 }
+
+DEFINE_HOOK(0x701CE5, TechnoClass_ReceiveDamage_AllowBerzerkOnAllies, 0x7)
```
NetsuNegi: as I known, it will not work when use with Ares. Have you ever test it ?

NetsuNegi: Ares hooked at 0x701BFE, you should check it

TaranDahl: > as I known, it will not work when use with Ares. Have you ever test it ?

No. This feature is requested by a user, so I'm counting on him to come and test.

TaranDahl: Now it should be possible to coexist peacefully with Ares.

src/Ext/Techno/Hooks.ReceiveDamage.cpp
```diff
@@ -427,3 +428,8 @@ DEFINE_HOOK(0x5F5480, ObjectClass_ReceiveDamage_FlashDuration, 0x6)
 
 	return SkipGameCode;
 }
+
+DEFINE_HOOK(0x701CFC, TechnoClass_ReceiveDamage_AllowBerzerkOnAllies, 0x5)
+{
+	return RulesExt::Global()->AllowBerzerkOnAllies ? 0x701D0B : 0;
```
NetsuNegi: maybe it's better to give a name for this address

NetsuNegi (approved):

Coronia:
@TaranDahl there's a problem that's occur when enabling friendly berzerk: it'll trigger an alarm when attacking a friendly miner even if the warhead ratio is 0% (setting AffectsAllies=no can prevent that). I guess this might be an issue of berzerk itself, but could you take a look into it?

TaranDahl:
> @TaranDahl there's a problem that's occur when enabling friendly berzerk: it'll trigger an alarm when attacking a friendly miner even if the warhead ratio is 0% (setting AffectsAllies=no can prevent that). I guess this might be an issue of berzerk itself, but could you take a look into it?

That EVA will be raised if the ReceiveDamage func is called. I suppose that is consistent with other WH with 0% versus.

PR #2069, opened 2026-01-28
Update contributing guide line

TaranDahl:

### DIFF
--- a/docs/Contributing.md
+++ b/docs/Contributing.md
@@ -108,7 +108,7 @@ Those roles are assigned based on the complexity of the PRs and the experience o
 
 ### Suggested reviewing amount
 
-The amount of code that maintainers should review should be roughly the same as the amount of code in the PRs they submit themselves. However, considering that we already have a large backlog of PRs at present, we suggest that maintainers review more code than they submit. Only in this way can we continuously reduce the number of backlogged PRs.
+The amount of code that maintainers should review should be roughly the same as the amount of code in the PRs they submit themselves. However, considering that we already have a large backlog of PRs at present, we suggest that maintainers review more code than they submit (f.ex., 125%-150% of they submit). Only in this way can we continuously reduce the number of backlogged PRs.
 
 At present this is not a hard rule, but we hope that each maintainer can try their best to achieve this.
 
--- a/docs/Contributing.md
+++ b/docs/Contributing.md
@@ -1,12 +1,9 @@
+
 # Contributing
-
-This page describes ways to help or contribute to Phobos and lists the contributing guidelines that are used in the project.
-
-## Ways to help (for everyone)
 
 Engine modding is a complicated process which is pretty hard to pull off, but there are also easier parts which don't require mastering the art of reverse-engineering or becoming a dank magician in C++.
 
-### Research and reverse-engineering
+## Research and reverse-engineering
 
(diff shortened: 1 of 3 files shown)


DeathFishAtEase:
Ignoring the layout, it seems that the actual changes are only these:

```diff
- Project guidelines and policies
+ Project guidelines and policies (for developers and maintainers)

+ 
+ ### Suggested reviewing amount
+ 
+ Considering that we already have a large backlog of PRs at present, we suggest that each maintainer should try to ensure that the amount of code they review is approximately 25% to 50% higher than the amount of code they submit (that needs to be reviewed). Only in this way can we continuously reduce the number of backlogged PRs.
+ This is not a hard and fast rule, but we hope that each maintainer can try their best to achieve this. + The lead will regularly remind maintainers who participate less in code reviews.

- ## Ways to help
+ ## Ways to help (for everyone)
```

However, there is doubt about whether to invert the order of the last two paragraphs, because the description of *Ways to help* is

> Engine modding is a complicated process which is pretty hard to pull off, but there are also easier parts which don't require mastering the art of reverse-engineering or becoming a dank magician in C++.

Therefore, it is placed at the end, indicating that this is apart from the main extension development work, a part that does not require much knowledge mentioned in Project guidelines and policies paragraph (the previous context).

So, I think this order change is unnecessary.

TaranDahl:
The reason why I changed the order is as follows: If I were a user without coding ability, when I read this document from top to bottom, I would lose patience in the instructions for coders at the beginning and then close the page before seeing "Ways to help".

docs/Contributing.md
```diff
@@ -2,7 +2,98 @@
...
+
+```{note}
+Please, provide screenshots, GIFs and videos in their natural size and without excess stuff or length.
+```
+
+### Promoting the work
+
+You can always help us by spreading the word about the project among people, whether you're an influential youtuber, a C&C related community leader or just an average player.
+
+## Project guidelines and policies (for developers and maintainers)
```
Metadorius: How about we make policies another page and just link them in "contributing features" and in ToC?

docs/Contributing.md
```diff
@@ -15,6 +106,12 @@ Due to the size of the project and varying complexity of the codebase, we have e
 
 Those roles are assigned based on the complexity of the PRs and the experience of the contributors. The roles are not fixed and can change based on the contributor's experience and the complexity of the PRs they are working on. Also only the commonly occured roles are established in the list, in case of need - individual permissions can be assigned to contributors by leads (for example, to help with documentation translations).
 
+### Suggested reviewing amount
+
+The amount of code that maintainers should review should be roughly the same as the amount of code in the PRs they submit themselves.
+However, considering that we already have a large backlog of PRs at present, we suggest that maintainers review an additional 25% to 50% of the code. Only in this way can we continuously reduce the number of backlogged PRs.
+This is not a hard and fast rule, but we hope that each maintainer can try their best to achieve this. The lead will regularly remind maintainers who participate less in code reviews.
```
Metadorius:
```suggestion
The amount of code that maintainers should review should be roughly the same as the amount of code in the PRs they submit themselves. However, considering that we already have a large backlog of PRs at present, we suggest that maintainers review more code than they submit. Only in this way can we continuously reduce the number of backlogged PRs.

At present this is not a hard rule, but we hope that each maintainer can try their best to achieve this.
```

TaranDahl: I think it's better to have a quantifiable reference value.

Metadorius: ok, but leave the rest of my amendments then

docs/Contributing.md
TaranDahl: Done.

docs/Project-guidelines-and-policies.md
Metadorius, line 1: This should be added to the TOC

Metadorius (approved):

PR #2071, opened 2026-01-28
Type conversion enhancement

TaranDahl:
- Allow techno conversion working on buildings.
- Fixed the issue that multiple attributes such as cloaking and sensor would not be updated correctly in techno conversion.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -682,6 +682,8 @@ This page lists all the individual contributions to the project by their author.
   - Fix an issue that the AI would set anger towards friendly houses, causing it to act stupidly
   - Fix an issue that the AI would look for the first house in the array as an enemy instead of the nearest one when there were no enemies
   - `AllowBerzerkOnAllies`
+  - Allow techno conversion working on buildings
+  - Fixed the issue that multiple attributes such as cloaking and sensor would not be updated correctly in techno conversion
 - **solar-III (凤九歌)**
   - Target scanning delay customization (documentation)
   - Skip target scanning function calling for unarmed technos (documentation)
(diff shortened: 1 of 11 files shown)


Coronia:
perhaps split the fix and building conversion to different PR. Could even commit the previous one to develop directly if you sure it works

TaranDahl:
Tester confirmed that most of the fixes work.

Refailure:
1.After the conversion of the building, there are residual issues with the TurretAnim and ActiveAnims, specifically with the Shape turret animation. The Voxel turret does not exhibit any residual effects.
2.Performing a conversion on a building that has an upgrade plugin installed will directly trigger an error.
3.Overlapping issue when surrounding available cells are insufficient.

Starkku:
All current building animations should be destroyed and recreated, don't just call `GameDelete`, there is a function for destroying animation in specific slot.

Upgrades should probably be cleared as well unless they are compatible with the new type.

Allowing changes to `Foundation` opens a massive can of worms and if it was possible to evaluate at time of parsing should be disallowed entirely but unfortunately it is not. Would discourage against doing such thing and eventual documentation probably at the very least should as well.

-----------

TaranDahl:
> perhaps split the fix and building conversion to different PR.

Will do after everything is fixed.

Coronia:
any progress on splitting the fix and  feature in this PR? Since it might still need more tweak and test to make building conversion work properly

DeathFishAtEase:
What is the necessity of splitting lambda in b0d2ef38faf48d301374c2b7ba7c7063af5dbb31, which causes a problem in `if (pToType->BalloonHover && pToType->DeployToLand && prevType->Locomotor != jjLoco && toLoco == jjLoco)`.

TaranDahl:
> What is the necessity of splitting lambda in https://github.com/Phobos-developers/Phobos/commit/b0d2ef38faf48d301374c2b7ba7c7063af5dbb31

For clarity

TaranDahl:
TODO:

- [ ] Split into PreProcess and PostProcess
- [ ] Add a registration callback mechanism for interaction with other engine extensions
- [ ] Complete necessary processing of Ares attributes
- [ ] Fix issues found in BunkerGeneral tests
- [ ] Fix issues found in Flactine tests

TaranDahl:
Closing this because I can not make it with a bug-free promise.

PR #2072, opened 2026-01-29
Vanilla Fix for Vehicle salesmen

CnCRAZER:
Are you tired of people abusing the sell mechanic on tanks when they're not supposed to? 

Are you tired of constantly being harassed by streamers abusing this bug? 

Look no further!!! 

With this pull request the functionality is still enabled but it's enforced to sell any vehicletype for 0 credits!

https://github.com/user-attachments/assets/ba297895-d672-4a7b-8182-878cf8195daf

Imagine the video above but the user gets 0 credits!! That's what happens when this PR gets accepted. Also free donuts. Don't ask. Just merge.

https://youtu.be/uPEFi4cte4U

### DIFF
--- a/src/Ext/Building/Hooks.Selling.cpp
+++ b/src/Ext/Building/Hooks.Selling.cpp
@@ -1,6 +1,7 @@
 #include "Body.h"
 
 #include <Misc/FlyingStrings.h>
+#include <Ext/Rules/Body.h>
 
 // SellSound and EVA dehardcode
 DEFINE_HOOK(0x4D9F7B, FootClass_Sell, 0x6)
@@ -8,7 +9,19 @@ DEFINE_HOOK(0x4D9F7B, FootClass_Sell, 0x6)
 	enum { ReadyToVanish = 0x4D9FCB };
 	GET(FootClass*, pThis, ESI);
 
-	const int money = pThis->GetRefund();
+	// Check if this is a vehicle and UnitsUnsellable is enabled
+	// If so, set refund to 0 (units being sold normally, not grinded)
+	int money = pThis->GetRefund();
+	
+	if (RulesExt::Global()->UnitsUnsellable.Get())
+	{
+		// Only apply to vehicles (not buildings, not infantry)
+		if (auto pUnit = abstract_cast<UnitClass*>(pThis))
+		{
+			money = 0;
+		}
+	}
+	
 	const auto pOwner = pThis->Owner;
 	pOwner->GiveMoney(money);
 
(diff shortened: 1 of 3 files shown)


src/Ext/Building/Hooks.Selling.cpp
```diff
@@ -1,14 +1,27 @@
 #include "Body.h"
 
 #include <Misc/FlyingStrings.h>
+#include <Ext/Rules/Body.h>
 
 // SellSound and EVA dehardcode
 DEFINE_HOOK(0x4D9F7B, FootClass_Sell, 0x6)
```
TaranDahl, line 7: Thinking that this should be solved at `0x4C6F5A`

TaranDahl (changes requested):

src/Ext/Building/Hooks.Selling.cpp
CnCRAZER: That was already there when I got here 🤷

TaranDahl: I mean your checking should be placed there. Not the hook.

TaranDahl: You should make a new hook there and check for related flags.

CnCRAZER: https://youtu.be/uPEFi4cte4U

Would changing the hook make it better somehow? It works at this one. Can change if necessary I guess

TaranDahl: Yeah I know it works, but this issue should not be solved like that.

TaranDahl: The root of the problem lies in the fact that the engine lacks a legality check for incoming events. The appropriate solution is to supplement such a check. 
Regarding the issue of selling units, before executing the event, it should be checked whether the unit is located on a building that allows unit sales (there is a flag, but I don't remember exactly what it is, you have to check it yourself), rather than allowing sales as long as there is a building under the unit.

TaranDahl: It is best to directly copy the decision - making method of Mouse Action.

TaranDahl: Oh, I checked the code, there should be no way to sell unit, except using a SW with Action=Sell.

TaranDahl: Stupid that such SW will not be fired. Thus we could never correct it unless we rewrite it. We should use an alternative now.
I suggest the following handling:
1. Reposition the checks around 0x4C6F5A
2. If the flag is true, then jump out if the building check failed.
image

PR #2073, opened 2026-01-29
Fix the bug where selected technos would lose their selection if their regular mind control was replaced with permanent mind control or with the control from the Psychic Dominator superweapon

NetsuNegi:
- In vanilla, if you use psychic dominator superweapon to control a techno which has beed controlled in yours, the techno will always deselect. Now it should keep select (if it is selected before you use superweapon).
- 在原版游戏中，如果你使用心灵控制超级武器去控制一个已经被你控制的科技类型，那这个科技类型总是会被取消选择。现在它应当会保持选中了（如果在使用超级武器前它已经被选中）.
- It is also same of Ares's `MindControl.Permanent`.
- 这对于Ares的 `MindControl.Permanent` 也有同样的问题和修复方案.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -487,6 +487,7 @@ This page lists all the individual contributions to the project by their author.
   - Fix the issue that warhead with `IsLocomotor=yes` can be used to vehicles who is in tank bunker
   - Fix the bug that if there's a tank in tank bunker, tank bunker use auto death by vanish will cause tank get stuck and game will crash quickly
   - Customize whether warhead can be used to targeting ironcurtained technos or not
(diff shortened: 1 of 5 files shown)


TaranDahl:
Previously, modifying Ares mostly involved re-implementing or fixing bugs. 
I am uncertain whether we should modify it for new features.
@Metadorius may need your opinion.

Metadorius:
no opinion, at least yet

also, the English docs seriously need more work, the spelling is pretty broken, sadly

TaranDahl (approved):
LGTM, except the doc Kerbiter addressed.

PR #2075, opened 2026-01-30
Fixed an issue that retaliation will make the unit keep switching among multiple targets with the same amount of threat

TaranDahl:


Coronia (approved):
codes look fine but not sure how to test it

TaranDahl:
> how to test it

1. Set all the threat xxxxcoefficient to zero.
2. Gather a group of E1 to attack one AI HTNK.

In vanilla the HTNK will continuously rotate its turret and fire at different targets.
With this fix, it should fire at one target to death.

mevitar:
I've noticed this issue when testing delayed fire logic, as AI units kept resetting their targets everytime they got shot at, and thus be unable to actually attack anything when under constant fire. Same problem happened with FireUp=, for infantry or vehicles.

It is indeed fixed now.

PR #2078, opened 2026-02-03
Allow disable an over-optimization in targeting

TaranDahl:
### Allow disable an over-optimization in targeting

- In vanilla, there is an optimization in targeting: if a unit finds a valid target within the 1/4 or 1/2 range, it will stop looking for other targets. Now you can disable it.
  - This optimization has a negligible effect on average performance, as most targeting calls fail to find a valid target.
  - At the same time, it can affect the gaming experience, as it will make units attack nearby targets while ignoring more threatening targets that are farther away.

In `rulesmd.ini`:
```ini
[General]
DisableOveroptimizationInTargeting=false  ; boolean
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -734,6 +734,7 @@ This page lists all the individual contributions to the project by their author.
   - Taunt warhead
   - Fix the bug where non-Teleporter miners would not return to work after minerals are depleted and then regenerated
   - Miners back to work when ore regenerated
+  - Allow disable an over-optimization in targeting
 - **solar-III (凤九歌)**
   - Target scanning delay customization (documentation)
   - Skip target scanning function calling for unarmed technos (documentation)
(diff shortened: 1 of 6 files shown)


TaranDahl:
If possible, it would be best to conduct a benchmark, that is, to see to what extent the performance would be affected if this optimization is disabled. 
However, I think this is optional.

Coronia:
instead of disabling, what if we can modify the range of this? Now it's forced to 1/2 and 1/4, could make it a customizable ratio or even a fixed range, so it can be further adjusted if someone want better performance or better experience

TaranDahl:
> instead of disabling, what if we can modify the range of this? Now it's forced to 1/2 and 1/4, could make it a customizable ratio or even a fixed range, so it can be further adjusted if someone want better performance or better experience

I doubt whether this "optimization" needs further improvement.
I believe that no performance optimization should come at the cost of significantly affecting the gaming experience. This optimization has made the units dumber.
In fact, I wanted to directly disable it without leaving a switch, treating it as a bug fix. However, I'm not sure if this would significantly reduce the game performance, so I didn't do it.

TaranDahl:
Confirm that this fix can correctly restore the enemy detection priority check.

TaranDahl:
image

NetsuNegi (approved):

TaranDahl:
I couldn't decide who to give the bounty to, so I sent 10 CNY to each person.

PR #2082, opened 2026-02-04
Customize several behavior of mission area guard

TaranDahl:
### Customize several behavior of mission area guard

- In vanilla, technos in area guard mission will use the focus as the center of targeting and stray.
  - Now you can change the center to itself by using `AreaGuard.UseSelfAsCenter=true`. (For stray, what is checked is the distance between the techno and the current target)
- In vanilla, technos in area guard mission will do targeting in an extended range (roughly 2 times of the guard range).
  - Now you can change the range to the guard range by using `AreaGuard.TargetingInRange=true`.
- In vanilla, technos in area guard mission will not stray unless it has no destination.
  - Now you can change the behavior by using `AreaGuard.StrayIgnoreDestination=true`.

In `rulesmd.ini`:
```ini
[General]
AreaGuard.UseSelfAsCenter=         ; boolean
AreaGuard.TargetingInRange=        ; boolean
AreaGuard.StrayIgnoreDestination=  ; boolean
```

```{note}
In vanilla, if `GuardRange` is not set, then the guard range will be automatically deduced based on the weapon range.
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -689,6 +689,7 @@ This page lists all the individual contributions to the project by their author.
   - Fixed an issue that retaliation will make the unit keep switching among multiple targets with the same amount of threat
   - Fix an issue where units recruited by a team with `AreTeamMembersRecruitable=false` cannot be recruited even if they have been liberated by that team
   - Global default value for `DefaultToGuardArea`
+  - Customize several behavior of mission area guard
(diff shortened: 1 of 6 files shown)


TaranDahl:
image

TaranDahl:
Closing this because no maintainer cares about it.

PR #2083, opened 2026-02-04
Add amount limit of `LimboKill`

NetsuNegi:
`LimboKill.Counts` sets how many buildings of each type will be removed. Value from position matching the position from `LimboKill.IDs` is used if found, or no limitation if not found. If list is empty, there's no limitation for all types.

In `rulesmd.ini`:
```ini
[SOMESW]                        ; SuperWeaponType
LimboKill.Counts=               ; List of counts
```


TaranDahl (approved):

TaranDahl:
Tested by 
QQ_1770306560816

PR #2084, opened 2026-02-04
Fixed the issue that the Jumpjet must end its movement before starting the next mission

TaranDahl:
- Fixed the issue that the Jumpjet must end its movement before starting the next mission.


TaranDahl:
~~It can be confirmed that it has fixed the issue where JJ cannot correctly execute AttackMove.~~
It can be confirmed that it fixes the issue where JJ fails to correctly switch the mission to Attack when a target is found while executing AttackMove.
However, whether there are other issues remains to be tested.

TaranDahl:
image

Coronia (approved):

TaranDahl:
@Coronia The bounty was transferred to your Alipay account a few hours ago.

PR #2085, opened 2026-02-04
Fixed the issue that the time for units in the area guard mission to reacquire targets after eliminating the target is significantly longer than that in other missions

TaranDahl:


TaranDahl:
image

Coronia:
not sure if this is really an issue. Looks like Westwood did it in this way deliberately. but idk their reason nor if there's any side effect of changing this

TaranDahl:
> not sure if this is really an issue. Looks like Westwood did it in this way deliberately. 

I'm more inclined to think that the code for the AreaGuard part was written by another coder. Due to unfamiliarity with the original enemy-detection process, he reinvented the wheel and also introduced new problems.

> but idk their reason nor if there's any side effect of changing this

I see no side effects so far. Mission AreaGuard seemingly only changes target, destination, etc.

TaranDahl:
> not sure if this is really an issue. Looks like Westwood did it in this way deliberately.

I checked the code of RA1.
In RA1, all target searching operations are completed within Missions (Move, Guard, Hunt, AreaGuard).
In RA2, the target searching functions of Move and Guard have both been integrated into the unit's own Update and managed by independent timers.
I believe they simply overlooked AreaGuard.

Coronia (approved):

TaranDahl:
Bonuty transferred.

PR #2086, opened 2026-02-04
Attached effect for transfering damage to invoker

Coronia:
Attached effect between object and invoker - Season 1

  - `TransferDamage` can be set to true to have any damage dealt to the object the effect is attached to be transfered to the invoker of the effect if it's alive.
    - `TransferDamage.SelfMultiplier` and `TransferDamage.InvokerMultiplier` are multuipliers of damage that'll be applied to the attached object and the invoker if a transfer happens. If an object has multiple effects with `TransferDamage`, the damage will be equally divided and calculated for each effect independently. `TransferDamage.Minimum` and `TransferDamage.Maximum` further restrict the bound of damage transfering to the invoker.
    - `TransferDamage.Warhead` determines which Warhead is used to deal the damage, defaults to `[CombatDamage] -> C4Warhead`. If `TransferDamage.Warhead.Detonate` is set to true, the Warhead is fully detonated instead of used to simply deal damage. If `TransferDamage.UseOriginalWarhead` is set to true, it'll reuse the Warhead that dealt damage to the object.
    - `TransferDamage.Chance` determines the chance of transfer, while `TransferDamage.AffectsHouse` customizes which houses can trigger the transfer damage. `TransferDamage.Invoker.AbovePercent` and `TransferDamage.Invoker.BelowPercent` determine if the transfer can happen when the invoker's health is above or below the given values.
    - Warheads can prevent transfer damage from occuring by setting `SuppressTransferDamage` to true. `SuppressTransferDamage.Types` can control which AttachEffectTypes' transfer damage is suppressed, if none are listed then all of them are suppressed. `SuppressTransferDamage.Groups` does the same thing but for all AttachEffectTypes in the listed groups.
    - `SuppressTransferDamage.SelfOwned` can be used to set the house and TechnoType that the object the effect is attached to as the transfered damage's owner & invoker. Otherwise, it'll be set to the original damage dealer that trigger this transfer.

In `rulesmd.ini`:
```ini
ReflectDamage.UseOriginalWarhead=false ; boolean
TransferDamage=false                               ; boolean
TransferDamage.SelfMultiplier=0.5                  ; floating point value
TransferDamage.InvokerMultiplier=0.5               ; floating point value
TransferDamage.Minimum=-2147483648                 ; integer
TransferDamage.Maximum=2147483647                  ; integer
TransferDamage.Warhead=                            ; WarheadType
TransferDamage.Warhead.Detonate=false              ; WarheadType
TransferDamage.UseOriginalWarhead=false            ; boolean
TransferDamage.AffectsHouse=all                    ; List of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
TransferDamage.Chance=1.0                          ; floating point value
TransferDamage.Invoker.AbovePercent=0.0            ; floating point value
TransferDamage.Invoker.BelowPercent=1.0            ; floating point value
TransferDamage.SelfOwned=false                     ; boolean

[SOMEWARHEAD]                                      ; WarheadType
SuppressTransferDamage=false                       ; boolean
SuppressTransferDamage.Types=                      ; List of AttachEffectTypes
SuppressTransferDamage.Groups=                     ; comma-separated list of strings (group IDs)
```

### DIFF
--- a/src/Ext/Techno/Hooks.ReceiveDamage.cpp
+++ b/src/Ext/Techno/Hooks.ReceiveDamage.cpp
@@ -16,7 +16,8 @@ DEFINE_HOOK(0x701900, TechnoClass_ReceiveDamage_Shield, 0x6)
 	GET(TechnoClass*, pThis, ECX);
 	LEA_STACK(args_ReceiveDamage*, args, 0x4);
 
-	const auto pWHExt = WarheadTypeExt::ExtMap.Find(args->WH);
+	const auto pWarhead = args->WH;
+	const auto pWHExt = WarheadTypeExt::ExtMap.Find(pWarhead);
 	int& damage = *args->Damage;
 
 	// AffectsAbove/BelowPercent & AffectsNeutral can ignore IgnoreDefenses like AffectsAllies/Enmies/Owner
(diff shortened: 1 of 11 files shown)


src/Ext/Techno/Hooks.ReceiveDamage.cpp
```diff
@@ -128,11 +130,93 @@ DEFINE_HOOK(0x701900, TechnoClass_ReceiveDamage_Shield, 0x6)
 		raiseCombatAlert();
 	}
 
-	// Shield Receive Damage
+	// Transfer Damage and Shield Receive Damage
 	if (!args->IgnoreDefenses)
 	{
 		int nDamageLeft = damage;
 
+		if (pExt->AE.TransferDamageCount > 0 && !pWHExt->Transfered)
```
TaranDahl, line 138: It is better to write a comment to explain that checking Transfered is to prevent circular calls.

TaranDahl (approved):
The damage trigger, the pre-damage trigger. Things are getting better and better XD.
I bet someone will encounter problems caused by the settlement order in the future.
Maybe it's time to add an overall explanation of the settlement order.

Metadorius:
Looking at all those coming PRs, it feels more and more like we should design a generic and flexible system instead of adding 123946123874238 different systems. It's just not feasible in the long run. We can add endless amount of things this way, because there will be endless amount of ideas to implement.

TaranDahl:
> Looking at all those coming PRs

A large part of them are my fixes to vanilla bullshit 😝

Metadorius:
> A large part of them are my fixes to vanilla bullshit 😝

I know. My point was more about @Coronia's additions like this one and the weapon one.

TaranDahl:
Add the comments mentioned above, and then we're good to go.

PR #2087, opened 2026-02-04
Extended auto-targeting

TaranDahl:
### Extended auto-targeting

- Now you can activate multiple auto-targeting optimizations by setting `ExtendedAutoTargeting=true`. These include:
  - When using stop, guard, or mouse commands, refresh the auto-targeting cooldown.
  - When the target becomes invalid, refresh the auto-targeting cooldown (In vanilla, this cooldown will be reduced to less than 10 frames).
  - When a unit is on a task that allows auto-targeting but already has a target, it will still auto-target. If a target with a threat level higher than the original target by more than `ExtendedAutoTargeting.SwitchTargetThreshold` is found, switch the target.

In `rulesmd.ini`:
```ini
[General]
ExtendedAutoTargeting=false                        ; boolean
ExtendedAutoTargeting.SwitchTargetThreshold=1000   ; integer
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -749,6 +749,7 @@ This page lists all the individual contributions to the project by their author.
   - Allow replacing vanilla repairing with togglable auto repairing
   - Fix an issue that the time for units in the area guard mission to reacquire targets after eliminating the target is significantly longer than that in other missions
   - Framework for dynamic sight
+  - Extended auto-targeting
 - **solar-III (凤九歌)**
   - Target scanning delay customization (documentation)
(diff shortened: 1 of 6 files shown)


TaranDahl:
Prerequisite: https://github.com/Phobos-developers/Phobos/pull/2078
The new feature of auto-targeting when having a target will not take effect unless this fix is enabled.

TaranDahl:
Demonstration of one of the features:

Vanilla:
对照组

ExtendedAutoTargeting=yes
实验组

Unit did auto-targeting and changed target.
换目标

TaranDahl:
image

PR #2081, opened 2026-02-04
Customize the keep-target-range of attackmove

TaranDahl:
### Customize the keep-target-range of attackmove

- In vanilla, the range of keep target in attackmove is determined by the range of weapon, or the range of the secondary weapon with `NeverUse=yes`.
- Now you can directly specify this distance by setting the following flag to a non-negative value.
  - `AttackMove.KeepTargetRange` is used to directly specify this distance. It takes effect when this flag is not less than 0.
  - `AttackMove.KeepTargetRangeMultiplier` and `AttackMove.KeepTargetRangeAddend` are used to automatically calculate this distance based on the guard range. It takes effect when the multiplier is not less than 0.
    - The final distance is guard range * multiplier + addend.

In `rulesmd.ini`:
```ini
[General]
AttackMove.KeepTargetRangeMultiplier=-1.0   ; float
AttackMove.KeepTargetRangeAddend=0.0        ; float, range in cell

[SOMETECHNO]                                ; TechnoType
AttackMove.KeepTargetRange=-1.0             ; float, range in cell
AttackMove.KeepTargetRangeMultiplier=       ; float
AttackMove.KeepTargetRangeAddend=           ; float, range in cell
```

PR #2088, opened 2026-02-04
Reimplement the code reverted in 35f5f3eb3758dc09d3a0ae3bffccd033245abe51

TaranDahl:
### Stop immediately if the target enters the range during ApproachTarget

- In vanilla, the ApproachTarget will simply exit and do nothing if the target is in range. This will cause your units to approach the target unnecessarily.
  - Now you can change this behavior by the following flag.
  - The setting can be customized per techno type; if not set on a type, the global setting is used as the fallback.
  - `AttackMove.StopWhenTargetAcquired` is read as a compatibility alias for `ApproachTarget.StopWhenInRange` (in both `[General]` and techno type sections).

In `rulesmd.ini`:
```ini
[General]
ApproachTarget.StopWhenInRange=false  ; boolean

[SOMETECHNO]
ApproachTarget.StopWhenInRange=       ; boolean, default to the same key in [General]
```

### Keep pursuing the target during ApproachTarget

- Now you can make a unit approach its target with the target itself as the destination, just like ZEP, instead of a point on the arc around the target whose radius is the unit's own weapon range.
  - If `ApproachTarget.StopWhenInRange` is set to `true`, the unit will stop as soon as the target enters its range and will not chase it further.
  - `AttackMove.PursuitTarget` is read as a compatibility alias for `ApproachTarget.PursuitTarget`.

In `rulesmd.ini`:
```ini
[SOMETECHNO]                   ; TechnoType
ApproachTarget.PursuitTarget=  ; boolean, default to false
```


TaranDahl:
The `AttackMove.StopWhenTargetAcquired` has been deprecated because it was implemented in an incorrect way. It should be `ApproachTarget` instead of `AttackMove` that needs to be modified. It has now been replaced by `ApproachTarget.StopWhenInRange`. The prerequisite for it to take full effect is https://github.com/Phobos-developers/Phobos/pull/2084.

The `AttackMove.PursuitTarget` has been deprecated because it was implemented in an incorrect way. It should be `InAttackMoveKeepRange` (implemented by https://github.com/Phobos-developers/Phobos/pull/2081) and `ApproachTarget` (implemented by https://github.com/Phobos-developers/Phobos/pull/1998) that need to be modified.

Coronia:
given there were quite a lot of controversial for our previous tag renaming (such as new Trajectory and the recent major renaming), I think we should still stick to the principal of no tag name change unless necessary. For this PR I think we don't have to depreciate those AttackMove tag, could just redirect their implementaion to the new one.

TaranDahl:
> deprecated because it was implemented in an incorrect way

I don't think we should keep something that is incorrect. This is not a simple tag renaming. They are removed because the function itself is incorrect.

TaranDahl:
image

docs/Whats-New.md
```diff
@@ -597,10 +590,7 @@ Vanilla fixes:
 - Fixed the bug where selected technos would lose their selection if their regular mind control was replaced with permanent mind control or with the control from the Psychic Dominator superweapon (by NetsuNegi)
 - Fixed an issue that retaliation will make the unit keep switching among multiple targets with the same amount of threat (by TaranDahl)
 - Fixed the issue where units recruited by a team with `AreTeamMembersRecruitable=false` cannot be recruited even if they have been liberated by that team (by TaranDahl)
-- Fixed the bug that cause technos teleport to cell 0,0 by ChronoSphere superweapon (by NetsuNegi)
-- Fixed the bug that techno in attack move will move to target if it cannot attack it (by NetsuNegi)
-- Fixed the bug in AI scripts 56 and 57 that forced the launch of superweapons with index numbers 3 and 4 (by FlyStar)
-- Buildings with `NeedsEngineer=true` are now considered to have threat value of 0 under ownership of `MultiplayPassive=true` houses regardless of their `ThreatPosed` value (by Starkku)
+- Allow the unit to stop immediately if the target enters the range during ApproachTarget (by TaranDahl)
```
DeathFishAtEase:
```suggestion
```

docs/Whats-New.md
```diff
@@ -524,24 +524,17 @@ New:
...
 - Option to scale `PowerSurplus` setting if enabled to current power drain with `PowerSurplus.ScaleToDrainAmount` (by Starkku)
 - Global default value for `DefaultToGuardArea` (by TaranDahl)
 - [Weapon range finding in cylinder](New-or-Enhanced-Logics.md#range-finding-in-cylinder) (by TaranDahl)
-- [Penetrates damage on transporter](New-or-Enhanced-Logics.md#penetrates-damage-on-transporter) (by NetsuNegi)
-- Added amount limit of `LimboKill` (by NetsuNegi)
-- [Customizations for techno type target scan/guard range](Fixed-or-Improved-Logics.md#target-scan-guard-range-customizations) (by Starkku)
-- [Spawns particle when spawns tiberium by terrain](Fixed-or-Improved-Logics.md#customizable-ore-spawners) (by NetsuNegi)
-- Allow jumpjet climbing ignore building height (by TaranDahl)
-- [Allow draw SuperWeapon timer as percentage](User-Interface.md#allow-draw-superweapon-timer-as-percentage) (by NetsuNegi)
-- Customize particle system of parasite logic (by NetsuNegi)
```
DeathFishAtEase, line 666:
```suggestion
- [Allow the unit to stop immediately if the target enters the range during ApproachTarget](Fixed-or-Improved-Logics.md#stop-immediately-if-the-target-enters-the-range-during-approachtarget) (by TaranDahl)

```

DeathFishAtEase (commented):
Changelog entry position error

DeathFishAtEase:
I see that it has already added the Tested label and is not a draft, so should this PR already be in a completed state?

> I don't think we should keep something that is incorrect. This is not a simple tag renaming. They are removed because the function itself is incorrect.

The documentation for the discarded INI flag and its related content does not seem to have been removed yet, which should also be part of the pending work for this PR, because I see that they have already been removed in the changes to the source code.

src/Utilities/TemplateDef.h
```diff
@@ -59,6 +59,7 @@
 #include <ParticleTypeClass.h>
 #include <CRT.h>
 #include <LocomotionClass.h>
+#include <JumpjetLocomotionClass.h>
```
DeathFishAtEase: Is this a remnant of some deprecated content?

YRpp
DeathFishAtEase, line 1: I don't understand the significance of this YRpp change; where is it used in this PR?

docs/Fixed-or-Improved-Logics.md
```diff
@@ -2162,6 +2162,17 @@ In `rulesmd.ini`:
...
 ```
 
+### Stop immediately if the target enters the range during ApproachTarget
+
+- In vanilla, the ApproachTarget will simply exit and do nothing if the target is in range. This will cause your units to approach the target unnecessarily.
+- Now you can change this behavior by the following flag.
+
+In `rulesmd.ini`:
+```ini
+[General]
```
DeathFishAtEase: I feel a little weird:

- if this behavior is wrong, should it be treated as a bugfix and have the change applied fixedly, leaving only a global flag to allow reverting the change?
- Currently it is in the form of a new feature, leaving the transition from unfixed to fixed for the user to enable.

The reason for this design is best explained in a Note block in the documentation, for example, what consequences enabling the fix may cause that might be unexpected.

docs/Whats-New.md
DeathFishAtEase, line 1: Since it removes two statements that were included in DevBuild#48 versions, it is best to explain this in the *Changes compared to inter-version builds / pre-releases* section on lines 61-68 of `Whats-New.md`.

docs/Fixed-or-Improved-Logics.md
TaranDahl: I don't think there will be any unintended effects.
However there are indeed some people who love every piece of shit from Westwood, so I left a toggle here.

src/Ext/Techno/Hooks.Firing.cpp
```diff
@@ -1248,3 +1248,17 @@ DEFINE_HOOK(0x737086, UnitClass_FiringAI_Gattling, 0x9)
...
+
+DEFINE_HOOK(0x4D5A34, FootClass_ApproachTarget_StopWhenInRange, 0x6)
+{
+	GET_STACK(const bool, closeEnough, STACK_OFFSET(0x158, -0x146));
+
+	if (RulesExt::Global()->ApproachTarget_StopWhenInRange && closeEnough)
+	{
+		GET(FootClass*, pThis, EBX);
+		pThis->StopMoving();
+		pThis->AbortMotion();
```
TaranDahl: Special care needs to be taken in handling this.
Any constructive review comments are welcome.

Coronia: looks fine to me to add AbortMotion here, also the previous AttackMove.StopWhenTargetAcquired has it too so the behavior should be consistent

btw, should InfantryClass_SetDestination_JJInfFix and UnitClass_SetDestination_JJVehFix use it too? Give they're also about stopping when target is in range

TaranDahl: Further investigation is required. It is expected to be completed within one week.

Coronia:
seems not working as great as before. Sometime the unit won't stop when facing a target in attack move
image

TaranDahl:
Can you post a gif?

src/Ext/Techno/Hooks.Firing.cpp
TaranDahl: I believe that with ApproachTarget.StopWhenInRange=true, UnitClass_SetDestination_JJVehFix is completely unnecessary to keep existing.
For InfantryClass_SetDestination_JJInfFix, ideally, yes. However, this involves more shit in the Mission system, and fixing them requires further investigation. Given the recent voting results, it is clear that no one wants to bear the cost of this investigation, so we had better just leave that shit as‑is.

fagonghaiwo:
bandicam 2026-08-07 18-30-06-340

old
[General]
;ApproachTarget.StopWhenInRange=1
AttackMove.StopWhenTargetAcquired=1


bandicam 2026-08-07 18-32-45-089

new
[General]
ApproachTarget.StopWhenInRange=1
;AttackMove.StopWhenTargetAcquired=1

doesnt work

TaranDahl:
image

This is not an issue with this feature. It is a problem with task switching. Hover has a similar issue to Jumpjet: their IsMovingNow is handled incorrectly, which causes incorrect evaluation of ReadyToNextMission. Even when AttackMove detects a target and queues an Attack mission, it cannot switch over.

Optional solutions:
1. Simply revert to the old code. I do not know why, but since it works in practice, it works. However, the old implementation is essentially piling more garbage onto a garbage heap, because the root cause of the failure to stop properly comes from ApproachTarget rather than AttackMove. Other tasks that need to call ApproachTarget will still be problematic.
2. Exclude Hover in FootClass_ReadyToNextMission_MovingCheck. In fact, I currently believe that the IsMovingNow check is unnecessary for any Locomotor. Side‑effects are unknown, though.
3. (Recommended) Ignore the Hover shit, push forward the merge of PR2042, and then replace all Hover instances with AdvancedDrive.Hover.

TaranDahl:
@Coronia The choice is in your hands.

Coronia:
First of all, I don't think the problem should be left unresolved even if AdvancedDrive is added. Hover is something that every mod has been used for age so it's not even feasible to replace them in a wide scale. Whatever AdvancedDrive did shouldn't be considered in the current topic

As of the problem itself, I think tweaking FootClass_ReadyToNextMission_MovingCheck _might be_ an acceptable solution since this is the root cause. But if we want to avoid potential side effect, maybe we could just do some special handling to Hover for ApproachTarget's case alone? In worst case we can just reuse the previous codes for Hover alone since it has proven to work

In any case, if we can't find a proper solution then I would still suggest reverting the old implementation for now

Coronia:
so far seems to be fine in testing, perhaps @Metadorius could take a look before the merge

Coronia:
there's still one remaining thing to optimize: using locomotion_cast to merely check if it's jumpjet or hover type is a little bit costy. I'll make a variable to cache it for such scenario, but that's not part of this PR anyway. Merge it first so that I can proceed the latter work

Coronia (approved):

PR #2089, opened 2026-02-05
[Vanilla Fix] Fix the bug that techno in attack move will move to target if it cannot attack it

NetsuNegi:


TaranDahl:
By "attack move", do you mean MegaMission == AttackMove?

NetsuNegi:
> By "attack move", do you mean MegaMission == AttackMove?

Ctrl+Shift

TaranDahl:
So what will it do now? Just stay and do nothing?

NetsuNegi:
> So what will it do now? Just stay and do nothing?

Current behavior is similar to a forced attack, which involves walking within range and then stopping

TaranDahl (approved):

TaranDahl:
This causes the cursor to issue an attack command when clicking on a selected friendly unit.

TaranDahl:
I don't think modifying MouseOverObject is as simple as I thought before. Because some positions that I think have nothing to do with the mouse (such as 0x4DC9A1) are calling it. Since this vfunc is in a very front position, I can't quickly locate whether there are other similar calls through searching.
I recommend to roll back this modification temporarily, or carefully investigate whether this modification will cause problems in those positions.

TaranDahl:
This also causes the use of NoMove to click on the tree (and Mirage Tank) to issue an attack command. This may affect the gameplay.

PR #2090, opened 2026-02-05
Separation of AutoTarget for `DeployFireWeapon`, `OpenTransportWeapon`, and `NoAmmoWeapon`

Fly-Star-him:
- When `DeployFireWeapon`, `OpenTransportWeapon`, and `NoAmmoWeapon` are set and available, their AutoTarget will only evaluate the current weapon's damage, similar to IFVs, rather than calculating the average damage from all weapons.
- 当`DeployFireWeapon`、`OpenTransportWeapon`和`NoAmmoWeapon`被设置且可用时，其自动索敌功能仅评估当前武器的伤害值（类似于步兵战车）而非计算所有武器的平均伤害值。


Fly-Star-him:
The scope of application for this PR is extremely limited; I honestly can't think of any practical scenario where it would be useful.

NetsuNegi (approved):

Fly-Star-him:
- Considering that direct fixes might affect the original behavior, I decided to add a new toggle option.
- 考虑到直接修复可能会影响到原本的行为，因此我决定新增一个开关选项。


---------

- In Vanilla, when `IsGattling=no` and `TurretCount` is less than or equal to 0, units will only repair/heal targets if the following condition is met: the total `Damage + AmbientDamage` of all weapons divided by the total number of weapons is less than 0. This may cause units equipped with both healing and lethal weapons to fail to properly execute `DeployFireWeapon`, `OpenTransportWeapon`, and `NoAmmoWeapon`. Now you can use a new flag `SeparateWeaponTypes` to overcome this problem.
  - When using a separate auto-aiming device, the effective range will use the current weapon's range rather than the unit's maximum weapon range, `GuardRange` remains effective. Whether ground or air targets can be attacked is also entirely determined by the current weapon.

In `rulesmd.ini`:
```ini
[SOMETECHNO]                ; TechnoType
SeparateWeaponTypes=none    ; List of SeparateWeaponType Enumeration (none|deployfireweapon|opentransport|noammo|all)
```

DeathFishAtEase:
Is the test conducted by user *bigothers* in [Phobos CN Forum/19](https://www.phoboscn.top/t/topic/19/20) valid? I see that you did not reply to his question about whether the test plan is correct below, and the PR is still marked as *Needs Testing*, it still appears to be in an unfinished testing state.

DeathFishAtEase:
Waiting for a response; it should have completed testing, but the previous volunteer's question about whether the testing method is correct needs confirmation or further guidance from the pull request author himself.

PR #2091, opened 2026-02-05
Group retaliate

TaranDahl:
### Group retaliate

- In vanilla, only the unit itself will counter - attack when it takes damage. This often makes units stupidly lured away one by one and eliminated.
- Now you can make units "seek support from nearby friendly units" when they take damage. They will retaliate the attacker together with the victim.
  - `GroupRetaliate.AllowAI` and `GroupRetaliate.AllowPlayer` control whether this effect applies to AI units and player units respectively.
  - The attacker's threat must be higher than the `GroupRetaliate.ThreatThreshold` of the unit's current target to trigger retaliation.
  - `GroupRetaliate.GroupRange` controls the range within which the victim can call for friendly forces.
  - `GroupRetaliate.TraceExtraRange` controls the distance for retaliating against the attacker. When the attacker is farther from the unit than the unit's guard range + TraceExtraRange, retaliation will not be triggered.
- An additional option `DisableVanillaRetaliateBehavior` is provided to turn off the vanilla retaliation behavior.

In `rulesmd.ini`:
```ini
[General]
GroupRetaliate.AllowAI=false             ; boolean
GroupRetaliate.AllowPlayer=false         ; boolean
GroupRetaliate.ThreatThreshold=1000.0    ; float, range in cell
GroupRetaliate.GroupRange=7.0            ; float, range in cell
GroupRetaliate.TraceExtraRange=20.0      ; float, range in cell
DisableVanillaRetaliateBehavior=false    ; boolean

GroupRetaliate.GroupRange=               ; float, range in cell, default to [General] -> GroupRetaliate.GroupRange
```

```{note}
The unit must be able to trigger the vanilla retaliation behavior in order to trigger group retaliate.
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -692,6 +692,7 @@ This page lists all the individual contributions to the project by their author.
   - Fix an issue where units recruited by a team with `AreTeamMembersRecruitable=false` cannot be recruited even if they have been liberated by that team
   - Global default value for `DefaultToGuardArea`
   - Weapon range finding in cylinder
+  - Group retaliate
 - **solar-III (凤九歌)**
   - Target scanning delay customization (documentation)
   - Skip target scanning function calling for unarmed technos (documentation)
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -1766,7 +1766,7 @@ InitialSpawnsNumber=      ; integer
 
 In `rulesmd.ini`:
 ```ini
-[General]
+[CombatDamage]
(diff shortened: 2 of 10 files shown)


deathreaperz:
What's the difference between Ares' TeamRetaliate and this one?

TaranDahl:
> What's the difference between Ares' TeamRetaliate and this one?

As the name implies, TeamRetaliate only takes effect on units with a Team, and only on units of the same Team. 
Here, `Group` does not refer to anything special. It refers to all friendly troops within a certain range around the unit.

PR #2092, opened 2026-02-05
Allow custom FPS values for each GameSpeed index

11EJDE11:
- Deprecates `CustomGS ini` keys
- Adds `EnableCustomFPS` and `CustomGameSpeedFPS`

Add to `RulesMD.ini`
```ini
;note: use 0 for the game's default fps for that index
[General]
EnableCustomFPS=True
CustomGameSpeedFPS.0=100
CustomGameSpeedFPS.1=90
CustomGameSpeedFPS.2=80
CustomGameSpeedFPS.3=70
CustomGameSpeedFPS.4=60
CustomGameSpeedFPS.5=0 ;default
CustomGameSpeedFPS.6=0 ;default
```

This PR allows for custom FPS values for each of the `GameSpeed` indices. Applies to Campaign, Skirmish, and Multiplayer.

For multiplayer we hook into `Queue_AI_Multiplayer` to override the hardcoded 60fps cap, and override the `GameSpeed` index to fps calculation.

For Skirmish and Campaign, we hook in main loop to set up the `NFTTimer` that is used in multiplayer. We then hook into `SyncDelay` and force skirmish/campaign down the multiplayer path instead.

### DIFF
--- a/docs/Miscellanous.md
+++ b/docs/Miscellanous.md
@@ -95,26 +95,9 @@ InsigniaType.PassengersN=                ; InsigniaType
 
 ## Game Speed
 
-### Single player game speed
+### Campaign default game speed
 
 - It is now possible to change the default (GS4/Fast/30FPS) campaign game speed with `CampaignDefaultGameSpeed`.
-- It is now possible to change the *values* of single player game speed, by inputing a pair of values. This feature must be enabled with `CustomGS=true`. **Only values between 10 and 60 FPS can be consistently achieved.**
-  - Custom game speed is achieved by periodically manipulating the delay between game frames, thus increasing or decreasing FPS.
-  - `CustomGSN.ChangeInterval` describes the frame interval between applying the effect. A value of 2 means "every other frame", 3 means "every 3 frames" etc. Increase of speedup/slowdown is approximately logarithmic.
-  - `CustomGSN.ChangeDelay` sets the delay (game speed number) to use every `CustomGSN.ChangeInterval` frames.
-  - `CustomGSN.DefaultDelay` sets the delay (game speed number) to use on other frames.
-  - Using game speed 6 (Fastest) in either `CustomGSN.ChangeDelay` or `CustomGSN.DefaultDelay` allows to set FPS above 60.
-    - **However, the resulting FPS may vary on different machines.**
-
-In `rulesmd.ini`:
-```ini
-[General]
-CustomGS=false              ; boolean
-CustomGSN.ChangeInterval=-1 ; integer >= 1
-CustomGSN.ChangeDelay=N     ; integer between 0 and 6
-CustomGSN.DefaultDelay=N    ; integer between 0 and 6
-; where N = 0, 1, 2, 3, 4, 5, 6
-```
 
 In `RA2MD.INI`:
 ```ini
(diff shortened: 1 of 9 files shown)


Metadorius:
...did you just make teh game be able to run on any FPS online?

11EJDE11:
Yeah. I got up to 140fps LAN.

src/Misc/Hooks.Gamespeed.cpp
```diff
@@ -1,97 +1,131 @@
...
-void SetNetworkFrameRate()
+struct FrameTimerStruct
 {
-	DEFINE_REFERENCE(int, PrecalcDesiredFrameRate, 0xA8B550u)
-	switch (GameOptionsClass::Instance.GameSpeed)
+	DWORD StartTime;
+	DWORD CurrentTime;
+	int   DelayTime;
+};
+DEFINE_REFERENCE(FrameTimerStruct, GameFrameTimer, 0x887348);
```
Metadorius: I think those may already exist in YRpp? There's some confusion with them there, maybe @ZivDero could help.

docs/Miscellanous.md
```diff
@@ -95,95 +95,36 @@ InsigniaType.PassengersN=                ; InsigniaType
...
-```{dropdown} Click to show the generator
-Enter desired FPS
-<div>
-<input id="customGameSpeedIn" type=number oninput="onInput()" style="width:100%";>
-</div>
+### Custom game speed
 
-Results (remember to replace N with your game speed number!):
+- Each of the 7 game speed slider positions (GameSpeed 0-6) can have a custom target FPS set independently. A value of `0` keeps that position at its vanilla FPS.
+- Works in skirmish, campaign, and multiplayer.
```
Metadorius: I think we should spell out explicitly that they will be unified now, so that the modders must disable the different game speed scales in client now.

docs/Miscellanous.md
```diff
@@ -95,95 +95,36 @@ InsigniaType.PassengersN=                ; InsigniaType
...
-
-In `rulesmd.ini`:
-```ini
-[General]
-CustomGS=false              ; boolean
-CustomGSN.ChangeInterval=-1 ; integer >= 1
-CustomGSN.ChangeDelay=N     ; integer between 0 and 6
-CustomGSN.DefaultDelay=N    ; integer between 0 and 6
-; where N = 0, 1, 2, 3, 4, 5, 6
-```
```
Metadorius: should add migration entries to docs and the migration script (see FIXME entries)

CnCRAZER:
> Yeah. I got up to 140fps LAN. 

LOL

I love it!

src/Misc/Hooks.Gamespeed.cpp
11EJDE11: I'm not seeing them there. There's Timer.h but that looks more like something you'd create rather than use when accessing existing ones?

docs/Miscellanous.md
11EJDE11: [b1e580f](https://github.com/Phobos-developers/Phobos/pull/2092/commits/b1e580f717f88d0c580fe1ccadfcc2e1dae3d7ab)

docs/Miscellanous.md
11EJDE11: [b1e580f](https://github.com/Phobos-developers/Phobos/pull/2092/commits/b1e580f717f88d0c580fe1ccadfcc2e1dae3d7ab)

Sorry, not seeing any FIXME?

Metadorius: Ctrl+F in migration script (see Phobos Supplementaries repo)

src/Misc/Hooks.Gamespeed.cpp
ZivDero: > I think those may already exist in YRpp? There's some confusion with them there, maybe @ZivDero could help.

Well the structs at least exist at
https://github.com/Phobos-developers/YRpp/blob/7a43c8bc7872d81c27ce470a4f0bf4d1c3a8e224/Timer.h

tomsons26:
I speculate NFT is for NetworkFrameTimer btw

src/Misc/Hooks.Gamespeed.cpp
ZivDero: The addresses don't seem to show up in YRpp so they're probably not in; Should be added there instead. 
Their names are `FrameTimer` (from RA) and the other one we called `NetworkFrameTimer`.

```
CDTimerClass<SystemTimerClass> FrameTimer;

CDTimerClass<MillisecondSystemTimerClass> NetworkFrameTimer;
```

`CDTimerClass` is what YRpp calls `TimerStruct`. `SystemTimerClass` is just `SystemTimer`.
`MillisecondSystemTimerClass` however is not present in YRpp it seems, and it is not the same as `SystemTimerClass` or `MillisecondTimerClass`.

I would provide the address but IDA isn't cooperating...

ZivDero: Starting at 5D5870, ctor, dtor, operator(), operator long

11EJDE11: Thanks, Ziv. Can you please check https://github.com/Phobos-developers/YRpp/pull/59

docs/Miscellanous.md
11EJDE11: Thanks. https://github.com/Phobos-developers/PhobosSupplementaries/pull/14

src/Misc/Hooks.Timers.cpp
```diff
@@ -18,7 +18,7 @@ DEFINE_HOOK(0x6D4B50, PrintTimerOnTactical_Start, 0x6)
 
 	if (Phobos::Config::RealTimeTimers_Adaptive
 		|| GameOptionsClass::Instance.GameSpeed == 0
-		|| (Phobos::Misc::CustomGS && !SessionClass::IsMultiplayer()))
+		|| (Phobos::Misc::EnableCustomFPS && Phobos::Misc::CustomGameSpeedFPS[GameOptionsClass::Instance.GameSpeed] > 0))
```
Metadorius, line 21: 1. I think you amending this check like this is actually making it worse. Since we have target FPS for all cases if custom game speed is on, we should actually *bail* from this branch if we have a custom game speed without adaptive timers set. So the check should read like "if adaptive OR vanilla FPS values AND game speed == 0".

src/Misc/PhobosToolTip.cpp
```diff
@@ -109,7 +109,7 @@ inline static int TickTimeToSeconds(int tickTime)
 
 	if (Phobos::Config::RealTimeTimers_Adaptive
 		|| GameOptionsClass::Instance.GameSpeed == 0
-		|| (Phobos::Misc::CustomGS && !SessionClass::IsMultiplayer()))
+		|| (Phobos::Misc::EnableCustomFPS && Phobos::Misc::CustomGameSpeedFPS[GameOptionsClass::Instance.GameSpeed] > 0))
```
Metadorius, line 112: should unify with the one I commented on as a common display value calc function (and any other instances if present)

docs/User-Interface.md
```diff
@@ -267,9 +288,8 @@ ShowPlacementPreview=yes   ; boolean
 ### Real time timers
 
 - Timers can now display values in real time, taking game speed into account. This can be enabled with `RealTimeTimers=true`.
-- By default, time is calculated relative to desired framerate. Enabling `RealTimeTimers.Adaptive` (always true for unlimited FPS and custom speeds) will calculate time relative to *current* FPS, accounting for lag.
-  - When playing with unlimited FPS (or custom speed above 60 FPS), the timers might constantly change value because of the unstable nature.
-- This option respects custom game speeds.
+- By default, time is calculated relative to the desired framerate for the current GameSpeed. Enabling `RealTimeTimers.Adaptive` (always forced on when GameSpeed is 0/Fastest or a custom game speed is active) will calculate time relative to *current* FPS, accounting for lag.
+  - When playing with a custom FPS above 60, the timers might constantly fluctuate due to frame-to-frame variance.
```
Metadorius, line 292: should amend this according to the comments

Metadorius (changes requested):

PR #2093, opened 2026-02-07
[Minor] Maintain compatibility withold tags

DeathFishAtEase:
*This is a **temporary** solution supplementing #1984.*

> [!Important]
> It is inelegant and inconvenient to maintain, but we need at least one workable solution ready before the next devbuild release, even if it will eventually be entirely replaced by other implementations.

---

There have been complaints about what the tag renaming PR did:

- Some consider it premature, believing it should have been done just before Release 0.5 to facilitate testing;
- Some think it didn't go far enough—certain tags that should have been renamed were not, but discussions on GitHub ceased after the previous PR was closed;
- Others strongly oppose changing what is already in use. However, a complete rollback of the previous PR’s changes would also create significant rework, forcing many who have already migrated to revert their changes.
- More elegant implementations have been considered, such as allowing "aliases" for INI tags, but they are far off, and it’s not even clear when work on them will begin.

Because the Phobos-chat channel keeps mentioning these issues, and maintainers like Ollerus are also concerned about the impact of whether to roll back on users already using the new or old versions, this PR, which was completed long ago but is obviously not perfect, has been opened here—at least when we have to face this problem, it is an option available to us.

> [!Caution]
> The branch of this PR will be synced with `develop` relatively frequently, but timely updates to the latest state of the `develop` branch are not guaranteed.


TaranDahl (approved):
I think you should add comments near the patch to facilitate future deletion.

TaranDahl:
@Metadorius I think there's no problem with this. Can it be merged now?

Metadorius (commented):
I still think that old tag names should be read with a log warning emitted. And the proper tag names should be prioritized, not vice versa.

TaranDahl:
> And the proper tag names should be prioritized, not vice versa.

As I understand it, it reads the old name first and then the new name. So, it should give priority to using the new name.

src/Ext/WarheadType/Body.cpp
```diff
@@ -162,14 +162,18 @@ void WarheadTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
...
+	this->Crit_AffectsHouse.Read(exINI, pSection, "Crit.AffectsHouses"); // Temporary solution for the INI tags renaming issue, see #2093
 	this->Crit_AffectsHouse.Read(exINI, pSection, "Crit.AffectsHouse");
 	this->Crit_AnimList.Read(exINI, pSection, "Crit.AnimList");
 	this->Crit_AnimList_PickRandom.Read(exINI, pSection, "Crit.AnimList.PickRandom");
 	this->Crit_AnimList_CreateAll.Read(exINI, pSection, "Crit.AnimList.CreateAll");
 	this->Crit_ActiveChanceAnims.Read(exINI, pSection, "Crit.ActiveChanceAnims");
 	this->Crit_AnimOnAffectedTargets.Read(exINI, pSection, "Crit.AnimOnAffectedTargets");
+	this->Crit_AffectsBelowPercent.Read(exINI, pSection, "Crit.AffectBelowPercent"); // Temporary solution for the INI tags renaming issue, see #2093
 	this->Crit_AffectsBelowPercent.Read(exINI, pSection, "Crit.AffectsBelowPercent");
+	this->Crit_AffectsBelowPercent.Read(exINI, pSection, "Crit.AffectAbovePercent"); // Temporary solution for the INI tags renaming issue, see #2093
```
Coronia: should be this->Crit_AffectsAbovePercent

Coronia (changes requested):

Metadorius:
> As I understand it, it reads the old name first and then the new name. So, it should give priority to using the new name.

Right, I think I've seen what @Coronia commented on and was thrown off because of this.

src/Ext/WarheadType/Body.cpp
DeathFishAtEase: fixed

Coronia (approved):

PR #2094, opened 2026-02-08
Improvements to stationary unit logic

Starkku:
- Extended logic e.g `Speed=0` to work on InfantryTypes.
- Speed lowered to 0 by multipliers will now prevent some of the logic from being executed (cursor, assigning destination etc. works but not actual movement).

-----------------

### Stationary units

- Infantry & vehicles with `Speed=0` or those that are prevented from moving on their cell by `MovementRestrictedTo` are now truly stationary will not attempt to move if issued movement commands. Should not be used on trainable/buildable units as they become stuck in factory. Following behaviours apply:
  - Cannot retaliate if target is beyond their weapon range.
  - Cannot acquire units outside their weapon range on `Area Guard` mission.
  - `Hunt` mission is reassigned to `Guard`.
  - Units rendered stationary by speed multipliers instead of intrinsic properties of TechnoType will respond to cursor actions e.g assigning destination but will not move while the multiplier is in effect.


src/Ext/Techno/Hooks.DisallowMoving.cpp
```diff
@@ -0,0 +1,378 @@
...
+
+		if (!pThis->IsCloseEnough(pTarget, weaponIndex))
+		{
+			pThis->SetTarget(nullptr);
+			return true;
+		}
+	}
+
+	return false;
+}
```
TaranDahl: I think this should be placed at 4D5761.

Coronia:
object info dump showed that these stationary units will sometime enter 'Move' mission for a short moment, then switch back to 'Guard' quickly without actually moving. Not sure if it's a desired behavior

Refailure:
After applying an AttachEffect with SpeedMultiplier=0 to the target unit, the target unit can still rotate its hull once before it stops rotating the hull and only rotates the turret. Hover units can always rotate their hull.

And units that have performed the single rotation cannot load passengers.

fagonghaiwo:
ee455ddb8d29529fb5544790238c3a48
turret still rotating

Starkku:
This isn't supposed to prevent turret from rotating.

src/Ext/Techno/Hooks.cpp
```diff
@@ -50,6 +50,27 @@ DEFINE_HOOK(0x4DA54E, FootClass_AI, 0x6)
 	if (pExt->AttackMoveFollowerTempCount)
 		pExt->AttackMoveFollowerTempCount--;
 
+	// Check speed multipliers for zero, update things like allowing
+	// locomotor to process destination again.
+	if (FootExt::GetCurrentSpeedMultiplier(pThis) > 0.0)
```
Coronia, line 55: is it possible to put this under where the movement actually happens, FootClass_Update_AfterLocomotorProcess maybe? So the speed multiplier won't have to be calculated for every FootClass per frame here

fagonghaiwo:
8cc24f424c2aa58491df17e2001b6ef8

aircraft can fly

76c099b471c91b7fb5c23860d33212e7

rotat when attack

src/Ext/Techno/Hooks.cpp
Starkku: 1. Moving the code between these two hooks would have negligible impact, only FootClass instances excluded would be those that were destroyed by radiation or in locomotor processing on that particular frame.
2. This is something that needs to be checked *before* locomotor processing for that frame anyway.

I'll investigate later if there's a way to improve how it is handled currently anyway, but your suggestion won't help.

Starkku:
> aircraft can fly

Had you read the documentation you would realize this isn't supposed to work for AircraftTypes to begin with.

> rotat when attack

This does not currently prevent rotation at all, actually. Maybe it should on turretless vehicles but the documentation so far only mentions no movement, nothing about rotating on place. Although if you ask me modifiers to rotation or its speed should perhaps be its own feature.

Starkku:
- Moved the temporary immobilization checks to after locomotor processing - there wasn't noticeable impact on functionality and presumably there is a minor performance benefit in edge cases where processing ceases early.
- Fixed an oversight that caused already moving infantry to not be fully immobilized by temporary speed multipliers. It is possible that there may be edge case issues stemming from this, however unlikely. I already checked the most obvious ones like amphibious infantry functioning correctly.

src/Ext/Foot/Body.cpp
```diff
@@ -865,6 +866,87 @@ void FootExt::HealthAutoConvertActions()
...
+		if (movementRestrictedTo != landType)
+			return true;
+	}
+
+	return false;
+}
+
+// Check speed multipliers for zero, update things like allowing
+// locomotor to process destination again.
+void FootExt::HandleTemporaryZeroSpeed()
```
Coronia, line 922: maybe it can be simplified when it's already Speed=0?

Coronia: actually I'm not sure. There might be a case that the unit is converted to 0 Speed/JumpjetSpeed, so it might still need to be handled in this case. But that's like going through a whole `FootExt::CannotMove` process which I'm not sure if it's worth to be executed with this much frequency

PR #2096, opened 2026-02-08
Skip drawing invisible voxel sections

Metadorius:
- Voxel drawing code now skips sections that are invisible (have all zeros in the transform matrix main diagonal, meaning that the scale is 0% on all axes), thus increasing drawing performance for some voxels (by Kerbiter, ZivDero)


src/Misc/Hooks.BugFixes.cpp
```diff
@@ -3014,3 +3015,26 @@ DEFINE_HOOK(0x6EA870, TeamClass_LiberateMember_Start, 0x6)
...
+	// stolen code
+	if (!pMotLib)
+		return 0x706FBD;
+
+	GET(int const, layer, EBX);
+	GET_STACK(unsigned int const, frame, STACK_OFFSET(0x13C, 0x18));
+
+	auto mtx = pMotLib->GetLayerMatrix(layer, frame);
+
+	if (CLOSE_ENOUGH(mtx.row[0][0], 0.0) && CLOSE_ENOUGH(mtx.row[1][1], 0.0) && CLOSE_ENOUGH(mtx.row[2][2], 0.0))
```
Starkku: Pretty sure we have equal to zero float comparisons without the epsilon all over the codebase and it is my understanding that major floating point standards represent zero as exact value, I don't know if the transformation matrix data has inaccuracies from tools that specifically warrant using it here.

Starkku (approved):
LGTM. Just one thing but it is basically 'who cares' kind of deal.

src/Misc/Hooks.BugFixes.cpp
Metadorius: I don't know either, just wanted to play it safe since I *think* I recall some 0,0023 or so values in the transform matrix somewhere. My memory may be playing tricks on me tho.

Metadorius: I don't think there's much practical difference anyway, I don't think anyone would want to deliberately make a dot out of their voxel.

TaranDahl:
I don't quite understand the drawing. Could you elaborate on which `some voxels` specifically refer to?

Coronia:
Tested it under SP and MP scenarios, no desync or wrong voxel displaying so far. Can be merged once it's finished

Metadorius:
> I don't quite understand the drawing. Could you elaborate on which `some voxels` specifically refer to?

it is mainly intended to improve performance on section swap animated voxels, most likely that's the only thing it will improve. I am waiting for the conclusive testing results from the community members who requested that and will most likely amend the doc/code.

Metadorius:
Waiting on feedback from Danielovich7 on Discord.

Metadorius:
image
Also please note the commit name.

docs/Fixed-or-Improved-Logics.md
```diff
@@ -307,6 +307,7 @@ This page describes all ingame logics that are fixed or improved in Phobos witho
 - Fixed the bug that `DeploysInto` and `UndeploysInto` will make damaged techno lose 1 health.
 - Fixed the issue that the Jumpjet must end its movement before starting the next mission.
 - Fixed an issue where parachute units would die upon landing if bridges were destroyed during their descent.
+- Voxel drawing code now skips sections that are invisible (have all zeros in the transform matrix main diagonal, meaning that the scale is 0% on all axes), thus increasing drawing performance for some voxels.
```
Metadorius, line 310:
```suggestion
- Voxel drawing code now skips sections that are invisible (have all zeros in the transform matrix main diagonal, meaning that the scale is 0% on all axes), thus increasing drawing performance for uncached voxels which are animated using section swap method.
```

src/Misc/Hooks.BugFixes.cpp
```diff
@@ -3164,3 +3165,26 @@ DEFINE_HOOK(0x7442AB, UnitClass_ReadyToNextMission_FallingDown, 0x6)
 	GET(FootClass*, pThis, ESI);
 	return pThis->IsFallingDown ? ReturnZero : 0;
 }
+
+DEFINE_HOOK(0x706F64, TechnoClass_RenderVoxelObject_SkipInvisibleSections, 0x0)
```
Metadorius, line 3170:
```suggestion
// sadly, useful for uncached voxels specifically, but no reason for the code to go to waste
DEFINE_HOOK(0x706F64, TechnoClass_RenderVoxelObject_SkipInvisibleSections, 0x0)
```

PR #2100, opened 2026-02-11
Fixed the issue that the move mission of the jumpjet does not end correctly

TaranDahl:
Which would prevent AttackMove from finishing.

### DIFF
--- a/src/Misc/Hooks.BugFixes.cpp
+++ b/src/Misc/Hooks.BugFixes.cpp
@@ -3094,7 +3094,7 @@ DEFINE_HOOK(0x6F833E, TechnoClass_CanAutoTargetObject_Garrisonable, 0x6)
 DEFINE_HOOK(0x4D4203, FootClass_MissionMove_EndCheckFix1, 0x6)
 {
 	GET(FootClass*, pThis, ESI);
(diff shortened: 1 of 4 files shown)


Coronia:
what's the difference between this and https://github.com/Phobos-developers/Phobos/pull/2084?

TaranDahl:
> what's the difference between this and #2084?

From a gameplay perspective, when the jumpjet uses AttackMove with a cell as the target, 2100 addresses the issue that the AttackMove task will not be cleared even after reaching the destination, and 2084 addresses the issue that the unit will not correctly switch to the attack task after acquiring the target.

Coronia (approved):

src/Misc/Hooks.BugFixes.cpp
```diff
@@ -3022,3 +3022,17 @@ DEFINE_HOOK(0x6EA870, TeamClass_LiberateMember_Start, 0x6)
 	pMember->RecruitableB = true;
 	return 0;
 }
+
+DEFINE_HOOK(0x4D4203, FootClass_MissionMove_EndCheckFix1, 0x6)
+{
+	GET(FootClass*, pThis, ESI);
+	R->EAX(pThis->Destination && pThis->DistanceFrom(pThis->Destination) > 256);
```
Coronia: maybe use Unsorted::LeptonsPerCell here for regulation

TaranDahl:
Tested by
image

PR #2104, opened 2026-02-12
DeployFire supports buildings

Fly-Star-him:
- Fixed an issue where `OmniFire` was ineffective on buildings with `Turret=yes`.
- Fixed an issue where setting a production building as `Primary` could cause it to enter an unload state.

---------

- Building types now also support using `DeployFire` and `DeployFireWeapon`.
  - If a building has other configurations such as `Factory`, `GapGenerator`, or `Passengers`, it will prioritize executing those deployment actions instead.
  - `DeployFireDelay` specifies the frame interval at which deployed weapons attempt to fire. If the weapon is unavailable due to `ROF`, `CanTarget`, or other reasons, the deployed weapon will not fire.
  
In `rulesmd.ini`:
```ini
[SOMEBUILDING]      ; BuildingType
DeployFireDelay=    ; integer, The default value is in the range of 14 to 16.
```


src/Ext/Building/Hooks.Unload.cpp
```diff
@@ -0,0 +1,122 @@
...
+					pThis->QueueMission(Mission::Guard, false);
+					pThis->NextMission();
+
+					return SkipGameCode;
+				}
+			}
+
+			auto const pTypeExt = BuildingTypeExt::ExtMap.Find(pType);
+			const int result = pTypeExt->DeployFireDelay.isset() ?
+				pTypeExt->DeployFireDelay : (ScenarioClass::Instance->Random.RandomRanged(0, 2) + 14);
```
TaranDahl: The Nullable template should be used like this: `nullableVar.Get(default)`.

src/Ext/Building/Hooks.Unload.cpp
```diff
@@ -0,0 +1,122 @@
...
+			}
+			else if (fireError == FireError::OK && pThis->Fire(pCell, weaponIndex))
+			{
+				auto const pWeapon = pThis->GetWeapon(weaponIndex)->WeaponType;
+
+				if (pWeapon->FireOnce)
+				{
+					// When Turret=yes, the Unload task may not be canceled, so this handling is performed.
+					pThis->QueueMission(Mission::Guard, false);
+					pThis->NextMission();
```
TaranDahl: Same

src/Ext/Building/Hooks.Unload.cpp
```diff
@@ -0,0 +1,122 @@
...
+			continue;
+
+		const Mission currentMission = pBuilding->CurrentMission;
+
+		if (currentMission == Mission::Construction || currentMission == Mission::Selling)
+			continue;
+
+		if (pBuilding->EMPLockRemaining > 0 || !pBuilding->WasOnline || pBuilding->BunkerLinkedItem)
+			continue;
+
```
TaranDahl, line 104: Should be wrapped into a function.

src/Ext/Techno/Hooks.Misc.cpp
```diff
@@ -874,19 +874,32 @@ DEFINE_HOOK(0x4C7512, EventClass_Execute_StopCommand, 0x6)
...
 	}
+	else if (auto const pBuilding = abstract_cast<BuildingClass*, true>(pThis))
+	{
+		auto const pType = pBuilding->Type;
+
+		if (pBuilding->CurrentMission == Mission::Unload && pType->DeployFire && !pType->WeaponsFactory)
+		{
+			pBuilding->SetTarget(nullptr);
+			pBuilding->QueueMission(Mission::Guard, false);
+			pBuilding->NextMission();
```
TaranDahl: Same

CREDITS.md
```diff
@@ -412,6 +412,9 @@ This page lists all the individual contributions to the project by their author.
   - Fix an issue where miners affected by `Passengers/DeployFire` were unable to unload minerals
   - Fix an issue where mining vehicles could not move after leaving a tank bunker
   - Fixed the bug in AI scripts 56 and 57 that forced the launch of superweapons with index numbers 3 and 4
+  - DeployFire supports buildings
+  - Fixed an issue where `OmniFire` was ineffective on buildings with `Turret=yes`
```
TaranDahl: `OmniFire` is actually a tag limited to UnitClass. AircraftClass uses `Fighter`, BuildingClass must face the target to fire, while InfantryClass ignores the facing.

src/Ext/Building/Hooks.Unload.cpp
```diff
@@ -0,0 +1,121 @@
...
+			const int deployFireWeapon = pType->DeployFireWeapon;
+			const int weaponIndex = deployFireWeapon >= 0 ? deployFireWeapon : pThis->SelectWeapon(pCell);
+			const FireError fireError = pThis->GetFireError(pCell, weaponIndex, true);
+
+			if (fireError == FireError::ILLEGAL)
+			{
+				// Do not allow the building to remain in the Unload task indefinitely.
+				pThis->QueueMission(Mission::Guard, false);
+				pThis->NextMission();
```
TaranDahl: Should use `ForceMission`

TaranDahl (changes requested):

NetsuNegi (approved):

src/Ext/Building/Hooks.Unload.cpp
TaranDahl: @Fly-Star-him

Coronia: if it's confirmed to be not null by isset(), there's no need to add default value here

TaranDahl (approved):

PR #2105, opened 2026-02-13
Buildable-upon OverlayTypes

Starkku:
### Buildable-upon OverlayTypes

- It is now possible to make OverlayTypes allow buildings to be placed on them by setting `CanBeBuiltOn` to true. This still requires the tile's landtype (which is changed to overlay's land type unless OverlayType has `NoUseTileLandType=false`) to allow buildings to be placed.
  - If `CanBeBuiltOn.Remove=true`, the overlay will be removed (if it is a wall owned by the player placing the building, it is sold) upon the building being placed. If this is not set to true, buildings with `Wall=true` cannot be placed on the overlay and neither does overlay with `Wall=true` allow buildings to be placed on itself regardless of other settings.
  - `Tiberium.CanBeBuiltOn` can be used to set the global default value for overlays with `Tiberium=true`.
  - `Wall.CanBeBuiltOn` can be used to set the global default value for overlays with `Wall=true`. This is checked after `Tiberium.CanBeBuiltOn`.
  - `Rock.CanBeBuiltOn` can be used to set the global default value for overlays with `IsARock=true`. This is checked after `Tiberium.CanBeBuiltOn` and `Wall.CanBeBuiltOn`.

In `rulesmd.ini`:
```ini
[General]
Tiberium.CanBeBuiltOn=false     ; boolean
Wall.CanBeBuiltOn=false         ; boolean
Rock.CanBeBuiltOn=false         ; boolean
CanBeBuiltOnOverlay.Remove=true ; boolean

[SOMEOVERLAY]                   ; OverlayType
CanBeBuiltOn=                   ; boolean, default to [General] -> Tiberium/Wall/Rock.CanBeBuiltOn depends on the overlay settings, and false otherwise
CanBeBuiltOn.Remove=            ; boolean, default to [General] -> CanBeBuiltOnOverlay.Remove
```

### DIFF
--- a/src/Ext/TerrainType/Hooks.Passable.cpp
+++ b/src/Ext/TerrainType/Hooks.Passable.cpp
@@ -267,8 +267,11 @@ DEFINE_HOOK(0x586780, MapClass_IsAreaFree, 0x7)
 				? (pCell->Passability != PassabilityType::Passable && pCell->Passability != PassabilityType::HasFreeSpots)
 				: (pCell->Passability != PassabilityType::Passable);
 
+			// Check if cell has overlay and if it is buildable on.
+			const bool invalidOverlay = pCell->OverlayTypeIndex != -1 && !OverlayTypeExt::CanPlaceBuildingOnOverlay(pCell->OverlayTypeIndex, nullptr, false);
+
 			if ((pCell->BaseSpacerOfHouses & mask) != 0
-				|| pCell->OverlayTypeIndex != -1
+				|| invalidOverlay
 				|| invalidPassability
 				|| pCell->SlopeIndex
 				|| pCell->GetBuilding())
--- a/src/Ext/TerrainType/Hooks.Passable.cpp
+++ b/src/Ext/TerrainType/Hooks.Passable.cpp
@@ -1,4 +1,5 @@
 #include "Body.h"
+#include <Ext/OverlayType/Body.h>
 
 #include <Ext/Rules/Body.h>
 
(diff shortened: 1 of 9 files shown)


TaranDahl (approved):

Coronia:
it seems that ore overlay can't be built on even with CanBeBuiltOn and CanBeBuiltOn.Remove set

Starkku:
> it seems that ore overlay can't be built on even with CanBeBuiltOn and CanBeBuiltOn.Remove set

`Tiberium` LandType has `Buildable=false` normally.

Coronia:
changing Tiberium to Buildable=yes still can't allow it to be built

Starkku:
image

Cannot reproduce the issue with ore overlays, as long as all of the overlays have `CanBeBuiltOn=true` and `Tiberium` landtype has `Buildable=true` I am able to place buildings on ore.

fagonghaiwo:
image

works

Coronia:
@DeathFishAtEase please check the current global default value settings, to see if they're correct or if more is needed

DeathFishAtEase (approved):

DeathFishAtEase:
> if more is needed

The global default values for the main grouped overlay types have already been added. If we still want to handle more, perhaps only Crate is worth considering, but it is harmless to ignore it. As for Veinhole and the like, they each have only one type, so defining them at the micro level is sufficient.

fagonghaiwo:
551b0cd8e7752d8a1eb927a6aa5ca39b

558d632877e47b98914402c17b86fce0


AI can deplay MCV but cant build

src/Ext/TerrainType/Hooks.Passable.cpp
Coronia, line 119: perhaps overlay buildable check should also be done here?

Coronia, line 184: same

Coronia, line 254: and here. Seems like this is related to AI buildable

Starkku: The code in game there doesn't specifically check for overlays in the first place.

Starkku: Does not check for overlays in game either.

Starkku: Not in slightest, this is about placing overlays on cells with terrain on them.

fagonghaiwo:
35f1881d41590353ded3dd487d4a178e
[General]
Wall.CanBeBuiltOn=1         ; boolean

AI still no build

fagonghaiwo:
e6b3cf08b3d756532e6a5716060d08d0
[General]
Rock.CanBeBuiltOn=1

cant

PR #2111, opened 2026-02-21
Fix desync caused by inconsistent shroud state

11EJDE11:
When Session.MPGameMode->SkipCheatCheck is active, there are two probes to check a hardcoded cell's shroud state (ShroudCounter != 1 || AltFlags & 0x18 at main loop around 0055ddc1 and HouseClass::AI around 004f88a1). If it passes the checks, the synced random gets advanced. 
Because Clear_Shroud skips the cell check but Reset_Shroud does not, a gap generator cycling near the top-left corner of the map (where the cell check falls) can leave the cell's shroud state inconsistent between players which causes a desync.

To fix, we're just setting that SkipCheatCheck to true so it never runs. In v1.006 they removed the check entirely.

Here is a test map. One player at top left. Both deploy, top left player sells powers and undeploys mcv > desync (note mcv gives power on this map, hence the undeploy).
[desynctest.zip](https://github.com/user-attachments/files/25455974/desynctest.zip)

Thanks to Razer for the map, Kerbiter and ZivDero for confirming and providing additional information.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -736,4 +736,6 @@ This page lists all the individual contributions to the project by their author.
 - **Damfoos** - extensive and thorough testing
 - **Dmitry Volkov** - extensive and thorough testing
 - **Rise of the East community** - extensive playtesting of in-dev features
-- **11EJDE11** - Prevent mpdebug number from being drawn when visibility toggled off
+- **11EJDE11**:
+  - Prevent mpdebug number from being drawn when visibility toggled off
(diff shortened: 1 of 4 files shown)


Starkku:
The `SkipCheatCheck ` function is also called at `0x501659` and setting the return value to true makes game allow paranoid AI's even if alliances are not enabled, I would patch out the call to prevent behaviour changes.

PR #2112, opened 2026-02-21
Taunt warhead

TaranDahl:
### Taunt warhead

- Now you can use the following tags to make the warhead "taunt" the target, override its current mission, and force it to attack the source of the damage.
  - The taunted target will behaves like doing retaliation.
  - If there is no source unit for the damage, the taunt will not take effect.

In `rulesmd.ini`:
```ini
[SOMEWARHEAD]    ; WarheadType
Taunt=false      ; boolean
```


Coronia:
feels like this needs a bit more customization. `AffectsHouse` is definitely needed to make it not trigger a taunt with friendly fire, and an `ImmuneToTaunt` tag could be given to techno or shield, and such

TaranDahl:
> feels like this needs a bit more customization. `AffectsHouse` is definitely needed to make it not trigger a taunt with friendly fire, and an `ImmuneToTaunt` tag could be given to techno or shield, and such

By using `ApplyPerTargetEffectsOnDetonate=false`, you can control them through vanilla and Ares damage controllers.

TaranDahl:
I admit that there may be situations where different effects require different controllers, but I think these situations can usually be achieved through `ExtraWarhead`. 
In short, I don't think it's necessary to reinvent the wheel of the controller.
In addition, I suggest that for the additional effects of warheads in the future, there is no need to reinvent the wheel of the controller.

Coronia (approved):

PR #2114, opened 2026-02-22
Extra threat

TaranDahl:
### Extra threat

- Now you can adjust the techno's evaluation of the threat posed by the target in more ways. This will help the techno in auto - targeting.
  - When the target poses a threat to the techno, it will receive an additional threat value defined by `ExtraThreat.IsThreat`.
    - Generally speaking, "Posing a threat" means the target can fire at the techno.
    - Using `AlwaysConsideredThreat` makes the target always considered to be a threat.
  - When the target is within the techno's range, it will receive an additional threat value defined by `ExtraThreat.InRange`.
  - When the target is within the techno's range, it will receive an additional threat value equal to `ExtraThreatCoefficient.InRangeDistance` multiplied by the distance (in cells) from the target to the techno.
    - Only considering in-range is because the vanilla flag `TargetDistanceCoefficientDefault` only considers outside-range. This flag is complementary to that.
  - The target will receive an additional threat value equal to `ExtraThreatCoefficient.Facing` multiplied by the difference in facing from the techno's current firing-facing to the target's facing.
    - "Firing-facing" refers to the facing the techno uses to check if it "is already facing the target and can fire". Infantry doesn't check the facing when firing, so this is also invalid for infantry.
    - The unit of facing is the in-game internal numerical scale. A full circle corresponds to 65536.
  - The target will receive an additional threat value equal to `ExtraThreatCoefficient.DistanceToLastTarget` multiplied by the distance (in cells) from the target to the techno's last target.
    - Each techno will record its current target as the "last target" per frame. This record will be retained for at most 15 frames after the target becomes invalid.
    - If the techno doesn't have a "last target", then this will not take effect.

In `rulesmd.ini`:
```ini
[General]
ExtraThreat.IsThreat=0.0                            ; double
ExtraThreat.InRange=0.0                             ; double
ExtraThreatCoefficient.InRangeDistance=0.0          ; double
ExtraThreatCoefficient.Facing=0.0                   ; double
ExtraThreatCoefficient.DistanceToLastTarget=0.0     ; double

[SOMETECHNO]                                        ; TechnoType
AlwaysConsideredThreat=false
ExtraThreat.IsThreat=                               ; double, default to the flag in [General] with same name
ExtraThreat.InRange=                                ; double, default to the flag in [General] with same name
ExtraThreatCoefficient.InRangeDistance=             ; double, default to the flag in [General] with same name
ExtraThreatCoefficient.Facing=                      ; double, default to the flag in [General] with same name
ExtraThreatCoefficient.DistanceToLastTarget=        ; double, default to the flag in [General] with same name
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -735,6 +735,7 @@ This page lists all the individual contributions to the project by their author.
   - Fix the bug where non-Teleporter miners would not return to work after minerals are depleted and then regenerated
   - Miners back to work when ore regenerated
   - Allow disable an over-optimization in targeting
+  - Extra threat
 - **solar-III (凤九歌)**
   - Target scanning delay customization (documentation)
   - Skip target scanning function calling for unarmed technos (documentation)
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -1397,6 +1397,17 @@ In `rulesmd.ini`:
 DisableOveroptimizationInTargeting=false  ; boolean
 ```
 
+### Allow techno type considered as other type when recruiting techno for teams
+
+- It is now possible to make techno type considered as other type when recruiting techno for teams, both for AI team recruitment and `Create Team` action.
+  - Only affect techno that's presented on the map. Cannot make AI produce this type of techno if it doesn't have any.
+
+In `rulesmd.ini`:
+```ini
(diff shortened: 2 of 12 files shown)


NetsuNegi (approved):

TaranDahl:
Tested by myself.

PR #2115, opened 2026-02-22
Fix inconsistency between GapGen and SpySat causing desyncs

Starkku:
This is an alternate fix to #2111 that addresses the root cause. SpySat skips certain cells in multiplayer when revealing shroud, this replicates the same check for resetting shroud as well as creating/deleting gap.

Do not merge both this and #2111.


Coronia:
@11EJDE11 please confirm that this PR can also fix the issue in https://github.com/Phobos-developers/Phobos/pull/2111

Starkku:
FYI this was supposed to merely be an alternate approach to the issue and not decisively better one, that would have warranted further discussion which I don't think reached a proper conclusion.

I did test it quickly in multiplayer with Kerbiter and the desync did seem to be gone, although having others test it too may not be the worst idea.

Coronia:
since it's confirmed to work in testing, maybe it's still better to merge it for now, and replace it with another 'better solution' if there's any in the future

Coronia (approved):

PR #2117, opened 2026-02-23
Fix the issue of health loss caused by conversion

CrimRecya:


TaranDahl (approved):

Coronia (approved):
resolve the compiler warning and we're all good

PR #2118, opened 2026-02-23
New bounty logic

NetsuNegi:
- Similar to [Ares' bounty logic](https://ares-developers.github.io/Ares-docs/new/bounty.html), but with more configurable options and easier to use.
- `Bounty.Enable` is the master switch for the new bounty logic; it must be turned on first to utilize the new bounty system.
- `Bounty.Enablers` specifies the TechnoTypes that grant the bounty capability, not limited to buildings.
  - If this list is empty, the bounty logic is enabled without requiring any prerequisite TechnoTypes.
- `Bounty.Multiplier` is the multiplier applied to the bounty a hunter receives when a victim is killed. This is defined on the victim.
- `Bounty.Value` is the fixed amount of funds a hunter receives when a victim is killed. This is also defined on the victim.
  - If `Bounty.Value` is not 0, the base value is the amount set by `Bounty.Value`; otherwise, the base bounty value is the victim's actual cost multiplied by `Bounty.Multiplier`.
- `Bounty.KillerMultiplier` is the multiplier applied to the bounty received when killing a target. This is defined on the hunter.
- `Bounty.Display` controls whether the bounty amount is shown, displayed on the hunter.

In `rulesmd.ini`:
```ini
[General]
Bounty.Enable=false                 ; boolean
Bounty.Enablers=                    ; list of TechnoTypes
Bounty.Default=false                ; boolean
Bounty.Multiplier=1.0               ; double
Bounty.KillerMultiplier=1.0         ; double

[AudioVisual]
Bounty.Display=false                ; boolean
Bounty.Display.Houses=all           ; List of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)

[SOMETECHNO]                        ; TechnoType
Bounty=                             ; boolean, default to [General] -> Bounty.Default
Bounty.Multiplier=                  ; double, default to [General] -> Bounty.Multiplier
Bounty.Multiplier.Veteran=          ; double, default to [TechnoType] -> Bounty.Multiplier
Bounty.Multiplier.Elite=            ; double, default to [TechnoType] -> Bounty.Multiplier
Bounty.Value=0                      ; integer
Bounty.Value.Veteran=               ; integer, default to [TechnoType] -> Bounty.Value
Bounty.Value.Elite=                 ; integer, default to [TechnoType] -> Bounty.Value
Bounty.KillerMultiplier=            ; double, default to [General] -> Bounty.KillerMultiplier
Bounty.KillerMultiplier.Veteran=    ; double, default to [TechnoType] -> Bounty.KillerMultiplier
Bounty.KillerMultiplier.Elite=      ; double, default to [TechnoType] -> Bounty.KillerMultiplier
Bounty.Display=                     ; boolean, default to [AudioVisual] -> Bounty.Display
Bounty.Display.Houses=              ; List of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all), default to [AudioVisual] -> Bounty.Display.Houses
Bounty.Display.Offset=0,0           ; X,Y, pixels relative to default
```

```{note}
When this new bounty logic is enabled, Ares' bounty logic will be disabled.
```


src/Ext/Techno/Body.cpp
```diff
@@ -883,6 +885,128 @@ void TechnoExt::ClickedApproachObject(FootClass* pThis, ObjectClass* pObject)
...
+
+	if (!value || !pKillerHouse->CanTransactMoney(value))
+		return;
+
+	pKillerHouse->TransactMoney(value);
+
+	if (pKillerTypeExt->Bounty_Display.Get(RulesExt::Global()->Bounty_Display))
+	{
+		const auto displayTo = pKillerTypeExt->Bounty_Display_Houses.Get(RulesExt::Global()->Bounty_Display_Houses);
+		FlyingStrings::AddMoneyString(value, pKiller, pKiller->Owner, displayTo, pKiller->Location, pKillerTypeExt->Bounty_Display_Offset);
```
TaranDahl: I think our Helper needs to be updated. This number should allow for custom colors.

docs/New-or-Enhanced-Logics.md
```diff
@@ -2167,6 +2170,51 @@ TiberiumEater.Anims.Tiberium3=    ; List of AnimationTypes
...
+
+- Similar to [Ares' bounty logic](https://ares-developers.github.io/Ares-docs/new/bounty.html), but with more configurable options and easier to use.
+- `Bounty.Enable` is the master switch for the new bounty logic; it must be turned on first to utilize the new bounty system.
+- `Bounty.Enablers` specifies the TechnoTypes that grant the bounty capability, not limited to buildings.
+  - If this list is empty, the bounty logic is enabled without requiring any prerequisite TechnoTypes.
+- `Bounty.Multiplier` is the multiplier applied to the bounty a hunter receives when a victim is killed. This is defined on the victim.
+- `Bounty.Value` is the fixed amount of funds a hunter receives when a victim is killed. This is also defined on the victim.
+  - If `Bounty.Multiplier` is not 0, the base bounty value is the victim's actual cost multiplied by this multiplier; otherwise, the base value is the amount set by `Bounty.Value`.
+- `Bounty.KillerMultiplier` is the multiplier applied to the bounty received when killing a target. This is defined on the hunter.
+- `Bounty.Display` controls whether the bounty amount is shown, displayed on the hunter.
```
TaranDahl, line 2281: I think it should be allowed to be displayed on the head of the prey.

docs/New-or-Enhanced-Logics.md
```diff
@@ -58,6 +58,8 @@ This page describes all the engine features that are either new and introduced b
   - `DisableWeapons` can be used to disable ability to fire any and all weapons.
     - On TechnoTypes with `OpenTopped=true`, `OpenTopped.CheckTransportDisableWeapons` can be set to true to make passengers not be able to fire out if transport's weapons are disabled by `DisableWeapons`.
   - `Unkillable` can be used to prevent the techno from being killed by taken damage (minimum health will be 1).
+  - `Bounty` can be used to override the option with the same name in TechnoTypes, enabling it for TechnoTypes that originally did not have the Bounty Hunter ability, or conversely, disabling it.
```
TaranDahl, line 62: It would be best if the value of the holder's bounty could also be adjusted through AE.

src/Ext/Techno/Hooks.cpp
```diff
@@ -1856,3 +1856,14 @@ DEFINE_FUNCTION_JUMP(VTABLE, 0x7F5F30, FootClass_GetThreatValue_Wrapper);  // Un
 DEFINE_FUNCTION_JUMP(VTABLE, 0x7E417C, Building_GetThreatValue_Wrapper);   // BuildingClass
 
 #pragma endregion
+
+DEFINE_HOOK(0x702E64, TechnoClass_RegisterDestruction_Bounty, 0x6)
+{
+	GET(TechnoClass*, pKiller, EDI);
```
TaranDahl, line 1937: It would be best to allow the acceptance of the Killer of HouseClass.

TaranDahl (approved):
Several non-essential suggestions.

mevitar:
If Phobos' bounty logic uses Bounty=yes/no on techno just like the Ares' one, doesn't that mean activating it on a specific techno automatically activates both? Not only that, both are also using Bounty.Value=, from which i assume i would get duplicated values by default.

How is someone supposed to use only Phobos bounty when both are using same tags?
This isn't like Phobos' AttachEffect, which required the modder to specify a separate AttachEffectType the object uses, so the modder could easily tell which one is being used and where.

Either disable Ares' bounty completely when Phobos one is enabled, or make a better way for the modder to choose which one is being used (like changing it to a separate type like Phobos' AttachEffect, or change name to something else, like Pillager or NewBounty).

And i think Bounty.Default= needs to be renamed. I understand it is the default value for all techno types, but the name is too generic and surely there has to be a better name for it (something like Bounty.EnableForAll= ?).

PR #2124, opened 2026-03-01
FallingDownDamage Extension

Fly-Star-him:
- Fixed an issue where hover vehicles could not be destroyed after malfunctioning on water surfaces.
- Fixed an issue where parachute units would die upon landing if bridges were destroyed during their descent.

 - `FallingDownDamage.AllowEMP` determines whether units incapacitated by EMP can use `FallingDownDamage/FallingDownDamage.Water`. The default value is `true`.

---------

运输船修复
- `HoverDrownable` allows customization of whether hover vehicles crash and cause death when deactivated.

In `rulesmd.ini`:
```ini
[SOMEVEHICLE]           ; VehicleType
HoverDrownable=true     ; boolean
```


docs/New-or-Enhanced-Logics.md
```diff
@@ -2228,6 +2228,16 @@ Ammo.DeployUnlockMinimumAmount=-1  ; integer
...
 ```
 
+### Custom hover vehicles shutdown drowning death
+
+- `HoverDrowned` allows customization of whether hover vehicles crash and cause death when deactivated.
+
+In `rulesmd.ini`:
+```ini
+[SOMEVEHICLE]           ; VehicleType
+HoverDrowned=true       ; boolean
```
TaranDahl: `HoverDrownable`

src/Ext/Techno/Body.h
```diff
@@ -100,6 +100,9 @@ class TechnoExt
 
 		bool JumpjetStraightAscend; // Is set to true jumpjet units will ascend straight and do not adjust rotation or position during it.
 
+		bool OnParachuted;
```
TaranDahl: I think this is not the correct way.
Vanilla treats every attached anim as a parachute. This is definitely a bug.
You should correct the maintenance method of `HasParachute` instead of reinventing the wheel.

TaranDahl (changes requested):

src/Ext/Techno/Body.h
Coronia: I think for this case it's still better to use a standalone condition. Even if `HasParachute` is applied weirdly, there's still a high chance that WW intended to use it for some cases (e.g. prevent Chronoshifted units to die from drop in some scenatios). So it's risky to simply change its usage unless we can sort them all out and handle them one by one, which is beyond this fix's range

And on the other hand, if we're trying to check if the unit is under this specific EMP'd falling status, simply using `HasParachute` or `IsUnderEMP()` alike is not 100% enough or correct. It has to be a condition that was set when the falling process is happening, instead of something like `HasParachute` which was set during the application of an anim

TaranDahl: All right. But at least mark the TODO near that property. 
For example, "This is just a temporary patch. TODO: fully check HasParachuted and correct its maintenance method."

TaranDahl (approved):

DeathFishAtEase:
> Fixed an issue where hover vehicles could not be destroyed after malfunctioning on water surfaces (by FlyStar)

@Fly-Star-him Since when was this Phobos bug introduced? Does it exist in v0.4? Or in DevBuild#48? If it exists in neither, then this log entry would be redundant. If this bug was introduced by d87d1661d6b46284c3287bd7b3a43defefdc8f0d, then it has already been rolled back, and this PR cannot fix a problem that no longer exists—it would also be redundant.

Fly-Star-him:
> > Fixed an issue where hover vehicles could not be destroyed after malfunctioning on water surfaces (by FlyStar)
> 
> @Fly-Star-him Since when was this Phobos bug introduced? Does it exist in v0.4? Or in DevBuild#48? If it exists in neither, then this log entry would be redundant. If this bug was introduced by [d87d166](https://github.com/Phobos-developers/Phobos/commit/d87d1661d6b46284c3287bd7b3a43defefdc8f0d), then it has already been rolled back, and this PR cannot fix a problem that no longer exists—it would also be redundant.

After the feature was introduced in DevBuild#47, the bug could be triggered. [d87d166](https://github.com/Phobos-developers/Phobos/commit/d87d1661d6b46284c3287bd7b3a43defefdc8f0d) merely made it more obvious.

PR #2128, opened 2026-03-04
Fix the bug where warhead flags are read twice

DeathFishAtEase:
Does anyone know why repeated loading is necessary? I discovered this while handling #2093 and suspect it is a bug.
By switching to the `develop` branch and adding log in `WarheadTypeExt::ExtData::LoadFromINIFile`, it can be observed that each warhead is loaded twice.
For example, based on 6035b7b86f5f51a1bf8c2d493886114315b2661b, make the following modifications:
```diff
DEFINE_HOOK(0x75DEA0, WarheadTypeClass_LoadFromINI, 0x5)
{
	GET(WarheadTypeClass*, pItem, ESI);
	GET_STACK(CCINIClass*, pINI, 0x150);

+	static std::map<uintptr_t, int> iniCallCount;
+	uintptr_t iniAddr = reinterpret_cast<uintptr_t>(pINI);
+	int count = ++iniCallCount[iniAddr];
+	Debug::Log("[DEBUG] LoadFromINIFile for [%s] count: %d\n", pItem->ID, count);

	WarheadTypeExt::ExtMap.LoadFromINI(pItem, pINI);

	return 0;
}
```

In `debug.log`, it can be observed:
image


TaranDahl:
What exactly causes it to be read twice?

DeathFishAtEase:
> What exactly causes it to be read twice?

See 0x66919B.

Later, it was changed to skip this part via `LJMP, 0x66919B, 0x6691B7`, which theoretically handles it more thoroughly than the current static check. However, after that, when adding a Log directly in the hook `WarheadTypeClass_LoadFromINI`, it was still output twice. Currently, a better solution is still being sought.

TaranDahl:
If it's not easy to solve, it might as well be left alone.
LoadFromINI isn't a performance-sensitive function.

DeathFishAtEase:
> If it's not easy to solve, it might as well be left alone. LoadFromINI isn't a performance-sensitive function.

If this problem is left unresolved, all the old flags that have been replaced on the warhead will create identical Log entries twice. If the problem exists, it would be better to resolve it as thoroughly as possible.

TaranDahl (approved):

PR #2131, opened 2026-03-10
Export interface for external call

TaranDahl:
### Interoperability

Phobos has opened the external interfaces of some key components. If you are also developing your own engine extension and wish to use Phobos at the same time, please check out [Interop](https://github.com/Phobos-developers/Phobos/tree/develop/src/Interop).


src/Interop/TechnoExt.cpp
```diff
@@ -0,0 +1,10 @@
+#include "TechnoExt.h"
+#include <Ext/Techno/Body.h>
+
+extern "C" __declspec(dllexport) bool __stdcall ConvertToType_Phobos(void* pThis, void* toType)
+{
+	auto pTechno = static_cast<FootClass*>(pThis);
+	auto pType = static_cast<TechnoTypeClass*>(toType);
+
+	return TechnoExt::ConvertToType(pTechno, pType);
```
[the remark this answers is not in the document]
TaranDahl: You've passed in a null pointer. What can I say?

Metadorius:
I am not sure why all pointers are `void*`?

ZivDero:
> I am not sure why all pointers are `void*`?

My bet would be that to make it so that this does not depend on YRpp (in case someone is using their own implementation of YRpp, perhaps)

TaranDahl:
> I am not sure why all pointers are `void*`?

Copilot said that if I use TechnoClass*, it will break the P/Invoke compatibility. When calling in C#, I can't directly use IntPtr. I have to do type mapping, otherwise the call will fail.

src/Interop/AttachEffect.cpp
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	if (attachInfo.AttachTypes.empty())
+		return 0;
+
+	AEAttachParams params;
+	params.DurationOverride = durationOverride;
+	params.Delay = delay;
+	params.InitialDelay = initialDelay;
+	params.RecreationDelay = recreationDelay;
```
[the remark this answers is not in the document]
TaranDahl: Changed

docs/Whats-New.md
```diff
@@ -545,6 +545,7 @@ New:
 - [Taunt warhead](New-or-Enhanced-Logics.md#taunt-warhead) (by TaranDahl)
 - [Custom hover vehicles shutdown drowning death](New-or-Enhanced-Logics.md#custom-hover-vehicles-shutdown-drowning-death) (by FlyStar)
 - [SHP turret vehicles support the use of `*tur.shp` files](New-or-Enhanced-Logics.md#independent-shp-vehicle-turret-files) (by FlyStar)
+- [Export interface for external call](../README.md#interoperability) (by TaranDahl)
```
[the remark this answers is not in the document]
TaranDahl: I don't understand Markdown. Is there anyone who can tell me if what Copilot said is correct?

Metadorius:
> Copilot said that if I use TechnoClass*, it will break the P/Invoke compatibility. When calling in C#, I can't directly use IntPtr. I have to do type mapping, otherwise the call will fail.

This sounds like AI hallucinations...
1. You can have `unsafe` methods in which you can do whatever like pointers etc
2. C# marshaller isn't as rigid as that usually

TaranDahl:
> > Copilot said that if I use TechnoClass*, it will break the P/Invoke compatibility. When calling in C#, I can't directly use IntPtr. I have to do type mapping, otherwise the call will fail.
> 
> This sounds like AI hallucinations...
> 
> 1. You can have `unsafe` methods in which you can do whatever like pointers etc
> 2. C# marshaller isn't as rigid as that usually

Changed

docs/Whats-New.md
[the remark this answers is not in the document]
DeathFishAtEase: > I don't understand Markdown. Is there anyone who can tell me if what Copilot said is correct?

It is correct.

README.md
```diff
@@ -95,6 +95,10 @@ Credits
 
 This project was founded by [@Belonit](https://github.com/Belonit) (Gluk-v48) and [@Metadorius](https://github.com/Metadorius) (Kerbiter) in 2020, with the first public stable release in 2021. Since then it has grown into a large community project with many contributors and maintainers.
 
+### Interoperability
+
+Phobos has opened the external interfaces of some key components. If you are also developing your own engine extension and wish to use Phobos at the same time, please check out [Interop](./src/Interop).
```
DeathFishAtEase: If you want to link in a document, absolute paths should be used here.
```suggestion
Phobos has opened the external interfaces of some key components. If you are also developing your own engine extension and wish to use Phobos at the same time, please check out [Interop](https://github.com/Phobos-developers/Phobos/tree/develop/src/Interop).
```

secsome:
Macros like PHOBOS_EXPORT and PHOBOS_DECL can make the code cleaner and such macros are widely used in most projects.
Just an example:
```cpp
#define PHOBOS_DECL __stdcall
#define PHOBOS_EXPORT extern "C" __declspec(dllexport) 
```

DeathFishAtEase:
> If you are also developing your own engine extension and wish to use Phobos at the same time

Would it be more appropriate to move the Changelog entry of this PR to the *Fixes / interactions with other extensions* section?

docs/Whats-New.md
```diff
@@ -668,6 +668,7 @@ Fixes / interactions with other extensions:
 - Fixed the bug that vehicle survivor can spawn on wrong position when transport has been destroyed (by NetsuNegi)
 - Fixed the bug that building with `Explodes=yes` use Ares's rubble logic will cause it's owner cannot defeat normally (by NetsuNegi)
 - Fixed ares hook which stopped OpenTopped transports from firing if cloaked. This can now be customized (by RAZER & Morton)
+- [Export interface for external call](../README.md#interoperability) (by TaranDahl)
```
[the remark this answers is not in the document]
DeathFishAtEase: IIRC, this issue has occurred and been addressed once before.

DeathFishAtEase:
A question: how is the *Tested* status of this PR determined, as there seems to be no one using another engine extension to try out the content of this PR or other testing information.

TaranDahl:
> A question: how is the _Tested_ status of this PR determined, as there seems to be no one using another engine extension to try out the content of this PR or other testing information.

I tested the usability of the initial version myself using DP. However, the current version has not been tested yet.

src/Utilities/Macro.h
```diff
@@ -3,6 +3,31 @@
 #include <ASMMacros.h>
 #include "Patch.h"
 
+// Export / calling-convention macros for public C ABI functions
+#ifndef PHOBOS_DECL
+#ifdef _WIN32
```
secsome: No need to check _WIN32 cause the target process is ofcourse a win32 target.

What you want maybe compiler check? But as we are only supporting MSVC now. a simple define would be enough.

secsome (changes requested):

src/Utilities/Macro.h
TaranDahl: Changed.

secsome (approved):
LGTM

DeathFishAtEase (approved):
Link redirection is correct.
You may consider filling in the Chinese translation before merging. I have created standard translation entries (don't worry about introducing errors; that will be handled subsequently by the Regular Revisions PR).

debug.log
DeathFishAtEase, line 1: Suspected of having extraneous files mixed in.

Metadorius:
I think one thing to consider would be the API versioning and backwards compatibility. Usually the APIs are somehow formatting. I am not saying this needs to be done regardless but we *may* need to employ sth like that.

Also, where docs?

TaranDahl:
> I think one thing to consider would be the API versioning and backwards compatibility. Usually the APIs are somehow formatting. I am not saying this needs to be done regardless but we may need to employ sth like that.

What exactly needs to be done?

> Also, where docs?

I assume that engine developers are smart enough to read interface declarations 🤔

Metadorius:
> What exactly needs to be done?

I mean, you tell me :P

I said this as an open question, something to think about.

> I assume that engine developers are smart enough to read interface declarations 🤔

don't be one of the "just read the code" people :P 

also interface declaration doesn't say anything about how to set up your project to be able to call Phobos, and we should provide that info

Metadorius:
One thing to note: you can write documentation as docstrings, and have Sphinx (doxygen?) generate the corresponding pages. I think there was such an ability. Would be cool to look into that.

Metadorius:
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html

DeathFishAtEase:
If using Sphinx to generate documentation is not convenient, you can first try using forum posts and comments as carriers on the Phobos CN forum and GitHub PR page.

TaranDahl:
@Metadorius I asked AI to write one. Take a look at it.

README.md
```diff
@@ -95,6 +95,10 @@ Credits
 
 This project was founded by [@Belonit](https://github.com/Belonit) (Gluk-v48) and [@Metadorius](https://github.com/Metadorius) (Kerbiter) in 2020, with the first public stable release in 2021. Since then it has grown into a large community project with many contributors and maintainers.
 
+### Interoperability
+
+Phobos has opened the external interfaces of some key components. If you are also developing your own engine extension and wish to use Phobos at the same time, please check out [Interoperability](/docs/Interoperability.md).
```
DeathFishAtEase:
```suggestion
Phobos has opened the external interfaces of some key components. If you are also developing your own engine extension and wish to use Phobos at the same time, please check out [Interoperability](Interoperability.md).
```

DeathFishAtEase (commented):
The current path is incorrect.

src/Interop/Version.cpp
```diff
@@ -0,0 +1,21 @@
+#include "Version.h"
+#include "../Phobos.version.h"
+#include "Utilities/Debug.h"
+
+DEFINE_EXPORT(unsigned int, GetInteropBuildNumber)
+{
+	return BUILD_NUMBER;
```
ZivDero: Seems too imprecise (given that it's just `48` right now). Should be a number that increments, yes, but more detailed.

Metadorius: I also dislike latching on to build numbers which will go away when/if we switch to the proposed versioning

TaranDahl: > but more detailed

Is it really necessary to be specific to nightly?

TaranDahl: > I also dislike latching on to build numbers which will go away when/if we switch to the proposed versioning

Then what number should be used?

Metadorius: usually APIs are versioned separately. like v1, v2, v3...

btw, deprecated means "it still works, but a newer method exists which is recommended", not "not working"

TaranDahl:
@secsome Could you help take a look at the newly added Callback things?

TaranDahl:
@Metadorius API versioning done. Please check it out.

Metadorius (commented):
maybe we should always return something like HRESULT? PHOBOS_RESULT or something like this, with 0 for success and other ints for error codes

TaranDahl:
> maybe we should always return something like HRESULT? PHOBOS_RESULT or something like this, with 0 for success and other ints for error codes

Whose return are you referring to? The API function or the version getter?

TaranDahl:
> maybe we should always return something like HRESULT?

I doubt whether it is necessary to do so. 
We are just a completely serial stupid video game with no access control or similar mechanisms. A single bool should suffice to answer most questions.
Also, we are completely open-sourced, developers can simply look into the source code and add logs.

TaranDahl:
@ZivDero what do you think

TaranDahl:
@Metadorius Please check the things about Copilot and Action.

TaranDahl:
@ZivDero @secsome @Metadorius 
Going to merge this tonight if no further instructions.

ZivDero:
> @ZivDero what do you think

I think an HRESULT-like thing wouldn't hurt. IMO the entire API is already overkill, but if it's happening, might as well future-proof it.

TaranDahl:
> > @ZivDero what do you think
> 
> I think an HRESULT-like thing wouldn't hurt. IMO the entire API is already overkill, but if it's happening, might as well future-proof it.

Alright then. Should I return all return values via the buffer?

ZivDero:
> > > @ZivDero what do you think
> > 
> > 
> > I think an HRESULT-like thing wouldn't hurt. IMO the entire API is already overkill, but if it's happening, might as well future-proof it.
> 
> Alright then. Should I return all return values via the buffer?

Buffer? If you mean the new enum, then yeah.
Well, anything that's a bool or anything like that. Not any pointers, OFC.

Metadorius:
> Buffer? If you mean the new enum, then yeah. Well, anything that's a bool or anything like that. Not any pointers, OFC.

No, he meant that the real, non-status return value is passed using parameters by reference

TaranDahl:
> > > > @ZivDero what do you think
> > > 
> > > 
> > > I think an HRESULT-like thing wouldn't hurt. IMO the entire API is already overkill, but if it's happening, might as well future-proof it.
> > 
> > 
> > Alright then. Should I return all return values via the buffer?
> 
> Buffer? If you mean the new enum, then yeah. Well, anything that's a bool or anything like that. Not any pointers, OFC.

How to return other results without using a buffer?
C++ only supports a single return value.

ZivDero:
> > Buffer? If you mean the new enum, then yeah. Well, anything that's a bool or anything like that. Not any pointers, OFC.
> 
> No, he meant that the real, non-status return value is passed using parameters by reference

Ah, in that sense... I guess then yeah.

TaranDahl:
Going to merge this next evening if no further instructions.
@Metadorius @ZivDero @secsome

PR #2095, opened 2026-02-08
Improve 2089

NetsuNegi:

PR #2134, opened 2026-03-12
Reapply #2095 and fix reconnection error cause by it

NetsuNegi:


src/Ext/Event/Body.cpp
```diff
@@ -35,16 +43,62 @@ bool EventExt::IsValidType(EventTypeExt type)
...
+	else
+	{
+		pSource->SendToFirstLink(RadioCommand::NotifyUnlink);
+	}
+
+	pSource->QueueUpToEnter = nullptr;
+	pSource->LastDestination = nullptr;
+
+	if (const auto pManager = pSource->SlaveManager)
+		reinterpret_cast<void(__thiscall*)(SlaveManagerClass*)>(0x6B0C80)(pManager); // pManager->AllGuard();
```
secsome: Add the corresponding function in YRpp is a better solution

src/Ext/Event/Body.cpp
```diff
@@ -35,16 +43,62 @@ bool EventExt::IsValidType(EventTypeExt type)
...
+	pSource->SetTarget(nullptr);
+	pSource->SetArchiveTarget(nullptr);
+
+	const auto pObject = this->ApproachObject.Target.As_Object();
+
+	if (!pObject)
+		return;
+
+	pSource->Target = pObject;
+	pSource->vt_entry_53C(0);
```
secsome: And this function

src/Ext/Event/Body.cpp
```diff
@@ -35,16 +43,62 @@ bool EventExt::IsValidType(EventTypeExt type)
...
+		const auto pLink = abstract_cast<BuildingClass*>(pSource->GetNthLink());
+
+		if (pLink && pLink->IsAlive && pLink->Type->DockUnload)
+		{
+			pSource->SendToFirstLink(RadioCommand::NotifyUnlink);
+			pSource->IsTether = false;
+		}
+	}
+	else
+	{
```
secsome, line 69: bracket style

secsome (changes requested):

src/Ext/Event/Body.cpp
NetsuNegi: what ?

secsome, line 69:
```cpp
else
{
	pSource->SendToFirstLink(RadioCommand::NotifyUnlink);
}
```
to
```cpp
else
	pSource->SendToFirstLink(RadioCommand::NotifyUnlink);
```

Coronia:
didn't desync in our test so far, and the previous issues are solved. More testing is appreciated tho

src/Ext/Event/Body.cpp
Coronia: I think it should be in line with the if statement, so the bracket could be kept

TaranDahl:
I think we should find out exactly what processing was missing before that caused the desync. 
This should be documented in the comments to remind other developers.

TaranDahl (approved):
The code lgtm. But the issue above should be addressed.

Coronia:
> I think we should find out exactly what processing was missing before that caused the desync. This should be documented in the comments to remind other developers.

We can't even figure out what exactly caused the desync yet, but just take reference of other event codes and add missing checks as much as possible (e.g. planning node, tether etc). So sadly there's no clear document as of now

NetsuNegi:
> I think we should find out exactly what processing was missing before that caused the desync. This should be documented in the comments to remind other developers.

The most important code to resolve this desync seems like `pSource->SetTarget(nullptr)`.
I don't known why, but when I removed it, tester said that desync happend again.

TaranDahl:
> > I think we should find out exactly what processing was missing before that caused the desync. This should be documented in the comments to remind other developers.
> 
> The most important code to resolve this desync seems like `pSource->SetTarget(nullptr)`. I don't known why, but when I removed it, tester said that desync happend again.

Should comment this in the code.

PR #1827, opened 2025-08-21
Add contribution process description

Metadorius:

PR #2143, opened 2026-03-22
[Vanilla Fix] Fixed the bug that crashing technos can be attack and gain expirence

NetsuNegi:


TaranDahl:
What is the difference between Infantry and Unit? 
Why does Infantry have this bug while Unit does not?
I think this correction should be placed at the position that causes Infantry and Unit to exhibit different behaviors. The current Hook position does not seem to distinguish between Infantry and Unit.

TaranDahl (changes requested):
Please address the issue above.

Coronia (approved):
@TaranDahl please review it again, it already has those issues commented

TaranDahl:
Reposted from [Phobos CN](https://www.phoboscn.top/t/topic/274/11?u=航味麻酱_tarandahl) :

[quote="NetsuNegi, post:10, topic:274"]
The infantry godmode code is at `0x51BC96`, and I don't know what it's for.
[/quote]

[quote="NetsuNegi, post:10, topic:274"]
However, when the vehicle sinks, it also locks at 1 HP.
[/quote]

Have you tried removing them?

NetsuNegi:
> Reposted from [Phobos CN](https://www.phoboscn.top/t/topic/274/11?u=%E8%88%AA%E5%91%B3%E9%BA%BB%E9%85%B1_tarandahl) :
> 
> [quote="NetsuNegi, post:10, topic:274"] The infantry godmode code is at `0x51BC96`, and I don't know what it's for. [/quote]
> 
> [quote="NetsuNegi, post:10, topic:274"] However, when the vehicle sinks, it also locks at 1 HP. [/quote]
> 
> Have you tried removing them?

No and I don't want try it.
I think this should open a separate pull request.

TaranDahl:
As mentioned above, I believe it fails to address the problem at its root cause, so the way of solving the problem is incorrect, and this is a relevant issue.
The author considers this an irrelevant issue and should be addressed elsewhere.
@Metadorius You decide whether this problem should be solved here.

src/Misc/Hooks.BugFixes.cpp
```diff
@@ -70,13 +70,21 @@ DEFINE_HOOK(0x5F53AA, ObjectClass_ReceiveDamage_DyingFix, 0x6)
...
+// See 0x5F578D and 0x5F579A
+// 1 HP sinking technos can pass inspection, but actually it should not
+DEFINE_HOOK(0x4D7413, FootClass_ReceiveDamage_DyingFix, 0x8)
 {
+	enum { ReturnFromFunction = 0x4D74CF };
+
 	GET(FootClass*, pThis, ESI);
-	GET(const DamageState, result, EAX);
 
-	if (result != DamageState::PostMortem && (pThis->IsSinking || (!pThis->IsAttackedByLocomotor && pThis->IsCrashing)))
```
Metadorius, line 78: why exactly did you remove this complicated condition?

Metadorius:
not sure why this got closed, I was simply asking

TaranDahl:
Suggestion, if you don't want to be forced to make major changes after submitting a PR:

> **talk with them about key design aspects before you even submit a contribution**

[In PR 1827.](https://github.com/Phobos-developers/Phobos/pull/1827/changes#diff-0474e525df6c10faef7f4f0dbee275b9f58d1312320b5a42d04be66850ec32a7R60)

NetsuNegi:
> not sure why this got closed, I was simply asking

The closure of this pull request is not related to your ask.

PR #2146, opened 2026-03-23
[Vanilla Fix] Fixed the bug that low-air taking off / landing objects will receive twice damage

NetsuNegi:


TaranDahl (changes requested):
I think this should not be implemented like that.
WW has established two separate Trackers, namely the FirstObject linked list and ATC, to track the army and air force respectively. The essential reason for this issue is that the maintenance of these two Trackers is independent, and their maintenance standards are also irrelevant.
I believe the correct implementation method is to rectify their maintenance approaches. Units with a value greater than or equal to 208 will enter the ATC. Units with a value greater than or equal to 0 and less than 208 will enter the cell linked list.

TaranDahl (approved):

tyuah8:
Does anyone know what is a reliable way to duplicate this bug?  I've tried detonating a large AOE projectile in air just when a flying unit is taking off, but the bug is not occurring.

PR #2149, opened 2026-03-24
[Vanilla Fix] Properly remove aircraft, jumpjet and rockets offmap when dead

CnCRAZER:
This pull request addresses a critical bug where units with Fly, Jumpjet, or Rocket locomotors that are destroyed while crashing off-map were not being properly cleaned up. This could lead to production slots being permanently blocked and unit limits being incorrectly counted. The fix ensures these units are now correctly removed from the game, and the documentation has been updated to reflect this change.

Bug fix implementation:

* Added hooks in `src/Misc/Hooks.BugFixes.cpp` to ensure that off-map crashing units using `FlyLocomotionClass`, `JumpjetLocomotionClass`, and `RocketLocomotionClass` are treated as having touched the ground, triggering proper cleanup and resource release.

Documentation updates:

* Updated `docs/Fixed-or-Improved-Logics.md` to describe the fix for units with Fly, Jumpjet, or Rocket locomotors being destroyed off-map and not being cleaned up, which could block production slots and unit limits.
* Added the fix to the changelog in `docs/Whats-New.md`, crediting the contributor.
* Updated `CREDITS.md` to include this bug fix in the list of individual contributions.


NetsuNegi (approved):

TaranDahl (changes requested):
What causes off-map or not to have different effects on crash behavior? The position you hooked doesn't seem to have such a difference.

CnCRAZER:
> What causes off-map or not to have different effects on crash behavior? The position you hooked doesn't seem to have such a difference.

AFAIK offmap aircraft types are only removed if they are spawned from spawner or from teamtypes exiting the map. Running uninit to clear the unit from map is also only checked if aircraft type is within radar bounds

CnCRAZER:
check the discord Phobos chat for the conversation regarding it

TaranDahl:
[Repost: ](https://discord.com/channels/338055131666972683/773636942775582720/1486204880237367426)

I checked this. I think the location of the fix in your PR is incorrect.
I believe the root cause of the problem is at 4CD748. If it is outside the map, loco will refuse to set the location. Setting the layer and Mark are also rejected. It is unclear why such a judgment is made.
I think the correct approach should be to make a judgment at 4CD797, treating a crash outside the map the same as touching the ground, and performing subsequent detonation and unit deletion operations.
In addition, I'm afraid the crash of jumpjet has a similar problem (54CBC2). You'd better fix it while you're at it.

TaranDahl:
Oh. And there's also Rocket (662FD5).

CnCRAZER:
Very interesting. I'll take a look at your suggestions

CnCRAZER:
> Oh. And there's also Rocket (662FD5).

Rocket and Jumpjet are now implemented at the addresses you suggested. 

Somebody tested the dll and sent me this link here https://drive.google.com/file/d/1MXKb1bCpaL-ZEyU6S_tM41eiYjPuFEMV/view

You can see the aircraft dying offmap and being able to rebuild them now

src/Misc/Hooks.BugFixes.cpp
```diff
@@ -2874,6 +2874,59 @@ DEFINE_HOOK(0x4DB874, FootClass_SetLocation_Extra, 0xA)
...
+// The fix: at each locomotor's height/health check, treat off-map as ground-touch.
+// FlyLocomotionClass::Process - height check after crash descent calculation.
+// If off-map, skip the height > 0 check and go straight to ground-touch cleanup.
+DEFINE_HOOK(0x4CD797, FlyLocomotionClass_CrashDescent_OffMap, 0x5)
+{
+	enum { GroundTouchCleanup = 0x4CD7AA };
+
+	GET(FlyLocomotionClass*, pThis, ESI);
+
+	if (!MapClass::Instance.IsWithinUsableArea(pThis->LinkedTo->GetCoords()))
```
TaranDahl: Is there no need to check IsCrashing here? I haven't carefully checked the previous code, so it can be ignored if it's actually not needed.

src/Misc/Hooks.BugFixes.cpp
```diff
@@ -2874,6 +2874,59 @@ DEFINE_HOOK(0x4DB874, FootClass_SetLocation_Extra, 0xA)
...
+	enum { IsCrashingCheck = 0x54CC36 };
+
+	GET(JumpjetLocomotionClass*, pThis, EDI);
+
+	if (!MapClass::Instance.IsWithinUsableArea(pThis->LinkedTo->GetCoords()))
+	{
+		// Replicate the stack init from the stolen bytes (mov byte ptr [esp+21h], 0)
+		// so the "fell on something" flag is properly zeroed for the crash path.
+		REF_STACK(BYTE, fellOnSomething, 0x21);
+		fellOnSomething = 0;
```
TaranDahl: Are you certain? Doubt if the stack offset calculation is correct. Also, why don't you use the STACK_OFFSET macro?

TaranDahl (changes requested):

src/Misc/Hooks.BugFixes.cpp
CnCRAZER: You're right, whoops!

src/Misc/Hooks.BugFixes.cpp
```diff
@@ -2905,7 +2909,7 @@ DEFINE_HOOK(0x54CC16, JumpjetLocomotionClass_CrashDescent_OffMap, 0x8)
 	{
 		// Replicate the stack init from the stolen bytes (mov byte ptr [esp+21h], 0)
 		// so the "fell on something" flag is properly zeroed for the crash path.
-		REF_STACK(BYTE, fellOnSomething, 0x21);
+		REF_STACK(BYTE, fellOnSomething, STACK_OFFSET(0x34, -0x13));
```
TaranDahl: I assume you want to change the var below.
I think it should be STACK_OFFSET(0x34, -0x23).
Image

CnCRAZER: Sure thing can try this 👍🏻

CnCRAZER: > I assume you want to change the var below. I think it should be STACK_OFFSET(0x34, -0x23). Image

I believe you are indeed correct on these values. I just updated in latest commit

TaranDahl (approved):

CnCRAZER:
Btw @TaranDahl thanks for the time you spent suggesting changes and reviewing this 👍🏻

TaranDahl:
You are welcome 😀

DeathFishAtEase:
> Oh. And there's also Rocket (662FD5).

As far as I know, the hook related to Rocket in this PR caused spawned missiles to detonate immediately when launched at the map boundary, like this.

image

*from [Phobos CN Forum/681](https://www.phoboscn.top/t/topic/681?u=noble_fish)*

To avoid this issue, Ollerus removed it in a recent commit 01b28abf554ca265dff9c2d5f22a1b9b317b2820 on the `develop` branch.

I would like to know why you proposed this change, or what kind of abnormal situation might occur after Ollerus removed it that the hook was intended to solve? Because I see the above discussion focused on issues caused by crashing after death, but missiles do not crash; they explode in mid-air when destroyed. So does it really have the same problem as other locomotives crashing?

Issue #1783, opened 2025-07-22
Soviet Harvester stops and does not continue to work


CCCP84:
### Description

After all available gold has been collected, the Soviet Harvester stops and will not continue working until it is forced to start again.
the alliance harvester does not have this problem. After all the gold is collected it also stops but when the drill extracts new gold the harvester automatically starts working.
We play long games on a map with islands and soviet players are forced to constantly control all islands every 5 minutes during a long game and resume mining this is a very routine and stupid work in addition players forget to control this during the battle and simply lose money because if the Harvester does not mine gold it does not accumulate due to a bad algorithm and is simply lost.
This is a very annoying problem that occurs constantly, players are tired of it.
I hope the solution to this problem is not very complicated and it will be solved.

### INI code

```ini

```

### Steps to reproduce

1.
2.
3.
...

### Expected behaviour

when the drill extracts new gold the harvester automatically starts working

### Actual behaviour

Soviet Harvester stops and will not continue working until it is forced to start again

### Checklist

- [x] The issue is not introduced by Phobos or any other engine extension, such as Ares, Kratos etc.
- [x] The issue wasn't fixed in the most recent version of Ares/Phobos yet.
- [x] I agree to elaborate the details if requested and provide thorough testing if the bugfix is implemented.
- [x] I added a very descriptive title to this issue.
- [x] I used the GitHub search and read the issue list to find a similar issue and didn't find it.
- [x] I have attached as much information as possible *(screenshots, gifs, videos, debug and exception logs, etc)*.

CCCP84:
Is there a way to fix this in the map file?
I would be happy to solve this problem at least temporarily, until it is solved globally.
@ZivDero

ahsy666:
[CMIN]
Teleporter=yes
------
This code make CMIN auto mining. When it stops BESIDE the dock, it will get chill and go back to work immediately. If you remove this, it will work like HARV, stop working permanently after all ores gathered.
Seems that there's no way to fix it with ini. I tried a lot.

CCCP84:
> ## [CMIN]
> Teleporter=yes
> This code make CMIN auto mining. When it stops BESIDE the dock, it will get chill and go back to work immediately. If you remove this, it will work like HARV, stop working permanently after all ores gathered. Seems that there's no way to fix it with ini. I tried a lot.

I added this and the harv started to stop after collecting ore

CCCP84:
Why did you decide that this line is the reason for resuming work?

[...]

PR #2156, opened 2026-03-27
Fixed the bug where non-Teleporter miners would not return to work after minerals are depleted and then regenerated.

TaranDahl:


TaranDahl:
Tested by https://github.com/Phobos-developers/Phobos/issues/1783#issuecomment-4144916263

CnCRAZER:
It is currently late for me as I am American but this weekend I can test and confirm this change 👍🏻

NetsuNegi (approved):

PR #2157, opened 2026-03-28
Miners back to work when ore regenerated

TaranDahl:
- Now miners will no longer withdraw from the Harvest mission due to mineral depletion and will periodically attempt to return to work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ out
 
 # Python virtual environment for docs
 .venv/
+debug.log
--- a/.gitignore
+++ b/.gitignore
@@ -27,4 +27,3 @@ out
 
 # Python virtual environment for docs
 .venv/
-debug.log
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -732,6 +732,7 @@ This page lists all the individual contributions to the project by their author.
   - Fix the issue that the Jumpjet must end its movement before starting the next mission
   - Taunt warhead
   - Fix the bug where non-Teleporter miners would not return to work after minerals are depleted and then regenerated
+  - Miners back to work when ore regenerated
 - **solar-III (凤九歌)**
   - Target scanning delay customization (documentation)
   - Skip target scanning function calling for unarmed technos (documentation)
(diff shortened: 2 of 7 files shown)


CCCP84:
I checked it works, the harvesters come to life one after the other, not at the same time, it looks very good

CCCP84:
I think this should be made permanently enabled and not so that players have to enable it in the map or something else. no one will want to disable this option. there is no reason for this

CCCP84:
Besides, it seems to solve this problem too: 

https://github.com/CnCNet/cncnet-yr-client-package/issues/303

in general this is a very significant improvement over the original game. There's always been a problem with these harvesters. In this regard, everyone has already gotten used to living with it. now the convenience will be significantly increased

TaranDahl:
> I think this should be made permanently enabled

Fine by me. What do you think? @Metadorius

CCCP84:
How does it work? Is there some kind of timer that checks for ore at certain times? If so, you can make the delay much longer because checking more frequently is unnecessary and could cause lags with a lot of harvesters. Or no? I think checking every one or 1.5 minutes would be fine.
Although apparently there is no timer here, right?

TaranDahl:
> How does it work? Is there some kind of timer that checks for ore at certain times? 

Yeah. The mission timer is used for the interval.

> If so, you can make the delay much longer because checking more frequently is unnecessary and could cause lags with a lot of harvesters. Or no?

I used a hardcoded value, 100 frames. Regarding whether it will cause lag, I think there's no need to worry. This check cycle is longer than the original check cycle of the "miner returns to work" behavior (which is about 1 second). If the original CMIN didn't cause lag, then the new behavior won't cause lag either.

Metadorius:
> Fine by me. What do you think? @Metadorius 

I agree

TaranDahl:
image
Fucked by git

.gitignore
```diff
@@ -27,3 +27,4 @@ out
 
 # Python virtual environment for docs
 .venv/
+debug.log
```
NetsuNegi: what's this ?

NetsuNegi (approved):

.gitignore
TaranDahl: I don't know. It should be something generated by VS, possibly related to AI. Anyway, it has nothing to do with Phobos.

NetsuNegi: Since it didn't do nothing with Phobos, I think it's better to delete it.

TaranDahl: Yeah, that is what I am doing. Modifying the gitignore is the thing to delete it from the repo.

TaranDahl: Deleted. It should be an issue with certain plugins.

PR #2160, opened 2026-04-01
Customize effects range of power plant enhancer

NetsuNegi:
In `rulesmd.ini`:
```ini
[AudioVisual]
ShowPowerPlantEnhancerRange=true   ; boolean

PowerPlantEnhancer.Range=0         ; integer, in leptons
```

In `RA2MD.INI`:
```ini
[Phobos]
ShowPowerPlantEnhancerRange=false  ; boolean
```

### DIFF
--- a/src/Ext/BuildingType/Body.cpp
+++ b/src/Ext/BuildingType/Body.cpp
@@ -40,7 +40,7 @@ int BuildingTypeExt::ExtData::GetSuperWeaponIndex(const int index) const
 	return -1;
 }
 
-std::pair<int, int> BuildingTypeExt::GetEnhancedPower(BuildingTypeClass* pBuilding, int output, HouseClass* pHouse, const CoordStruct* pCoords)
+std::pair<int, int> BuildingTypeExt::GetEnhancedPower(BuildingTypeClass* pBuilding, int output, HouseClass* pHouse, BuildingClass* pPowerPlant)
 {
 	const auto pHouseExt = HouseExt::ExtMap.Find(pHouse);
 	int amount = 0;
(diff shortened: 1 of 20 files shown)


src/Ext/BuildingType/Body.cpp
```diff
@@ -40,22 +40,42 @@ int BuildingTypeExt::ExtData::GetSuperWeaponIndex(const int index) const
...
+		const auto pEnhancerType = pEnhancer->Type;
+		const auto pEnhancerTypeExt = BuildingTypeExt::ExtMap.Find(pEnhancerType);
+
+		if (!pEnhancerTypeExt->PowerPlantEnhancer_Buildings.Contains(pBuilding))
+			continue;
 
-		if (pTypeExt->PowerPlantEnhancer_Buildings.Contains(pBuilding))
+		const int range = pEnhancerTypeExt->PowerPlantEnhancer_Range.Get();
+
+		if (range > 0 && (!pPowerPlant || pPowerPlant->DistanceFrom(pEnhancer) > range))
```
TaranDahl: I think this distance calculation is incorrect. It does not match the way you draw the indicator.
The range shown by your indicator does not take the Enhancer foundation into account, whereas the calculation here does.
When the Enhancer has a large foundation, this will result in a mismatch between the visual display and the actual logic. For example, if you set the range of GACNST to 1, the indicator will show a range that lies entirely within the foundation area, while the actual effective range is a circle around GACNST.
I believe pEnhancer->DistanceFrom(pPowerPlant) is the correct approach.

TaranDahl (changes requested):

TaranDahl (approved):

PR #2167, opened 2026-04-10
Fixed the incorrect mission switching in infantry EnterIdleMode

TaranDahl:

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -737,6 +737,7 @@ This page lists all the individual contributions to the project by their author.
   - Miners back to work when ore regenerated
   - Allow disable an over-optimization in targeting
   - Extra threat
+  - Fixed the incorrect mission switching in infantry EnterIdleMode
 - **solar-III (凤九歌)**
   - Target scanning delay customization (documentation)
   - Skip target scanning function calling for unarmed technos (documentation)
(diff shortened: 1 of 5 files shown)


TaranDahl:
This should fix the bug you mentioned in https://github.com/Phobos-developers/Phobos/commit/819afa2d1ddb2621ce6387befc3a8d22f941504f#commitcomment-181924072 .
Considering that WW's mission system is a total mess, it's best to conduct more tests.
@Coronia

TaranDahl:
I have checked EnterIdleMode for RA1, and infantry units also use MissionAttack.
The comments indicate that EnterIdleMode for both Unit and Infantry was implemented by the same developer, ruling out inconsistencies caused by the author being unfamiliar with existing code. However, the specific reason for this design remains unknown.

Coronia:
the issue of flying infantry indeed resolved by this, and so far no weird behavior of infantry has occur in skirmish and mission scenarios. But given this is a generic change to infantry behavior, there might still be some cases that haven't covered yet, which needs to be figured out with more testing

Starkku:
Maybe figure out why the jumpjet change leads to this being a problem in the first place.

TaranDahl:
> Maybe figure out why the jumpjet change leads to this being a problem in the first place.

In vanilla, MissionMove does not end naturally, so EnterIdleMode is not triggered, which naturally does not cause this issue.
After the fix, a stationary Jumpjet will complete MissionMove and then trigger EnterIdleMode. If it has a target at this point, it will enter MissionAttack due to this issue. If not, it will enter MissionMove because of the Destination and return to its initial state.

TaranDahl:
> In vanilla, MissionMove does not end naturally

BTW, I think this should also be regarded as a bug. A unit that finished a move mission should enter guard or guardarea mission, instead of being stuck in the move mission.
However, this requires more effort.

Coronia (approved):
so far seems to be correct in both logic and testing

PR #2172, opened 2026-04-15
Fix AI team recruitment inconsistency causing underfilled teams

handama:
In certain situations, the AI incorrectly assumes that some units on the map are recruitable when forming a team, leading to teams that can never reach full strength. For example, when there are 3 GI on the map with recruitable flags (RecruitableA, RecruitableB) set to 1,1; 1,0; and 0,0, and an AI team with Autocreate enabled is created via trigger action 4 (Create Team), only the units with 1,1 should be considered recruitable, and the AI should produce 2 additional GIs to fill the team. However, during production planning (0x4DA230), the AI incorrectly treats units with 1,0 or 0,0 as recruitable and therefore only produces 1 GI. During actual recruitment (0x6EA610), these GIs are skipped due to stricter conditions, leaving the team permanently underfilled. Notably, if the intermediate GI (1,0) is destroyed, the AI will then correctly produce the missing GI and the team can finally form. 
Snipaste_2026-04-15_15-12-31
[testmap.zip](https://github.com/user-attachments/files/26739838/all01umd.zip)

This PR addresses it by modifying 0x4DA230 so that its conditions match those of 0x6EA610 as closely as possible to avoid such inconsistencies.


Coronia:
as far as I know, if Recruitable A and B are both set to 0, the unit will not be considered as occupying a team member. AI will still skip it when recruiting a team and build sufficient units instead

Making it as 1,0 is indeed problematic but maybe a toggle is needed for this fix, since it'll alter the AI behaviors of existing missions, making their AI unintentionally stronger

handama:
> as far as I know, if Recruitable A and B are both set to 0, the unit will not be considered as occupying a team member. AI will still skip it when recruiting a team and build sufficient units instead
> 
> Making it as 1,0 is indeed problematic but maybe a toggle is needed for this fix, since it'll alter the AI behaviors of existing missions, making their AI unintentionally stronger

I understand the concern about potentially affecting AI strength, but I don’t think a toggle is necessary here. This change mainly fixes a case which is more of a bug than intended behavior. The situations where this occurs should also be relatively limited, so the overall impact on existing missions is likely small. What do you think?

TaranDahl (approved):
The code looks good to me.
I have a shallow understanding of the Team system. So I think I am not qualified for judging the design aspect.

src/Ext/Techno/Body.cpp
```diff
@@ -884,6 +884,35 @@ void TechnoExt::ClickedApproachObject(FootClass* pThis, ObjectClass* pObject)
 	event.AddEvent();
 }
 
+bool TechnoExt::CanBeRecruitedFix(FootClass* pThis, HouseClass* pHouse)
+{
+	if (!pThis || !pHouse) return false;
```
NetsuNegi: no need check `pThis` here

TaranDahl:
@Metadorius @Starkku @ZivDero can anyone help judging the design concept?

DeathFishAtEase:
Tested by: https://www.phoboscn.top/t/topic/361/9
image

TaranDahl:
Given that no one is willing to make further comments, this PR will be merged in 24 hours.

DeathFishAtEase:
Strange, ef120cd is just a documentation change (resolve merge conflicts), but Nightly failed to build successfully. However, when I pull the branch of that PR locally for testing, it builds successfully.

> error C1083: Cannot open include file: 'atlbase.h': No such file or directory [D:\a\Phobos\Phobos\Phobos.vcxproj]

TaranDahl:
Could it possibly be this one that's causing the problem?
https://github.com/Phobos-developers/Phobos/commit/055cdc22592b26d8fd8da0d8ed224ed6e8611783

DeathFishAtEase:
> Could it possibly be this one that's causing the problem? [055cdc2](https://github.com/Phobos-developers/Phobos/commit/055cdc22592b26d8fd8da0d8ed224ed6e8611783)

Not that one. I also suspected this, so I tested in another repository. The result was that whether or not Revert that commit, [the build succeeded](https://github.com/FreeWindor/PBS_R1/pull/5#issuecomment-4380519310). I even reset the `FreeWindor:develop` branch of that repository to be exactly the same as `Phobos-developers:develop` (without Revert or Reapply) for testing, but the build also succeeded.

Additionally, that 055cdc22592b26d8fd8da0d8ed224ed6e8611783 had already been merged into the `develop` branch during the last synchronization (dea4751a09c30c64c243aad8145b9a3fc1503167), and at that time it did not cause any problems.

After further testing, it was found that as long as the synchronization target of the branch where this PR is located is set to commit 0b21108094784ade80b4e600ebb07fda3a8d4c51 and later commits, this abnormal behavior will occur, but the cause is not clear.

TaranDahl:
@Metadorius Help

DeathFishAtEase:
The branch has been restored to the state before a series of tests were conducted, waiting for others to handle it.

Starkku (changes requested):
After review, investigation and some research it appears this PR addresses a real inconsistency and correctly fixes the problem - adjusting the recruitment-stage check to require both recruitment flags (which as an aside could be renamed in YRpp to better match their actual purpose, A is for `Autocreate=false` teams and B for true ones) to be set to true produces no false-positive results for `Autocreate=true` teams and has no impact on `Autocreate=false` teams which didn't have this issue to begin with. 

Although I am unsure how the both flags set to false case return true from `0x4DA230` when the only code path in it that can requires RecruitableA to be set to true.

Aside from that I have some nitpicks.

src/Ext/Techno/Body.cpp
```diff
@@ -884,6 +884,33 @@ void TechnoExt::ClickedApproachObject(FootClass* pThis, ObjectClass* pObject)
...
 }
 
+bool TechnoExt::CanBeRecruitedFix(FootClass* pThis, HouseClass* pHouse)
+{
+	const bool inTeam = pThis->Team != nullptr;
+	const bool available = pThis->IsAlive && pThis->Health > 0 && !pThis->InLimbo;
+	const bool wrongOwner = pThis->Owner != pHouse;
+
+	if (inTeam || !available || wrongOwner)
+		return false;
```
Starkku: Didn't check if the compiler is smart enough to make use of short-circuit evaluation here but if it isn't, could probably help it along even if it is less clean to read. Not high impact issue though.

EDIT: Compiler does optimize it quite well so it can be left as it is since it is probably cleaner to read.

src/Ext/Techno/Body.cpp
```diff
@@ -884,6 +884,33 @@ void TechnoExt::ClickedApproachObject(FootClass* pThis, ObjectClass* pObject)
...
+
+	const bool canRecruit = pThis->RecruitableA && pThis->RecruitableB;
+	if (!canRecruit)
+		return false;
```
Starkku: Probably unnecessary to make variable for these since it is a short bit of code that fits nicely on single row inside `if`. Compiler will likely optimize it away anyway but it is the principle of things.

Starkku (changes requested):

Starkku (approved):
Single row `if` statements (the condition is on single row) don't need the curly braces per project code style specs and omitting them is usually cleaner but them being there isn't a problem. LGTM I guess.

> Although I am unsure how the both flags set to false case return true from 0x4DA230 when the only code path in it that can requires RecruitableA to be set to true.

If someone is able to enlighten me on how this happened though it would be nice for understanding this part of the game better.

handama:
> Single row `if` statements (the condition is on single row) don't need the curly braces per project code style specs and omitting them is usually cleaner but them being there isn't a problem. LGTM I guess.
> 
> > Although I am unsure how the both flags set to false case return true from 0x4DA230 when the only code path in it that can requires RecruitableA to be set to true.
> 
> If someone is able to enlighten me on how this happened though it would be nice for understanding this part of the game better.

My mistake. The issue only occurs in the 1,0 case. I didn’t thoroughly double-check during the translation.

PR #2182, opened 2026-04-25
Allows defining whether passengers of an OpenTopped transport unit can fire when it is affected by a locomotor warhead

DeathFishAtEase:
~~Allow using `OpenTopped.AllowFiringIfDeactivated=false` for the case where a transport unit is being attacked by a weapon that uses a warhead with `IsLocomotor=true`.~~

Use `OpenTopped.AllowFiringIfAttackedByLocomotor` to control whether the passengers of a non-building transport unit can fire when the unit is being attacked by a weapon whose warhead has `IsLocomotor=true`

### DIFF
--- a/docs/New-or-Enhanced-Logics.md
+++ b/docs/New-or-Enhanced-Logics.md
@@ -1566,20 +1566,22 @@ DrainMoneyDisplay.OnTarget.UseDisplayIncome=        ; boolean
 In `rulesmd.ini`:
 ```ini
 [General]
-OpenTopped.DecloakToFire=true             ; boolean
+OpenTopped.AllowFiringIfAttackedByLocomotor=true  ; boolean
+OpenTopped.DecloakToFire=true                     ; boolean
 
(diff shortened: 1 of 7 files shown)


DeathFishAtEase:
Tested by https://www.phoboscn.top/t/topic/375/3
image

TaranDahl (approved):

Coronia:
I'm not sure if this needs a standalone tag, since locomotor change is a different scenario from EMP/deactivated

DeathFishAtEase:
> I'm not sure if this needs a standalone tag, since locomotor change is a different scenario from EMP/deactivated

Considering the accuracy issue of the INI flag name, it is indeed best to add a separate flag and allow setting a global default value to change the original behavior, rather than simply treating both as a deactivated state to handle them equivalently.

Starkku:
Realistically the locomotor WH specifically disabling firing could be considered a special case of 'deactivation' but semantically I suppose it is clearer to split the two because of, as you put it, the 'accuracy issues' with the INI key name.

Coronia (approved):

PR #2186, opened 2026-04-30
Fix infantry with Engineer=yes ignores cell occupier in AreaGuard, Parol and Guard mission

TaranDahl:
To be tested: If I recall correctly, the AI Engineer will automatically repair damaged surrounding buildings when in guard or area guard mission. It is necessary to confirm that this fix does not break this behavior.


TaranDahl:
Closing this because no maintainer cares about it.

PR #2191, opened 2026-05-03
Berzerk / Psychedelic duration stacking customization

Starkku:
### Berzerk (Psychedelic) duration stacking customization

- By default `Psychedelic` warheads override the current duration of the berzerk effect regardless of if the new duration is higher or lower than the current one. This can now be customized with `Psychedelic.StackingMode`, with both global setting under `[CombatDamage]` and per-Warhead customization.

In `rulesmd.ini`:
```ini
[CombatDamage]
Psychedelic.StackingMode=override  ; Stacking mode enum (override|setifzero|min|max|add|subtract|multiply|divide)

[SOMEWARHEAD]                      ; WarheadType
Psychedelic.StackingMode=          ; Stacking mode enum (override|setifzero|min|max|add|subtract|multiply|divide)
```

---------------

Dev note: The enum was generalized for potential (future) use elsewhere but is also possibly over-engineered for these kind of purposes and mixes two things together (when and how) which might not be ideal, can simplify it if need be I suppose.


TaranDahl (approved):

Coronia:
'add' option doesn't work in our test for some reason, which still behaves like override

fagonghaiwo:
all are override

Starkku:
Fixed the `[CombatDamage]` default not being correctly applied.

fagonghaiwo:
works, but min|subtract|multiply|divide need unit been Berzerk, maybe shouldnt in [CombatDamage]

Coronia:
> works, but min|subtract|multiply|divide need unit been Berzerk, maybe shouldnt in [CombatDamage]

this case should be mentioned in the doc that it'll make Berzerk can't be applied normally if the global tag is set to these values

Coronia:
there's still one remaining thing: even if we use these stacking values to make the duration of berzerk becomes 0, it'll still retint for a frame. This make it look bad when we want to use it to clear berzerk effect

Coronia:
> there's still one remaining thing: even if we use these stacking values to make the duration of berzerk becomes 0, it'll still retint for a frame. This make it look bad when we want to use it to clear berzerk effect

done and tested to work, and since 0 Damage can't make it enter berzerk already, I made it simply bail out if the duration is not positive. I don't think a customizable tag is needed here but correct me if I'm wrong

PR #2193, opened 2026-05-03
Fix the filtering mechanism of weapon range bonus effect in AE

DeathFishAtEase:
Fixed the bug where `WeaponRange.AllowWeapons` and `WeaponRange.DisallowWeapons` only support weapons listed in the `[WeaponTypes]` list


src/Utilities/TemplateDef.h
```diff
@@ -1679,6 +1679,23 @@ void __declspec(noinline) ValueableVector<T>::Read(INI_EX& parser, const char* p
 	}
 }
 
+// Specialization: use FindOrAllocate for weapon vectors, avoiding dependency on [WeaponTypes] registration
```
TaranDahl, line 1682: Why not for every abstracttype?

DeathFishAtEase: Out of the principle of minimizing impact
Or rather, I'm not sure how large a scope it is safe to expand to, and if you think it would be better to do so, I'm afraid I would have to ask for your help.

Starkku:
If you're gonna do this might as well do same for WarheadTypeClass, WW(P) used FindOrAllocate for single WH values pretty much everywhere.

Strictly speaking wasn't a bug, more of a technical limitation that is even explicitly listed in docs for some things (may need to go over those now).

DeathFishAtEase:
> Strictly speaking wasn't a bug

Ah, initially it was a user who reported to me as a bug that those two INI flags were not working, so I also followed that description.

> may need to go over those now

Specifically, what do you think should be done?

Starkku:
> Specifically, what do you think should be done?

Making sure that whatever you change the docs accurately reflect that. Check where WeaponTypes are parsed into vectors in codebase and check ifthe docs for the appropriate features mention this limitation that's now been lifted.

DeathFishAtEase:
> > Specifically, what do you think should be done?
> 
> Making sure that whatever you change the docs accurately reflect that. Check where WeaponTypes are parsed into vectors in codebase and check ifthe docs for the appropriate features mention this limitation that's now been lifted.

Currently, aside from the two AE filter flags, there are only two that use `ValueableVector<WeaponTypeClass*>` , namely `SuppressKillWeapons_Types` and `SuppressRevengeWeapons_Types` , and these two do not rely on the list being complete to function, because they compare an already existing weapon pointer (like `KillWeapon` or `RevengeWeapon` ) against the list items. Even if a weapon name is dropped from the list due to missing `[WeaponTypes]` entry, the comparison can still succeed as long as that weapon object itself was created elsewhere via `FindOrAllocate` . The list is only used as a filter, and the pointer being compared is always valid.

This is why users consider it a bug that the two filter flags of AE cannot support weapons not listed in `[WeaponTypes]` — among the only four flags that fill in weapon lists, only this pair used for the same function fails to work.

DeathFishAtEase:
> https://www.phoboscn.top/t/topic/404/2

For resources such as test videos and images, see the posts in this topic.

TaranDahl (approved):

PR #2198, opened 2026-05-05
Fix shield condition check in `IsRedSP` function

shatyuka:
This function is not being used. Fix it to prevent issues if anyone calls it in the future.


shatyuka:
Can not add label :(

Coronia (approved):

PR #2200, opened 2026-05-05
Fix tint color RGB mode conversion

shatyuka:
The game's code looks weird because the switch case forgot to add break.

Has very little impact on the vanilla game:
| TintColor | RGB565 Before | RGB565 After | RGB24 Before | RGB24 After |
| - | - | - | - | - |
| IronCurtain | 0x0000 | 0x0000 | `#000000` | `#000000` |
| ForceShield | 0x0018 | 0x0018 | `#0000C6` | `#0000C6` |
| Berserk | 0xE000 | 0xC000 | `#E70000` | `#C60000` |

Vanilla Berserk tint in game:
 Before | After 
---|---
before | after

---

Update: Custom tint color for better comparison, R=21, G=21, B=0

| RGB565 Before | RGB565 After | RGB24 Before | RGB24 After |
| - | - | - | - |
| 0xFFE0 | 0xAAA0 | `#FFFF00` | `#AD5500` |

Custom tint in game:
 Before | After 
---|---
before | after

### DIFF
--- a/src/Utilities/GeneralUtils.cpp
+++ b/src/Utilities/GeneralUtils.cpp
@@ -288,10 +288,10 @@ int GeneralUtils::GetColorFromColorAdd(int colorIndex)
 		colorValue |= (red << 6 | green) << 5 | blue;
 		break;
 	case RGBMode::RGB556:
-		colorValue |= (red << 5 | green) << 6 | blue;
+		colorValue |= (red << 5 | green >> 1) << 6 | blue;
 		break;
 	default:
-		colorValue |= (red << 5 | green) << 5 | blue;
+		colorValue |= (red << 5 | green >> 1) << 5 | blue;
 		break;
 	}
 
--- a/src/Utilities/GeneralUtils.cpp
+++ b/src/Utilities/GeneralUtils.cpp
@@ -282,13 +282,18 @@ int GeneralUtils::GetColorFromColorAdd(int colorIndex)
 	const int green = color.G;
 	const int blue = color.B;
 
-	if (Drawing::ColorMode == RGBMode::RGB565)
+	switch (Drawing::ColorMode)
+	{
+	case RGBMode::RGB565:
 		colorValue |= blue | (32 * (green | (red << 6)));
-
-	if (Drawing::ColorMode != RGBMode::RGB655)
+		break;
+	case RGBMode::RGB556:
 		colorValue |= blue | (((32 * red) | (green >> 1)) << 6);
-
-	colorValue |= blue | (32 * ((32 * red) | (green >> 1)));
+		break;
+	default:
+		colorValue |= blue | (32 * ((32 * red) | (green >> 1)));
+		break;
+	}
 
 	return colorValue;
 }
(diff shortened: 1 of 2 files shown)


Coronia (approved):

DeathFishAtEase:
How was this PR tested? I don't seem to see a comparison of Berserk in vanilla YR before and after the changes in this PR, although I'm afraid we need to compare the color codes in the game screen to see the difference.

shatyuka:
> How was this PR tested? I don't seem to see a comparison of Berserk in vanilla YR before and after the changes in this PR, although I'm afraid we need to compare the color codes in the game screen to see the difference.

Comparison attached.

src/Utilities/GeneralUtils.cpp
```diff
@@ -282,13 +282,18 @@ int GeneralUtils::GetColorFromColorAdd(int colorIndex)
...
-	if (Drawing::ColorMode != RGBMode::RGB655)
+		break;
+	case RGBMode::RGB556:
 		colorValue |= blue | (((32 * red) | (green >> 1)) << 6);
-
-	colorValue |= blue | (32 * ((32 * red) | (green >> 1)));
+		break;
+	default:
+		colorValue |= blue | (32 * ((32 * red) | (green >> 1)));
+		break;
```
TaranDahl, line 295: Why use both bit shifting and multiplication?
Wouldn't it be better to replace all multiplications with bit shifting?

shatyuka: I just didn't touch the original code.
Of course.

TaranDahl (approved):

Coronia:
> The game's code looks weird because the switch case forgot to add break.
> 
> Has very little impact on the vanilla game:
> 
> TintColor	RGB565 Before	RGB565 After	RGB24 Before	RGB24 After
> IronCurtain	0x0000	0x0000	`#000000`	`#000000`
> ForceShield	0x0018	0x0018	`#0000C6`	`#0000C6`
> Berserk	0xE000	0xC000	`#E70000`	`#C60000`
> Vanilla Berserk tint in game:
> 
> Before	After
> before	after
> Update: Custom tint color for better comparison, R=21, G=21, B=0
> 
> RGB565 Before	RGB565 After	RGB24 Before	RGB24 After
> 0xFFE0	0xAAA0	`#FFFF00`	`#AD5500`
> Custom tint in game:
> 
> Before	After
> before	after

if R and G have the same values, then the final RGB24 should still have R and G being the same I assume? While #FFFF00 is too deep of a yellow color, think it should at least be a shallow yellow after the fix?

shatyuka:
> if R and G have the same values, then the final RGB24 should still have R and G being the same I assume? While #FFFF00 is too deep of a yellow color, think it should at least be a shallow yellow after the fix?

No, it's RGB565, and the range of the green value is twice that of the red and blue.

Comment in [ColorAdd] section:
> At the level that the blitters opperate, colors are 16 bit numbers, using the 5,6,5 RGB scheme. This
means that values of red range from 0 to 31, values of green from 0 to 63, and values of blue from 0
to 31.

NetsuNegi (approved):

PR #2202, opened 2026-05-09
Aux technos and TechLevel requirement of superweapon

NetsuNegi:
In `rulesmd.ini`:
```ini
[SOMESW]        ; SuperWeaponType
SW.AuxTechnos=  ; List of TechnoTypes
SW.NegTechnos=  ; List of TechnoTypes
SW.TechLevel=1  ; integer
```


src/Ext/SWType/SWHelpers.cpp
```diff
@@ -170,45 +172,54 @@ std::pair<double, double> SWTypeExt::ExtData::GetLaunchSiteRange(BuildingClass*
...
+		return false;
+
+	auto IsTechnoPresent = [pHouse](TechnoTypeClass* pType)
+		{
+			const auto pBuildingType = abstract_cast<BuildingTypeClass*, true>(pType);
+
+			if (pBuildingType && (!BuildingTypeExt::ExtMap.Find(pBuildingType)->PowersUp_Buildings.empty() || BuildingTypeClass::Find(pBuildingType->PowersUpBuilding)))
+				return BuildingTypeExt::GetUpgradesAmount(pBuildingType, pHouse) > 0;
+
+			return pHouse->CountOwnedAndPresent(pType) > 0;
```
TaranDahl: IIRC the counters in ActiveXXTypes (namely those checked by this function) will exclude units that are in Limbo.

TaranDahl: I think it should be CountOwnedNow.

NetsuNegi: This is not a good choice

NetsuNegi: use `CountOwnedNow` will cause technos who is producing in factory will also be considered.

NetsuNegi: best is `CountOwnedPresentAndLimboed()`

TaranDahl (approved):

src/Ext/SWType/SWHelpers.cpp
```diff
@@ -170,45 +172,54 @@ std::pair<double, double> SWTypeExt::ExtData::GetLaunchSiteRange(BuildingClass*
...
+	if (!(this->SW_RequiredHouses & ownerBits) || (this->SW_ForbiddenHouses & ownerBits))
 		return false;
 
 	// check that any aux building exist and no neg building
-	auto IsBuildingPresent = [pHouse](BuildingTypeClass* pType)
-		{
-			if (pType)
-			{
-				if (BuildingTypeExt::ExtMap.Find(pType)->PowersUp_Buildings.size() > 0 || BuildingTypeClass::Find(pType->PowersUpBuilding))
-					return BuildingTypeExt::GetUpgradesAmount(pType, pHouse) > 0;
```
Coronia, line 196: now that we've taken over Ares' SW requirement code, this part should probably be kept so SW.Aux/NegBuildings can support building upgrade for real

src/Ext/SWType/SWHelpers.cpp
```diff
@@ -170,45 +172,54 @@ std::pair<double, double> SWTypeExt::ExtData::GetLaunchSiteRange(BuildingClass*
...
+				return BuildingTypeExt::GetUpgradesAmount(pBuildingType, pHouse) > 0;
+
+			return HouseExt::ExtMap.Find(pHouse)->CountOwnedPresentAndLimboed(pType) > 0;
+		};
 
 	// check whether the optional aux building exists
-	if (pThis->AuxBuilding)
-	{
-		if ((BuildingTypeExt::ExtMap.Find(pThis->AuxBuilding)->PowersUp_Buildings.size() > 0 || BuildingTypeClass::Find(pThis->AuxBuilding->PowersUpBuilding))
-			&& BuildingTypeExt::GetUpgradesAmount(pThis->AuxBuilding, pHouse) <= 0)
```
Coronia, line 178: same

src/Ext/SWType/SWHelpers.cpp
DeathFishAtEase: In fact, it was later confirmed that #1505 did not achieve the intended goal of making upgrades work as valid elements of the `SW.AuxBuildings` list, and this was also verified in [the tests of this PR on topic/430 of the Phobos CN Forum](https://www.phoboscn.top/t/topic/430/55?u=noble_fish).
- The changes made by this PR truly enable upgrades to reach this goal.
- Information update: The portion of the conversation that was tested on # 1505 has been split into [a new topic/458](https://www.phoboscn.top/t/topic/458/9).

Of course, this does not mean that the current code is standard or requires no changes, but I doubt whether this part should be reverted—if we want to complete the previously unfinished work within this PR, we can improve it to achieve standardization; if we want to exclude that part of the work from the changes in this PR, then it needs to be reverted.
- However, even if reverted, it would not achieve the purpose of "`SW.Aux/NegBuildings` can support building upgrade"—that would require additional work without the changes from this PR.

Coronia: If I understand the implementation of this PR correctly, it already takes over Ares' SW condition logic. Hence, the previously incomplete upgrade check should've been completed with it too. That is to say, it can make SW.Aux/NegBuildings able to work on upgrade without doing additional work already

TaranDahl:
Is this fixed? https://www.phoboscn.top/t/topic/430/70

DeathFishAtEase:
Changes later than 6df9224de6adc588dfa91af5d3efac9aa990ab42 (05/14 @ 20:38 (+8)) did not handle the mechanism of `SW.Aux/NegTechnos`. The user feedback time is 05/15 @ 23:25 (+8), so I think this bug was not properly addressed before merging.

PR #2205, opened 2026-05-09
Radar / Power Outage warhead

WingHwy:
- Now you can use the following tags to make the warhead cause a radar or power outage to the affected houses for a specified duration.  
- `RadarOutage` disables the radar display (identical to a Lightning Storm), while `PowerOutage` causes a blackout (identical to spy infiltration on a Power Plant).  
- If `Duration` is **positive**, it stacks onto the current outage timer, unless `Cap` is **negative** — then the effect does not stack and only the longest single duration is kept.  
- If `Duration` is **negative**, it reduces the current outage timer.  
- When `Cap` is **positive**, it acts as a ceiling for stacking or a floor for reduction.  
- Setting `Cap` to **0** with a negative `Duration` removes the outage entirely.  
- The final duration never goes below zero.

In `rulesmd.ini`:
```ini
[SOMEWARHEAD]                     ; WarheadType
RadarOutage.Duration=0            ; integer, game frames
RadarOutage.Cap=0                 ; integer
RadarOutage.AffectsHouse=enemies  ; Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
PowerOutage.Duration=0            ; integer, game frames
PowerOutage.Cap=0                 ; integer
PowerOutage.AffectsHouse=enemies  ; Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|all)
```


CCCP84:
Is it possible to use this feature so that a spy enters a Spysat and turns it off for a certain time, losing map visibility for the player?

DeathFishAtEase:
About docs:
1. Please ensure that the comments within the ini block are aligned.
2. Please place descriptive content outside the ini block, especially content that has already been elaborated in the descriptive text is not suitable to be repeated within the ini block.

WingHwy:
> About docs:
> 
> 1. Please ensure that the comments within the ini block are aligned.
> 2. Please place descriptive content outside the ini block, especially content that has already been elaborated in the descriptive text is not suitable to be repeated within the ini block.

Okay, got it. Thanks for correcting me.

WingHwy:
> Is it possible to use this feature so that a spy enters a Spysat and turns it off for a certain time, losing map visibility for the player?

Do you mean turning off the radar screen (like a Lightning Storm), or literally removing the map reveal so everything goes back to black fog? Our RadarOutage feature only does the first one. If you need the full map vision removed, that would require a different approach, such as triggering a Create Gap effect or adding a new tag to disable SpySat – maybe something we can suggest for a future update.

src/Ext/WarheadType/Body.cpp
```diff
@@ -308,6 +308,13 @@ void WarheadTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 	this->JumpjetNoWobbles.Read(exINI, pSection, "JumpjetNoWobbles");
 	this->JumpjetDeviation.Read(exINI, pSection, "JumpjetDeviation");
 
+	this->RadarOutage_Duration.Read(exINI, pSection, "RadarOutage.Duration");
+	this->RadarOutage_Max.Read(exINI, pSection, "RadarOutage.Max");
+	this->RadarOutage_AffectsHouse.Read(exINI, pSection, "RadarOutage.AffectsHouse");
+	this->PowerOutage_Duration.Read(exINI, pSection, "PowerOutage.Duration");
+	this->PowerOutage_Max.Read(exINI, pSection, "PowerOutage.Max");
+	this->PowerOutage_AffectsHouse.Read(exINI, pSection, "PowerOutage.AffectsHouse");
```
TaranDahl, line 342: I recall it seems to be named "AffectedHouses" elsewhere? Please check the manual and keep the ini flag name consistent.

TaranDahl (approved):

TaranDahl:
Tested by https://www.phoboscn.top/t/topic/439/5

TaranDahl:
> [Needs MP testing](https://github.com/Phobos-developers/Phobos/issues?q=state%3Aopen%20label%3A%22Needs%20MP%20testing%22)

@DeathFishAtEase for what reason?

src/Ext/WarheadType/Body.cpp
DeathFishAtEase: They should have been unified into `AffectsHouse` together with `AffectHouses` and others in #1984.

DeathFishAtEase:
I think I mistakenly thought that "losing map visibility" was also part of the PR's planned content, so I added that label. As it currently appears, it might not be necessary.

WingHwy:
Renamed RadarOutage.Max / PowerOutage.Max to Cap and switched to Helpers::Alex::getCappedDuration, adding support for non‑stacking mode and reduction floors via the Cap value while preserving the existing stacking behavior.

src/Ext/WarheadType/Body.cpp
WingHwy: Done

WingHwy (commented):
Done

Coronia (commented):
If I understand correctly, this will apply power/radar outage effect to all houses whenever the warhead detonates. Maybe it's better to allow it to affect specific house as well? Could add a toggle for that which will make it only work when the warhead hit and can affect a target, in which case it'll only affect the target's owner

That's just my own thought, would like some more opinions

TaranDahl:
> If I understand correctly, this will apply power/radar outage effect to all houses whenever the warhead detonates. Maybe it's better to allow it to affect specific house as well? Could add a toggle for that which will make it only work when the warhead hit and can affect a target, in which case it'll only affect the target's owner
> 
> That's just my own thought, would like some more opinions

Reasonable

PR #2210, opened 2026-05-13
Z-depth adjustment of EBolt visual effect

DeathFishAtEase:
Added more freedom for Z-depth adjustment of WeaponType and LaserTrailType.

The code is not much so they are combined in the same PR; if someone thinks splitting is necessary, please add the `Needs splitting` label.


DeathFishAtEase:
The group on the south side is the default behavior, and the group on the north side is the behavior with `EBolt.ZAdjust`/`Bolt.ZAdjust` set to `-999`.

https://github.com/user-attachments/assets/0c030b55-d115-43b4-8e0f-65da545393b1

TaranDahl (approved):

PR #2211, opened 2026-05-16
AE function PeriodicWeapon

AGAGUGUG:
The function of this property is to periodically fire weapons at all targets within its range. Currently, it determines which targets are locked by polling the units within its range. Currently, there are the following key-value pairs PeriodicWeapon=
PeriodicWeapon.Delay=
PeriodicWeapon.Range=
PeriodicWeapon.AffectsHouses=a
PeriodicWeapon.UseInvokerAsOwner=
PeriodicWeapon.FireAll=
When the value is 'no', it will only fire at the nearest target within range that can be filtered; when this value is 'yes', it will fire at all targets within range.

### DIFF
--- a/src/New/Entity/AttachEffectClass.cpp
+++ b/src/New/Entity/AttachEffectClass.cpp
@@ -1,10 +1,45 @@
 #include "AttachEffectClass.h"
 
+#include <Utilities/EnumFunctions.h>
+#include <algorithm>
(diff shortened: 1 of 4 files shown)


AGAGUGUG:
[Phobos.zip](https://github.com/user-attachments/files/27852737/Phobos.zip)

PR #2214, opened 2026-05-17
Attempt to fix/improve fixed angle Parabola trajectory

Starkku:
This change specifically concerns `Trajectory=Parabola` combined with `Trajectory.Parabola.OpenFireMode=Angle`.

Examples of faulty/corrected trajectory behaviour courtesy of Gen.Ari:
[Before PR changes](https://github.com/user-attachments/assets/18103587-8145-4818-bba2-08467844a8c7)
[After PR changes](https://github.com/user-attachments/assets/13afe39b-2cc7-4320-b4e1-3d873d6d9c11)

The previous implementation for finding velocity for a fixed-angle trajectory involved using [Newton's method](https://en.wikipedia.org/wiki/Newton%27s_method) to iteratively find approximations for roots. It is useful for approximating solutions to complex equations which might be impossible to solve algebraically. However, we're talking about parabolic arc with almost no external constraints or additional factors like air resistance here, just standard inputs for kinematic equations like horizontal and vertical distance, angle and gravity. Pretty much all sane inputs have valid algebraic solutions.

 Or at least should have, to my understanding.

So apparently what the old method is essentially doing is using more unstable approximation method to guess solutions that converge in same results as an algebraic formula would give, which is evidenced by the fact that comparing the two methods yields identical results in cases where Newton's method doesn't just flat-out fail. This PR just ditches it in favour of just directly calculating the solutions with a formula. It will still fail under certain nonsense conditions like vertical launch angles but will also output an error message in log in such cases rather than fail completely silently.

`[Phobos] Failed to solve fixed angle Parabola trajectory velocity: BulletType: [MarauderGunProj] - Source: 17000,22865,279 - Target: 13440,23680,936 - Angle: 90.00 - Gravity: 6.00`

Because:

And my knowledge is mostly constrained to algebra and basic calculus and ability to do research on the Internet, someone should probably double check that any of this shit is making sense. In addition to testing that it works as expected in-game.

`No Documentation Needed` is currently here per still current policy of not writing changes to changelog for changes within same version (Parabola trajectory was introduced after 0.4 and patches)

### DIFF
--- a/src/Ext/Bullet/Trajectories/ParabolaTrajectory.cpp
+++ b/src/Ext/Bullet/Trajectories/ParabolaTrajectory.cpp
@@ -581,8 +581,18 @@ void ParabolaTrajectory::CalculateBulletVelocityRightNow(BulletClass* pBullet, C
 		// Step 1: Read the appropriate fire angle
 		const auto radian = pType->LaunchAngle * Math::Pi / 180.0;
 
-		// Step 2: Using Newton Iteration Method to determine the projectile velocity
-		const auto velocity = (radian >= Math::HalfPi || radian <= -Math::HalfPi) ? 100.0 : this->SearchVelocity(horizontalDistance, distanceCoords.Z, radian, gravity);
+		// Step 2: Find projectile velocity
+		const auto velocity = this->SolveFixedAngleVelocity(horizontalDistance, distanceCoords.Z, radian, gravity);
+
+		if (velocity < 0.0)
+		{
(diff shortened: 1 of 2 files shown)


CrimRecya (approved):
The analytical formula seems correct: v = √((g*x²) / (2·cos²θ·(x·tanθ - z))).
And I would be extremely grateful if you rewrite the comments of `ParabolaTrajectory::SearchFixedHeightMeetTime` and `ParabolaTrajectory::SearchFixedAngleMeetTime`.

PR #2218, opened 2026-05-18
Allow Laser drawing position update

DeathFishAtEase:
- Now you can define whether the endpoints of a laser drawing are updated during its duration.
  - `None`: No update.
  - `Firer`: The start point follows the firer's FLH; if the firer dies, the update stops.
    - Since the FLH of `DiskLaser` actually determines the center of the ring, in this scenario, during the update process, the direction of the line connecting the beam's starting point to the center is fixed relative to the ground and the distance is constant - that is, the starting point will still remain on the ring.
  - `Target`: The end point follows the target; if the target object dies, the update stops.
  - `All`: Equivalent to specifying both `Firer` and `Target`.

https://github.com/user-attachments/assets/5abd91fe-b988-49fd-b94c-902087f62fe4

*Used together with [Projectile interception logic](https://phobos.readthedocs.io/en/latest/New-or-Enhanced-Logics.html#projectile-interception-logic)*

> [!Note]
> In vanilla, an occupied building uses the weapons of its occupants in turn. Thus, if an infantry's occupy weapon uses `Firer` , the laser's starting point will update to the position specified by the most recently used `MuzzleFlashX` after each shot.

> [!Note]
> For a sub-weapon created by `ShrapnelWeapon` or `AirburstWeapon`, its start point is the position where the parent weapon detonates, not the firer's FLH.
> - If `Firer` is set, it will be treated as `None`.
> - If `All` is set, it will be treated as `Target`.

In `rulesmd.ini`:
```ini
[SOMEWEAPON]             ; WeaponType with IsLaser=yes or DiskLaser=yes
LaserPositionUpdate=none ; Position Follow Enumeration (none|firer|target|all)
```

> [!Warning]
> If the weapon sets this logic to a non-`None` value while also using other logics that change the drawing position, such as `FlakScatter`, then after initially drawing the laser according to those other logics, the drawing position will be forced to change due to the update rules.


src/Misc/Hooks.LaserDraw.cpp
```diff
@@ -53,3 +54,289 @@ DEFINE_HOOK(0x6FD3FD, TechnoClass_LaserZap_ZAdjust, 0x5)
...
+		(info.FollowMode == PositionFollow::Firer || info.FollowMode == PositionFollow::All);
+	const bool followTarget = (info.Target != nullptr) &&
+		(info.FollowMode == PositionFollow::Target || info.FollowMode == PositionFollow::All);
+
+	if (followFirer)
+	{
+		if (ObjectClass::Array.FindItemIndex(info.Shooter) != -1 && !info.Shooter->IsDead())
+		{
+			CoordStruct flh;
+			info.Shooter->GetFLH(&flh, info.WeaponIndex, CoordStruct { 0, 0, 0 });
```
TaranDahl: Should consider Burst

src/Misc/Hooks.LaserDraw.cpp
```diff
@@ -53,3 +54,289 @@ DEFINE_HOOK(0x6FD3FD, TechnoClass_LaserZap_ZAdjust, 0x5)
...
+				data.FollowMode = mode;
+				LaserRT::g_Trackers[pLaser] = data;
+			}
+		}
+	}
+	return 0;
+}
+
+// Per‑frame coordinate update
+DEFINE_HOOK(0x550173, LaserDrawClass_UpdateCoords_Update, 6)
```
TaranDahl: LaserDrawClass_Update_UpdateCoords

src/Misc/Hooks.LaserDraw.cpp
```diff
@@ -53,3 +54,289 @@ DEFINE_HOOK(0x6FD3FD, TechnoClass_LaserZap_ZAdjust, 0x5)
...
+
+	auto& info = it->second;
+	const bool followFirer = (info.Shooter != nullptr) &&
+		(info.FollowMode == PositionFollow::Firer || info.FollowMode == PositionFollow::All);
+	const bool followTarget = (info.Target != nullptr) &&
+		(info.FollowMode == PositionFollow::Target || info.FollowMode == PositionFollow::All);
+
+	if (followFirer)
+	{
+		if (ObjectClass::Array.FindItemIndex(info.Shooter) != -1 && !info.Shooter->IsDead())
```
TaranDahl: IsDead may not be a safe way to check for invalid pointer.
Some WW shit DTOR the objects immediately.
PointerExpired is needed.

TaranDahl (changes requested):

docs/New-or-Enhanced-Logics.md
```diff
@@ -2984,6 +2984,31 @@ UnlimboDetonate.KeepSelected=false     ; boolean
...
+```{note}
+For a sub-weapon created by `ShrapnelWeapon`, its start point is the position where the parent weapon detonates, not the firer's FLH.
+- If `Firer` is set, it will be treated as `None`.
+- If `All` is set, it will be treated as `Target`.
+```
+
+In `rulesmd.ini`:
+```ini
+[SOMEWEAPON]             ; WeaponType with IsLaser=yes or DiskLaser=yes
+LaserPositionUpdate=none ; Position Follow Enumeration (none|firer|target|all)
```
TaranDahl: Maybe a global value would be useful?

DeathFishAtEase: Not sure, I've [gone to the corresponding topic on Forum](https://www.phoboscn.top/t/topic/478/2?u=noble_fish) and started a poll to see.

DeathFishAtEase:
Since a certain error introduced during local testing of Phobos Build#46 - Phobos DevBuild#47 may cause `TargetCoordOffset` of BuildingType to be not respected, and the functionality of this PR may mask that error, I will first focus on investigating that existing error and optimizing this PR, rather than immediately completing and merging it.

Z%Q7X1P ITXH~IKWE{RXP51

*96c6550b9f8534a209910da14e76509bbeab36df*

VZ6@DKS8 9KZ1NIF9(FY(YC

*838c2e7b7997e9e771e0f93ac7070c93e83ba472 (#1553)*

Resolved in #2222.

TaranDahl:
I thought those were supposed to be several GIFs

DeathFishAtEase:
I think it's that after further refining the scope, I forgot to remove the image of Ares 3.0p1 without Phobos, so perhaps the three images arranged together earlier looked a bit confusing. :P
However, I have located it!

TaranDahl (approved):

DeathFishAtEase:
Tested by: https://www.phoboscn.top/t/topic/478/72?u=noble_fish

image

image

image

Metadorius:
How does this mesh into the idea of a potential rewrite of engrave trajectory? The one we thought should be implemented similar to `BurstFLH` but instead you'd define offsets not of source, but of target.

DeathFishAtEase:
Whether engrave trajectory will be added to the official version remains unknown.

> you'd define offsets not of source, but of target

I don't quite understand the conflict described here; could you elaborate?

Coronia (approved):

Starkku:
This is currently performing expensive hash table lookups (`std::unordered_map.find()`) on every frame for every `LaserDrawClass` instance with zero mitigations in place f.ex disabling hooks after all parsing passes if tracking feature is not enabled, or even a simple skip if no tracking laser weapons exist. Ideally `LaserDrawExt` would be created for this purpose. Unlike `EBolt`, `LaserDrawClass` technically has enough padding to house extension pointer but it is not contiguous which would either require adding additional logic to the extension template to allow parsing pointer from non-contiguous offsets or a bespoke implementation.

Otamaa:
> This is currently performing expensive hash table lookups (`std::unordered_map.find()`) on every frame for every `LaserDrawClass` instance with zero mitigations in place f.ex disabling hooks after all parsing passes if tracking feature is not enabled, or even a simple skip if no tracking laser weapons exist. Ideally `LaserDrawExt` would be created for this purpose. Unlike `EBolt`, `LaserDrawClass` technically has enough padding to house extension pointer but it is not contiguous which would either require adding additional logic to the extension template to allow parsing pointer from non-contiguous offsets or a bespoke implementation.

Technically it is possible to store extension pointer to the separate padding (2 BYTE each) and reconstruct it for example , here is what i do with mine  , altho the safety is kind a questionable but it is doable: 
```
class LaserDrawClass:

	void* GetPptrFromPad() {
		uint32_t full_address = (static_cast<uint32_t>(align_22) << 16) | align_4A;
		return reinterpret_cast<void*>(full_address);
	}

	void SetPadToPtr(void* ptr){
		uint32_t full_address = reinterpret_cast<uint32_t>(ptr);
		// Extract the top 16 bits
		align_22 = static_cast<uint16_t>(full_address >> 16);
		// Extract the bottom 16 bits
		align_4A = static_cast<uint16_t>(full_address & 0xFFFF);
	}
};
```

Starkku:
Never meant to imply it isn't doable, only that what we currently have doesn't do it out of the box.

Coronia:
> or even a simple skip if no tracking laser weapons exist.

but it does, the per frame check has an early bail out if the map is empty

anyway, this is not *that* bad in performance at least according to my profiler. It was even more problematic before due to pointer invalidate, which got optimized later on. If we want to further improve its performance, maybe the better way is to nuke those hooks when not being used

PR #2226, opened 2026-05-26
Allow *Harvester counter* to display only the total number or the number currently working

DeathFishAtEase:
- You can use `Sidebar.HarvesterCounter.HideMaxValue` to allow the counter to display only the number of active harvesters. If this value is `false`, you can use `Sidebar.HarvesterCounter.OnlyMaxValue` to allow the counter to display only the number of total harvesters.
  - This does not disable the color change of the counter text; to adapt accordingly, you need to configure the counter color yourself for the side that uses this feature.

In `rulesmd.ini`:
```ini
[SOMESIDE]                                      ; Side
Sidebar.HarvesterCounter.HideMaxValue=false     ; boolean
Sidebar.HarvesterCounter.OnlyMaxValue=false     ; boolean
```

HideMaxValue

*HideMaxValue*

OnlyMaxValue

*OnlyMaxValue*


Coronia (approved):

CnCRAZER:
Ohhh very nice 👍🏻

PR #2227, opened 2026-05-26
[Minor] WeaponRange.Maximum/Minimum

Coronia:
There're cases when we don't want weapon range AE or other range modifiers to change weapon's range infinitely. F.ex. if it's longer than the weapon's `ProjectileRange` it'll fail to hit the target beyond that. Adding a setting on weapon for theses cases to define the maximum/minimum value of its range


TaranDahl:
> F.ex. if it's longer than the weapon's ProjectileRange it'll fail to hit the target beyond that

The proper approach should be to make ProjectileRange support RangeMultiplier.

docs/New-or-Enhanced-Logics.md
```diff
@@ -172,6 +173,8 @@ AttachEffect.DisallowedMinCounts=                  ; integer - minimum disallowe
 AttachEffect.DisallowedMaxCounts=                  ; integer - maximum disallowed instance count (comma-separated) for cumulative types in order from first to last.
 AttachEffect.IgnoreFromSameSource=false            ; boolean
 AttachEffect.CheckOnFirer=false                    ; boolean
+Range.Maximum=2147483647                           ; integer
```
DeathFishAtEase: Perhaps they are suitable for using the same series of prefixes in tag names?

docs/New-or-Enhanced-Logics.md
```diff
@@ -47,6 +47,7 @@ This page describes all the engine features that are either new and introduced b
   - `WeaponRange.Multiplier` and `WeaponRange.ExtraRange` can be used to multiply the weapon firing range of the object the effect is attached to, or give it an increase / decrease (measured in cells), respectively. `ExtraRange` is cumulatively applied from all attached effects after all `Multiplier` values have been applied.
     - `WeaponRange.AllowWeapons` can be used to list only weapons that can benefit from this range bonus and `WeaponRange.DisallowWeapons` weapons that are not allowed to, respectively.
     - On TechnoTypes with `OpenTopped=true`, `OpenTopped.UseTransportRangeModifiers` can be set to true to make passengers firing out use the transport's active range bonuses instead.
+    - The weapon range with all sort of range modifiers can't be greater than `WeaponRange.Maximum` or smaller than `WeaponRange.Minimum` of the affected weapon. Notice that `Range=-512` is a special case and will be ignored.
```
DeathFishAtEase: I think the `-512` here refers to the special value `-2` set by specifying a lepton value in `gamemd.exe`. Considering that in the log, these values are compared with `[Weapon] -> Range= (float)`, perhaps it is better to clarify this point and the fact that the units are different.

TaranDahl (approved):

PR #2228, opened 2026-05-26
Fixed voxel projectile and animation lighting issues

TaranDahl:
In vanilla, they are darker than voxel technos.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -760,6 +760,7 @@ This page lists all the individual contributions to the project by their author.
   - Allow replacing vanilla repairing with togglable auto repairing
   - Fix an issue that the time for units in the area guard mission to reacquire targets after eliminating the target is significantly longer than that in other missions
   - Framework for dynamic sight
+  - Fixed voxel projectile and animation lighting issues
 - **solar-III (凤九歌)**
   - Target scanning delay customization (documentation)
   - Skip target scanning function calling for unarmed technos (documentation)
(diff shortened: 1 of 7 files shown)


NetsuNegi (approved):
Looks good

Metadorius:
can we get a visual?

TaranDahl:
Vanilla:
VoxelProjFix-Vanilla
Fixed:
VoxelProjFix-Fixed

DeathFishAtEase:
Is the performance impact caused by this fix such that it necessitates adding a switch?

TaranDahl:
> Is the performance impact caused by this fix such that it necessitates adding a switch?

I suppose bullets and voxelanims(debris) stay for a very short time on the map, thus they make a minor impact on performance?

DeathFishAtEase:
I think your point is very reasonable, but I was considering some special scenarios at that time - our users often prefer to use `AirburstSpread` to create a large number of projectiles at once. In retrospect, I was overthinking it.
In the scenario where I used `[ZEP]` from vanilla YR as the image combined with `AirburstSpread=10` , `ROF=20` , I compared `compiled-dll-4c1b896a93a469b2616d18e29e89cb122812780d.zip` with `compiled-dll-5518bd557c2c223af152a0e3ba0a9dafcbe9b1d4.zip` , and the stable FPS value only dropped from 18 to 14. So I agree that there might not be a need to worry too much about performance issues.
~~Although this change causes my maximum FPS to drop by a few hundred, which is somewhat annoying :/~~

TaranDahl:
> Although this change causes my maximum FPS to drop by a few hundred, which is somewhat annoying :/

Very strange. This should have no impact when there are no proj or debris on the map.

DeathFishAtEase:
I just spent some time making it as precise as possible, running several comparisons between the PR-built Nightly and f8c8cc41253e76b7340626ee93c36dc5c86a369a on the vanilla YR map "The Alamo (2)". After extending the sample to a million frames, the avg difference was only in the double digits, and in one case the avg of the PR scenario was even higher.

> This should have no impact when there are no proj or debris on the map.

I think it might just be that my maximum FPS baseline is relatively large and there were some small disturbances at the time, so we can temporarily ignore it.

PR #2230, opened 2026-05-28
Allow weapon range modifiers to apply on ProjectileRange

Starkku:
`ProjectileRange` now has weapon range modifiers applied to it if greater than 0 and unless `ProjectileRange.ApplyModifiers` is set to false on the WeaponType.

----------------------

Summary of changes:
- Patches Ares code to skip `Ranged=true` projectile checks and updating projectile position to new coordinates.
- Add new hook at `0x467BA4` (Ares hook return address) with updated projectile range check as well as the skipped position update.
- New mechanism treats `BulletClass::Range` as immutable initial max range (as opposed to distance that the projectile can still travel that is deducted from every frame) to preserve ability to check effective range with dynamic modifiers, distance traveled so far is stored in new field `BulletExt::DistanceTraveled`.

----------------------

Fixes the main underlying reason for #2227's existence, whether or not it has any reason to remain after addressing this I do not know.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -302,6 +302,7 @@ This page lists all the individual contributions to the project by their author.
   - Animation theater/tile palette toggle
   - Animatable template
   - Tank Bunker improvements
+  - `ProjectileRange` weapon range modifiers interaction fix
 - **Morton (MortonPL)**:
   - `XDrawOffset` for animations
   - Shield passthrough & absorption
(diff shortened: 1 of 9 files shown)


TaranDahl (approved):

Starkku:
Behaviour of projectiles immolated by Firestorm Wall should be tested, there's a high likelihood that this was broken here (something I realized as an afterthought). I will adjust the patches and hook accordingly if this turns out to be the case.

Coronia:
tested it and firestorm wall seemed to behave like what it was before: can't intercept the projectile directly but changing its destination to a location near the wall. Not sure if this need to be adjusted

and there're other issues: it can't affect range bonus from high ground, opentopped transport, Tank Bunker etc. These cases can also resulted in failure of projectile with range limit

in addition, the game now becomes super laggy when a unit with `ProjectileRange` weapon and `WeaponRange.Multiplier`  AE entering a Tank Bunker. It won't be laggy if it's not inside Bunker or using `WeaponRange.ExtraRange` instead of multiplier

Starkku:
Presumably fixed the issue with processing order of updating bullet coords vs. Ares' Firestorm Wall code.

Cannot reproduce the issue with Tank Bunkers, need more info.

fagonghaiwo:
19fd338c03e858ea7d785bcdd30a9bea

about firestorm wall
seems work


11ac3a0b2ad1b67b8b47e946764fd808

cant instead of 2227

PR #2233, opened 2026-05-30
Allow users to customize whether the condition of `DiscardOn=move` uses `Is_Really_Moving_Now()` or `Is_Moving()`

DeathFishAtEase:
During DevBuild#41-44, `DiscardOn=move` considered hovering jumpjet objects etc., which could be used to achieve some effects. But after the commit 56583d8ccd83c2ae71aa009714bc07181f3aad17 immediately following the release of DevBuild#44, they were broken due to changes in the moving check method of `DiscardOn`. Therefore I add a switch here.

ConsiderHoverAsMoving

*Behavior before DevBuild#44 and when the switch added by this PR is enabled.*

ConsiderHoverAsStationary

*Behavior in dev*


TaranDahl (approved):

Coronia:
Is the only difference between `Is_Moving_Now()` and `Is_Really_Moving_Now()` being hover in the air condition? Not sure if this change will only affect jumpjets or some other units, hovering or not

DeathFishAtEase:
An insightful comment, the original statement is indeed somewhat misleading.

TaranDahl:
> Is the only difference between `Is_Moving_Now()` and `Is_Really_Moving_Now()` being hover in the air condition? Not sure if this change will only affect jumpjets or some other units, hovering or not

Basically, `Is_Really_Moving_Now` is the same as `Is_Moving_Now`. IIRC they only differ in WalkLoco.

PR #2237, opened 2026-06-02
Allow warhead to only affect invoker

DeathFishAtEase:
- In the past, modders used many methods such as `Dropping` and extremely small `CellSpread` to prevent a weapon fired by a unit from affecting others, thereby achieving complex combination logic. However, existing methods merely reduced the range of affected objects through various means. Now, you can directly use the following flag to check whether the victim is the unit itself:
  - `AffectsInvokerOnly` can be used to make the affected object only the unit itself.
  - `AffectsInvokerOnly.Reverse` can be used to exclude the unit from the affected objects.
  - `AffectsInvokerOnly.IgnoreInvokerState` can be used to determine whether to consistently avoid affecting other objects regardless of the unit's state.

```ini
[CombatDamage]
AffectsInvokerOnly.IgnoreInvokerState=true  ; boolean

AffectsInvokerOnly=false                    ; boolean
AffectsInvokerOnly.Reverse=false            ; boolean
AffectsInvokerOnly.IgnoreInvokerState=      ; boolean, default to [CombatDamage] -> AffectsInvokerOnly.IgnoreInvokerState
```

> [!Tip]
> If you have enabled [`ApplyPerTargetEffectsOnDetonate`](https://phobos.readthedocs.io/en/latest/New-or-Enhanced-Logics.html#toggle-per-target-warhead-effects-apply-timing) to make affects handling completely follow damage propagation, then most of the time you can just use [`DamageSelf`](https://modenc.renegadeprojects.com/DamageSelf) and [`AllowDamageOnSelf`](https://phobos.readthedocs.io/en/latest/Fixed-or-Improved-Logics.html#allowing-damage-dealt-to-firer) without needing `AffectsInvokerOnly.Reverse`.


Otamaa:
Just an reminder , if you do keep old invoker pointer always make sure that is really valid invoker , because in the middle of damaging/detonating state the invoker itself may get killed or deleted and need to be cleanup , if you keep it like this that mean you will potentially referencing an dangling pointer.

TaranDahl:
> Just an reminder , if you do keep old invoker pointer always make sure that is really valid invoker , because in the middle of damaging/detonating state the invoker itself may get killed or deleted and need to be cleanup , if you keep it like this that mean you will potentially referencing an dangling pointer.

From my understanding, this step is unnecessary.
IIRC UnInit will be invoked when an entity dies from damage. It marks the entity with IsAlive=false and adds it to the pending deletion list instead of deleting it immediately. All such entities will be cleaned up collectively at the end of the frame.

DeathFishAtEase:
During testing, it was found that in versions no later than DevBuild#48, the *[Customizable Warhead trigger conditions](https://phobos.readthedocs.io/en/latest/Fixed-or-Improved-Logics.html#customizable-warhead-trigger-conditions)* feature had an issue where 1 HP would be deducted when the conditions were not met. This PR uses a similar execution flow and is therefore affected as well, so it is postponed.
See https://www.phoboscn.top/t/topic/573

DeathFishAtEase:
After subsequent confirmation, this abnormal situation does not occur when the target is an entity rather than the ground, so this PR resumes maintenance. Although... testers need to pay more attention to actual data to avoid erroneous conclusions.

DeathFishAtEase:
Tested by: https://www.phoboscn.top/t/topic/541/10?u=noble_fish

image

```ini
[NukePayload]
;RadLevel=500

[NUKE]
AffectsInvokerOnly=yes
```

image

```ini
[SA]
CellSpread=10
AffectsInvokerOnly=yes
AllowDamageOnSelf=yes
Convert.From=E2
Convert.To=E1
```

image

```ini
[SA]
CellSpread=10
AffectsInvokerOnly=yes
AllowDamageOnSelf=yes
AttachEffect.AttachTypes=AE1

[AttachEffectTypes]
0=AE1

[AE1]
Duration=-1
Animation=MININUKE
```

---

Other test (Mark through Tint AE)

Work On `DiskLaser`
_}(~2@~E8NGWL7TMG~`1BX6

`Reverse` with `ApplyPerTargetEffectsOnDetonate=false`, `AllowDamageOnSelf=false`
image

Work on sub weapon's warhead (the parent weapon uses a regular Warhead)
$9UZF63_T$BKUBK%A0I}F`Q

NetsuNegi (approved):

PR #2239, opened 2026-06-03
Allow infantry to perform type conversion when deploying and undeploying

DeathFishAtEase:
Now infantry can perform type conversion when executing the `Deploy` and `Undeploy` sequences.

Convert (Un)deploy

*E1↔GGI, YURI↔YURIPR, E2→BORIS*

In `rulesmd.ini`:
```ini
[SOMEINFANTRY]             ; InfantryType
Convert.Deploy=            ; InfantryType
Convert.Undeploy=          ; InfantryType
```

> [!Note]
> For the sake of logical clarity, this feature does not allow the converted unit to trigger another conversion before the current sequence ends. If continuous conversion is needed, please use `Convert.Land/Water`.

### DIFF
--- a/src/Ext/Techno/Hooks.cpp
+++ b/src/Ext/Techno/Hooks.cpp
@@ -61,6 +61,7 @@ DEFINE_HOOK(0x4DA54E, FootClass_AI, 0x6)
 	pExt->UpdateWarpInDelay();
 	pExt->UpdateTiberiumEater();
 	pExt->AmmoAutoConvertActions();
+	pExt->DeployConvertAction();
 
 	return 0;
 }
--- a/src/Ext/Techno/Hooks.cpp
+++ b/src/Ext/Techno/Hooks.cpp
@@ -49,7 +49,6 @@ DEFINE_HOOK(0x4DA54E, FootClass_AI, 0x6)
 	pExt->UpdateWarpInDelay();
 	pExt->UpdateTiberiumEater();
 	pExt->AmmoAutoConvertActions();
-	pExt->DeployConvertAction();
 
 	return 0;
 }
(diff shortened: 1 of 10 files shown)


NetsuNegi (approved):

src/Ext/Techno/Hooks.cpp
```diff
@@ -61,6 +61,7 @@ DEFINE_HOOK(0x4DA54E, FootClass_AI, 0x6)
 	pExt->UpdateWarpInDelay();
 	pExt->UpdateTiberiumEater();
 	pExt->AmmoAutoConvertActions();
+	pExt->DeployConvertAction();
```
TaranDahl: I think it is incorrect that you put it here.
It should be handled within MissionUnload.

TaranDahl (changes requested):

src/Ext/Techno/Hooks.cpp
DeathFishAtEase: Not sure if checking `Mission::Unload` is better; it involves too many things. I need to first investigate the overlap between `Mission::Unload` and `Sequence::Deploy` /`Sequence::Undeploy` .

TaranDahl: @Metadorius do you think this is ok to be ignored?

Starkku: I am not @Metadorius but I don't think it is OK. It should be handled in `InfantryClass::Mission_Unload()` or `InfantryClass::Doing_AI() (0x520AE0)`. UnitClass version is done in Unload mission handler so that would make sense.

Adding one-time checks that logically belong elsewhere into top-level per-frame logic should be avoided.

DeathFishAtEase: About `Mission::Unload`:
image

```diff
	pExt->AmmoAutoConvertActions();
-	pExt->DeployConvertAction();

+	if (pThis->CurrentMission == Mission::Unload)
+		Debug::Log("DEBUG [%s]: Mission is Unload\n", pThis->get_ID());
+	if (pThis->CurrentMission == Mission::Move)
+		Debug::Log("DEBUG [%s]: Mission is Move\n", pThis->get_ID());

	return 0;
```
image

TaranDahl (approved):

PR #2240, opened 2026-06-04
[Bugfix][Aircraft] Only apply unlandable penalty when truly idle

KiLiMe:
Prevent the unlandable penalty (damage or crash) from triggering while the aircraft still has a valid target or destination.


Coronia:
I think it's intended to make it as the way it is? @CrimRecya

CrimRecya (commented):
This is clearly not a bug fix. My purpose of making this is to prevent aircraft from always flying by using planning path without an airport.

PR #2243, opened 2026-06-06
[Vanilla Enhancement] 2 customizeble of `DiskLaser`

NetsuNegi:
- `DiskLaser.SimulateFire` determines whether a `DiskLaser=true` weapon simulates firing, avoiding the limitations of the original hardcoded kill method.
- `DiskLaser.ChargeUp` customizes the charge‑up sound for `DiskLaser=true` weapons.

In `rulesmd.ini`:
```ini
[SOMEWEAPON]                  ; WeaponType, with DiskLaser=yes
DiskLaser.SimulateFire=false  ; boolean
DiskLaser.ChargeUp=           ; sound entry, default to [AudioVisual] -> DiskLaserChargeUp
```


src/Ext/WeaponType/Hooks.DiskLaser.cpp
```diff
@@ -0,0 +1,130 @@
...
+	if (const auto pBullet = pWeapon->Projectile->CreateBullet(pDiskLaser->Target, pOwner, damage, pWeapon->Warhead, bulletSpeed, pWeapon->Bright))
+	{
+		BulletExt::ExtMap.Find(pBullet)->FirepowerMult = pData->FirepowerMultiplier;
+		BulletExt::SimulatedFiringUnlimbo(pBullet, pOwner->Owner, pWeapon, fireCoords, false);
+		BulletExt::SimulatedFiringEffects(pBullet, pOwner->Owner, pOwner, true, false);
+	}
+
+	pDiskLaser->unknown_30 = static_cast<DWORD>(-1); // Restrore overriden instruction and skip Ares's hook here
+	return ReturnFromFunction;
+}
```
TaranDahl, line 130: Since you want to simulate firing, why not directly call TechnoClass::Fire when DiskLaser fires (and skip the processing for the DiskLaser part)?

TaranDahl: Maybe the rof part either

NetsuNegi: I don't want too many unnecessary operations mixed in; And calling the Fire() function directly will use the current Owner's firepower multiplier, which is an incorrect behavior. If a hook is added for this, why call Fire()? The current one is already sufficient

TaranDahl: @Metadorius what do you think?

TaranDahl: @Starkku @ZivDero or anyone?

ZivDero: The premise of the PR is unclear. What limitations are there? What does this circumvent?

Metadorius: from my understanding per @Starkku's words, there are probably some troubles with just calling ::Fire here as well?

> why call Fire()

to have a unified path for all weapon firing? imagine Kratos or some other DLL having some special hooks in that function

TaranDahl: > The premise of the PR is unclear. What limitations are there? What does this circumvent?

Vanilla's DiskLaser is not a genuine weapon. When it fires, it creates an instance of DiskLaserClass. At the end of its lifecycle, this instance directly detonates the warhead via DamageArea, skipping numerous processes such as firing, projectile updating and projectile detonation. SimulateFire enables DiskLaser to invoke a more complete firing sequence upon termination.

TaranDahl: > from my understanding per @Starkku's words, there are probably some troubles with just calling ::Fire here as well?

Ah, what troubles (and what words also) exactly?


> > why call Fire()
> 
> to have a unified path for all weapon firing? imagine Kratos or some other DLL having some special hooks in that function

Yeah. I believe that actions with identical logic should be handled by calling the same function. Avoid reinventing the wheel unless absolutely necessary.

Metadorius:
I mean just because I commented there's no need to close the PR immediately.

> And calling the Fire() function directly will use the current Owner's firepower multiplier, which is an incorrect behavior

can you elaborate?

NetsuNegi:
> I mean just because I commented there's no need to close the PR immediately.
> 
> > And calling the Fire() function directly will use the current Owner's firepower multiplier, which is an incorrect behavior
> 
> can you elaborate?

not because you

PR #2248, opened 2026-06-09
Adjust recruitable status on team member discharge

TaranDahl:
This is an extension to https://github.com/Phobos-developers/Phobos/pull/2022

### Adjust recruitable status on team member liberate

- In vanilla, when a unit is added to a team, its `RecruitableB` flag is overwritten by the team's `AreTeamMembersRecruitable` setting. When the unit is liberated from the team, the flag is not restored. The following settings allow a team to reset this flag when liberating its members.
  - If set to a value **greater than 0**, the liberated unit is forcibly marked as recruitable.
  - If set to **0**, the liberated unit is forcibly marked as not recruitable.
  - If set to a value **less than 0** (default: `-1`), the original game behavior is preserved.

In `rulesmd.ini`:
```ini
[General]
SetRecruitableOnLiberate=-1  ; integer

[SOMETEAMTYPE]                               ; TeamType
SetRecruitableOnLiberate=     ; integer, default to [General] -> SetRecruitableOnLiberate
```


Coronia (approved):
my only concern is the tag name. We usually use 'destroy team' to describe such discharge behavior in mapping, but calling it 'DestroyedMember` will also be misleading. Not sure what is a better name here for now

TaranDahl:
Could change the tag name if a better one comes in 24 hours.

DeathFishAtEase:
Scheme (keeping existing style): What about `DischargedMemberRecruitable`? `Is...` does not match the value type of this INI flag, while `Autocreate` is not closely related to the resetting of `RecruitableB` and may cause misunderstandings.
Considering other styles: What about `SetRecruitableOnDischarge`?

---

Furthermore, according to existing reverse-engineering annotations, we have a function named `TeamClass::LiberateMember`; perhaps `SetRecruitableOnLiberate` would be better. Just for reference.

PR #2249, opened 2026-06-10
Dehardcode of parasites unlimboing after killing naval targets

DeathFishAtEase:
- In vanilla, parasites with `Naval=false` perform a series of additional checks on the current cell after killing a target, including a series of determinations such as whether there is a bridge when the cell's LandType is `Water`, `Beach`, or `Rock`, which prevents them from normally unlimbo in open water; while parasites with `Naval=true` skip these checks. Now, you can customize the behavior of parasites after killing a target in water:
  - If not set, the original behavior is performed by default.
  - If set to `true`, they can unlimbo normally even if `Naval=false`.
  - If set to `false`, they cannot unlimbo normally even if `Naval=true`.

In `rulesmd.ini`:
```ini
[General]
Parasite.AllowWaterExit=  ; boolean

Parasite.AllowWaterExit=  ; boolean, defaults to [General] -> Parasite.AllowWaterExit
```

> [!Note]
> Setting `Parasite.AllowWaterExit` to `true` does not skip the original check of whether a BuildingType exists on the cell.

Parasite_AllowWaterExit

*`[DRON] -> Parasite.AllowWaterExit=true` & `[SQD] -> Parasite.AllowWaterExit=false`*

### DIFF
--- a/src/Ext/Techno/Hooks.cpp
+++ b/src/Ext/Techno/Hooks.cpp
@@ -842,6 +842,21 @@ DEFINE_HOOK(0x6298CC, ParasiteClass_AI_GrippleAnim, 0x5)
 	return SkipGameCode;
 }
 
+DEFINE_HOOK(0x62AB71, ParasiteClass_CanExistOnVictimCell_WaterExit, 0x6)
+{
+	GET(ParasiteClass*, pParasite, ESI);
+	auto const pOwner = pParasite->Owner;
+	if (pOwner)
+	{
+		auto const pTypeExt = TechnoTypeExt::ExtMap.Find(pOwner->GetTechnoType());
+		if (!pTypeExt->Parasite_WaterExit_RequireNaval.Get(RulesExt::Global()->Parasite_WaterExit_RequireNaval))
+		{
+			return 0x62AC0F;
+		}
+	}
+	return 0;
+}
+
 #pragma region RadarDrawing
 
 DEFINE_HOOK(0x655DDD, RadarClass_ProcessPoint_RadarInvisible, 0x6)
--- a/src/Ext/Techno/Hooks.cpp
+++ b/src/Ext/Techno/Hooks.cpp
@@ -842,16 +842,30 @@ DEFINE_HOOK(0x6298CC, ParasiteClass_AI_GrippleAnim, 0x5)
 	return SkipGameCode;
 }
 
-DEFINE_HOOK(0x62AB71, ParasiteClass_CanExistOnVictimCell_WaterExit, 0x6)
+DEFINE_HOOK(0x62AB6F, ParasiteClass_CanExistOnVictimCell_AllowWaterExit, 0x8)
 {
 	GET(ParasiteClass*, pParasite, ESI);
 	auto const pOwner = pParasite->Owner;
 	if (pOwner)
 	{
 		auto const pTypeExt = TechnoTypeExt::ExtMap.Find(pOwner->GetTechnoType());
-		if (!pTypeExt->Parasite_SkipNavalCheck.Get(RulesExt::Global()->Parasite_SkipNavalCheck))
+		const auto& globalVal = RulesExt::Global()->Parasite_AllowWaterExit;
+
+		if (pTypeExt->Parasite_AllowWaterExit.isset())
 		{
-			return 0x62AC0F;
+			return pTypeExt->Parasite_AllowWaterExit.Get() ? 0x62AC0F : 0x62AB4B;
+		}
+		else if (globalVal.isset())
+		{
+			return globalVal.Get() ? 0x62AC0F : 0x62AB4B;
+		}
+		else
+		{
+			auto const pType = pOwner->GetTechnoType();
+			if (pType->Naval)
+				return 0x62AC0F;
+			else
+				return 0;
 		}
 	}
 	return 0;
--- a/src/Ext/Techno/Hooks.cpp
+++ b/src/Ext/Techno/Hooks.cpp
@@ -844,17 +844,19 @@ DEFINE_HOOK(0x6298CC, ParasiteClass_AI_GrippleAnim, 0x5)
 
 DEFINE_HOOK(0x62AB69, ParasiteClass_CanExistOnVictimCell_AllowWaterExit, 0x6)
 {
+	enum { SkipChecks = 0x62AC0F, ForceFail = 0x62AB4B, ContinueVanilla = 0x62AB77 };
+
 	GET(TechnoTypeClass*, pType, EAX);
 
 	auto const pTypeExt = TechnoTypeExt::ExtMap.Find(pType);
 	const auto& globalVal = RulesExt::Global()->Parasite_AllowWaterExit;
 
 	if (pTypeExt->Parasite_AllowWaterExit.isset())
-		return pTypeExt->Parasite_AllowWaterExit.Get() ? 0x62AC0F : 0x62AB4B;
+		return pTypeExt->Parasite_AllowWaterExit.Get() ? SkipChecks : ForceFail;
 	else if (globalVal.isset())
-		return globalVal.Get() ? 0x62AC0F : 0x62AB4B;
+		return globalVal.Get() ? SkipChecks : ForceFail;
 	else
-		return pType->Naval ? 0x62AC0F : 0x62AB77;
+		return pType->Naval ? SkipChecks : ContinueVanilla;
 }
 
 #pragma region RadarDrawing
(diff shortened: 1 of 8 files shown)


Metadorius:
The name is a bit of a mouthful. also where credits entry/why does this have "no docs needed" tag?

Metadorius:
otherwise looks good from a glance

DeathFishAtEase:
A classic issue: how to obtain an INI flag name that is general, accurate, and not too long.

Metadorius:
> A classic issue: how to obtain an INI flag name that is general, accurate, and not too long.

`Parasite.AllowNavalExit=(defaults to Naval yes/no)`?

DeathFishAtEase:
Hmmm... I'm more expressing that this new flag can free parasites from dependence on `Naval` or not. `AllowNavalExit` sounds like it's used to prevent `[SQD]` from exiting normally when set to false... And this also involves the `Naval` flag on the unit and whether the target is on water, which might introduce some ambiguity.

Metadorius:
> it's used to prevent [SQD] from exiting normally when set to false

I mean who knows, maybe someone would want to make a one-use parasite or something, so that would be desired.

TaranDahl:
> I'm more expressing that this new flag can free parasites from dependence on Naval or not.

Taking Naval into account in pathfinding is simply a poorly thought-out design. I believe this should even be fixed as a bug.
I think the correct approach is to find the nearest position where IsCellOccupied==Move::OK within a certain range around the target unit, and then move the unit there via UnLimbo. WW is reinventing the wheel for checking grid cell availability here, which is really silly.

DeathFishAtEase:
> maybe someone would want to make a one-use parasite or something, so that would be desired.

Ah, that design which causes the LimboLuanch weapon to inevitably fail at unlimbo - I remember many people have wanted a similar mechanism to implement suicide planes. But that is even more unrelated to Naval or water. Well, this is indeed a requirement; I'll mark it down for now.

> Taking Naval into account in pathfinding is simply a poorly thought-out design. I believe this should even be fixed as a bug.

Haven't thought of a good enough design for now.

> WW is reinventing the wheel for checking grid cell availability here, which is really silly.

:+1:

TaranDahl:
I think many of WW's designs are clearly done on the spur of the moment, lacking refinement and careful consideration.
You should spend more time thinking about "How would I build this from scratch?" instead of "How do I fix this mess?"

TaranDahl:
> Haven't thought of a good enough design for now.

AllowWaterExit=true ➡️ The target position must be on Water or Beach.
AllowWaterExit=false ➡️ The target position must not be on Water or Beach.
!isset➡️Vanilla behavior.

Maybe you should also investigate why they took Rock into consideration, and judge if it is reasonable.

DeathFishAtEase:
Removing this dependency does not equate to solving WW's reinventing-the-wheel problem; this function's logic also includes checks for whether the cell has a TerrainType or ground bridge overlay. So improving this handling would require designing a system that covers all existing checks and is better—I lack interest in expanding this area unless there is high demand from the community. This PR focuses on solving the pain point of Attack Dog and Terror Drone attacking water targets.

TaranDahl:
> Removing this dependency does not equate to solving WW's reinventing-the-wheel problem; this function's logic also includes checks for whether the cell has a TerrainType or ground bridge overlay. So improving this handling would require designing a system that covers all existing checks and is better—I lack interest in expanding this area unless there is high demand from the community. This PR focuses on solving the pain point of Attack Dog and Terror Drone attacking water targets.

I didn't ask you to solve that reinventing-the-wheel problem.
What I mean is:

HOOK(0x62AB5E)
if (!isset)
  return 0
if (flag==true)
  return 0x62AB77
if (flag==false)
  return 0x62AC0F

Metadorius:
alternative idea:
Parasite.ExitAnywhere=yes/no

DeathFishAtEase:
> I didn't ask you to solve that reinventing-the-wheel problem.

That is an explanation of "a good enough", as it originally responded to your criticism of WW's design issues.

DeathFishAtEase:
> Parasite.ExitAnywhere=yes/no

In fact, this is not "Anywhere", because I intend to preserve the original behavior as much as possible—for example, if there is a BuildingType on the cell, it still cannot appear. Of course, I would be happy to see someone bulldoze this whole set of mechanics in the future. :P

Moreover, if "Anywhere" means land & water, then combining it with TaranDahl's proposal (where leaving it empty means no processing, true means it can only normally attack water targets, false means it can only normally attack land targets) would make it impossible to handle with a simple bool value.

These current proposals exceed the original design goal, and it might become something slightly complicated.

TaranDahl:
> In fact, this is not "Anywhere", because I intend to preserve the original behavior as much as possible

I also think Anywhere is not a good idea.

> Moreover, if "Anywhere" means land & water, then combining it with TaranDahl's proposal (where leaving it empty means no processing, true means it can only normally attack water targets, false means it can only normally attack land targets) would make it impossible to handle with a simple bool value.

That is how Nullable\<bool\> goes.

DeathFishAtEase:
Default behavior, land & water, land only, water only are 4 cases; !isset, true, false are insufficient.

> HOOK(0x62AB5E)
> if (!isset)
>   return 0
> if (flag==true)
>   return 0x62AB77
> if (flag==false)
>   return 0x62AC0F

Hmm... It looks a bit strange. If the hook is placed at 0x62AB5E, it should cover `.text:0062AB63 FF 92 84 00 00 00 call dword ptr [edx+84h]`. Are you sure it's better than the currently used 0x62AB71?

Can returning 0x62AB77 or 0x62AC0F really achieve the purpose of “The target position must not be on Water or Beach.”? For the function currently being handled, even if you do nothing and directly return 0x62AB4B, it would only make it fail regardless of Naval but would not affect behavior on land.

TaranDahl:
> Default behavior, land & water, land only, water only are 4 cases; !isset, true, false are insufficient.

Default case; Naval=true-like case; Naval=false-like case. 3 cases.

> Hmm... It looks a bit strange. If the hook is placed at 0x62AB5E, it should cover .text:0062AB63 FF 92 84 00 00 00 call dword ptr [edx+84h]. Are you sure it's better than the currently used 0x62AB71?

Hook it wherever you like. 0x62AB69 is also valid. For 0x62AB71, I am not sure about it. Some jump code is recoverable while some is not. I do not understand the asm things very well.

> Can returning 0x62AB77 or 0x62AC0F really achieve the purpose of “The target position must not be on Water or Beach.”?

I thought you wanted to minimize the code changes, so I adjusted the requirements. Now it only handles matters related to Naval.

DeathFishAtEase:
> Default case; Naval=true-like case; Naval=false-like case. 3 cases.

It sounds like just an additional case of "force fail regardless of Naval" compared to the default case and the case after skipping the check?

TaranDahl:
> > Default case; Naval=true-like case; Naval=false-like case. 3 cases.
> 
> It sounds like just an additional case of "force fail regardless of Naval" compared to the default case and the case after skipping the check?

Yes. Intuitively, the behavior here should not be bound to Naval=\<bool\> (as it also governs production), so create a new key to take over its function.

DeathFishAtEase:
I have now renamed it to `Parasite.AllowWaterExit` and adopted the design proposed by TaranDahl
> Default case; Naval=true-like case; Naval=false-like case.

TaranDahl (approved):

Metadorius:
just realized, will also forbid land exit, or not? if yes -- this is kinda not reflected in the naming, could drop `Allow`

DeathFishAtEase:
not.

src/Ext/Techno/Hooks.cpp
```diff
@@ -842,6 +842,35 @@ DEFINE_HOOK(0x6298CC, ParasiteClass_AI_GrippleAnim, 0x5)
 	return SkipGameCode;
 }
 
+DEFINE_HOOK(0x62AB6F, ParasiteClass_CanExistOnVictimCell_AllowWaterExit, 0x8)
+{
+	GET(ParasiteClass*, pParasite, ESI);
+	auto const pOwner = pParasite->Owner;
+	if (pOwner)
```
NetsuNegi: Actually, there's no need sanity check for `pOwner` here.

src/Ext/Techno/Hooks.cpp
```diff
@@ -842,6 +842,35 @@ DEFINE_HOOK(0x6298CC, ParasiteClass_AI_GrippleAnim, 0x5)
...
+	{
+		auto const pTypeExt = TechnoTypeExt::ExtMap.Find(pOwner->GetTechnoType());
```
NetsuNegi: you can just `GET(TechnoTypeClass*, pOwnerType, EAX);` instead of `pOwner->GetTechnoType()`

NetsuNegi (approved):

PR #2250, opened 2026-06-11
[Retry] Allow infantry to perform type conversion when deploying and undeploying

DeathFishAtEase:
Reimplement #2239.
- Used a more performant method
- The execution timing of convert was changed from the start of Sequence to after its end

Retry  Convert (Un)deploy

*E1↔GGI, YURI↔YURIPR, E2↔BORIS*


Coronia:
The performance cost is now resolved
image

TaranDahl (approved):
@Starkku Going to merge this in 24h if you have no further instructions.

Starkku (approved):
Looks correct to me.

PR #2253, opened 2026-06-13
Allow customizing whether the creation of shrapnel weapon is controlled by the new target check on the warhead of the parent weapon

DeathFishAtEase:
During testing of #2237, it was found that warhead effect checks such as `AffectsBelowPercent` affect the triggering of shrapnel weapons, so I added this switch.

Old

*`AffectsBelowPercent=.5`*

New

*`AffectsBelowPercent=.5` & `Shrapnel.ObeyWarheadTriggerConditions=false`*

---

`Shrapnel.ObeyWarheadTriggerConditions` can be used to determine whether to use [Customizable Warhead trigger conditions](https://phobos.readthedocs.io/en/latest/Fixed-or-Improved-Logics.html#customizable-warhead-trigger-conditions) and #2237 when hitting units to decide if it triggers.

```ini
[CombatDamage]
Shrapnel.ObeyWarheadTriggerConditions=true  ; boolean

[SOMEPROJECTILE]                            ; Projectile
Shrapnel.ObeyWarheadTriggerConditions=      ; boolean, defaults to [CombatDamage] -> Shrapnel.ObeyWarheadTriggerConditions
```


NetsuNegi (approved):

TaranDahl:
I don't have time right now. Remind me to review this in a few days.

TaranDahl (approved):

docs/Fixed-or-Improved-Logics.md
```diff
@@ -1434,17 +1434,20 @@ BallisticScatter.Max= ; floating point value, distance in cells
...
 [CombatDamage]
 Shrapnel.IgnoreHitBuildings=false  ; boolean
+Shrapnel.UseNewWarheadChecks=true  ; boolean
 
 [SOMEPROJECTILE]                   ; Projectile
 Shrapnel.AffectsGround=false       ; boolean
 Shrapnel.AffectsBuildings=false    ; boolean
 Shrapnel.UseWeaponTargeting=false  ; boolean
 Shrapnel.IgnoreHitBuildings=       ; boolean
+Shrapnel.UseNewWarheadChecks=      ; boolean, defaults to [CombatDamage] -> Shrapnel.UseNewWarheadChecks
```
Thrifinesma: Compared to `New`, I would recommend using `Weapon` or `Sharpnel`. `Weapon` is consistent with existing tags, while `Sharpnel` has a clearer meaning but may make the tags look somewhat repetitive.

DeathFishAtEase: For consistency, I used function names as statement names when they basically describe the behavior. `New` is hard to omit because it specifically refers to the 4 new checks added by Phobos. Actually, I would prefer to use `CreateObeyWarheadChecks`, but it's a bit too long.

DeathFishAtEase: Also, the four items such as `AffectsBelowPercent` are checks on the warhead, not checks on the weapon. Checks on the weapon affect target selection, not whether it can hit and create shrapnel weapon.

Thrifinesma (approved):

PR #2254, opened 2026-06-14
Customize `Tiled` drawing interval and centering

DeathFishAtEase:
- In vanilla, the drawing interval of an animation with `Tiled=yes` is determined by the height of the rectangle formed by the non-transparent pixels of the first frame in the Shape resource file. Now you can customize it.
  - If `Tiled.Interval` is greater than `0`, the specified value is used; otherwise, the default rule applies.
  - `Tiled.AlignToCenter` can be used to change the alignment of the Shape resource file coordinates from the bottom center to the canvas center for the Animation entity's center.

In `artmd.ini`:
```ini
[SOMEANIM]                 ; AnimationType, with Tiled=yes
Tiled.Interval=0           ; integer, pixels
Tiled.AlignToCenter=false  ; boolean
```

---
Vanilla default behavior:
image

**Now:**

https://github.com/user-attachments/assets/eb775be2-db69-4ec7-9923-ed459af97987

(Of course, it could also be done before having this feature, but it would be much more cumbersome :P)


NetsuNegi (approved):

CREDITS.md
```diff
@@ -795,7 +795,7 @@ This page lists all the individual contributions to the project by their author.
 - **E1 Elite** - TileSet 255 and above bridge repair fix
 - **AutoGavy** - interceptor logic, Warhead critical hit logic
 - **Chasheen (Chasheenburg)** - CN docs help for Build#24
-- **Noble Fish** - some minor improvements and fixes, established Community Chinese docs, took over and completely rewrite the Official Chinese docs during Build#46, Dehardcode of parasites unlimboing after killing naval targets
+- **Noble Fish** - some minor improvements and fixes, established Community Chinese docs, took over and completely rewrite the Official Chinese docs during Build#46, Dehardcode of parasites unlimboing after killing naval targets, Customize the drawing interval for `Tiled`
```
TaranDahl: Formation issue

TaranDahl (approved):

src/Ext/Anim/Hooks.cpp
```diff
@@ -370,6 +370,39 @@ DEFINE_HOOK(0x4236F0, AnimClass_DrawIt_Tiled_Palette, 0x6)
...
+	R->ECX(*pValue);
+	return 0x423659;
+}
+
+DEFINE_HOOK(0x423660, AnimClass_DrawIt_Tiled_Center, 0x7)
+{
+	GET(AnimClass*, pThis, ESI);
+	GET(int, height, EAX); 
+	GET(int*, pVtable, ESI);
+	R->EDX(*pVtable);
```
NetsuNegi: should use `VTable::Get()` to get virtual table

NetsuNegi (approved):

PR #2257, opened 2026-06-16
Customize whether technos with `Locomotor=Fly` wobble

DeathFishAtEase:
- In vanilla, if technos use `Locomotor=Fly` and do not have `IsDropship=true`, they will have a hardcoded wobble effect. However, using `IsDropship=true` also introduces a series of hardcoded behaviors associated with it. Now, you can customize whether to disable this behavior, and it can also be used to enable this behavior for technos with `IsDropship=true`.
  - If not defined, the original behavior is used by default.

In `rulesmd.ini`:
```ini
[AudioVisual]
FlyNoWobbles=  ; boolean

[SOMETECHNO]   ; TechnoType with Locomotor=Fly
FlyNoWobbles=  ; boolean, defaults to [AudioVisual] -> FlyNoWobbles
```

---

Wobble

*FlyLocomotorVehicle*

NoWobble

*FlyLocomotorVehicle with `FlyNoWobbles=true`*

### DIFF
--- a/src/Ext/Techno/Hooks.cpp
+++ b/src/Ext/Techno/Hooks.cpp
@@ -2065,29 +2065,15 @@ DEFINE_HOOK(0x70AFEF, TechnoClass_UpdateSight_DynamicSight2, 0x6)
 
 #pragma endregion
 
-DEFINE_HOOK(0x4CF8B1, FlyNoWobbles, 0x6)
+DEFINE_HOOK(0x4CF8B1, FlyLocomotionClass_Draw_Point_NoWobbles, 0x6)
 {
-	enum { Continue = 0x4CF8B7 };
-	GET(TechnoTypeClass*, pType, EAX);
+    enum { Continue = 0x4CF8B7 };
+    GET(TechnoTypeClass*, pType, EAX);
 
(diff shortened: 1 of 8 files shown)


TaranDahl (approved):

src/Ext/Techno/Hooks.cpp
```diff
@@ -2065,6 +2065,31 @@ DEFINE_HOOK(0x70AFEF, TechnoClass_UpdateSight_DynamicSight2, 0x6)
...
 
+DEFINE_HOOK(0x4CF8B1, FlyNoWobbles, 0x6)
+{
+	enum { Continue = 0x4CF8B7 };
+	GET(TechnoTypeClass*, pType, EAX);
+
+	auto const pTypeExt = TechnoTypeExt::ExtMap.Find(pType);
+	const auto& globalVal = RulesExt::Global()->FlyNoWobbles;
+
+	if (pTypeExt->FlyNoWobbles.isset())
```
NetsuNegi: I think `R->CL(pTypeExt->FlyNoWobbles.Get(RulesExt::Global()->FlyNoWobbles.Get(pType->IsDropship)));` is better than current code.

NetsuNegi (approved):
LGTM

PR #2258, opened 2026-06-17
Fix: the banner render position not clamped to screen bounds

Chang-zhi:
- Clamp PCX/SHP/CSF banner render position to the visible area after centering, preventing off-screen drawing
- Prevent a potential crash when a PCX banner exceeds the top screen edge
- Add corresponding documentation and credits


Chang-zhi:
This is a before-and-after comparison.
BANNERPR

Coronia (approved):

DeathFishAtEase:
This is some opinions from ststl-s, one of the original authors of the feature:

> I think it's fine. You could consider adding a global feature, a tag that defaults to true, to control whether the Banner is always displayed within the ViewBounds or still partly outside, but it's probably not necessary. Map authors can just change the assets themselves.
> > 我认为是没什么问题的，可以考虑加一个全局功能，默认为true的标签，控制是否让Banner固定显示在ViewBounds内，还是仍然一部分在外面，不过应该没什么必要，地图作者自己改素材罢（

Chang-zhi:
> This is some opinions from ststl-s, one of the original authors of the feature:
> 
> > I think it's fine. You could consider adding a global feature, a tag that defaults to true, to control whether the Banner is always displayed within the ViewBounds or still partly outside, but it's probably not necessary. Map authors can just change the assets themselves.
> > > 我认为是没什么问题的，可以考虑加一个全局功能，默认为true的标签，控制是否让Banner固定显示在ViewBounds内，还是仍然一部分在外面，不过应该没什么必要，地图作者自己改素材罢（

I see your point, and I'm a bit worried. If a mapmaker turns the tag off, the crash would still happen when a PCX banner goes above the screen. I'm not sure that's acceptable. Unfortunately, I'm not familiar enough with reverse engineering to add a new hook that modifies the game's PCX rendering logic.

DeathFishAtEase:
> I see your point, and I'm a bit worried. If a mapmaker turns the tag off, the crash would still happen when a PCX banner goes above the screen. I'm not sure that's acceptable. Unfortunately, I'm not familiar enough with reverse engineering to add a new hook that modifies the game's PCX rendering logic.

From ststl-s:

> Mainly, if someone tries to make a large image keep moving toward the screen center, forcibly restricting it would require the author to crop a lot of assets.
> It depends on whether this crash is acceptable to users and reviewers; for example, it might be possible to clearly state in the documentation that specific situations will cause a crash,
> so that modders can avoid it by reading the documentation.
> > 主要是如果有人试图做一个很大的图像一直向屏幕内移动的话如果强行限制了的话作者就需要剪非常多的素材
> > 需要看崩溃是否能接受的，例如说明书里写特定情况会崩溃
> > 这种不这样做就不会崩溃的或许可以

PR #2262, opened 2026-06-20
Enhanced Berzerk behavior

TaranDahl:
### Enhanced Berzerk behavior

- In vanilla, when a unit enters Berzerk state, the game assigns it a `Hunt` mission, which may cause the unit to choose a very distant target and simply move toward it without firing during the Berzerk duration.
- Now you can customize which mission the unit uses when entering Berzerk state.
  - Additionally, this enhancement adds the necessary stop handling (clearing destination), fixing jumpjet vehicles behaving incorrectly when getting Berzerk'd while moving.

In `rulesmd.ini`:
```ini
[CombatDamage]
BerzerkMission=Hunt               ; MissionType
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -799,6 +799,7 @@ This page lists all the individual contributions to the project by their author.
   - Allow disable an over-optimization in targeting
   - Extra threat
   - Fix the incorrect mission switching in infantry EnterIdleMode
+  - Customizable Berzerk mission
   - Fix the issue that `BombSight` not being updated correctly in techno conversion
   - Fix BalloonHover incorrectly considering ground factors when pathfinding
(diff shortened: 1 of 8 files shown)


Coronia:
would it be better to make it an enum tag that choose the mission of units under berzerk? Like EMPAIRecoverMission, AIParadropMission etc

TaranDahl:
> would it be better to make it an enum tag that choose the mission of units under berzerk? Like EMPAIRecoverMission, AIParadropMission etc

Done.

Coronia (approved):
maybe it's better to place the hook in Hooks.ReceiveDamage.cpp

TaranDahl:
> maybe it's better to place the hook in Hooks.ReceiveDamage.cpp

Done.

PR #2179, opened 2026-04-24
Knock-up, Traction, and forced displacement locomotor

TaranDahl:
### Knock-up warhead

- You can now launch targets in an arc with projectiles. The target will be knocked up in the opposite direction of its current facing.
  - `KnockUp` controls whether to enable this logic.
  - `KnockUp.Range` defines the knock-up distance.
  - `KnockUp.Speed` defines the horizontal knock-up velocity.
  - `KnockUp.Angle` defines the knock-up angle.

In `rulesmd.ini`:

```ini
[SOMEWARHEAD]            ; WarheadType
KnockUp=false            ; boolean
KnockUp.Range=0.0        ; floating point, cells
KnockUp.Speed=0.0        ; floating point, cells/frame (horizontal component)
KnockUp.Angle=45.0       ; floating point, degrees
```

### Traction warhead

- `Traction` pulls affected targets toward the explosion point.
  - `Traction` controls whether to enable this logic.
  - `Traction.Range` is treated as a movement "budget" (floating point, in cells) used to limit how far the pull can progress.
  - `Traction.Speed` controls the linear movement speed applied when the pull occurs (cells/frame).

In `rulesmd.ini`:

```ini
[SOMEWARHEAD]            ; WarheadType
Traction=false           ; boolean
Traction.Range=0.0       ; floating point, cells
Traction.Speed=0.0       ; floating point, cells/frame
```

PR #2265, opened 2026-06-22
Port Vector effect system from KratosAI to PhobosAI

KenosisM:
Ported from Kratos code base, the phobos ver of the all mighty vector attach effect.  All by vibe coding (including this PR) so human review is very much needed.
preview gifs (kratos based):
https://www.bilibili.com/opus/1214163469064994817
https://www.bilibili.com/opus/1214494881641136144
https://www.bilibili.com/opus/1215222921544335378
The incredibly magical Vector class mechanism, capable of directly displacing projectiles and units.
Ported from Kratos by Deepusheek.

**Function Description**
Vector is a pure movement effector that forcibly modifies the coordinates of the attached object during the Duration.
Although it also works on Techno, its greatest use lies in:

Programmable, scriptable projectile trajectories.
You can use Vector to create incredibly stunning projectile flight paths.
After the Vector effect ends, the projectile will proceed to attack its original target. Therefore, it is recommended to use only ROT=100.
Each Vector can use a Next statement, allowing other Vectors to be activated after its own effect ends, enabling you to freely combine effects as you wish.

Vector abstracts "displacement" into five independent modes, with only one mode active per AE at a time.
Priority-based matching: Freeze > Circle > MoveTo > ReachTarget > Speed

**Freeze**: Frozen in mid-air, motionless.
**Circle**: Draws a circle. The center of the circle itself can also continue moving.
**MoveTo**: Directly provides a set of coordinates, moving frame by frame. For example, MoveTo=0,0,-10 is equivalent to the vanilla Vertical. If the effect ends halfway, the projectile will drop first, then charge toward the target.
**ReachTarget**: Takes the current coordinate difference and Duration, ensuring the projectile reaches the target exactly when Duration ends, ignoring speed settings. An arc height can be set. Adding an arc height makes it equivalent to the vanilla Arcing-type projectile.
**Speed**: Uses a custom speed setting or reads the original speed for straight-line flight.

The following image demonstrates the effects of the ReachTarget mode combined with random arc height, random arc rotation, and random arc vertices.
B

All tags are written under [AttachEffectSection], prefixed with "Vector."
As an AE, it needs to be registered under AttachEffectTypes.
Due to Phobos's AE underlying architecture, projectiles do not accept AEs. You can only write the Vector on the firer, indicating that projectiles fired by this firer will be subject to displacement.
For example:
[FV]
AttachEffect.AttachTypes=SomeVector


src/Ext/Bullet/Body.cpp
```diff
@@ -511,6 +551,14 @@ BulletExt::ExtContainer::ExtContainer() : Container("BulletClass") { }
 
 BulletExt::ExtContainer::~ExtContainer() = default;
 
+#include <Ext/Techno/Body.h>
+#include <New/Entity/VectorState.h>
```
NetsuNegi: Please place `#include` at the file beginning

src/New/Entity/VectorState.cpp
```diff
@@ -0,0 +1,752 @@
...
+	}
+	else
+	{
+		pObject->SetLocation({ currentPos.X + moveDisp.X, currentPos.Y + moveDisp.Y, currentPos.Z + moveDisp.Z });
+		if (pType->Vector_Freeze)
+		{
+			if (auto const pFoot = abstract_cast<FootClass*>(pObject))
+				pFoot->StopMoving();
+		}
+	}
```
TaranDahl, line 750: Pretty sure these are not enough. Simply SetLocation may be good for BulletClass, but not for FootClass.
I suggest creating a Locomotor to handle this. From my understanding, many parts of the game access the Locomotor of FootClass. Moving characters without going through the Locomotor may cause inconsistencies at these access points.
See https://github.com/Phobos-developers/Phobos/pull/2179. Due to the complexity of this section, I cannot guarantee that it is completely correct.

KenosisM: ofc. BTW I don't think vector on techno is good - maybe just keep the bullet part, or only support jumpjet&aircraft

Metadorius: yes, @TaranDahl is completely correct. if you want to manipulate FootClass movement - you must do it through locomotors. Otherwise you're opening a can of worms and fighting the game design.

I did Techno Attachment this way first, it went completely bonkers on many locomotors and had inexplicable bugs, freezes and crashes. When I remade it to be a locomotor - everything got fine.

KenosisM: Actually I dunno what the heck Deepusheek has written. It needs ofc serious redo - or at least give it stable examples of how things should work.

KenosisM:
[Phobos_Vector_Manual.html](https://github.com/user-attachments/files/29405080/Phobos_Vector_Manual.html)
[Phobos_Vector_Manual_CN.html](https://github.com/user-attachments/files/29405081/Phobos_Vector_Manual_CN.html)
Added manual files

mevitar:
I think it's better if this is separated from AttachEffect and instead made its own section, with a vector list also separated from AttachEffectTypes (and call it something like AttachVector and AttachVectorTypes).

KenosisM:
> I think it's better if this is separated from AttachEffect and instead made its own section, with a vector list also separated from AttachEffectTypes (and call it something like AttachVector and AttachVectorTypes).

Perhaps. In kratos prototype it merges perfectly into AE infrastructure so the ported version is also started as AE. Maybe a new list, or better no listing at all. The more important part is let projectiles directly accept vector.

PR #2266, opened 2026-06-23
Improvements to tank bunker logic

Starkku:
- Introduced alternate code paths for odd-sized foundations in relevant places to allow odd-sized tank bunkers f.ex 3x3 ones.
  - The logic to make vehicles enter center of building on even-sized foundation hinges on function (`Force_Track()`) not implemented on all locomotors and does not behave in visually consistent manner on odd-sized foundations. Conversely the simpler approach with discrete center cells present cannot be used with even-sized foundations.
  - Non-Drive locomotors should hypothetically work better with the discrete center cell approach but since there's no guarantee of parity I have elected to not lift the hardcoded locomotor restrictions.
- Added customization for the logic update delay that can be used to make tank bunkers more responsive on unit attempting to enter.
- Documentation with prior changes to tank bunker logic was consolidated under single header since there's not a large amount of changes.

-------------------------------------

### Tank Bunker improvements

- `Bunker=true` and associated logic should now work more reliably on odd-sized foundations. It still expects size on both X and Y axis to be equal e.g `2x2` or `3x3` and will not behave correctly otherwise.
- It is now possible to change the tank bunker logic update delay by setting `BunkerStateUpdateDelay`. Defaults to `[General]` -> `BunkerStateUpdateDelay`. Lowering the value from the default value of 15 will make units trying to bunker up spend less time waiting.

In `rulesmd.ini`:
```ini
[General]
BunkerStateUpdateDelay=15  ; integer, game frames

[SOMEBUILDING]             ; BuildingType
BunkerStateUpdateDelay=    ; integer, game frames
```


TaranDahl (approved):

Coronia:
tested entering, attacking, unloading and bunker penetration on 1x1, 3x3 and 10x10 bunkers (still use the same image). All seem to work properly
image
image
image

DeathFishAtEase:
Hey, guys, I just saw this `BunkerStateUpdateDelay` definition while checking this week's new additions, and it reminded me of an edge case: `[BuildingType] -> ConstructionYard=true` would give Tank Bunker an effect similar to skipping the update delay. Did you investigate this during the development and testing of this PR?

@Starkku @Coronia 

---
Engine environment: DevBuild#48

BunkerTime

In `rulesmd.ini`:
```ini
[BuildingTypes]
+=NATBNK_2
[NATBNK_2]:[NATBNK]
Image=NATBNK
ConstructionYard=true
```

Starkku:
> Did you investigate this during the development and testing of this PR?

Nothing to investigate. `BuildingClass_Mission_Repair()` has separate code path for `ConstructionYard=true` building types that is checked after bunker update but before everything else that always returns with 1 frame delay until next mission logic check. `ConstructionYard` affects other things so I wouldn't recommend using it for this purpose. Only reason I made the delay for bunkers specifically customizable is cause it affects game balance if the default is changed, I did initially just hardcode it to 1 but made it customizable as the last thing I changed code-wise for this PR.

PR #2268, opened 2026-06-23
Auto-remove earliest beacon

TaranDahl:
### Auto-remove earliest beacon

- In vanilla, each player can place up to 3 beacons. When all 3 beacon slots are occupied, attempting to place a new beacon does nothing. Now you can make the game automatically remove the earliest-placed beacon to free up a slot for the new one.
  - `[General] -> AutoRemoveEarliestBeacon` controls this behavior.

In `rulesmd.ini`:
```ini
[General]
AutoRemoveEarliestBeacon=no   ; boolean
```

### DIFF
--- a/src/Ext/House/Hooks.cpp
+++ b/src/Ext/House/Hooks.cpp
@@ -6,7 +6,11 @@
 #include "Ext/Building/Body.h"
 #include <Ext/Event/Body.h>
 
+#include <BeaconManagerClass.h>
+
 #include <unordered_map>
+#include <algorithm>
+#include <utility>
 
 // Trigger power recalculation on gain/loss of any techno, not just buildings.
 DEFINE_HOOK_AGAIN(0x5025F0, HouseClass_RegisterGain, 0x5) // RegisterLoss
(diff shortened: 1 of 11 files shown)


CnCRAZER:
Fucking finally. This PR needs to be expedited to top of the priority list 😆

src/Ext/House/Hooks.cpp
```diff
@@ -681,3 +684,67 @@ DEFINE_HOOK(0x45063F, BuildingClass_UpdateRepairSell_PlayerAutoRepair, 0x6)
...
+
+DEFINE_HOOK(0x43131B, BeaconManagerClass_DeleteBeacon_RecordOrder, 0x5)
+{
+	GET(int, beaconIdx, EBX);
+	GET(int, houseIdx, ECX);
+
+	const auto pHouse = HouseClass::Array.GetItem(houseIdx);
+	const auto pExt = HouseExt::ExtMap.Find(pHouse);
+
+	const int oldValue = pExt->BeaconsPlacedOrder[beaconIdx];
```
Otamaa: can use `std::exchange` here i suppose :
`const int oldValue = std::exchange(pExt->BeaconsPlacedOrder[beaconIdx], 0);`

TaranDahl: Done.

TaranDahl:
Should test whether this feature works properly in multiplayer games, as well as its interaction with the beacon deletion function.

TaranDahl:
@CnCRAZER Can you test if this works correctly in MP?

QinJie29:
Tested in MP
动画11
There is a bug where, if a player attempts to place a new beacon after already placing three beacons, the oldest beacon is removed. However, if the placement is cancelled, the removed beacon does not reappear for that player, even though it remains visible to other players.

TaranDahl:
@QinJie29 You may try it again.

QinJie29:
Tested in MP.The issue appears to be fixed.
动画0
@TaranDahl @Coronia @DeathFishAtEase What else needs to be tested?

TaranDahl:
> What else needs to be tested?

No more, I think.

DeathFishAtEase (commented):
It seems that the hook function does not check the value of `AutoRemoveEarliestBeacon`.

DeathFishAtEase (approved):

PR #2269, opened 2026-06-24
Allow chat box in singleplayer

TaranDahl:
### Allow chat box in singleplayer

- In vanilla, the in-game chat box is disabled in singleplayer scenarios. You can now enable it by setting `[General] -> AllowChatBoxInSinglePlayer` to true.
  - `AllowChatBoxInSinglePlayer` controls whether the chat box is available in singleplayer mode. Defaults to `false`.

In `rulesmd.ini`:
```ini
[General]
AllowChatBoxInSinglePlayer=false  ; boolean
```


Starkku:
Not really an issue per se but could have maybe put up something in discussions or elsewhere that describes the roadmap/endgame for these multiplayer -> singleplayer feature changes to clue rest of the developers and others in. As standalone changes it isn't particularly clear what the objective is.

The actual individual changes can stay as separate PR's of course.

TaranDahl:
> Not really an issue per se but could have maybe put up something in discussions or elsewhere that describes the roadmap/endgame for these multiplayer -> singleplayer feature changes to clue rest of the developers and others in. As standalone changes it isn't particularly clear what the objective is.
> 
> The actual individual changes can stay as separate PR's of course.

I suspect that the objective is unachievable on the develop branch.
The primary objective is to build an agentic AI system that enables LLMs to control units and production in Red Alert 2. Players can communicate with the agent via text input, beacons (and possibly voice commands).
The roadmap is as follows: enable the chatbox and beacons → develop a full tactical command agent → integrate the agent into Red Alert 2. 
Given Phobos's current efficiency in reviewing PRs of Tier 2 and above, I believe the second stage is unlikely (and perhaps ought not) to be merged into Phobos. Therefore, the Phobos-related roadmap essentially ends here. Do you think I should start a Discussion or something similar regarding this full yet unfeasible roadmap? I will do that if you think it should be done.

Starkku:
> Do you think I should start a Discussion or something similar regarding this full yet unfeasible roadmap?

I don't know

Metadorius:
> I suspect that the objective is unachievable on the develop branch. The primary objective is to build an agentic AI system that enables LLMs to control units and production in Red Alert 2. Players can communicate with the agent via text input, beacons (and possibly voice commands). The roadmap is as follows: enable the chatbox and beacons → develop a full tactical command agent → integrate the agent into Red Alert 2. Given Phobos's current efficiency in reviewing PRs of Tier 2 and above, I believe the second stage is unlikely (and perhaps ought not) to be merged into Phobos. Therefore, the Phobos-related roadmap essentially ends here. Do you think I should start a Discussion or something similar regarding this full yet unfeasible roadmap? I will do that if you think it should be done.

Better proposal: implement Age of Empires 2 AI chat commands or something like this.

11EJDE11:
You can set `ForceMultiplayer=True` in the spawner to get chat too btw.

IMO LLMs wouldn't be great for this. See https://openra-rl.dev/ which is lucky to get any kills. Something like OpenAI Five would be killer if we had the budget :(

CnCRAZER:
Side note: I think it would be pretty cool for testing/debugging maps to have the ability to turn on the chat option and while in-game to change rules values.

For Example: Type in `/[General] BuildSpeed=0` to change buildspeed while In-game(skirmish only I suppose)

TaranDahl:
> Side note: I think it would be pretty cool for testing/debugging maps to have the ability to turn on the chat option and while in-game to change rules values.
> 
> For Example: Type in `/[General] BuildSpeed=0` to change buildspeed while In-game(skirmish only I suppose)

Maybe this is the thing you are looking for. https://github.com/Xkein/YRDynamicPatcher#examples
Note: This feature usually only works properly on ini keys that do not require initialization. Modifying keys such as Locomotor within the game will likely lead to undesirable outcomes.

TaranDahl:
> You can set `ForceMultiplayer=True` in the spawner to get chat too btw.

I've looked through the source code of Spawner. IsSinglePlayer (also known as ForceMultiplayer) controls more elements than I need, and its implementation approach differs as well. I believe there should be no conflicts in terms of functional design and code implementation.

> IMO LLMs wouldn't be great for this. See https://openra-rl.dev/ which is lucky to get any kills. Something like OpenAI Five would be killer if we had the budget :(

The core goal of my concept is to make AI conversable and commandable, rather than making it extremely intelligent enough to defeat powerful opponents. LLMs are indeed not good at playing games, yet they excel at conversation. Therefore, my plan is to encapsulate the game operation components as much as possible, leaving the AI primarily responsible for receiving players' commands and converting them into decisions, and only directly controlling game units when necessary (upon explicit requests from players).

Coronia:
before doing the grand LLM thing, maybe we can make a 'cheat code' system with it first, like input 'whoisyourdaddy' to make all units immune to damage, like many other modern RTS did. This should be much easier to do than LLM theoratically and is indeed useful

TaranDahl:
So, first of all, who is going to review this code? 🤣

QinJie29:
Tested in campaign
动画
Tested in skirmish
动画2
@TaranDahl @Coronia @DeathFishAtEase What else needs to be tested?

DeathFishAtEase (approved):
LGTM

PR #2270, opened 2026-06-25
Dropship Loadout remake

FS-21:
dropship_loadout_default_view

*Note: Isn't a logic restoration, is a full remake of the unfinished TS logic but expanded (why not).*

*Note 2: Due to the long list of variables for customizing this logic please check the documentation of this PR for updated content.*

General Features of this single player logic:

- Can be opened before the game start, during the game with a map action & with a SW.

- You can choose a fixed budget in this window or use the player's own economy.

- UI feedback on cameos (blue for dragging, purple to indicate restricted/forbidden, red for selling, green for buying, yellow to indicate the last unit purchased).

- Drag-and-drop mechanics: Drag units from the right panel to the transports to buy them. Drag them from the transports back to the right panel to sell them.

- Starting units can be set to appear in the transport at no cost to the player.

- Fixed units can be set so they cannot be sold.

- Each transport can have size limits to prevent heavy units from being loaded.

- All SHP graphics can be replaced with their PCX versions.

- Tooltips with unit information, including name, availability, cost, and description.

- Countries can have their own graphics customization and unit lists that overwrite the map's [Basic] configuration. This doesn't include the Superweapons case.

- Map Action 900 to open the Dropship Loadout window during the match. A different list of available units can be displayed each time this map action is called.

- Map Action 901 to launch the specified transport (if available). Regarding the initial Team configuration: Although the Taskforce is dynamically generated with the selected transport, the Team must be configured by indicating the starting waypoint used to spawn the transport, and the user must configure the script that the transport will execute.

- Superweapons can be used to open this Dropship Loadout window (capped to only 1 transport), and the transport launch can be executed via a separate SW from the one used to open the window. Note that the initial screen & Map actions doesn't share the graphics and unit lists with the superweapons data, both methods are independent.

- Can be used any air/ground/Subterranean transport.


src/Misc/Hooks.DropshipLoadout.cpp
```diff
@@ -0,0 +1,4209 @@
...
+				if (j < (int)pSWTypeExt->DropshipLoadout_DropshipCameosLocations.size())
+				{
+					offsetX = pSWTypeExt->DropshipLoadout_DropshipCameosLocations[j].X;
+					offsetY = pSWTypeExt->DropshipLoadout_DropshipCameosLocations[j].Y;
+				}
+				int cameoX = backgroundX + offsetX;
+				int cameoY = backgroundY + offsetY;
+				list.push_back({ cameoX, cameoY, cameoWidth, cameoHeight });
+			}
+			dropshipBayCameLocations.push_back(list);
```
Otamaa: Prefer to use `emplace_back` to reduce copy-move 
you can do : 
```c++
auto& list = dropshipBayCameLocations.emplace_back(); // create the empty list 

//... 
list.emplace_back(cameoX, cameoY, cameoWidth, cameoHeight); // crate rectangle to the vector
```

Otamaa (dismissed):
Pleae review and apply the change requested to the whole code , since the pattern are repeated all over the place , thanks .

src/Misc/Hooks.DropshipLoadout.cpp
```diff
@@ -0,0 +1,4209 @@
...
+}
+
+void DropshipLoadout_OnMouseWheelDown()
+{
+	pendingScrolls++;
+}
+
+static ShapeButtonClass* CreateShapeButton(unsigned int nID, int nX, int nY, int nWidth, int nHeight, bool bIsAlpha)
+{
+	auto const pButton = GameAllocator<ShapeButtonClass>().allocate(1);
```
Otamaa, line 105: i dont really understand this part , why dont use `GameCreate<ShapeButtonClass>` ? , also for the `0x69DD30` are you sure that is correct args ? , the IDB it written like this : 
`ShapeButtonClass(ShapeButtonClass *this, int id, ShapeHeaderStruct *shape, int x, int y, int w, int h, char a8)`

FS-21:
Sorry but I added a new variable that I consider useful for missions:

Now each transport can have customizable slots istead of following a generic slots size (DropshipLoadout.DropshipCameosCount).

So one transport can have 1 slot, another 4, another 8...
If DropshipLoadout.DropshipN.CameosCount isn't set then DropshipLoadout.DropshipCameosCount will be used as generic fallback.

src/Misc/Hooks.DropshipLoadout.cpp
FS-21: Applied feedback

src/Misc/Hooks.DropshipLoadout.cpp
FS-21: Because when I tried to compile in Release the game always crashed (I only could compile & run in Debug). I had to use Gemini AI for finding the annoying crash problem:
image
image

FS-21:
I really forgot unharcode the texts with CSF strings:
- For customizing or localizing the interface strings, you can add the following entries to your `.csf` file:
  - `TXT_DROPSHIP_CREDITS` - The text for the credits/money label. Default value: `Credits: %d`
  - `TXT_DROPSHIP_START_MISSION` - The text prompting the player to start the mission. Default value: `Press SPACE to start the mission`
  - `TXT_DROPSHIP_AVAILABLE` - The label text for availability in the tooltip. Default value: `Available: `
  - `TXT_DROPSHIP_COST` - The label text for unit cost in the tooltip. Default value: `Cost: `
  
  Now I really feel I can't add more features.

src/Misc/Hooks.DropshipLoadout.cpp
Otamaa: Yeah , YRPP has the bugged CTOR for sure, but iam not sure where the AI assume it to be the `ConverClass* pDrawer` since in our IDB there is no such construtor avaible but instead at the `0x69DD30 -> ShapeButtonClass::ShapeButtonClass(int id, ShapeHeaderStruct *shape, int x, int y, int w, int h, char a8)` , which is the correct args count of 7 .

docs/User-Interface.md
```diff
@@ -563,6 +563,149 @@ In `uimd.ini`:
...
+- `DropshipLoadout.DropshipN.FixedUnits` (or `DropshipLoadout.FixedUnits.DropshipN` for backward compatibility) - A list of TechnoTypes pre-placed as fixed cargo in dropship index `N`. Players cannot sell or remove these units.
+- `DropshipLoadout.DropshipN.InitialUnits` - A list of TechnoTypes pre-placed as customizable cargo in dropship index `N`. Players can sell or remove these units to refund their cost. Unlike preloaded cargo, their cost is not deducted from starting money. These units are one-time use: once the dropship is generated, this list is cleared for the rest of the match.
+- `DropshipLoadout.BuyClickSound` - Sound played when purchasing a unit cameo.
+- `DropshipLoadout.SellClickSound` - Sound played when selling/removing a unit cameo.
+- `DropshipLoadout.ArrowsClickSound` - Sound played when clicking navigation arrows.
+- `DropshipLoadout.StartingDragDropSound` / `DropshipLoadout.EndingDragDropSound` - Sounds played when starting or ending dragging/dropping cameo items.
+
+In `rulesmd.ini` (under country sections):
```
DeathFishAtEase, line 695: Huh, is this AI-written documentation? And there are alignment issues with the comments inside the ini block.

docs/New-or-Enhanced-Logics.md
```diff
@@ -1173,6 +1173,111 @@ In `rulesmd.ini`:
...
+In `rulesmd.ini`:
+```ini
+[SOMESW]                              ; SuperWeaponType
+DropshipLoadout.Launch=false          ; boolean
+DropshipLoadout.Carrier=              ; TechnoType
+DropshipLoadout.PersistentCargo=false  ; boolean
+DropshipLoadout.VeteranLevel=1        ; integer
+```
+
+---
```
DeathFishAtEase, line 1275: Strange formatting

docs/New-or-Enhanced-Logics.md
```diff
@@ -1173,6 +1173,111 @@ In `rulesmd.ini`:
...
+
+#### Open Window
+
+These parameters configure the interactive Dropship Loadout purchase window that opens when the superweapon is fired:
+
+* `DropshipLoadout.OpenWindow`: If set to `true`, firing this superweapon opens the interactive Dropship Loadout configuration window.
```
DeathFishAtEase, line 1281: Our unordered list does not use `* ` but uses `- `.

DeathFishAtEase (commented):
Insane amount of code! Anyway, thanks for the contribution and maintenance. I've always liked this design, but come to think of it, the last time I experienced it was back in the Npatch era :P

docs/User-Interface.md
FS-21: Yes! Too little free time for documenting it so I asked to study the format around the target and replicate it. Let me check this Monday that I'll be able to continue with the reported feedback.

I'll check manually the docs

docs/New-or-Enhanced-Logics.md
FS-21: I think I fixed all these mistakes.

docs/New-or-Enhanced-Logics.md
FS-21: Yes, fixed.

PR #2273, opened 2026-06-27
Custom cruise missiles

DeathFishAtEase:
From RockPatch to Ares, custom missiles have never had a way to enter cruise missile mode, so neither RP's `MissileRaiseRate` nor Ares' `Missile.RaiseRate` have ever been effective, and modders cannot create another type of cruise missile beyond the type set in `[General] -> CMislType=`. Now, you can customize whether a custom missile is a cruise missile.

Missile Raise

- The take-off animation of a cruise missile is actually the continuously created trail smoke during the ascent phase, rather than the animation created only once at launch like a conventional missile. Now, the creation interval of this animation can be customized via `Missile.TakeOffSeparation`.

In `rulesmd.ini`:
```ini
[SOMEAIRCRAFT]                ; AircraftType with Missile.Custom=yes
Missile.Cruise=false          ; boolean
Missile.TakeOffSeparation=24  ; integer
```

> [!Note]
> Like Ares' `Missile.TrailerSeparation`, `Missile.TakeOffSeparation` also works without requiring `Missile.Custom=true`, which means it can be directly applied to the unit with `[General] -> CMislType=` without completely rewriting it as a custom missile.

> ## :page_facing_up: Wiki & Documentation (See Also)
> - [MissileSpawn Control on ModEnc](https://modenc.renegadeprojects.com/MissileSpawn_Control)
> - [Custom Missiles - Ares documentation](http://ares-developers.github.io/Ares-docs/new/custommissiles.html)

---

The original implementation comes from my modified Ares, thanks to @NetsuNegi for the help provided during the porting process.


TaranDahl:
Currently unavailable. Will review this later.

src/Ext/TechnoType/Body.cpp
```diff
@@ -1204,9 +1212,15 @@ void TechnoTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 
 	this->LandingAnim.Read(exINI, pSection, "LandingAnim");
 
+	this->IsCruiseMissile.Read(exINI, pSection, "Missile.Cruise");
```
TaranDahl: It is best to keep variable names consistent with key names.

NetsuNegi (approved):

TaranDahl (approved):

src/Ext/TechnoType/Body.cpp
DeathFishAtEase: Done

PR #2276, opened 2026-06-28
AutoDeath based on player power status and player credits

Flactine:
Add new auto-death conditions based on the owner's power status and credits.

- `PlayerPowerState`: The object will die if its owner's power status matches the configured state.
  - `low` / `consumer`: Trigger when the owner is in low power.
  - `full` / `normal`: Trigger when the owner is not in low power.
- `PlayerMoney.Max` / `PlayerMoney.Min`: The object will die based on the owner's available credits.
  - If only `PlayerMoney.Max` is set, triggers when money is not above this value.
  - If only `PlayerMoney.Min` is set, triggers when money is not below this value.
  - If both are set, triggers when money is **inside the range**.
---
- `PlayerPowerState`：当对象所属方的电力状态符合设定条件时，则对象死亡。
  - `low` / `consumer`：当对象所属方处于电力不足状态时触发。
  - `full` / `normal`：当对象所属方未处于电力不足状态时触发。
- `PlayerMoney.Max` ／ `PlayerMoney.Min`：根据对象所属方的资金决定对象是否死亡。
  - 仅设置 `PlayerMoney.Max` 时，当资金不高于该值时触发。
  - 仅设置 `PlayerMoney.Min` 时，当资金不低于该值时触发。
  - 同时设置时，资金处于**区间内**时触发。

In `rulesmd.ini`:
```ini
[SOMETECHNO]                         ; TechnoType
AutoDeath.PlayerPowerState=none     ; enumeration (none | low/consumer | full/normal)
AutoDeath.PlayerMoney.Max=-1         ; integer
AutoDeath.PlayerMoney.Min=-1         ; integer
```

QQ20260628-232702


TaranDahl (approved):

TaranDahl:
Note: You can use AI skill for generating Chinese docs.

src/Ext/Techno/Body.Update.cpp
```diff
@@ -332,6 +332,18 @@ bool TechnoExt::ExtData::CheckDeathConditions(bool isInLimbo)
 				: std::all_of(vTypes.begin(), vTypes.end(), existSingleType);
 		};
 
+	// death if player power status is not as specified
+	if (pTypeExt->AutoDeath_PlayerPowerStatus != PlayerPowerStatus::None)
+	{
+		const bool isLowPower = pOwner->PowerDrain > pOwner->PowerOutput;
+
+		if ((pTypeExt->AutoDeath_PlayerPowerStatus == PlayerPowerStatus::Normal && !isLowPower) || (pTypeExt->AutoDeath_PlayerPowerStatus == PlayerPowerStatus::Consumer && isLowPower))
```
TaranDahl: You can reuse the function 4FCE30.

Coronia (commented):
would be better to rename the tag to none/low/full since it's been widely used in many places already

src/Ext/Techno/Body.Update.cpp
```diff
@@ -332,12 +332,42 @@ bool TechnoExt::ExtData::CheckDeathConditions(bool isInLimbo)
...
+				shouldDie = (money < lessThan || money > moreThan);
+		}
+		else if (lessThan != -1)
+		{
+			shouldDie = (money < lessThan);
+		}
+		else
+		{
+			shouldDie = (money > moreThan);
+		}
```
TaranDahl: shouldDie = (money < lessThan) || (moreThan != -1 && money > moreThan)

src/Utilities/Enum.h
```diff
@@ -188,6 +188,13 @@ enum class AutoDeathBehavior
 	Sell = 2,     // buildings only
 };
 
+enum class PlayerPowerStatus
+{
+	None = 0,
+	Normal = 1,   // not low power
+	Low = 2, // low power
+};
```
Coronia, line 207: PlayerPowerStatus -> PowerStatus
Normal -> Full

src/Utilities/TemplateDef.h
```diff
@@ -1625,6 +1625,33 @@ if(_strcmpi(parser.value(), #name) == 0){ value = __uuidof(name ## LocomotionCla
...
+	inline bool read<PlayerPowerStatus>(PlayerPowerStatus& value, INI_EX& parser, const char* pSection, const char* pKey)
+	{
+		if (parser.ReadString(pSection, pKey))
+		{
+			static const std::pair<const char*, PlayerPowerStatus> Names[] =
+			{
+				{"none", PlayerPowerStatus::None},
+				{"consumer", PlayerPowerStatus::Low},
+				{"low", PlayerPowerStatus::Low},
+				{"normal", PlayerPowerStatus::Normal},
```
Coronia: remove consumer since it's not a clear term here
change normal to full

src/Ext/TechnoType/Body.cpp
```diff
@@ -806,6 +806,14 @@ void TechnoTypeExt::ExtData::LoadFromINIFile(CCINIClass* const pINI)
 	this->AutoDeath_TechnosExist_Any.Read(exINI, pSection, "AutoDeath.TechnosExist.Any");
 	this->AutoDeath_TechnosExist_AllowLimboed.Read(exINI, pSection, "AutoDeath.TechnosExist.AllowLimboed");
 	this->AutoDeath_TechnosExist_Houses.Read(exINI, pSection, "AutoDeath.TechnosExist.Houses");
+	this->AutoDeath_PlayerPowerStatus.Read(exINI, pSection, "AutoDeath.PlayerPowerStatus");
+	this->AutoDeath_PlayerMoney_Max.Read(exINI, pSection, "AutoDeath.PlayerMoney.Max");
+	this->AutoDeath_PlayerMoney_Min.Read(exINI, pSection, "AutoDeath.PlayerMoney.Min");
```
Coronia, line 833: rename variables names to be identical with tag names

Coronia (changes requested):
for code style and reusablity, some tag/variable names should be changed, especially `PowerStatus` which already has common terminologies and is likely to be reused in the future

src/Utilities/TemplateDef.h
DeathFishAtEase: I think there is no need to worry about the expression, because the naming design here is aligned with Ares' `Degrade.AmountNormal` and `Degrade.AmountConsumer`, so it is not a case of "using this term for the first time".

DeathFishAtEase:
@Coronia I think the content of Request changes has been satisfied; it is enough to resolve conflicts and let testers test again. (The latest test post under the corresponding forum topic was compiled by the user themselves because no new Nightly Build was generated, rather than being generated by the workflow.)

Coronia (approved):

PR #2279, opened 2026-07-01
Customize the step limit of the credits indicator and disable the credits indicator smooth transition

DeathFishAtEase:
- In vanilla, the Credits Indicator in the sidebar increases by at most 143 points per frame, and now you can customize this limit.
  - If set to a value less than or equal to 0, the upper limit will be removed.

In `uimd.ini`:
```ini
[Sidebar]
CreditsIndicator.MaxStep=143  ; integer
```

- In vanilla, the Credits Indicator has a smooth transition effect when the displayed credit value approaches the actual credit value. Now you can disable this effect, allowing the change step to always follow the value of [`CreditsIndicator.MaxStep`](User-Interface.md#customize-the-step-limit-of-the-credits-indicator).

In `uimd.ini`:
```ini
[Sidebar]
CreditsIndicator.Smooth=true  ; boolean
```

> [!Tip]
> For example, if you want to make the credits update immediately without a transition process by setting `CreditsIndicator.MaxStep=0`, then you should also set `CreditsIndicator.Smooth=false`; otherwise, the displayed credit value will still change in the form of an exponential approximation function.

### DIFF
--- a/docs/User-Interface.md
+++ b/docs/User-Interface.md
@@ -83,6 +83,10 @@ In `uimd.ini`:
 CreditsIndicator.MaxStep=143  ; integer
 ```
 
+```{note}
+The game first takes the absolute value of the difference between the actual credits and the current value, then divides it by 8, and then clamps this value to the limit set here. If you wish to disable this smoothing effect, set [`CreditsIndicator.Smooth=false`](User-Interface.md#disable-the-credits-indicator-smooth-transition).
+```
+
 ### Digital display
 
 ![image](_static/images/digital_display_shapes.png)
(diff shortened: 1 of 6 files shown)


docs/User-Interface.md
```diff
@@ -72,6 +72,17 @@ HealthBar.Permanent=false            ; boolean
 HealthBar.Permanent.PipScale=false   ; boolean
 ```
 
+### Customize the step limit of the credits indicator
+
+- In vanilla, the Credits Indicator in the sidebar increases by at most 143 points per frame, and now you can customize this limit.
+  - If set to a value less than or equal to 0, it will continue to execute the vanilla behavior (limited to 143).
```
TaranDahl: Why not parse an invalid value as unlimited?

TaranDahl (approved):

docs/User-Interface.md
DeathFishAtEase: Done

NetsuNegi (approved):

PR #2281, opened 2026-07-06
Fix HouseClass_ShouldDisableCameo_GreyCameo hook address.

AdjWang:
Hook address of `HouseClass_ShouldDisableCameo_GreyCameo` should be at `0x50B370` instead of `0x50B669`. Ares has the same hook, loading Phobos with Ares is fine, but without Ares would crash because of this.

### DIFF
--- a/src/Ext/House/Hooks.cpp
+++ b/src/Ext/House/Hooks.cpp
@@ -350,7 +350,7 @@ static inline bool CheckShouldDisableDefensesCameo(HouseClass* pHouse, TechnoTyp
 	return false;
 }
 
-DEFINE_HOOK(0x50B669, HouseClass_ShouldDisableCameo_GreyCameo, 0x5)
+DEFINE_HOOK(0x50B370, HouseClass_ShouldDisableCameo_GreyCameo, 0x5)
 {
 	GET(HouseClass*, pThis, ECX);
 	GET_STACK(TechnoTypeClass*, pType, 0x4);


TaranDahl (changes requested):
In the most used Syringe's rule, Ares.dll's hook will take effect before Phobos.dll's.
Ares hooks at the func head and returns to the func ending. So I suppose what you have done is disable the Phobos hook.

AdjWang:
So this hook point intends to chain after Ares.dll's? If so, when without ares, stack at `esp+4` is not actually original `TechnoTypeClass*`, the value has been corrupted by `mov byte ptr ss:[esp+0x1C], al` at `50B399`.

TaranDahl:
You may hook at the start and record the arg. If Ares is present, that arg could be extracted from the stack. Otherwise it can be extracted from your record.

AdjWang:
It makes sense, but phobos announced to be independent of Ares: "[Phobos is independent of Ares and does NOT require Ares specifically to function](https://phobos.readthedocs.io/en/latest/)". Does it mean loading `Phobos.dll` without `Ares.dll`? Or do I have to create my own fork to patch it?

TaranDahl:
I'm talking about the solution. IDK what you are confused about.

AdjWang:
I don't know if it's safe to load `Phobos.dll` without `Ares.dll`. Btw, since this pr is not what that hook designed for, I'm closing it.

Metadorius:
> I don't know if it's safe to load `Phobos.dll` without `Ares.dll`. Btw, since this pr is not what that hook designed for, I'm closing it.

@TaranDahl didn't day the bug is not real, just that you have to fix it differently, I think.

PR #2284, opened 2026-07-06
Add a defining for AutoDeath on whether it can trigger when in Limbo state or as a passenger

DeathFishAtEase:
> ## https://www.phoboscn.top/t/topic/609
> A unit designed with `AutoDeath.OnOwnerChange`, when serving as a passenger of a transport with `OpenTopped=true`, happens to need `Passengers.SyncOwner` to sync the passenger's owner change to prevent them from continuing to attack the transport's new owner, but they do not want this handling to trigger the passenger's self-destruction.

- `AutoDeath.AllowLimboed` can be used to define whether AutoDeath is triggered when the unit is in limbo state.
- `OnOwnerChange.IgnoreRevertOnExit`: Whether to ignore the case where `Passengers.SyncOwner.RevertOnExit=true` causes ownership change.

In `rulesmd.ini`:
```ini
[CombatDamage]
AutoDeath.AllowLimboed=true                       ; boolean
AutoDeath.OnOwnerChange.IgnoreRevertOnExit=false  ; boolean

[SOMETECHNO]                                      ; TechnoType
AutoDeath.AllowLimboed=                           ; boolean, defaults to [CombatDamage] -> AutoDeath.AllowLimboed
AutoDeath.OnOwnerChange.IgnoreRevertOnExit=       ; boolean, defaults to [CombatDamage] -> AutoDeath.OnOwnerChange.IgnoreRevertOnExit
```


TaranDahl (approved):

DeathFishAtEase:
After more exploration, it was found that the original handling's support for objects in Limbo state was not complete (I previously thought #830 had handled this part, but it was actually support for LimboDelivery) and this leads to the fact that even if the implementation is completed, the two flags cannot be set to true by default, otherwise it will break existing behavior; this PR is shelved for now.

DeathFishAtEase:
In any case, supporting Limbo scenarios for more AutoDeath conditions is something that should have been done but has not been completed. This PR effectively fulfills the corresponding feature request, and passengers inside `OpenTopped=true` transports are also not affected by the original defects. This PR only provides the toggle statements; how the toggles should behave when on or off should be addressed outside of this PR.

DeathFishAtEase:
Tested by: https://www.phoboscn.top/t/topic/647/4?u=noble_fish

image
```ini
[DRON]
AutoDeath.Behavior=kill
AutoDeath.AfterDelay=200
AutoDeath.AllowLimboed=no

[SQD]
AutoDeath.Behavior=kill
AutoDeath.AfterDelay=200
AutoDeath.AllowLimboed=no
```

image
```ini
[E1]
AutoDeath.Behavior=kill
AutoDeath.OnOwnerChange=yes
AutoDeath.AllowLimboed=no
AutoDeath.AllowPassenger=no

[BFRT]
InitialPayload.Types=E1
InitialPayload.Nums=5
Passengers.SyncOwner=yes
```

src/Ext/Techno/Body.Update.cpp
```diff
@@ -282,11 +282,24 @@ bool TechnoExt::ExtData::CheckDeathConditions(bool isInLimbo)
...
+
 	auto const pThis = this->OwnerObject();
 
+	if (pThis->Transporter && !pTypeExt->AutoDeath_AllowPassenger)
+		return false;
+
 	// Self-destruction must be enabled
 	const auto howToDie = pTypeExt->AutoDeath_Behavior.Get();
 
+	// You're already dead
```
TaranDahl: Should comment specific design rationale

TaranDahl:
Should check if AutoDeath can still work when technos are in non-opentopped vehicles before merging.
Also, the comment issue mentioned above should be addressed.

src/Ext/House/Hooks.cpp
```diff
@@ -216,7 +216,7 @@ DEFINE_HOOK(0x7015C9, TechnoClass_Captured_UpdateTracking, 0x6)
 	auto const pTypeExt = pExt->TypeExtData;
 	const bool isInLimbo = !pThis->IsInLogic && pThis->IsAlive;
 
-	if (pTypeExt->AutoDeath_Behavior.isset() && !(isInLimbo && !pTypeExt->AutoDeath_AllowLimboed.Get(RulesExt::Global()->AutoDeath_AllowLimboed)) && !(pThis->Transporter && !pTypeExt->AutoDeath_AllowPassenger.Get(RulesExt::Global()->AutoDeath_AllowPassenger)))
+	if (pTypeExt->AutoDeath_Behavior.isset())
```
Coronia, line 218: not sure if this should be removed as it'll still execute the line and set ShouldBeDead to true, which might make it die after leaving the transport

DeathFishAtEase: Perhaps we could add something like `AutoDeath.OnOwnerChange.IgnoreRevertOnExit`?

DeathFishAtEase: Done

src/Ext/Techno/Body.Update.cpp
DeathFishAtEase: Are the changes in [cbabdf3](https://github.com/Phobos-developers/Phobos/commit/cbabdf3b05ef5777ec7c38f01c22139ea045aa79#diff-824aae1ff40105bb3e13ff23a008005a39bc3d88aa5b70b55563ea2c649bc0c3R190) sufficient?

PR #2293, opened 2026-07-17
New exception handler with crash dialog, reports and minidumps

ZivDero:
Ports Vinifera's exception handler to Phobos, replacing the game's — and Ares' — crash handling with a hardened one that produces useful crash artifacts.

**What it does on a crash:**
- Writes a detailed `EXCEPT_*.TXT` report (exception info, registers, EBP + `StackWalk64` backtraces, raw stack scan, loaded module list, memory status, Phobos/Ares versions) and a `CRASHDUMP_*.DMP` minidump to the `debug` folder.
- Shows a themed dialog with the report and **Quit** / **Debug** / **Save Full Dump** buttons.
- Cleans up artifacts older than 5 days.

**Details:**
- Hardened: dedicated dumper thread, thread suspension for a coherent snapshot, stack-overflow survivability, and recursion guards.
- Optional `gamemd.pdb` / `Phobos.pdb` in the game dir resolve symbols in the report; an optional `gamemd.edb` (Vinifera-compatible format) annotates known crash addresses — a starter `gamemd.edb` with the documented Yuri's Revenge internal-error addresses is included.
- Activates Common Controls v6 so Phobos/game Win32 dialogs render with modern visual styles (gamemd ships no manifest).
- Gated by the existing `-ExceptionHandler` toggle (now on by default in all builds); new `-FullCrashDump` makes the automatic dump a full-memory dump.

image


src/Misc/ExceptionHandler.Dialog.cpp
```diff
@@ -0,0 +1,237 @@
...
+		if (DSurface::Tile != nullptr)
+			DSurface::Tile->Fill(0);
+
+		// Leave DirectDraw exclusive fullscreen and restore the desktop
+		// display mode - inside a 16-bit exclusive mode the dialog would
+		// render uncomposited (classic frames) at the game's resolution.
+		if (DirectDrawWrap::lpDD != nullptr)
+		{
+			DirectDrawWrap::lpDD->SetCooperativeLevel(Game::hWnd, eDDCoopLevel::DDSCL_NORMAL);
+			DirectDrawWrap::lpDD->RestoreDisplayMode();
```
Otamaa, line 53: this will crash on TS-DDRAW OpenGL .

TaranDahl (approved):
I see no issue, but I have few knowledge about the exception handler.
@IronHammer-Std maybe you can help?

src/Misc/ExceptionHandler.Dialog.cpp
ZivDero: Good point; Should be good now.

PR #2294, opened 2026-07-18
New Feature: Support Convert to another type.

frg2089:

## Description

This PR allows Phobos TechnoType conversion logic to convert between different TechnoType categories.

### Changes

- Generalized `TechnoExt::ConvertToType` to accept `TechnoClass` instead of only `FootClass`.
- Added support for conversions between infantry, vehicles, aircraft and buildings.
- Added support for conversions involving `BuildingType`, including building-to-building conversions.
- Same-category conversions between non-building TechnoTypes continue to use the existing in-place conversion logic.
- Cross-category conversions and conversions involving buildings create a new object, transfer applicable state, and remove the original object.
- Transferable state includes:
  - Health percentage
  - Veterancy
  - Mind control
  - Shield state
  - Invulnerability
  - Attached effects
  - Armor and firepower multipliers
  - Target
  - Destination
  - Current mission
  - Selection state
  - FootClass speed multiplier where applicable
- Added building placement and coordinate handling for converted buildings.
- Preserved building deployment and buildup initialization when the conversion result is a building.
- Reused the conversion logic for `Convert.Deploy`, `Convert.Undeploy`, `Ammo.AutoConvertType` and `Convert.Health`.
- Moved health-based automatic conversion into the common `TechnoClass` update path.
- Unified deployment placement checks for vehicles and infantry.
- `Convert.Deploy` conversions to buildings now validate building placement rules before execution.
- Cross-category conversions do not depend on Ares 3.0 and use Phobos' own conversion implementation.

### Notes

Because cross-category conversions create a new map object, some state or associations attached to the original object may not be preserved.

### DIFF
--- a/src/Ext/Techno/Body.cpp
+++ b/src/Ext/Techno/Body.cpp
@@ -372,8 +372,9 @@ static bool XConvertToType(TechnoClass* pFromTechno, TechnoTypeClass* pToType)
 	auto const pToInfantry = abstract_cast<InfantryClass*>(pToTechno);
 	if (pFromBuilding)
 	{
-		if (pFromBuilding->CurrentMission == Mission::Selling)
+		if (pFromBuilding->CurrentMission == Mission::None)
 			return false;
+		pFromBuilding->ForceMission(Mission::None);
 
 		auto const buildingCell = CellClass::Coord2Cell(location);
 		auto const tmpCell = BuildingTypeExt::GetBuildingTopLeftCellFromDeployCell(pFromBuilding->Type, buildingCell);
--- a/src/Ext/Techno/Body.cpp
+++ b/src/Ext/Techno/Body.cpp
@@ -406,7 +406,7 @@ static bool XConvertToType(TechnoClass* pFromTechno, TechnoTypeClass* pToType)
 
 	if (pFromFoot && pFromFoot->IsDeploying)
 	{
-		if (pToInfantry->Type->Sequence->GetSequence(Sequence::Deployed))
+		if (pToInfantry->Type->Sequence->GetSequence(Sequence::Deployed).CountFrames)
 		{
 			pToInfantry->PlayAnim(Sequence::Deployed, true);
 		}
--- a/src/Ext/Techno/Body.cpp
+++ b/src/Ext/Techno/Body.cpp
@@ -351,7 +351,7 @@ bool TechnoExt::AllowedTargetByZone(TechnoClass* pThis, TechnoClass* pTarget, Ta
 static bool XConvertToType(TechnoClass* pFromTechno, TechnoTypeClass* pToType)
 {
 	auto const pToObject = pToType->CreateObject(pFromTechno->Owner);
-	auto const pToTechno = abstract_cast<TechnoClass*>(pToObject);
+	auto const pToTechno = static_cast<TechnoClass*>(pToObject);
 	if (!pToTechno)
 	{
 		if (pToObject)
@@ -441,8 +441,9 @@ bool TechnoExt::ConvertToType(TechnoClass* pThisTechno, TechnoTypeClass* pToType
 		return false;
 	}
 
-	if (pType->WhatAmI() != pToType->WhatAmI()
-		|| (pType->WhatAmI() == pToType->WhatAmI() && pType->WhatAmI() == AbstractType::BuildingType))
+    const auto absType = pType->WhatAmI();
+    
+	if (absType != pToType->WhatAmI() || absType == AbstractType::BuildingType)
 		return XConvertToType(pThisTechno, pToType);
 
 	auto const pThis = abstract_cast<FootClass*>(pThisTechno);
@@ -589,8 +590,12 @@ void TechnoExt::TransferStatus(TechnoClass* pFrom, TechnoClass* pTo)
 	pTo->ArmorMultiplier = pFrom->ArmorMultiplier;
 	pTo->FirepowerMultiplier = pFrom->FirepowerMultiplier;
 	if (auto const pFromFoot = abstract_cast<FootClass*>(pFrom))
+	{
 		if (auto const pToFoot = abstract_cast<FootClass*>(pTo))
+		{
 			pToFoot->SpeedMultiplier = pFromFoot->SpeedMultiplier;
+		}
+	}
 }
 
 void TechnoExt::TransferMindControlOnDeploy(TechnoClass* pTechnoFrom, TechnoClass* pTechnoTo)
@@ -599,13 +604,13 @@ void TechnoExt::TransferMindControlOnDeploy(TechnoClass* pTechnoFrom, TechnoClas
 		? pTechnoFrom->MindControlRingAnim->Type
 		: TechnoExt::Fetch(pTechnoFrom)->MindControlRingAnimType;
 
-	if (const auto Controller = pTechnoFrom->MindControlledBy)
+	if (const auto pController = pTechnoFrom->MindControlledBy)
 	{
-		if (const auto Manager = Controller->CaptureManager)
+		if (const auto pManager = pController->CaptureManager)
 		{
-			CaptureManagerExt::FreeUnit(Manager, pTechnoFrom, true);
+			CaptureManagerExt::FreeUnit(pManager, pTechnoFrom, true);
 
-			if (CaptureManagerExt::CaptureUnit(Manager, pTechnoTo, false, pAnimType, true))
+			if (CaptureManagerExt::CaptureUnit(pManager, pTechnoTo, false, pAnimType, true))
 			{
 				if (const auto pBld = abstract_cast<BuildingClass*, true>(pTechnoTo))
 				{
@@ -629,9 +634,9 @@ void TechnoExt::TransferMindControlOnDeploy(TechnoClass* pTechnoFrom, TechnoClas
 			}
 		}
 	}
-	else if (const auto MCHouse = pTechnoFrom->MindControlledByHouse)
+	else if (const auto phMC = pTechnoFrom->MindControlledByHouse)
 	{
-		pTechnoTo->MindControlledByHouse = MCHouse;
+		pTechnoTo->MindControlledByHouse = phMC;
 		pTechnoFrom->MindControlledByHouse = nullptr;
 	}
 	else if (pTechnoFrom->MindControlledByAUnit) // Perma MC
--- a/src/Ext/Techno/Body.cpp
+++ b/src/Ext/Techno/Body.cpp
@@ -2,6 +2,7 @@
 #include <Ext/Anim/Body.h>
 #include <Ext/Building/Body.h>
 #include <Ext/BuildingType/Body.h>
+#include <Ext/CaptureManager/Body.h>
 #include <Ext/House/Body.h>
 #include <Ext/Infantry/Body.h>
 #include <Ext/Unit/Body.h>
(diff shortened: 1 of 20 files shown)


src/Ext/Techno/Body.cpp
```diff
@@ -467,6 +468,215 @@ bool TechnoExt::ConvertToType(FootClass* pThis, TechnoTypeClass* pToType)
...
+
+				return false;
+			}
+			
+			pThis->UnInit();
+
+			// Carry over the vehicle's move order
+			pUnit->SetDestination(savedDestination, true);
+			pUnit->QueueMission(savedMission, false);
+			pUnit->NextMission();
```
TaranDahl: Why no Target?

src/Ext/Techno/Body.cpp
```diff
@@ -467,6 +468,215 @@ bool TechnoExt::ConvertToType(FootClass* pThis, TechnoTypeClass* pToType)
...
+			mapCoords += CellStruct { -1, -1 };
+
+		pThis->Limbo();
+
+		if (!pBuildingType->CanCreateHere(mapCoords, pHouse))
+		{
+			Debug::Log("ConvertToTypeEx: Building %s cannot be placed at unit %s location\n", pBuildingType->get_ID(), pThis->get_ID());
+
+			if (pThis->IsOwnedByCurrentPlayer)
+				VoxClass::PlayIndex(VoxClass::FindIndex(GameStrings::EVA_CannotDeployHere));
```
TaranDahl: This should not be here. (Maybe also should not exist.)

src/Ext/Techno/Body.cpp
```diff
@@ -467,6 +468,215 @@ bool TechnoExt::ConvertToType(FootClass* pThis, TechnoTypeClass* pToType)
...
+				pThis->Select();
+
+			return false;
+		}
+
+		auto const pBuilding = static_cast<BuildingClass*>(pBuildingType->CreateObject(pHouse));
+		if (!pBuilding)
+		{
+			if (pThis->IsOwnedByCurrentPlayer)
+				VoxClass::PlayIndex(VoxClass::FindIndex(GameStrings::EVA_CannotDeployHere));
```
TaranDahl: Same

src/Ext/Techno/Body.cpp
```diff
@@ -349,8 +351,7 @@ bool TechnoExt::ConvertToType(FootClass* pThis, TechnoTypeClass* pToType)
 	// It really should be at the beginning.
 	if (pType == pToType || pType->WhatAmI() != pToType->WhatAmI())
 	{
-		Debug::Log("Incompatible types between %s and %s\n", pThis->get_ID(), pToType->get_ID());
-		return false;
+		return ConvertToTypeEx(pThis, pToType);
 	}
```
TaranDahl: I don't think the pType==pToType case should be processed.

src/Ext/Unit/Hooks.DeploysInto.cpp
```diff
@@ -101,11 +34,20 @@ DEFINE_HOOK(0x44A03C, BuildingClass_Mi_Selling_Transfer, 0x6)
...
 
-	TransferMindControlOnDeploy(pStructure, pUnit);
+	TechnoExt::TransferMindControlOnDeploy(pStructure, pUnit);
 	ShieldClass::SyncShieldToAnother(pStructure, pUnit);
 	TechnoExt::SyncInvulnerability(pStructure, pUnit);
 	AttachEffectClass::TransferAttachedEffects(pStructure, pUnit);
 
+	auto const pTypeExt = BuildingTypeExt::ExtMap.Find(pStructure->Type);
+
+	// If UndeploysInto.ConvertTo is set, defer unit conversion to next frame
```
TaranDahl: Should comment about the reason you must defer it.

src/Ext/Techno/Body.Update.cpp
```diff
@@ -41,6 +41,19 @@ void TechnoExt::ExtData::OnEarlyUpdate()
...
+	// Handle pending conversion
+	if (this->PendingConvertTechnoType)
+	{
+		auto const pTechnoType = this->PendingConvertTechnoType;
+		this->PendingConvertTechnoType = nullptr;
+
+		if (const auto pFoot = abstract_cast<FootClass*, true>(this->OwnerObject()))
+		{
+			if (TechnoExt::ConvertToType(pFoot, pTechnoType))
+				return; // Unit destroyed, skip rest of AI
```
TaranDahl: I think it should not be skipped if the type conversion did not destroy the unit.

docs/New-or-Enhanced-Logics.md
```diff
@@ -728,17 +728,37 @@ SpyEffect.VictimSuperWeapon=       ; SuperWeaponType
...
+```
+
 ## Infantry
 
 ### Allow infantry to perform type conversion when deploying and undeploying
 
-- Now infantry can perform type conversion immediately after the `Deploy` or `Undeploy` sequence action has been executed.
+- Infantry can now perform type conversion when executing the `Deploy` or `Undeploy` sequence action.
+  - `Convert.Deploy` and `Convert.Undeploy` accept any `TechnoType` as target.
+  - Same-type conversion (e.g. `InfantryType` → `InfantryType`, `UnitType` → `UnitType`, `AircraftType` → `AircraftType`) keeps the existing object and updates its type. This is the only way to convert to/from `AircraftType`.
```
TaranDahl: Is there any reason to exclude aircraft?

TaranDahl (changes requested):

src/Ext/Techno/Body.cpp
frg2089: I think it is necessary to prompt "Cannot Deploy Here." when it cannot be deployed as a building.
Or can I find a way to disable the deployment cursor?

TaranDahl: Even if you need to play EVA, it should be implemented in functions related to deployment rather than the Convert function. Convert is a general-purpose utility and should not contain this logic.

TaranDahl: If you want to support deploying infantry into buildings, handling the cursor is a thing that must be done.

frg2089: got it.

src/Ext/Unit/Hooks.DeploysInto.cpp
frg2089: After testing, I found that the vehicle created by undeploying a building is not fully initialized immediately (possibly still in a limbo state). Converting it to an infantry unit right away causes the new unit to lose states such as the current target and selection status.

TaranDahl: You should comment that in the code

src/Ext/Techno/Body.cpp
```diff
@@ -470,13 +476,211 @@ bool TechnoExt::ConvertToType(FootClass* pThis, TechnoTypeClass* pToType)
...
+				return false;
+			}
+
+			pThis->UnInit();
+
+			// Carry over the vehicle's move order
+			pUnit->SetDestination(savedDestination, true);
+			pUnit->QueueMission(savedMission, false);
+			pUnit->NextMission();
+			pUnit->SetTarget(savedTarget);
```
TaranDahl: Not quite sure about the inheritance status section. MissionStatus may also be considered. And whether buildings should skip the inheritance of some attributes.

TaranDahl: @NetsuNegi @Coronia @DeathFishAtEase

DeathFishAtEase: Huh,

- Q: Should a building that has `UndeploysInto` inherit the destination?

This kind of issue might need a toggle.

TaranDahl: Building has no destination.

DeathFishAtEase: > no destination

That is an example of behavior such as a moving vehicle immediately undeploying after conversion and continuing to the target location.

Alternatively: if a unit becomes a building under unload mission, should the building inherit the unload mission to attack the ground directly below... and so on.

TaranDahl (approved):
I see no other issues.

frg2089:
喵？

TaranDahl:
> 喵？

Since this is your first PR submission, your code will need reviews from more maintainers before it can be merged. This is the rule of Phobos.

frg2089:
> Since this is your first PR submission, your code will need reviews from more maintainers before it can be merged. This is the rule of Phobos.

OK, I will stop resolving the conflict before there are further results.

src/Ext/Techno/Body.cpp
```diff
@@ -347,19 +348,107 @@ bool TechnoExt::AllowedTargetByZone(TechnoClass* pThis, TechnoClass* pTarget, Ta
 	return true;
 }
 
+static bool XConvertToType(TechnoClass* pFromTechno, TechnoTypeClass* pToType)
+{
+	auto const pToObject = pToType->CreateObject(pFromTechno->Owner);
+	auto const pToTechno = abstract_cast<TechnoClass*>(pToObject);
```
NetsuNegi: I think it's ok to use `static_cast`

src/Ext/Techno/Body.cpp
```diff
@@ -347,19 +348,107 @@ bool TechnoExt::AllowedTargetByZone(TechnoClass* pThis, TechnoClass* pTarget, Ta
...
-	if (pType == pToType || pType->WhatAmI() != pToType->WhatAmI())
+	if (pType == pToType)
 	{
-		Debug::Log("Incompatible types between %s and %s\n", pThis->get_ID(), pToType->get_ID());
+		Debug::Log("Incompatible types between %s and %s\n", pThisTechno->get_ID(), pToType->get_ID());
 		return false;
 	}
 
+	if (pType->WhatAmI() != pToType->WhatAmI()
+		|| (pType->WhatAmI() == pToType->WhatAmI() && pType->WhatAmI() == AbstractType::BuildingType))
```
NetsuNegi:
```suggestion
    const auto absType = pType->WhatAmI();
    
	if (absType != pToType->WhatAmI()
		|| absType == AbstractType::BuildingType)
```

src/Ext/Techno/Body.cpp
```diff
@@ -470,6 +559,107 @@ bool TechnoExt::ConvertToType(FootClass* pThis, TechnoTypeClass* pToType)
...
+{
+	pTo->SetHealthPercentage(pFrom->GetHealthPercentage());
+	pTo->Veterancy = pFrom->Veterancy;
+	TransferMindControlOnDeploy(pFrom, pTo);
+	ShieldClass::SyncShieldToAnother(pFrom, pTo);
+	TechnoExt::SyncInvulnerability(pFrom, pTo);
+	AttachEffectClass::TransferAttachedEffects(pFrom, pTo);
+	pTo->ArmorMultiplier = pFrom->ArmorMultiplier;
+	pTo->FirepowerMultiplier = pFrom->FirepowerMultiplier;
+	if (auto const pFromFoot = abstract_cast<FootClass*>(pFrom))
```
NetsuNegi, line 615: brackets plz

src/Ext/Techno/Body.cpp
```diff
@@ -470,6 +559,107 @@ bool TechnoExt::ConvertToType(FootClass* pThis, TechnoTypeClass* pToType)
...
+			pToFoot->SpeedMultiplier = pFromFoot->SpeedMultiplier;
+}
+
+void TechnoExt::TransferMindControlOnDeploy(TechnoClass* pTechnoFrom, TechnoClass* pTechnoTo)
+{
+	const auto pAnimType = pTechnoFrom->MindControlRingAnim
+		? pTechnoFrom->MindControlRingAnim->Type
+		: TechnoExt::Fetch(pTechnoFrom)->MindControlRingAnimType;
+
+	if (const auto Controller = pTechnoFrom->MindControlledBy)
```
NetsuNegi: Suggest using the Hungarian naming convention uniformly for pointer local variables

src/Ext/Techno/Body.cpp
```diff
@@ -470,6 +559,107 @@ bool TechnoExt::ConvertToType(FootClass* pThis, TechnoTypeClass* pToType)
...
+	{
+		if (const auto Manager = Controller->CaptureManager)
```
NetsuNegi: same

src/Ext/Techno/Body.cpp
```diff
@@ -470,6 +559,107 @@ bool TechnoExt::ConvertToType(FootClass* pThis, TechnoTypeClass* pToType)
...
+				int nSound = pTechnoTo->GetTechnoType()->MindClearedSound;
+				if (nSound == -1)
+					nSound = RulesClass::Instance->MindClearedSound;
+
+				if (nSound != -1)
+					VocClass::PlayIndexAtPos(nSound, pTechnoTo->Location);
+			}
+		}
+	}
+	else if (const auto MCHouse = pTechnoFrom->MindControlledByHouse)
```
NetsuNegi: same

src/Ext/Techno/Body.cpp
```diff
@@ -470,6 +559,107 @@ bool TechnoExt::ConvertToType(FootClass* pThis, TechnoTypeClass* pToType)
...
+		pTechnoTo->MindControlledByAUnit = true;
+
+		const auto pBuilding = abstract_cast<BuildingClass*, true>(pTechnoTo);
+		CoordStruct location = pTechnoTo->GetCoords();
+
+		location.Z += pBuilding
+			? pBuilding->Type->Height * Unsorted::LevelHeight
+			: pTechnoTo->GetTechnoType()->MindControlRingOffset;
+
+		const auto pAnim = pAnimType
```
NetsuNegi, line 676: just check `pAnimType`, then you dont need to check `pAnim` because game will crash when `GameCreate` failed

frg2089: This code was not written by me. I am not familiar with its functions. To be safe, I keep the logic of this method as it is and only rename it.

NetsuNegi (approved):

PR #2295, opened 2026-07-18
Move Unit/Infantry/Aircraft[Type] fields & methods from Techno[Type]Ext

ZivDero:
Moves various Unit/Infantry/Aircraft[Type] fields & methods from Techno[Type]Ext to theire respective extensions


TaranDahl (approved):
Only reviewed part of the changes as there were too many of them. Given that this modification is relatively simple in terms of design, I believe it is free of issues.

PR #2297, opened 2026-07-18
Add type selection by double/triple-click

leosnake2208:
## Feature

Adds the modern-RTS multi-click selection gesture, complementing the existing type-select hotkey (hold `T` + click):

- **Double-click** a unit → select every unit of the same selection group currently **on screen**.
- **Triple-click** a unit → select every unit of that group across the **whole map**.

The "selection group" reuses the existing Phobos concept (`TechnoTypeExt::GetSelectionGroupID` / `HasSelectionGroupID`, i.e. the `[TechnoType] -> GroupAs` tag with the type ID as fallback), so it stays consistent with the hotkey-driven type select in `Selection.cpp`. Only the player's own selectable mobile units are affected.

## Config

Gated behind a new user setting, **default off**:

```ini
[Phobos]
TypeSelectByMultiClick=false  ; boolean
```

## Implementation

- New `src/Misc/MultiClickTypeSelect.cpp`. A single hook in the tactical `LBUTTONUP` handler — after the click's own selection has been applied, on the single-click dispatch path only (completed rubber-band selections exit earlier) — tracks the click streak via `GetTickCount` / `GetCursorPos` / `GetDoubleClickTime` (±4 px tolerance) and expands the freshly-clicked unit's selection to its group: on screen for a double-click, map-wide for a triple-click.
- The map-wide case is the only genuinely new capability; the vanilla/hotkey type select is screen-only.

## Docs & credits

- `docs/User-Interface.md` (new "Type selection by multi-click" section), `docs/Whats-New.md` (new setting + changelog entry), `CREDITS.md`.

Built clean (Debug/Release, 0 warnings) and tested in-game (skirmish) on the CnCNet Yuri's Revenge build.

### DIFF
--- a/src/Misc/MultiClickTypeSelect.cpp
+++ b/src/Misc/MultiClickTypeSelect.cpp
@@ -1,0 +1,111 @@
+#include <Phobos.h>
+
+#include <FootClass.h>
+#include <TacticalClass.h>
+#include <Ext/Techno/Body.h>
+#include <Ext/TechnoType/Body.h>
+
+// Multi-click type selection.
+//
+// Vanilla Yuri's Revenge only offers "type select" through the type-select hotkey (hold T,
+// then click/drag) - there is no double-click equivalent; a double-click in the tactical
+// area falls through to the message handler's default case. This adds the modern-RTS gesture:
+//   * double-click a unit  -> select every unit of the same selection group on screen;
+//   * triple-click a unit  -> select every unit of the same selection group across the map.
+//
+// "Selection group" reuses Phobos' existing concept (TechnoTypeExt::GetSelectionGroupID /
+// HasSelectionGroupID, i.e. the [TechnoType]GroupAs tag with the type ID as fallback), so it
+// stays consistent with the hotkey-driven type select in Selection.cpp.
+//
+// Gated behind [Phobos]TypeSelectByMultiClick (default off).
+
+namespace MultiClickTypeSelect
+{
+	// Clicks within GetDoubleClickTime() and this many screen pixels of the previous one
+	// count as part of the same streak.
+	static constexpr int PositionTolerance = 4;
+
+	static DWORD LastClickTick = 0;
+	static POINT LastClickPos = { -9999, -9999 };
+	static int ClickStreak = 0;
+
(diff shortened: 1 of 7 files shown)


src/Misc/MultiClickTypeSelect.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+	// Add to the current selection every own, selectable mobile unit sharing the just-clicked
+	// unit's selection group. onScreenOnly limits it to units drawn in the tactical viewport.
+	static void SelectSameGroup(bool onScreenOnly)
+	{
+		if (ObjectClass::CurrentObjects.Count < 1)
+			return;
+
+		// The single click that preceded this streak left the clicked unit as the selection.
+		const auto pClicked = abstract_cast<FootClass*, true>(ObjectClass::CurrentObjects.GetItem(0));
```
TaranDahl, line 50: Have you tested this? I doubt if this is working as you said in your docs when you are selecting multiple units.

> Double-clicking a unit selects every unit of the same selection group currently on screen

TaranDahl (commented):
I suspect the entire design is flawed. In other games such as SC2, you can double-click the left mouse button to select all units, as the left click only serves to select units in that game. However, in RA2, the left mouse button has multiple functions, and double-clicking a unit will immediately command it to perform corresponding actions.

TAK02:
If I may make a suggestion, as I rather recently got my hands on E:BfD's 1.009 patch:

After selecting a deployable techno (MCV), you have 1-2s where deploy-order seems blocked (at least via mouse, haven't tried D-key); after that duration passes cursor switches to deploy; during that block-time can do a type-select via dbl-click in E:BfD.

Also, on this topic: can we make a customisable range for typeSelect via dbl-click as opposed to screen-wide selection?
World in Conflict had this: dbl-tapping a unit for type-select worked for "immediate" vicinity only, not full screen. Couldn't tell you rn what that range is, tho... maybe make it 10 cells around clicked unit be default for this feature as starting value?

src/Misc/MultiClickTypeSelect.cpp
leosnake2208: You are right, I did not think the design through, and I split the original modern control idea into two separate ones, which was stupid. So I am closing #2297 and the type selection by multi-click moves to #2298 together with the right click to command.

PR #2298, opened 2026-07-18
Add right-click-to-command control scheme and type selection by multi-click

leosnake2208:
Replaces #2297. Type selection by multi-click is part of this PR now: a double click can only select units once the left button stops giving orders, otherwise it deploys an MCV or an Allied GI instead.

## Feature

An **opt-in** control scheme matching modern RTS games:

- **Right mouse button** issues orders to the current selection — move, attack, enter, capture, harvest, guard, and so on.
- **Left mouse button** only selects, box-selects and self-deploys, and **deselects** when clicking empty ground.
- **Double click** on a unit selects the same units nearby, **triple click** selects them on the whole map. Optional, and only works with the scheme on.
- **The minimap** follows the same scheme: the right button gives orders, the left button moves the view. In vanilla it is the other way around.

Special left-click modes are preserved on both buttons: while placing a building, repairing, selling, toggling power, planting a beacon, planning a path, or targeting a superweapon, the left button performs that action and the right button cancels it — exactly as in vanilla. Only mouse behaviour changes; no keyboard hotkeys are rebound.

## Config

Gated behind a new user setting, **default off**:

```ini
[Phobos]
RightClickCommand=false                 ; boolean
TypeSelectByMultiClick=false            ; boolean, needs RightClickCommand
TypeSelectByMultiClick.Range=-1         ; integer, cells, negative means the whole screen
TypeSelectByMultiClick.DeployDelay=500  ; integer, milliseconds
```

`TypeSelectByMultiClick` is turned off with a log line if `RightClickCommand` is not on. `Range` is how far the double click reaches, `DeployDelay` is how long the left button will not deploy after a click selected something, so the second click of a double click selects instead of unpacking the unit. Both were asked for by @TAK02 on #2297, the delay works like the one in Emperor: Battle for Dune 1.009.

## Implementation

The tactical mouse message handler (`MouseClass` @ `0x6930A0`) dispatches window mouse messages through a jump table. `RBUTTONUP` and `LBUTTONUP` are two cases of the same function, sharing one stack frame and `this`.

- **RMB → command** (`0x693397`): the RBUTTONUP handler, past its drag-flag gate, populates the view-relative click point and jumps into the shared LMB-up command dispatch at `0x69323E`, so the order flows through the game's own `DecideAction` → apply path — reusing 100% of the command and **network** logic (no argument/packet reconstruction). A flag tells the LBUTTONUP hook to let this RMB-issued order pass unmodified.
- **LMB → select/deploy only** (`0x6931B4`, `0x693276`): right after `DecideAction`, any command action is downgraded to `Action::None`, so the left button only selects/box-selects/deploys. On empty ground a would-be move becomes a deselect (`MapClass::UnselectAll`). Completed band-selects exit earlier and never reach this point.
- **Multi-click type select** (`0x693290`): after the click's own selection was applied, on the single-click path only. The clicked unit is read from `ProcessClickCoords`' output slot, not from `CurrentObjects[0]` — that was the bug @TaranDahl found on #2297, it picks the wrong unit whenever several are selected. Grouping reuses `TechnoTypeExt::GetSelectionGroupID`, so it matches the type select in `Selection.cpp`.
- **Minimap** (`0x6539D3`): `RadarClass::GetMouseAction` reads the buttons out of its flags argument, so swapping the press/held/release bits at the top of it swaps the two buttons for the radar. Vanilla already gives orders from the minimap with the left button, through the same applier the tactical view uses; the hover bits are left alone so the radar cursor still updates.
- A special-mode guard (`InSpecialLeftClickMode`) keeps repair/sell/placement/beacon/planning/superweapon-targeting on their vanilla buttons.

Validated with the repo's `check-hooks` skill — no conflicts with Phobos or Ares 3.0p1; hook sizes and return targets check out. Built clean (Debug/Release, 0 warnings). The scheme was developed and extensively tested in-game on the CnCNet Yuri's Revenge build.

## Notes

- Net-safe: client-side input mapping only; orders reuse the game's own networked dispatch.
- Could later be split into finer toggles (RMB-to-command vs LMB-select-only) if that is preferred.


Coronia:
rename the tag to RightClickCommand since ModernControl is an ambigious name that can refer to many things

leosnake2208:
Done — renamed the tag to `RightClickCommand` (and updated the config field, source file, namespace, hook names, docs and CREDITS to match). PR title/body updated too. Thanks!

Metadorius:
hint: we'd appreciate (even broken) real human English more than an LLM text ;)

leosnake2208:
> hint: we'd appreciate (even broken) real human English more than an LLM text ;)

mea culpa. i wasn't thinking much and didn't have any idea about the type of dinamics here. it's good to talk to real people.

leosnake2208:
Thanks for the comments to all. The bugs you found and the notes about where it does not work are taken into account. So the idea itself is reworked into a full remap of the mouse buttons, like in modern strategy games (right button - orders, left button - select and deselect + multi-click). There is also a fix for clicks on the radar, so they work the same way.

TAK02:
"TypeSelectByMultiClick=false            ; boolean, needs RightClickCommand"
E:BfD 1.009 was LeftClickCommand-game but had TypeSelectByMultiClick via LMB; TypeSelectByMultiClick should be independent of RightClickCommand.

MantisMO:
This PR had absolutely no expected effect.

Coronia:
I think it would be better to just disable left click deploy if `RightClickCommand` is enabled, especially now that it's used for double click selection

Coronia:
> I think it would be better to just disable left click deploy if `RightClickCommand` is enabled, especially now that it's used for double click selection

done, and move range tag to [GlobalControls] to follow the principal that balance related stuff should be in rules

btw, there's another unresolved issue: if using right mouse drag to move the view during RightClickCommand, both drag and the action will be executed. Not sure how to solve it yet, but the ideal way is to make it not executing right click command if holding right and moving

PR #2300, opened 2026-07-19
remove per frame type update check

Coronia:
changing TechnoType is a very rare case compared with the grand per frame per techno check. Ideally this should only be done when the conversion of type is happening. Attempted to remove such check and only attributed them to conversion (already have that) and (Un)DeploysInto. Ingame test suggested a very noticeable perf improvement but not sure if there're still errors or if I miss some other cases of type change. Hence I make a PR for more test and review


NetsuNegi (approved):

Coronia:
It's now confirmed that (Un)DeploysInto won't go throught this update process, but'll reset TypeExtData through Init process. Conversion is the only scenario that'll need to update TypeExtData according to several code review and test.

I'm assuming the per frame check was made before mostly because we wanted to make it compatible with Ares' conversion, which is now being taken over by Phobos already. Hence, it *should* be save to remove it for now, unless we want to support conversion from another dll

And given the above conclusion, UpdateTypeData_Foot is no longer needed to be a standalone function. Now that we've made it only be called during conversion, it can just be merged into UpdateTypeData

PR #2301, opened 2026-07-21
Implement country-defined crew

Sovietianqi:
Now you can specify the crew for each country by adding Crew=infantry type under the corresponding [country] section.
For example:

```
[CountryName]
Crew=Infantry
```

This allows you to define which infantry type serves as the crew for that country.


src/Ext/HouseType/Hooks.cpp
```diff
@@ -35,3 +35,23 @@ DEFINE_HOOK(0x68AD0C, ScenarioClass_ReadMap_SetEVAIndex, 0x7)
...
+	GET(TechnoClass*, pThis, ECX);
+
+	if (!pThis)
+		return 0;
+
+	HouseClass* pHouse = pThis->Owner;
+	if (!pHouse)
+		return 0;
+
+	auto pHouseTypeExt = HouseTypeExt::ExtMap.Find(pHouse->Type);
```
TaranDahl: Find func is now deprecated.

src/Ext/HouseType/Hooks.cpp
```diff
@@ -35,3 +35,23 @@ DEFINE_HOOK(0x68AD0C, ScenarioClass_ReadMap_SetEVAIndex, 0x7)
...
+	if (pHouseTypeExt && pHouseTypeExt->Crew.isset())
+	{
+		R->EAX(pHouseTypeExt->Crew.Get());
+	}
```
TaranDahl: Incorrect Nullable template usage.

src/Ext/HouseType/Hooks.cpp
```diff
@@ -35,3 +35,23 @@ DEFINE_HOOK(0x68AD0C, ScenarioClass_ReadMap_SetEVAIndex, 0x7)
...
+DEFINE_HOOK(0x707DCF, TechnoClass_GetCrew_NationalOverride, 0x5)
+{
+	GET(TechnoClass*, pThis, ECX);
+
+	if (!pThis)
+		return 0;
+
+	HouseClass* pHouse = pThis->Owner;
+	if (!pHouse)
+		return 0;
```
TaranDahl: Redundant checks.

TaranDahl (changes requested):
Docs needed. You can ask your agent to generate it.

DeathFishAtEase:
Actually, you only need to set it like this on the basis of b175ec7bb785bf8312cdeae0c3902f0ed9ae17d2 to avoid the issue where the side settings get overwritten.
```diff
.\src\Ext\HouseType\Body.h#L34
- , Crew { nullptr }
+ , Crew {}
```
Of course, a better method might be to use `Valueable` and check whether it is `nullptr`.

DeathFishAtEase:
Reopened in the form of new PR - #2342.

Has been merged as commit 62324c5378cd9cc74bbe4da9ab907f2493428816.

PR #2303, opened 2026-07-21
`GameHandle` infrastructure for replacing `PointerExpired` broadcasts and improving performance

TaranDahl:

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -834,6 +834,7 @@ This page lists all the individual contributions to the project by their author.
   - Allow beacon placement hotkey in single player
   - Adjust recruitable status on team member discharge
   - RA1-Style multi-turret and multi-barrel
+  - `GameHandle` infrastructure for replacing `PointerExpired` broadcasts and improving performance
 - **solar-III (凤九歌)**
   - Target scanning delay customization (documentation)
   - Skip target scanning function calling for unarmed technos (documentation)
--- a/Phobos.vcxproj
+++ b/Phobos.vcxproj
@@ -309,6 +309,7 @@
     <ClCompile Include="src\Utilities\ShapeTextPrinter.cpp" />
     <ClCompile Include="src\Utilities\SpawnerHelper.cpp" />
     <ClCompile Include="src\Utilities\Stream.cpp" />
+    <ClCompile Include="src\Utilities\GameHandle.cpp" />
   </ItemGroup>
   <!-- Header files -->
   <ItemGroup>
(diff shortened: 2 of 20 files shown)


TaranDahl:
It is necessary to test whether SL can function normally. This should be carried out after the SL issues on the develop branch are resolved.

ZivDero:
I still don't like that this kind of solution has no callback to its owner class.

TaranDahl:
> I still don't like that this kind of solution has no callback to its owner class.

Wdym? Additional handling required when the pointer becomes invalid?

ZivDero:
> > I still don't like that this kind of solution has no callback to its owner class.
> 
> Wdym? Additional handling required when the pointer becomes invalid?

Yeah. E.g. 
```cpp
void ObjectClass::Detach(AbstractClass const * target, bool all)
{
	if (Tag == target) {
		if (Tag) {
			Tag->AttachCount--;
			Tag = NULL;
		}
	}

	if (all) {
		if (target == Next) {
			if (target) {
				Next = Next->Next;
			}
		}
	}
}
```

Such logic is currently impossible to reproduce with this solution.

Metadorius:
I haven't read into the code, but keep in mind that callbacks either might not be serializable (unverified) or need special handling

TaranDahl:
> I haven't read into the code, but keep in mind that callbacks either might not be serializable (unverified) or need special handling

Should be no such things to be swizzled, I think.

TaranDahl:
> > > I still don't like that this kind of solution has no callback to its owner class.
> > 
> > 
> > Wdym? Additional handling required when the pointer becomes invalid?
> 
> Yeah. E.g.
> 
> ```c++
> void ObjectClass::Detach(AbstractClass const * target, bool all)
> {
> 	if (Tag == target) {
> 		if (Tag) {
> 			Tag->AttachCount--;
> 			Tag = NULL;
> 		}
> 	}
> 
> 	if (all) {
> 		if (target == Next) {
> 			if (target) {
> 				Next = Next->Next;
> 			}
> 		}
> 	}
> }
> ```
> 
> Such logic is currently impossible to reproduce with this solution.

I know it. This will not replace every PointerExpired. Broadcast is still needed for such fields.

ZivDero:
> > > > I still don't like that this kind of solution has no callback to its owner class.
> > > 
> > > 
> > > Wdym? Additional handling required when the pointer becomes invalid?
> > 
> > 
> > Yeah. E.g.
> > ```c++
> > void ObjectClass::Detach(AbstractClass const * target, bool all)
> > {
> > 	if (Tag == target) {
> > 		if (Tag) {
> > 			Tag->AttachCount--;
> > 			Tag = NULL;
> > 		}
> > 	}
> > 
> > 	if (all) {
> > 		if (target == Next) {
> > 			if (target) {
> > 				Next = Next->Next;
> > 			}
> > 		}
> > 	}
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Such logic is currently impossible to reproduce with this solution.
> 
> I know it. This will not replace every PointerExpired. Broadcast is still needed for such fields.

Then IMO it would be premature optimization to do this now. Phobos barely uses PointerExpired and I've already replaced the old broadcast-to-all solution with one that only broadcasts to relevant types.

TaranDahl:
@ZivDero But if we reference AbstractClass on a large number of Ext instances such as CellExt, the overhead of PointerExpired will become non‑negligible.

PR #2307, opened 2026-07-23
Show game time

Coronia:
ported from https://github.com/CnCNet/yrpp-spawner/pull/64 since this is more of a standalone function than a part of disabling save, and is less relevant to Spawner

- A timer can be displayed to show how many time has passed since game starts.
  - The timer will be shown in the format of `TXT_GAMETIME hh:mm:ss`. For localization add `TXT_GAMETIME` into your `.csf` file.
  - This will show the ingame time by default. It can also be set to show real time by [real time timers](#real-time-timers)
  - `ShowGameTime.BoardOpacity` can be used to set the opacitiy of background for game time display.
  - Observer can't see this timer since they've already gotten one on the top of sidebar.

In `RA2MD.INI`:
```ini
[Phobos]
ShowGameTime=false             ; boolean
ShowGameTime.BoardOpacity=40   ; integer
```

by Trsdy


chaserli:
If we are doing this on Phobos, it should be implemented differently. The hook was copied from https://github.com/Phobos-developers/Phobos/blob/develop/src/Phobos.cpp#L305

Personally I think we can fake this as a SW timer and show it along with all other SW timers/mission timer/ (in TacticalClass::Draw iirc)

TaranDahl (approved):

Coronia:
> If we are doing this on Phobos, it should be implemented differently. The hook was copied from https://github.com/Phobos-developers/Phobos/blob/develop/src/Phobos.cpp#L305
> 
> Personally I think we can fake this as a SW timer and show it along with all other SW timers/mission timer/ (in TacticalClass::Draw iirc)

It'll indeed overwrite Phobos version text for Debug and DevBuild if enabled since the hook will be skipped. But I'm not sure making it a SW timer is a good idea either since it'll be too clustered

I personally think it's fair to overwrite it since they also take the same place in screen. If we keep them both them there'll be an overlap. But maybe this should be decided by @Metadorius

Metadorius:
can't we write some of this in the bottom left?

Coronia:
> can't we write some of this in the bottom left?

bottom left is also occupied by release note and FPS counter. Besides, if we want to make an UI for details of selected objects, that's also the reserved place
image

Metadorius:
then how about using top right, but making stacking of "messages"

Coronia:
> then how about using top right, but making stacking of "messages"

done
image

PR #2312, opened 2026-07-26
Customize whether aircraft is a cargo plane

TaranDahl:
### Customize whether aircraft is a cargo plane

- In vanilla, the `IsALoaner` flag of an aircraft is determined by the game's own logic during `UnLimbo`. Now, you can explicitly control this flag per AircraftType.
  - The key `IsALoaner` determines the value of the aircraft's `IsALoaner` flag when it is unlimboed. If left unset, the vanilla behavior is preserved.

In `rulesmd.ini`:
```ini
[SOMEAIRCRAFT]      ; AircraftType
IsALoaner=       ; boolean
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -851,6 +851,7 @@ This page lists all the individual contributions to the project by their author.
   - Adjust recruitable status on team member discharge
   - RA1-Style multi-turret and multi-barrel
   - New hotkey to select the units within the current screen that are captured by non-permanent mind-controller
+  - Customize whether aircraft is a cargo plane
 - **solar-III (凤九歌)**
   - Target scanning delay customization (documentation)
   - Skip target scanning function calling for unarmed technos (documentation)
--- a/docs/Fixed-or-Improved-Logics.md
+++ b/docs/Fixed-or-Improved-Logics.md
@@ -844,6 +844,17 @@ AircraftFiringForceScatter=true     ; boolean
 FiringForceScatter=                 ; boolean, default to [General] -> AircraftFiringForceScatter
 ```
 
+### Customize whether aircraft is a cargo plane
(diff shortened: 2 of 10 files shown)


Coronia:
so the `Spawned` tag from vanilla is not working for this case? Maybe it's better to resume it if it's not doing its job intendedly

TaranDahl:
The ini key `Spawned` has a group of hardcoded functions, and none of them is related to `TechnoClass::Spawned`.
`TechnoClass::Spawned` is not a reasonable name. It is named IsALoaner in RA1. In fact, it has nothing to do with the ini key `Spawned`.

Coronia:
since this is not just about literal cargo plane, maybe the tag name should be changed to avoid confusion. Either reuse IsAloaner or something like IsSpawned, or such

Also, given Spawned tag are already widely used for spawnees, which is what this tag aiming at, maybe this should be default to it

TaranDahl:
> Also, given Spawned tag are already widely used for spawnees, which is what this tag aiming at, maybe this should be default to it

I think you are still misunderstanding something. The sequence of behaviors caused by TechnoClass::Spawned=true is exactly the root cause of the bug mentioned at https://www.phoboscn.top/t/topic/679/10. In other words, any unit with TechnoTypeClass::Spawned=true should normally have TechnoClass::Spawned=false.

Coronia:
in this case it would still deserve a smart default regarding to Spawned tag, something like this:
```
if (pTypeExt->IsCargoPlane.isset() || pType->Spawned)
{
    pThis->Spawned = pTypeExt->IsCargoPlane.Get(false);
    return SkipGameCode;
}
```

Coronia (approved):

PR #2314, opened 2026-07-27
Fix the issue of Ares' EMP not suspending the production of AI factories

CrimRecya:
As the title says.


NetsuNegi (approved):

Coronia:
test suggests that EMP a War Factory when the gate is open will result in units stuck in it, but that's irrelevant to this PR since it'll also happen before. Otherwise fine

PR #2316, opened 2026-07-29
Remove the restriction that prohibits InfantryTypes from using the InitialPayload logic

DeathFishAtEase:
In the past, in order to use `OpenTopped` with Infantry as a transport, it was necessary to graft/temporarily substitute with an `Abductor` weapon whose invoker is that Infantry; now you can directly use the InitialPayload logic to create passengers, just like other FootClass objects.

image


TaranDahl (commented):
Maybe this should be commit to Antares

DeathFishAtEase:
> Maybe this should be commit to Antares

Not sure. And many other patch Ares contents are still in Phobos currently. If we are to move to Antares, it might be necessary to perform a migration later to port over all Ares-interaction-related contents? After all, Phobos and Antares are currently not guaranteed to be compatible.

> Microsoft Visual C++:
> ideally antares should not be developed at all
> bug fixes & interop only

---

Anyway, I will submit it to Antares, but I think adding them to the Phobos repository is not unnecessary at this time.

NetsuNegi (approved):

PR #2318, opened 2026-07-31
Improve #2218

DeathFishAtEase:
> ## [Starkku commented Jul 30, 2026, 18:56 GMT+8](https://github.com/Phobos-developers/Phobos/pull/2218#issuecomment-5129925246)
> This is currently performing expensive hash table lookups (std::unordered_map.find()) on every frame for every LaserDrawClass instance with zero mitigations in place f.ex disabling hooks after all parsing passes if tracking feature is not enabled, or even a simple skip if no tracking laser weapons exist. Ideally LaserDrawExt would be created for this purpose. Unlike EBolt, LaserDrawClass technically has enough padding to house extension pointer but it is not contiguous which would either require adding additional logic to the extension template to allow parsing pointer from non-contiguous offsets or a bespoke implementation.

> [!Note]
> The `LaserRT::TrackingMap[pLaser] = LaserRT::TrackingData {};` in the original implementation's CTOR hook cannot be directly removed for optimization, otherwise the laser ring of `DiskLaser` during the charging phase may have its start point on ring A and end point on ring B.


src/Ext/WeaponType/Body.cpp
```diff
@@ -506,6 +509,10 @@ DEFINE_HOOK(0x77311D, WeaponTypeClass_SDDTOR, 0x6)
 {
 	GET(WeaponTypeClass*, pItem, ESI);
 
+	auto pExt = WeaponTypeExt::ExtMap.Find(pItem);
+	if (pExt && pExt->LaserPositionUpdate != PositionFollow::None)
```
Coronia: use TryFind if it can be nullptr, or remove the sanity check if it can't

Starkku: It would crash before entering the hook if it was nullptr, `TryFetch` and the sanity check are both unnecessary here.

Coronia (approved):
if we're not hurry for a LaserDrawExt I'd suggest merge this first to solve the immediate problem

PR #2319, opened 2026-08-01
Disable AlphaImage during Buildup

DeathFishAtEase:
- Now you can disable AlphaImage before the building is completed.

In `rulesmd.ini`:
```ini
[AudioVisual]
NoAlphaImageOnBuildup=false  ; boolean

[SOMEBUILDING]               ; BuildingType
NoAlphaImageOnBuildup=       ; boolean, defaults to [AudioVisual] -> NoAlphaImageOnBuildup
```


Coronia (approved):

src/Misc/Hooks.AlphaImage.cpp
```diff
@@ -72,8 +72,16 @@ static void __fastcall UpdateAlphaShape(ObjectClass* pSource)
...
 	const auto pBuilding = abstract_cast<BuildingClass*, true>(pSource);
 
-	if (pBuilding && !inactive && pBuilding->GetCurrentMission() != Mission::Construction)
-		inactive |= !pBuilding->IsPowerOnline() || BuildingExt::Fetch(pBuilding)->LimboID != -1;
+	if (pBuilding && !inactive)
+	{
+		if (pBuilding->GetCurrentMission() == Mission::Construction)
+		{
+			if (BuildingTypeExt::Fetch(pBuilding->Type)->NoAlphaImageOnBuildup.Get(RulesExt::Global()->NoAlphaImageOnBuildup))
+				inactive = true;
```
Coronia: should we make it fall back to the old calculation if NoAlphaImageOnBuildup is set to false?

DeathFishAtEase: The current logic is: when `NoAlphaImageOnBuildup` is `false`, the new branch condition will not trigger, and the inactive state remains unchanged, so the display behavior of AlphaImage is completely consistent with the original logic. Therefore, no additional fallback code is needed here; false itself means fallback.

DeathFishAtEase:
Tested by: https://www.phoboscn.top/t/topic/688/2?u=noble_fish

image

PR #2320, opened 2026-08-02
Automatic conversion based on HP

obsidian14:
### Automatic conversion based on HP

- Units can now be converted into another unit by health percentage.
- `Convert.Health.AbovePercent`determines the minimal health percentage at which a unit converts automatically.
- `Convert.Health.BelowPercent`determines the maximum health percentage at which a unit converts automatically.
- `Convert.Health` specify the new techno after the conversion. This unit must be of the same type of the original (infantry -> infantry, vehicle -> vehicle or aircraft -> aircraft).
- Setting a negative number will disable the HP check, and when both checks are disabled, conversion will not occur.

In `rulesmd.ini`:
```ini
[SOMETECHNO]                         ; TechnoType, before conversion
Convert.Health.AbovePercent=-1.0     ; floating point value, percents or absolute
Convert.Health.BelowPercent=-1.0     ; floating point value, percents or absolute
Convert.Health=                      ; TechnoType, after conversion
```


src/Ext/Techno/Body.h
```diff
@@ -179,6 +179,7 @@ class TechnoExt : public RadioExt, public Detach::Listener<AirstrikeClass>
 	void ResetDelayedFireTimer();
 	void UpdateTintValues();
 
+	void AutoConvertActions();
```
Coronia: should be put in FootExt

Coronia (changes requested):
This implementation is very problematic in performance. If you have a conversion type set it'll be executed every frame once the health condition meets. Should at least make it only happen when there's an actual change of health conditions, and make it bail out as early as possible by checking its current type and such

obsidian14:
I did my best to meet your requests. This is my solution. Could you please take a look?

Coronia (approved):
did some regulation to tag names, code style and such. Notice that we already have a universal IsHealthInThreshold function to handle health condition, so no need to reimplement it again

TaranDahl (approved):

TaranDahl:
It would be best to write the Chinese documentation as well. We have a skill for this.

PR #2328, opened 2026-08-05
Add per-weapon VXL turret switching for MultiWeapon units

NukeAtty:
Adds support for swapping VXL turret and barrel models per weapon index for units using `MultiWeapon=yes` together with `TurretCount>0`.

## Summary

- Units with `MultiWeapon=yes` + `TurretCount>0` now swap their VXL turret and barrel based on the currently selected weapon.
- Extra models are loaded by naming convention (`<ImageFile>turN.vxl` / `<ImageFile>barlN.vxl`); no `artmd.ini` changes required.
- The number of effective turrets is the smaller of `WeaponCount` and `TurretCount`; weapons beyond that range fall back to the default turret.
- Elite weapons follow the same per-index rule.
- Turret and barrel shadows are drawn as usual when `UseTurretShadow=yes`.

## Implementation

- `UnitTypeExt` loads the extra VXL/HVA pairs (`LoadMultiWeaponTurrets`) and frees them on destruction.
- `UnitClass_AI` sets `CurrentTurretNumber` from `SelectWeapon`.
- Draw (0x73BA12) and shadow (0x73C47A) hooks route the turret/barrel voxel per weapon index, bypassing the voxel cache for custom turrets.
- Defensive skip in the charge turret hook (0x6FA540).

## Notes

- Does not apply to units with `IsChargeTurret=yes`, `IsGattling=yes` or `Gunner=yes`.


TaranDahl:
The unit will lose its target when you issue an order with a mouse click, so I suspect this will lead to very strange in‑game behavior.
What specific use case are you trying to achieve with it?

NukeAtty:
> The unit will lose its target when you issue an order with a mouse click, so I suspect this will lead to very strange in‑game behavior. What specific use case are you trying to achieve with it?

e.g. Use AA turret when firing AA weapon.
swap-turret

TaranDahl:
I think a timer could be set so that the turret remains active for a period of time after the unit loses its target.
Looking forward to other maintainers' opinions about the whole design.

docs/Whats-New.md
```diff
@@ -72,6 +72,7 @@ This serves as a changelog for when you just need to drop the new version in wit
   - `PassengerDeletion.SoylentMultiplier`: `0.0` -> `1.0`
   - `PassengerDeletion.SoylentAllowedHouses`: `none` -> `enemies`
   - `PassengerDeletion.DisplaySoylentOffset`: `none` -> `all`
+- VXL vehicles can now swap their turret and barrel models per weapon index by using `MultiWeapon` together with `TurretCount`. (by Nuke)
```
DeathFishAtEase: Changelog position is incorrect.

docs/New-or-Enhanced-Logics.md
```diff
@@ -2620,6 +2620,33 @@ BurstPerTurret=0            ; integer
 This is essentially designed to restore multi-turrets in RA1, not to help you make a Zantos. Therefore, features like different turrets using different weapons/targeting separately/calculating cooldown separately/having separate health are not within the scope of this function. Please use techno attachment for such requirements.
 ```
 
+### Switch Voxel Turret Depends on MultiWeapon
```
DeathFishAtEase: If it does not require any new statements and is used in conjunction with *MultiWeapon*, then it is best to merge it into the paragraph where the *MultiWeapon* feature resides.

NukeAtty:
> I think a timer could be set so that the turret remains active for a period of time after the unit loses its target. Looking forward to other maintainers' opinions about the whole design.

I think it will be better to keep the turret of lastest used weapon, so no timer needs to be set.

PR #2332, opened 2026-08-09
Experiment with removing moving check from ReadyToNextMission

Coronia:
This is a follow up of https://github.com/Phobos-developers/Phobos/pull/2088. Before a techno is going to switch to another mission, it'll check if it's currently moving. While it's problematic for Jumpjet and Hover already, there's also the question that whether this is really needed

Make a PR for testing the behavior of its removal first, and then we can decide if we could just remove this check directly, or make it an option with an ini toggle, or not doing it at all. No issue has been found in my local test so far, but more test is needed under different scenarios

- Fixed the issue that the move mission doesn't end when the techno is moving, which is mostly problematic for jumpjet and hover techno. Set `[General] -> ReadyToNextMission.MovingCheck` to true to disable the fix.


Coronia:
add a toggle `[General] -> ReadyToNextMission.MovingCheck` to restore it to the vanilla behavior. It's set to false as of now for easier test, and is likely to keep this way if no problem occurs in testing

NetsuNegi (approved):

src/Ext/Techno/Hooks.Firing.cpp
Coronia, line 1291: @TaranDahl should we also check Limbo at here as well as some other ApproachTarget hooks?

TaranDahl: There is no such check in vanilla pursuit (0x7415F2).
But I think it is reasonable to add it here.

TaranDahl (approved):
Should make the final decision at the time of the v0.5 stable release.

Coronia:
the Jumpjet handling of ApproachTarget.StopWhenInRange has been restored to the old version, otherwise it'll rotate when executing the stop process

the OpenTopped issue seems to be fixed, and no other issue is found other than these 2 so far. Gonna merge it first after some more test to solve the immediate problem, and then we can observe more to decide whether we should turn this change on or off by default

TaranDahl:
I think it's best to only call StopMoving here. That extra handling should actually be a new Feature of JumpjetLocomotion (more code is required to make it work fully correctly elsewhere).
However, this can be addressed at a later time.

PR #2334, opened 2026-08-10
Add veteran and elite reload time customization with \RELOAD\ and \EMPTY_RELOAD\ abilities

NukeAtty:
Adds two new veterancy abilities and their duration multipliers:

- `RELOAD` - scales the duration of normal reload cycles.
- `EMPTY_RELOAD` - scales the duration of the empty-clip reload cycle, which is only used when the TechnoType has a positive `EmptyReload`.

Both abilities are read from the vanilla `VeteranAbilities` / `EliteAbilities` lists and accumulate the same way as vanilla abilities (Veteran abilities keep working at elite level, Elite abilities are added on top).

New INI tags:

```ini
[General]
VeteranReload=1.0        ; floating point value, multiplier
VeteranEmptyReload=      ; floating point value, multiplier, default to [General] -> VeteranReload

VeteranReload=           ; floating point value, multiplier, default to [General] -> VeteranReload
VeteranEmptyReload=      ; floating point value, multiplier, default to [General] -> VeteranEmptyReload
VeteranAbilities=        ; Ability, `RELOAD` and `EMPTY_RELOAD` in addition to the vanilla abilities
EliteAbilities=          ; Ability, `RELOAD` and `EMPTY_RELOAD` in addition to the vanilla abilities
```

The multipliers are applied to the final reload duration computed by the game, so other reload modifiers such as `ReloadIncrement` are still taken into account. They only affect newly started reload cycles, and a multiplier of `1.0` (the default) leaves vanilla behavior unchanged.

Tested in-game: both `RELOAD` and `EMPTY_RELOAD` verified for Veteran/Elite ranks, and mixed lists with vanilla abilities (e.g. `VeteranAbilities=STRONGER,RELOAD`) confirmed to keep the vanilla abilities working.

Documentation, changelog and credits updated.


docs/New-or-Enhanced-Logics.md
```diff
@@ -2454,6 +2454,32 @@ TiberiumEater.AnimMove=true       ; boolean
...
+In `rulesmd.ini`:
+```ini
+[General]
+VeteranReload=1.0        ; floating point value, multiplier
+VeteranEmptyReload=1.0   ; floating point value, multiplier
+
+[SOMETECHNO]             ; TechnoType
+VeteranReload=           ; floating point value, multiplier, defaults to [General] -> VeteranReload
+VeteranEmptyReload=      ; floating point value, multiplier, defaults to [General] -> VeteranEmptyReload
+VeteranAbilities=RELOAD  ; Ability, `RELOAD` and `EMPTY_RELOAD` in addition to the vanilla abilities
```
DeathFishAtEase: Format error, see any other INI block that has default values.

docs/New-or-Enhanced-Logics.md
```diff
@@ -2454,6 +2454,32 @@ TiberiumEater.AnimMove=true       ; boolean
...
+EliteAbilities=EMPTY_RELOAD ; Ability
```
DeathFishAtEase: Also; the description text of the comments is not standardized, and the semicolons are not aligned.

src/Ext/Rules/Body.cpp
```diff
@@ -352,6 +354,22 @@ void RulesExt::ExtData::LoadBeforeTypeData(RulesClass* pThis, CCINIClass* pINI)
...
 	this->NoReload_Temporal.Read(exINI, GameStrings::General, "NoReload.Temporal");
+
+	this->VeteranReload.Read(exINI, GameStrings::General, "VeteranReload");
+	this->VeteranEmptyReload.Read(exINI, GameStrings::General, "VeteranEmptyReload");
+
+	const auto validateReloadMultiplier = [](const char* pKey, Valueable<double>& value)
+	{
+		if (!std::isfinite(value.Get()) || value.Get() <= 0.0)
+		{
+			Debug::INIParseFailed(GameStrings::General, pKey, "<invalid>", "Expected a finite value greater than 0.0");
```
DeathFishAtEase: What is the necessity of doing this... Another PR has a similar style; that was done because LLM told him to do it that way.

src/Ext/TechnoType/Body.cpp
```diff
@@ -12,6 +12,46 @@
...
+			}
+		}
+	}
+
+	void ValidateReloadMultiplier(const char* pSection, const char* pKey, Nullable<double>& value)
+	{
+		if (value.isset() && (!std::isfinite(value.Get()) || value.Get() <= 0.0))
+		{
+			Debug::INIParseFailed(pSection, pKey, "<invalid>", "Expected a finite value greater than 0.0");
+			value.Reset();
```
DeathFishAtEase: It appears to be re-implementing the reading mentioned in the previous item.

src/Ext/Techno/Body.Update.cpp
```diff
@@ -29,6 +33,45 @@ void TechnoExt::OnEarlyUpdate()
...
+
+	const auto pTypeExt = this->TypeExtData;
+	const auto pRulesExt = RulesExt::Global();
+
+	const double multiplier = isEmptyReload
+		? pTypeExt->VeteranEmptyReload.Get(pRulesExt->VeteranEmptyReload)
+		: pTypeExt->VeteranReload.Get(pRulesExt->VeteranReload);
+
+	// A non-positive or non-finite multiplier must not create an invalid timer.
+	if (!std::isfinite(multiplier) || multiplier <= 0.0)
```
DeathFishAtEase: I think the value has already been processed into a legal value when loading.

src/Ext/Techno/Body.Update.cpp
```diff
@@ -7,13 +7,17 @@
...
 #include <New/Type/Affiliated/TypeConvertGroup.h>
 
+#include <cmath>
+
 
 // TechnoClass_AI_0x6F9E50
 // It's not recommended to do anything more here it could have a better place for performance consideration
 void TechnoExt::OnEarlyUpdate()
 {
+	this->ApplyPendingReloadVeterancy();
```
TaranDahl: I suppose this should be applied like the ROF multiplier. Why do you implement it like that?

TaranDahl (approved):

TaranDahl (changes requested):
Wrong decision in the last review.

TaranDahl (approved):

DeathFishAtEase (approved):

docs/New-or-Enhanced-Logics.md
```diff
@@ -2343,6 +2343,33 @@ RecountBurst=false  ; boolean
...
+VeteranReload=1.0        ; floating point value, multiplier
+VeteranEmptyReload=      ; floating point value, multiplier, default to [General] -> VeteranReload
+
+[SOMETECHNO]             ; TechnoType, with VeteranAbilities and/or EliteAbilities containing RELOAD or EMPTY_RELOAD.
+VeteranReload=           ; floating point value, multiplier, default to [General] -> VeteranReload
+VeteranEmptyReload=      ; floating point value, multiplier, default to [General] -> VeteranEmptyReload
+```
+
+```{note}
+For `EMPTY_RELOAD` to have any effect, the TechnoType must have a positive `EmptyReload` value, otherwise the empty clip uses the regular `Reload` duration instead.
```
DeathFishAtEase: This part seems to be an extension of the description of vanilla behavior, but according to 0x6FB0B0, it should be checking `EmptyReload == -1`.

src/Ext/TechnoType/Body.cpp
```diff
@@ -12,6 +12,43 @@
...
+		if (!_stricmp(value.c_str(), "RELOAD"))
+		{
+			result.set(static_cast<size_t>(AdditionalAbility::Reload));
+		}
+		else if (!_stricmp(value.c_str(), "EMPTY_RELOAD"))
+		{
+			result.set(static_cast<size_t>(AdditionalAbility::EmptyReload));
+		}
+	}
+}
```
DeathFishAtEase: Is this better? Because `ReadAdditionalAbilities` is only used within this file, it does not need to be defined in the global scope.

```suggestion
namespace
{
	constexpr std::pair<const char*, AdditionalAbility> AbilityTokens[] = {
		{ "RELOAD",       AdditionalAbility::Reload },
		{ "EMPTY_RELOAD", AdditionalAbility::EmptyReload },
	};

	void ReadAdditionalAbilities(
			INI_EX& parser,
			const char* section,
			const char* key,
			std::bitset<AdditionalAbilityCount>& result)
	{
		std::vector<std::string> values;

		if (!parser.ParseStringList(values, section, key))
			return;

		// When the key is present, fully replace the previous value with this
		// list so that map INIs can override rules values.
		result.reset();

		for (const auto& value : values)
		{
			for (const auto& [name, ability] : AbilityTokens)
			{
				if (!_stricmp(value.c_str(), name))
				{
					result.set(static_cast<size_t>(ability));
					break;
				}
			}
		}
	}
}
```

DeathFishAtEase (commented):
Two optional suggestions regarding wording precision and writing improvements.

PR #2336, opened 2026-08-11
Refactor LaserPositionUpdate as LaserDrawExt and further optimize it

DeathFishAtEase:
Todo:
- [x] Refactor LaserPositionUpdate in the form of LaserDrawExt
- [x] Fix the bug where FLH calculation is incorrect when turning sideways after firing.
- [ ] Confirm the coordination of Nuke with the built-in configuration rules in multi-map files.
  - Especially the issue of whether Nuke handling is correct in scenarios where built-in INI configuration changes occur while entering and exiting multiple maps in succession (e.g., consecutive levels).


src/Ext/LaserDraw/Hooks.cpp
```diff
@@ -0,0 +1,276 @@
+#include <Ext/LaserDraw/Body.h>
+
+#include <Ext/Techno/Body.h>
+#include <Ext/WeaponType/Body.h>
+#include <DiskLaserClass.h>
+#include <Helpers/Macro.h>
+#include <Utilities/Patch.h>
+
+static void NukeTrackingHooks()
```
Coronia: put this in Phobos::ApplyOptimizations() like all the other hook nuking functions

DeathFishAtEase: Done

PR #2337, opened 2026-08-11
[Vanilla Fix] Fixed the issue of significant lagging caused by frequent lighting updates due to the accumulation of a large amount of radsite in a short time

NetsuNegi:


src/Ext/RadSite/Body.cpp
Coronia: might as well just remove this variable and make it use false in the below functions directly

Coronia (approved):

PR #2338, opened 2026-08-11
Fix the bug where `Range=-2` ignores `MinimumRange`

DeathFishAtEase:
In vanilla, weapons with `Range=-2` ignore `MinimumRange`; this has now been fixed.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -739,6 +739,7 @@ This page lists all the individual contributions to the project by their author.
   - Remove the restriction that prohibits InfantryTypes from using the InitialPayload logic
   - Add `ammo`, `health`, `mission`, `landtype` and `sequence` conditions to `DiscardOn`
   - Disable AlphaImage during Buildup
+  - Fix the bug where `Range=-2` ignores `MinimumRange`
 - **Ollerus**:
   - Build limit group enhancement
   - Customizable rocker amplitude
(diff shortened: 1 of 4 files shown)


DeathFishAtEase:
Because negative `Range` values require additional maintenance overhead to behave like positive values in distance calculations required in some places such as `MinimumRange` behavior, and adding support without maintaining a separate set of `MinimumRange` calculations would cause other checks like `SubjectToCliffs` and `SubjectToWalls` to also take effect, it's better to keep it as is. If needed, use a positive value like `Range=513` that is large enough to cover the maximum distance for all map sizes.

DmitryVolkov666:
> If needed, use a positive value like `Range=513` that is large enough to cover the maximum distance for all map sizes. 

I don't know if I should mention this or not, but,
in my experience, using a weapon with such a long range and auto attack will cause the game to lags in real time.

DeathFishAtEase:
> in my experience, using a weapon with such a long range and auto attack will cause the game to lags in real time.

You can change `CanPassiveAquire` or `GuardRange`

DmitryVolkov666:
I know. I just wanted to clarify

although in the case of GuardRange, it isn't very versatile 😅

DeathFishAtEase:
`GuardRange` is obtained from the weapon's `Range` by default; `Range=-2` does not directly negate the scanning of a large area; you can also set a value like `GuardRange=-3` to disable it.

DmitryVolkov666:
off-topic
I miss the days when GuardRange directly boosted range FlakScatter 😂 (This gets fixed in one of the Phobos builds).

PR #2340, opened 2026-08-13
Allow customizing the default value of `[Warhead] -> PreventScatter` via `[CombatDamage] -> Warhead.PreventScatter`

DeathFishAtEase:
```ini
[CombatDamage]
PlayerScatter=true
Warhead.PreventScatter=true

[IQ]
Scatter=0

[Warheads]
+=Test::SA

[Test::SA]:[SA]

[M1Carbine]
Warhead=Test::SA

[SA]
PreventScatter=false
```


TaranDahl (approved):
May also commit to Antares

PR #2346, opened 2026-08-16
Ares' `KeepAlive` adds global tags

Fly-Star-him:
Ares has hardcoded the default value of `KeepAlive` for buildings with `Insignificant=no` and `DontScore=no` as `yes`, while for other types of technologies, it is set to `no`. Now, five global tags have been added for customization.

Ares 将`Insignificant=no`、`DontScore=no`的建筑的`KeepAlive`默认值设为`yes`，其他科技类型设为`no`。现添加了五个全局标签用于自定义设置。

---------

In `rulesmd.ini`:
```ini
[General]
KeepAlive.Buildings=true      ; boolean
KeepAlive.Defenses=true       ; boolean
KeepAlive.Infantry=false      ; boolean
KeepAlive.Units=false         ; boolean
KeepAlive.Aircraft=false      ; boolean
```


TaranDahl:
I think the original automatic value is already smart enough. 
What are the specific use cases for this enhancement?

Fly-Star-him:
> I think the original automatic value is already smart enough. What are the specific use cases for this enhancement?

You no longer need to frantically add `KeepAlive=yes` for each unit in the map file to achieve specific effects. Now, you only need to modify a maximum of four global tags.

Coronia (approved):
modified the tag names a bit to make it inline with other tags that works for different abstract types

PR #2347, opened 2026-08-16
Add EmptyReload InfoType for digital display

NukeAtty:
Adds a new `InfoType=EmptyReload` for digital display.

- New `DisplayInfoType::EmptyReload` enumeration value (appended as `21` to preserve save-game compatibility).
- New INI tag parsing for `InfoType=EmptyReload`.
- Displays the empty reload countdown while the Techno has no ammo left, and `0` otherwise.
- Falls back to `Reload` duration when `EmptyReload` is not positive.
- Existing `InfoType=Reload` behavior is unchanged.


docs/Whats-New.md
```diff
@@ -419,6 +419,7 @@ HideShakeEffects=false           ; boolean
 :open:
 
 #### New:
+- [New `InfoType=EmptyReload` for digital display](User-Interface.md#digital-display) (by NukeAtty)
```
DeathFishAtEase: I suggest that after LLM helps complete the PR, human developers should review it independently first before opening the pull request.

Coronia:
would it be better if this just use `Reload` infotype and handle `EmptyReload` specifically? Not sure if there's any case that only need to show EmptyReload seperatedly

NukeAtty:
> would it be better if this just use `Reload` infotype and handle `EmptyReload` specifically? Not sure if there's any case that only need to show EmptyReload seperatedly

Oh I see, silly of me, now this PR is kinda meaningless. I'd close it now :)

NukeAtty:
Closing: EmptyReload behavior is already covered by the existing Reload InfoType.

PR #2349, opened 2026-08-16
Fix the bug where a building with `Factory=BuildingType` owned by the AI did not play `ProductionAnim` when placing a produced building.

DeathFishAtEase:

## Description

https://github.com/user-attachments/assets/3bc25195-7ad1-4f6f-bcf4-1e029a9b78d3

### DIFF
--- a/src/Ext/Building/Hooks.cpp
+++ b/src/Ext/Building/Hooks.cpp
@@ -1084,7 +1084,7 @@ DEFINE_HOOK(0x4501AF, AI_ConYard_CompleteProduction_ProductionAnim, 0x5)
 	GET(BuildingClass*, pBuilding, ESI);
 	GET(TechnoClass*, pObject, EDI);
 
-	if (pBuilding->Owner->IsHumanPlayer)
+	if (pBuilding->Owner->IsControlledByHuman())
 		return 0;
 
 	if (!pObject || pObject->WhatAmI() != AbstractType::Building)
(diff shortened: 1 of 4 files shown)


DeathFishAtEase:
Tested by: https://www.phoboscn.top/t/topic/720/2?u=noble_fish

image

src/Ext/Building/Hooks.cpp
```diff
@@ -1079,6 +1079,22 @@ DEFINE_HOOK(0x444D11, BuildingClass_ExitObject_ProductionAnimForInfantryFactory,
 	return 0;
 }
 
+DEFINE_HOOK(0x4501AF, AI_ConYard_CompleteProduction_ProductionAnim, 0x5)
+{
+	GET(BuildingClass*, pBuilding, ESI);
+	GET(TechnoClass*, pObject, EDI);
+
+	if (pBuilding->Owner->IsHumanPlayer)
```
Coronia: does the issue also happens on player controlled ConYard of another house? If not then IsControlledByHuman() should be used

DeathFishAtEase: I think that scenario is effectively a non-issue - the AI production path itself is gated by the same check. `HouseClass::Update` skips AI production and AI base construction for any house where `IsControlledByHuman()` returns true, which includes a player-controlled non-player house in campaign. Such a house therefore never produces buildings autonomously, so the issue cannot arise there. That said, your suggestion is still valuable - `IsControlledByHuman()` is self-documenting and saves future maintainers from having to reverse-engineer this part of the code to confirm the distinction. I accept the change.

Coronia (approved):

PR #2358, opened 2026-08-20
Fix issue 2351

TaranDahl:


TaranDahl:
I thought it would apply the label automatically 🤔
image

Coronia (approved):
I‘m not sure if it's necessary to use guard here, but the method is correct anyway

Coronia:
maybe InApplyCrit should still be processed in save load in case of doing so in edge moment?

PR #2360, opened 2026-08-21
Extended jumpjet hovering

TaranDahl:

### Extended Jumpjet Hovering

- In vanilla, jumpjets scatter while hovering by randomly setting a destination to a nearby cell. This approach is inefficient when many jumpjets are crowded around, and it forces the unit to turn, which interferes with firing for non-omni-firing units.
- Now you can enable a new hovering scatter behaviour via the following switch, which disperses units more efficiently and without requiring the unit to turn.
  - `ExtendedJumpjetHovering` controls whether the new hovering scatter behaviour is enabled.

In `rulesmd.ini`:
```ini
[JumpjetControls]                ; global section
ExtendedJumpjetHovering=false   ; boolean
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -881,6 +881,7 @@ This page lists all the individual contributions to the project by their author.
   - Customize whether aircraft is a cargo plane
   - Allow the unit to stop immediately if the target enters the range during ApproachTarget
   - Allow the unit to keep pursuing the target during ApproachTarget
+  - Extended Jumpjet Hovering
 - **solar-III (凤九歌)**
   - Target scanning delay customization (documentation)
   - Skip target scanning function calling for unarmed technos (documentation)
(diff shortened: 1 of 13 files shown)


Coronia:
with this PR jumpjets will scatter in a pattern like infantry: 3 of them will try to occupy a cell. Guess this is intended

https://github.com/user-attachments/assets/256ef84b-6931-4379-b578-226f0b02b0a1

this is not always a better behavior tho, since it'll make jumpjets more vulnerable to aoe damage and make it harder to pick a jumpjet from a cluster. It would be better to split the '3 cells scattering' and 'no need to turn' into 2 different tags so there's more choice here

TaranDahl:
That is not the case.
The final scatter result is: for any units A and B, the distance between A and B shall be no less than the sum of their collision radii, where the collision radius is 64 (for infantry) or 128 (for vehicles or aircraft).
(But it does not seem to have met the design goals. I will check it.)

PR #2365, opened 2026-08-23
Detonate Ivan Bombs placed by the attacker using custom warhead.

dh381-1:

## Description

- Now you can detonate planted Ivan bombs using custom werhead.
  - `IvanBomb.Detonate.SameInvokerOnly` can be used to configure whether the warhead can detonate ivan bombs only if they come from the same invoker.
  - `IvanBomb.Detonate.PenetratesTransport` can be used to configure whether the warhead can detonate ivan bombs on a unit that is inside a transport. The bomb will explode after the unit is unloaded.
  - `IvanBomb.Detonate.PenetratesGarrison` can be used to configure whether the warhead can detonate ivan bombs on a unit that is inside a building. The bomb will explode after the unit leaves the building.
  - IvanBomb.Detonate.AffectsParasite` can be used to configure whether the warhead can detonate ivan bombs on a parasite. The bomb will explode after the parasite leaves the victim.
  - `IvanBomb.Detonate.AffectTypes` can be used to configure Ivan bombs from which TechnoType can be detonated by warhead, use empty for all types.

In `rulesmd.ini`:
```ini
[SOMEWARHEAD]                                ; WarheadType
IvanBomb.Detonate=false                      ; boolean
IvanBomb.Detonate.SameInvokerOnly=true       ; boolean
IvanBomb.Detonate.PenetratesTransport=false  ; boolean
IvanBomb.Detonate.PenetratesGarrison=false   ; boolean
IvanBomb.Detonate.AffectsParasite=false       ; boolean
IvanBomb.Detonate.AffectTypes=               ; List of TechnoTypes
```

> [!Note]
> `IvanBomb.Detonate.AffectTypes` doesn't work if the owner of the Ivan bomb is dead. This may change in future.


src/Ext/Techno/Hooks.WeaponEffects.cpp
```diff
@@ -340,3 +340,32 @@ DEFINE_HOOK(0x762D57, WaveClass_AI_TargetUnset, 0x6)
 
 	return 0;
 }
+
+DEFINE_HOOK(0x702D19, TechnoClass_ReceiveDamage_IvanBombDetonate, 0x6)
+{
+	GET(ObjectClass*, pThis, ESI);
```
NetsuNegi: `pThis` should be `TechnoClass*`. Besides, register esi has been popped here, I think you should not use it.
`pThis` 应该是 `TechnoClass*`。另外，ESI寄存器在此处已被弹出，我认为你不该用它。

dh381-1: After changing the hook location,register esi can be used
换了一个hook位置，ESI寄存器现在就可以用了

docs/New-or-Enhanced-Logics.md
```diff
@@ -3682,3 +3682,15 @@ CanTargetVeterancy=all      ; List of Affected Veterancy Enumeration (none|rooki
...
+### Manually detonate Ivan bomb
+
+![image](_static/images/IvanBombDetonate.gif)
+
+- Now you can detonate planted Ivan bombs using custom weapon. The bomb attached to the targeted unit will explode immediately, provided that it was planted by the attacker.
+
+In `rulesmd.ini`:
+```ini
+[SOMEWEAPON]               ; WeaponType
+IvanBomb.Detonate=true     ; Enable bomb detonation (true/false), default: false.
```
DeathFishAtEase:
```suggestion
IvanBomb.Detonate=false  ; boolean
```
If there is a fixed default value, it should be placed as the value on the right side of the equals sign; only the value type should be written after the semicolon.

src/Ext/Techno/Hooks.ReceiveDamage.cpp
```diff
@@ -566,3 +566,30 @@ DEFINE_HOOK(0x737E6E, UnitClass_ReceiveDamage_SkipExplode, 0xA)
...
+	GET(TechnoClass*, pThis, ESI);                   
+    GET_STACK(TechnoClass*, pSource, STACK_OFFSET(0xC4, 0x10));
+	GET_STACK(WarheadTypeClass*, pWH, STACK_OFFSET(0xC4, 0xC));
+
+    if (!pSource)
+        return 0; 
+
+	if(auto pBomb = pThis->AttachedBomb)
+	{
+		if(auto pSourceExt = TechnoExt::TryFetch(pSource))
```
Coronia: no need to use TryFetch here since it has already been checked above

src/Ext/WeaponType/Body.h
```diff
@@ -118,6 +118,7 @@ class WeaponTypeExt final : public AbstractTypeExt
 	bool SkipWeaponPicking;
 
 	Nullable<bool> CylinderRangefinding;
+
```
NetsuNegi, line 121: what's this ?

src/Ext/Techno/Hooks.ReceiveDamage.cpp
NetsuNegi:
```suggestion
		if(auto pSourceExt = TechnoExt::Fetch(pSource))
```
No need use `TryFetch` here, you had checked `pSource` before.
这里不需要使用 `TryFetch`，你在前面已经检查过 `pSource` 了。

NetsuNegi (changes requested):

NetsuNegi:
Other codes looking good

Coronia (commented):
I think this doesn't have to be restricted to the owner's bomb. It would also be useful if another unit could detonate it. Something like this:

IvanBomb.Detonate.InvokerOnly -> whether it can only affect your own bomb or not
IvanBomb.Detonate.AffectsType -> TechnoTypes in this list can have their ivan bomb detonated by this warhead. Use empty list for all types

src/Ext/Techno/Hooks.ReceiveDamage.cpp
dh381-1: Now can use `IvanBomb.Detonate.InvokerOnly` and `IvanBomb.Detonate.AffectsType`

src/Ext/Techno/Hooks.ReceiveDamage.cpp
```diff
@@ -566,3 +566,40 @@ DEFINE_HOOK(0x737E6E, UnitClass_ReceiveDamage_SkipExplode, 0xA)
...
+	GET(TechnoClass*, pThis, ESI);                   
+    GET_STACK(TechnoClass*, pSource, STACK_OFFSET(0xC4, 0x10));
+	GET_STACK(WarheadTypeClass*, pWH, STACK_OFFSET(0xC4, 0xC));
+
+    if (!pSource)
+        return 0; 
+
+	if(auto pBomb = pThis->AttachedBomb)
+	{
+		if(auto pSourceExt = TechnoExt::Fetch(pSource))
```
NetsuNegi: this sanity check is not nesseary

src/Ext/Techno/Hooks.ReceiveDamage.cpp
```diff
@@ -566,3 +566,40 @@ DEFINE_HOOK(0x737E6E, UnitClass_ReceiveDamage_SkipExplode, 0xA)
...
+		{
+			if(auto pWHExt = WarheadTypeExt::TryFetch(pWH))
+			{
+				if(!pWHExt->IvanBomb_Detonate)
+					return 0;
+
+				bool CanAffects = pWHExt->IvanBomb_Detonate_AffectsType.Contains(pThis->GetTechnoType())
```
NetsuNegi: I think it will be better to check empty first

docs/New-or-Enhanced-Logics.md
```diff
@@ -3682,3 +3682,19 @@ CanTargetVeterancy=all      ; List of Affected Veterancy Enumeration (none|rooki
...
+- Now you can detonate planted Ivan bombs using custom werhead. The bomb attached to the targeted unit will explode immediately, provided that it was planted by the attacker.
+- Use `IvanBomb.Detonate.InvokerOnly` to configure whether the warhead can detonate Ivan bombs from other sources.
+- Use `IvanBomb.Detonate.AffectsType` to configure which targets' Ivan bombs can be detonated by warhead, use empty for all types.
+
+In `rulesmd.ini`:
+```ini
+[SOMEWARHEAD]                             ; WarheadType
+IvanBomb.Detonate=true                    ; boolean
+IvanBomb.Detonate.InvokerOnly=false       ; boolean
+IvanBomb.Detonate.AffectsType=HTNK,E1     ; List of Registration Name, use empty list for all types
```
TaranDahl, line 3703: It is unreasonable to reinvent a filter here.

src/Ext/Techno/Hooks.ReceiveDamage.cpp
```diff
@@ -566,3 +566,40 @@ DEFINE_HOOK(0x737E6E, UnitClass_ReceiveDamage_SkipExplode, 0xA)
 	R->EAX(pThis->GetHeight());
 	return ContinueCheck;
 }
+
+DEFINE_HOOK(0x701DFF, TechnoClass_ReceiveDamage_IvanBombDetonate, 0x7)
```
TaranDahl: This hook is redundant.
You should process this in WarheadTypeExt::DetonateOnOneUnit.

src/Ext/Techno/Hooks.ReceiveDamage.cpp
```diff
@@ -566,3 +566,40 @@ DEFINE_HOOK(0x737E6E, UnitClass_ReceiveDamage_SkipExplode, 0xA)
...
+				if(!pWHExt->IvanBomb_Detonate)
+					return 0;
+
+				bool CanAffects = pWHExt->IvanBomb_Detonate_AffectsType.Contains(pThis->GetTechnoType()) 
+								  || pWHExt->IvanBomb_Detonate_AffectsType.empty();
+				
+				if(pWHExt->IvanBomb_Detonate_InvokerOnly)
+				{
+					if(pBomb->Owner == pSource && CanAffects)
+						pBomb->Detonate();
```
TaranDahl: Invoking a damage in another damage process is dangerous.
IMO resetting the timer would be better.

TaranDahl (changes requested):

NetsuNegi (commented):
uh, NVM this comment

src/Ext/Techno/Hooks.ReceiveDamage.cpp
NetsuNegi: > This hook is redundant. You should process this in WarheadTypeExt::DetonateOnOneUnit.

This is his reason: https://www.phoboscn.top/t/topic/739/12

TaranDahl: I have checked it. It makes no sense. This must be changed.

docs/New-or-Enhanced-Logics.md
```diff
@@ -3682,3 +3682,23 @@ CanTargetVeterancy=all      ; List of Affected Veterancy Enumeration (none|rooki
...
+- Use `IvanBomb.Detonate.PenetrateBuilding` to configure whether the warhead can detonate Ivan bombs on a unit that is inside a building. The bomb will explode after the unit leaves the building.
+- Use `IvanBomb.Detonate.AffectsType` to configure which targets' Ivan bombs can be detonated by warhead, use empty for all types.
+
+In `rulesmd.ini`:
+```ini
+[SOMEWARHEAD]                             ; WarheadType
+IvanBomb.Detonate=true                    ; boolean
+IvanBomb.Detonate.InvokerOnly=false       ; boolean
+IvanBomb.Detonate.PenetrateTransport=true ; boolean
+IvanBomb.Detonate.PenetrateBuilding=true  ; boolean
```
Coronia, line 3702: rename this to `PenetratesGarrison` since it can only affect garrisoned buildings, while not the other 'enterable' buildings such as Bio Reactor and Ares' tunnel

NetsuNegi (approved):

docs/New-or-Enhanced-Logics.md
```diff
@@ -3799,6 +3799,30 @@ OmniFire.TurnToTarget=false  ; boolean
 OmniFire.TurnToTarget=       ; boolean, default to [General] -> OmniFire.TurnToTarget
 ```
 
+### Manually detonate Ivan bomb
```
DeathFishAtEase: 1. This is a warhead feature, but it is currently placed under the weapon category section.
2. Manually does not match the semantics of its actual use: this is not manually triggered, but triggered by warhead detonation.

```suggestion
### Detonate ivan bomb on impact
```

docs/New-or-Enhanced-Logics.md
```diff
@@ -3799,6 +3799,30 @@ OmniFire.TurnToTarget=false  ; boolean
...
+- Use `IvanBomb.Detonate.AffectTypes` to configure Ivan bombs from which TechnoType can be detonated by warhead, use empty for all types.
+
+In `rulesmd.ini`:
+```ini
+[SOMEWARHEAD]                              ; WarheadType
+IvanBomb.Detonate=true                     ; boolean
+IvanBomb.Detonate.InvokerOnly=false        ; boolean
+IvanBomb.Detonate.PenetratesTransport=true ; boolean
+IvanBomb.Detonate.PenetratesGarrison=true  ; boolean
+IvanBomb.Detonate.AffectTypes=             ; List of TechnoTypes
```
DeathFishAtEase: In the INI block, the default values of the statements should be accurately reflected; the current "true" and "false" do not match the initialization values in the source code.

BTW, I would rather recommend that the order of variables in the source code be consistent with the ordering in the INI block; this helps save the energy spent on cross-checking.

docs/New-or-Enhanced-Logics.md
```diff
@@ -3799,6 +3799,30 @@ OmniFire.TurnToTarget=false  ; boolean
 OmniFire.TurnToTarget=       ; boolean, default to [General] -> OmniFire.TurnToTarget
 ```
 
+### Manually detonate Ivan bomb
+
+![image](_static/images/IvanBombDetonate.gif)
+
+- Now you can detonate planted Ivan bombs using custom werhead. The bomb attached to the targeted unit will explode immediately, provided that it was planted by the attacker.
+- Use `IvanBomb.Detonate.InvokerOnly` to configure whether the warhead can detonate Ivan bombs from other sources.
```
DeathFishAtEase: "InvokerOnly" is ambiguous; for example, we already have a statement such as `AffectsInvokerOnly`. It needs to be ensured that the semantics are "granted by the same invoker" rather than other meanings such as "on the invoker".

DeathFishAtEase (changes requested):
Some documentation issues

docs/New-or-Enhanced-Logics.md
DeathFishAtEase: `AffectsInvokerOnly` means that the effect "only affects the invoker", which is not "only affects ivan bombs granted by the same invoker".

> Use `IvanBomb.Detonate.AffectsInvokerOnly` to configure whether the warhead can detonate Ivan bombs from other sources.

The wording here is also problematic - when `true`, it means "only affects ivan bombs granted by the same invoker" rather than "can affect ivan bombs not granted by the invoker".

If you need some help with naming, here are two options for reference:

- `IvanBomb.Detonate.SameInvokerOnly` - an improvement to the current proposal, but the flag description issue in the documentation still needs to be corrected.
- `IvanBomb.Detonate.AllowExternalSources` - a reverse naming scheme, meaning "whether it can detonate ivan bombs not from the invoker", consistent with the documentation, but the default value needs to be inverted.

DeathFishAtEase (approved):

docs/New-or-Enhanced-Logics.md
TaranDahl: ?

TaranDahl: @FS-21

FS-21: Tell me what happens

TaranDahl: The PR is merged with this issue unaddressed. This should not happen.

PR #2371, opened 2026-08-25
Customize `DefaultToGuardArea` per gunner mode

DeathFishAtEase:

## Description

- Technos with `Gunner=yes` can now restrict the `DefaultToGuardArea` and the `GUARD_AREA` promotion ability to specific gunner modes.

In `rulesmd.ini`:
```ini
[SOMETECHNO]                     ; TechnoType, with Gunner=yes
DefaultToGuardArea.Modes=-1      ; List of integers, IFVMode
DefaultToGuardArea.AIModes=-1    ; List of integers, IFVMode
```


src/Ext/Techno/Hooks.Misc.cpp
```diff
@@ -1103,3 +1104,25 @@ DEFINE_HOOK(0x4C6CF0, EventClass_RespondToEvent_CheckControllability, 0x8)  // P
...
+		auto const pTypeExt = TechnoTypeExt::Fetch(pType);
+		bool const hasExplicit = pThis->Owner->IQLevel2 >= RulesClass::Instance->GuardArea || pType->DefaultToGuardArea || pThis->HasAbility(Ability::GuardArea);
+		auto const& modes = pThis->Owner->IsControlledByHuman() ? pTypeExt->DefaultToGuardArea_Modes : pTypeExt->DefaultToGuardArea_AIModes;
+
+		if (hasExplicit && !modes.empty() && (modes.size() != 1 || modes[0] != -1))
+			return modes.Contains(pThis->CurrentWeaponNumber) ? GoAreaGuardDecision : GoGuard;
+	}
+
+	R->EAX(pThis->Owner);
+	return Continue;
```
NetsuNegi:
```suggestion
	return 0;
```

NetsuNegi (approved):

Coronia:
is it also possible to remove the hardcoded range of area guard for repair IFV?

TaranDahl (changes requested):
I suppose this won't work if you order a passenger to enter a vehicle that has already been idle.

DeathFishAtEase:
> remove the hardcoded range of area guard for repair IFV

According to `0x6F9037`, under `Area_Guard` mission, there is originally no `512` leptons limit; that is for `Guard` mission.

> that has already been idle.

`DefaultToGuardArea` works like this: even if you directly create a unit with `DefaultToGuardArea=true`, or have such a unit already in the `Guard` mission gain the `GUARD_AREA` promotion ability through external means, it still waits until the next command finishes and returns to idle before making the judgment.

If someone wants passengers to switch to the `Area_Guard` mission immediately upon entering, that should be implemented as a separate feature.

src/Ext/Techno/Hooks.Targeting.cpp
```diff
@@ -90,3 +90,8 @@ DEFINE_HOOK(0x6F9AF4, TechnoClass_SelectAutoTarget_DisableStupid, 0x6)
 {
 	return RulesExt::Global()->DisableOveroptimizationInTargeting ? 0x6F9B1B : 0;
 }
+
+DEFINE_HOOK(0x6F9039, TechnoClass_SelectAutoTarget_HealGuardRange, 0x5)
```
Coronia, line 94: could use DEFINE_JUMP here?

DeathFishAtEase:
Is it necessary to force-write a `GuardRange` based on parameters such as the weapon's `Range` when using a repair weapon? A difference in terms of Mission: if a unit's `GuardRange` exceeds the weapon's `Range`, under `Area_Guard` they will approach the enemy, while under `Guard` they will just stand there staring.

Added a max(512, range) handling to be compatible with the vanilla combination of weapon range and guard radius.

src/Ext/Techno/Hooks.Targeting.cpp
DeathFishAtEase: Done

TaranDahl:
> even if you directly create a unit with `DefaultToGuardArea=true`, or have such a unit already in the `Guard` mission gain the `GUARD_AREA` promotion ability through external means, it still waits until the next command finishes and returns to idle before making the judgment.

That should be considered a bug. 
Bug of the mission system, tho.

Coronia:
so how's the DefaultToGuardArea fix? If it's not doable recently, perhaps merge this one first since it's tested and approved, while doing that in another PR?

src/Ext/Techno/Hooks.Targeting.cpp
```diff
@@ -90,3 +90,56 @@ DEFINE_HOOK(0x6F9AF4, TechnoClass_SelectAutoTarget_DisableStupid, 0x6)
...
+		pHealWeapon = pThis->GetWeapon(pThis->CurrentWeaponNumber)->WeaponType;
+	}
+	else
+	{
+		for (int i = 0; i < count; i++)
+		{
+			auto pWeapon = (isElite ? pType->GetEliteWeapon(i) : pThis->GetWeapon(i))->WeaponType;
+
+			if (!pWeapon && isElite)
+				pWeapon = pThis->GetWeapon(i)->WeaponType;
```
Coronia: perhaps use existing TechnoTypeExt::GetWeaponType for these

src/Ext/Techno/Hooks.Targeting.cpp
```diff
@@ -90,3 +90,56 @@ DEFINE_HOOK(0x6F9AF4, TechnoClass_SelectAutoTarget_DisableStupid, 0x6)
...
+
+			if (isHealWeapon(pWeapon))
+			{
+				pHealWeapon = pWeapon;
+				break;
```
Coronia: not sure what would be a better option if a unit has more than 1 heal weapon. Maybe the one with the longest range should be picked?

src/Ext/Techno/Hooks.Targeting.cpp
```diff
@@ -90,3 +90,56 @@ DEFINE_HOOK(0x6F9AF4, TechnoClass_SelectAutoTarget_DisableStupid, 0x6)
...
 }
+
+DEFINE_HOOK(0x6F9039, TechnoClass_SelectAutoTarget_HealGuardRange, 0x5)
+{
+	GET(TechnoClass* const, pThis, ESI);
+
+	int rangeLeptons = 512; // fallback to the vanilla value if no heal weapon is found
+
+	const auto pType = pThis->GetTechnoType();
+	const auto pTypeExt = TechnoExt::Fetch(pThis)->TypeExtData;
```
Coronia, line 100: const auto pTypeExt = TechnoExt::Fetch(pThis)->TypeExtData;
const auto pType = pTypeExt->OwnerObject();

src/Ext/Techno/Hooks.Targeting.cpp
```diff
@@ -90,3 +90,56 @@ DEFINE_HOOK(0x6F9AF4, TechnoClass_SelectAutoTarget_DisableStupid, 0x6)
...
+
+	const bool isElite = pThis->Veterancy.IsElite();
+	int count = 2;
+	bool useCurrentWeapon = false;
+
+	if (pTypeExt->MultiWeapon.Get() && !pType->IsGattling && (!pType->HasMultipleTurrets() || !pType->Gunner))
+		count = pType->WeaponCount;
+	else if (pThis->WhatAmI() == AbstractType::Unit && !pType->IsGattling && pType->TurretCount > 0 && (pType->Gunner || !pTypeExt->MultiWeapon.Get()))
```
Coronia: no need to check WhatAmI here

PR #2373, opened 2026-08-27
Set mission timer type

Coronia:

### `513` Set mission timer properties

- Set the method to display mission timer and whether the timer should be displayed reversedly based on given value.
  - For behaviours `0`, `1`, and `2`, the second parameter decides the base value for percentage timer type or reversed timer that'll be used for calculation.
  - For behaviours `3` and `4`, the second parameter decides the index of variable that'll be displayed. If the local/global variable does not exist, 0 will be displayed.

In `mycampaign.map`:
```ini
[Actions]
...
ID=ActionCount,[Action1],521,0,0,[MissionTimerType],[BaseValue or VariableIndex],[ReverseTimer],0,A,[ActionX]
...
```

| *Behaviour* | *Description*                              |
| :---------: | :----------------------------------------- |
| 0           | Show normal timer                          |
| 1           | show percentage based on the given value   |
| 2           | Show timer in digit form                   |
| 3           | Show value of the selected local variable  |
| 4           | Show value of the selected global variable |

```{note}
This won't affect how the global mission timer really ticks, so it'll still end after the mission tiemr expired.
```


docs/AI-Scripting-and-Mapping.md
```diff
@@ -716,6 +716,32 @@ ID=ActionCount,[Action1],512,0,0,[FollowerIndex],0,0,0,A,[ActionX]
 ...
 ```
 
+### `521` Set mission timer properties
```
DeathFishAtEase: Does choosing 521 instead of 513 or any other number have any special meaning?

Coronia: it's just in case that some PRs already take the consecutive ID, so I made a room here

DeathFishAtEase: Considering the conflict management mechanism before merging, I think the confusion caused by this discontinuity is not necessarily better than the sorting work for some old content from the past that may not be able to move forward anytime soon - especially since some pull requests may contain multiple new actions at once, and reserving empty slots in advance cannot prevent a PR with 4 new actions from being merged when only 3 empty slots remain; in that case, they would have to be placed after rather than into the empty slots, which instead causes all similar situations to require maintaining two lists with different starting points on the same sequence.

Of course, these are just some management opinions. After all, our newly added Trigger Actions already have many skipped numbers now, and no one knows when or what will fill the vacant positions.

docs/AI-Scripting-and-Mapping.md
```diff
@@ -716,6 +716,32 @@ ID=ActionCount,[Action1],512,0,0,[FollowerIndex],0,0,0,A,[ActionX]
...
+In `mycampaign.map`:
+```ini
+[Actions]
+...
+ID=ActionCount,[Action1],521,0,0,[MissionTimerType],[BaseValue or VariableIndex],[ReverseTimer],0,A,[ActionX]
+...
+```
+
+| *Behaviour* | *Description*                              |
+| :---------: | :----------------------------------------: |
```
DeathFishAtEase: If you want the actual rendered output to be left-aligned as you wrote in Markdown, then this should be changed; the current writing is center-aligned.
```suggestion
| :---------: | :----------------------------------------- |
```

src/Ext/TAction/Body.cpp
```diff
@@ -777,6 +781,21 @@ bool TActionExt::SetNextScanario(TActionClass* const pThis, HouseClass* const pH
 	return true;
 }
 
+bool TActionExt::SetMissionTimer(TActionClass* const pThis, HouseClass* const pHouse, ObjectClass* const pObject, TriggerClass* const pTrigger, const CellStruct& location)
+{
+	const int type = pThis->Param3;
+	const int reverse = pThis->Param5;
+	ScenarioExt::Global()->MissionTimer_Variable = pThis->Param4;
+
```
NetsuNegi: Excess space

NetsuNegi (approved):

src/Misc/Hooks.UI.cpp
```diff
@@ -546,12 +547,61 @@ DEFINE_HOOK(0x6D3D10, TacticalClass_Render_BeforeAll, 0x6)
...
+		DrawTimerTemp::IsDigit = true;
+		break;
+	}
+	case 3:
+	{
+		DrawTimerTemp::IsDigit = true;
+		const auto& variables = ScenarioExt::Global()->Variables[0];
+		const auto& it = variables.find(ScenarioExt::Global()->MissionTimer_Variable);
+		if (it != variables.end())
+			DrawTimerTemp::TimeLeft = it->second.Value;
```
DeathFishAtEase, line 585: > If the local/global variable does not exist, 0 will be displayed.

If there is a case where an undefined index is referenced, it is recommended to add
```cpp
else
    DrawTimerTemp::TimeLeft = 0;
```
This is a low-cost defense.

DeathFishAtEase (approved):

PR #2377, opened 2026-08-30
Customize `IdleActionFrequency`

DeathFishAtEase:

## Description

- Now `IdleActionFrequency` can be customized on each infantry.
  - With a single value, the interval is a random value between `IdleActionFrequency * 450` and `IdleActionFrequency * 1800` frames, which customizes the frequency the same way the global value does.
  - With two values, the first one directly sets the lower bound and the second one the upper bound of the random interval in frames.

In `rulesmd.ini`:
```ini
[SOMEINFANTRY]                        ; InfantryType
IdleActionFrequency=                  ; a single floating point value, or a pair of integers defining the random delay range in frames (min, max), defaults to [AudioVisual] -> IdleActionFrequency
```


DeathFishAtEase:
Tested by: https://www.phoboscn.top/t/topic/749/2?u=noble_fish

image

*GIF with `[E2] -> IdleActionFrequency=999,9999`*

image

Coronia (approved):

PR #2382, opened 2026-09-01
Extend CanTargetHouses to allow targeting civilian, special and neutral houses

Rinimabi:

## Description

Adds `civilian`, `special` and `neutral` to the `AffectedHouse` enumeration so weapons can specifically target the civilian, special and neutral houses via `CanTargetHouses`.

- `civilian` matches the first house on the `Civilian` side, same as `CreateUnit.Owner=civilian`.
- `special` matches the house whose country is `Special`.
- `neutral` matches the house whose country is `Neutral`.
- These three houses are not allied with other houses, so they are still matched by `enemies` as before; existing behaviour is unchanged.
- `all` now covers the three new categories as well.

In `rulesmd.ini`:
```ini
[SOMEWEAPON]             ; WeaponType
CanTargetHouses=civilian ; List of Affected House Enumeration (none|owner/self|allies/ally|team|enemies/enemy|civilian|special|neutral|all)
```

Note: `AffectedHouse` is a shared enumeration also used by other tags (e.g. `Interceptor.CanTargetHouses`, `VisibleToHouses`, `*.AffectsHouse`), so those tags accept the new tokens as well.

---
添加了 `civilian`、`special` 和 `neutral` 到 `AffectedHouse` 枚举，允许武器通过 `CanTargetHouses` 专门瞄准平民、特殊和中立所属方。

- `civilian` 匹配 `Civilian` 阵营的第一个所属方，与 `CreateUnit.Owner=civilian` 一致。
- `special` 匹配国家为 `Special` 的所属方。
- `neutral` 匹配国家为 `Neutral` 的所属方。
- 这三个所属方与其他所属方不结盟，因此仍会像之前一样被 `enemies` 命中，现有行为不变。
- `all` 现在也涵盖这三个新类别。

注意：`AffectedHouse` 是共享枚举，也被其他标签使用（如 `Interceptor.CanTargetHouses`、`VisibleToHouses`、`*.AffectsHouse`），因此这些标签也会接受新词条。

## Test

Tested in-game: CanTargetHouses=civilian correctly restricts targeting to the civilian-side house; enemies/all behaviour unchanged.


TaranDahl:
So they were always treated as enemies in IsAlliedWith?
I have not investigated those houses carefully. Maybe you can give a brief?

Starkku:
> So they were always treated as enemies in IsAlliedWith? I have not investigated those houses carefully. Maybe you can give a brief?

Neutral is considered enemy. Special is mutually allied with all players, although some older spawners broke this behaviour and made it enemy as well. Newer (YRpp-based) spawner has it toggleable in spawn.ini defaulting to the old vanilla behaviour of allying with players.

TaranDahl:
@Rinimabi So your premise is wrong. This design is problematic.

Rinimabi:
> @Rinimabi So what is your design when the Ally/Enemy and Civilian/Special/Neutral are in conflict?

I've updated the implementation to match civilian/special/neutral by house identity (FindCivilianSide()/FindSpecial()/FindNeutral()) before the alliance check.

Coronia:
honestly I'm not a fan of this kind of change. House checking is a functions that'll be regularly called, and adding several virtual function call with string comparison is not really an performance friendly way. On the other hand, I don't think it's necessary to make 3 different categories here since it's already a seldom case to filter neutral

I would suggest to only keep one neutral filter which checks `IsNeutral()`. In this way it'll match how other civilian houses are processed in Ares and Phobos and being better in performance. This can also deal with the case when there're more than just 2 civilian sides, since FindCivilianSide can only return the first in the list, and FindNeutral/Special can only match one given name

Rinimabi:
> honestly I'm not a fan of this kind of change. House checking is a functions that'll be regularly called, and adding several virtual function call with string comparison is not really an performance friendly way. On the other hand, I don't think it's necessary to make 3 different categories here since it's already a seldom case to filter neutral
> 
> I would suggest to only keep one neutral filter which checks `IsNeutral()`. In this way it'll match how other civilian houses are processed in Ares and Phobos and being better in performance. This can also deal with the case when there're more than just 2 civilian sides, since FindCivilianSide can only return the first in the list, and FindNeutral/Special can only match one given name

seems better than what i do now, i was just want to filter civilian houses only

Coronia (approved):

PR #2387, opened 2026-09-03
(SW) Cooldown Groups

FS-21:
- Superweapons can be grouped together using `SW.CooldownGroup`. When any superweapon belonging to a group is launched, all other currently active (`IsPresent`) superweapons belonging to that group for the firer house will have their countdown timers reset (`RechargeTimer`), implementing a shared cooldown mechanism.
- `SW.CooldownGroup` accepts a comma-separated list of group names, allowing a superweapon to belong to multiple groups simultaneously.
  - The reset is non-transitive: launching a superweapon only resets superweapons that share at least one group with the *launched* superweapon. Superweapons reset collaterally do not propagate resets to other groups.
- `SW.CooldownGroup.SyncLongest` controls whether all present superweapons in the affected group(s) synchronize their cooldown to the longest `RechargeTime` among the superweapons currently active (built) in the group for that player. If any active superweapon in the group has this set to `true`, the synchronized longest cooldown applies to the entire group. If all active superweapons in the group have this set to `false`, each superweapon resets to its own individual recharge time.
- If the AI has multiple superweapons in the same group ready to fire simultaneously, it automatically and fairly selects one candidate at random each frame rather than always defaulting to the first superweapons defined in `[SuperWeaponTypes]`.

In `rulesmd.ini`:
```ini
[SOMESW]                                ; SuperWeaponType
SW.CooldownGroup=                       ; List of group names (strings separated by commas)
SW.CooldownGroup.SyncLongest=false      ; boolean
```


Coronia:
I think `SW.Link` has already achieved similar thing? Maybe not as convenient as a group though, but it would be better to make the implementation based on it instead of reinventing the wheel

FS-21:
> I think `SW.Link` has already achieved similar thing? Maybe not as convenient as a group though, but it would be better to make the implementation based on it instead of reinventing the wheel

Yeah, `SW.Link.Reset=true` can reset another weapon, but trying to simulate cooldown groups with it is a massive pain and misses several key things (Typed with AI for clean explanation):
- **Setup & maintenance**: If you have 4 weapons in a group, you have to cross-link all of them manually (A links to B, C, D; B links to A, C, D, etc.), which gets messy very quickly. With groups you just give them the same group name tag.
- **No `SyncLongest`**: `SW.Link` just resets each weapon to its own individual timer. It can't check which structures are currently on the map and sync all of them to the longest recharge time.
- **No UI / Tooltip info**: It doesn't update the cameo tooltips, so the player has no idea the weapon is actually tied to a longer group cooldown.
- **AI selection**: For the AI, `SW.Link` only runs *after* a weapon is already launched. It doesn't solve the issue where the AI always prioritizes the first weapon in `[SuperWeaponTypes]` when several are ready at the same time.
So while you can technically do a basic mutual reset with `SW.Link`, having a proper group tag is much cleaner to configure and covers the actual gameplay and UI details that `SW.Link` can't do.

PR #2388, opened 2026-09-03
Some shield respawn/self heal issue fix

Coronia:

## Description

```{note}
When using `Convert` or `(Un)DeploysInto` to change TechnoType, shield interaction will be like the following:
- If the previous TechnoType doesn't have a shield, while the new TechnoType has `ShieldType` defined, the new TechnoType's shield will be applied immediately.
- If the previous TechnoType has a shield with `AllowTransfer=true`, it will be transferred to the new TechnoType whether it's broken or not, and whether the new TechnoType has `ShieldType` defined or not. If `AllowTransfer` for old shield type is set to false, it will be removed immediately, and changed to the new TechnoType's `ShieldType` if defined.
- If the previous TechnoType has a shield that is broken and `AllowTransfer=false`, while the new TechnoType has `ShieldType` defined, the new shield will be created but still in broken status, which can be respawned if the new shield type has `Respawn` set or being affected by a warhead with Respawn modifiers.
  - If the old shield type also has Respawn, the new shield's Respawn timer will be adjusted to the same percentage through before and after. Otherwise the timer will be reset to the new shield's `Respawn.Rate`. If the Respawn timer is applied via warhead, then it will continue to tick without value change.
  - If the old shield type is during the delay of `Respawn.RestartInCombatDelay` when the change of TechnoType is happening, it will prevent the Respawn from happening until it's over.
  - The shield's `SelfHealing` properties also follow an identical transfer rule of `Respawn` when the change of TechnoType is happening.
- Whenever a shield is passed between TechnoTypes, its current strength will be adjusted to the same percentage through before and after, and most of the properties will be transferred. Notice that the timers for warhead modifiers and combat delay will not be reset to the same percentage; they keep their remaining time as‑is, for simplicity.
```


Coronia:
change:
- when a warhead with shield modifier is about to restart the Respawn/SelfHealing process of a shield, it won't happen if it's during `Respawn/SelfHealing.RestartInCombatDelay`
- if `Shield.Respawn/SelfHealing.Duration` is not set, the warhead won't change the shield's Respawn/SelfHealing properties. That is to say: `Shield.Respawn/SelfHealing.RestartTimer` can't reset then alone, which also matches the previous behavior since it'll be reverted in the next frame if there's no duration
- when tweaking the timer of Respawn/SelfHealing, the timer is no longer required to be `InProgress()`. This is in case that the timer is stopped or paused by `RestartInCombatDelay`, `Temporal`, `Powered` etc
- if a techno with shield that can be transferred is changing to another techno with shield, its Respawn/SelfHealing timer will be recalculated based on the ratio of their `Respawn/SelfHealing.Rate`, if they're not under the effect of warhead modifier
- specially, if the old shield doesn't have Respawn/SelfHealing while the new one has it, the timer will be restarted properly
- if the old shield type has Powered=true while new type doesn't, it'll now be handled properly
- sync properties when transferring attached effects and shield during (Un)DeploysInto
- non positive damage and Berzerk warhead shouldn't trigger Respawn/SelfHealing.RestartInCombat now, like Ares's `SelfHealing.CombatDelay` did
- Respawn/SelfHealing.RestartInCombat will be processed as long as a shield entity exists
- removed some useless check. Available is not needed at all since it'll never be false

Coronia:
Fixed as much as issues I could find and tested. There might still be hidden issues for edged cases, especially when the Respawn/SelfHealing timer, the warhead modifier timer and the combat restart timer convoluted together. Honestly I don't even know why warhead modifier is designed like this as it created many code branches which is hard to maintain and prone to error, and the status overriding is not intuitive either

Anyway, since this PR contains many pieces of changes here and there which are for specific cases, guess it'll take too much while gain too little to review it. I'll merge it today even if there's no review, unless someone like to do so or pose an objection

PR #2389, opened 2026-09-04
[Customization] Allow customize that whether `Temporal=yes` warhead will cause target building animation poweroff

NetsuNegi:

## Description

In `rulesmd.ini`:
```ini
[General]
Temporal.KillPoweredAnim=true   ; boolean
```

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -572,6 +572,7 @@ This page lists all the individual contributions to the project by their author.
   - Fix the issue where vehicles always finish turret resetting first before turn to a new attack target, now it should turn to new target immediately
   - Fix a bug that computer player record cannot be log normally in non English mode
   - Cloak Enhancement
+  - Fix the bug that `Temporal=yes` warhead will cause target building animation poweroff
 - **Apollo** - Translucent SHP drawing patches
 - **ststl**:
   - Customizable `ShowTimer` priority of superweapons
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -572,7 +572,7 @@ This page lists all the individual contributions to the project by their author.
   - Fix the issue where vehicles always finish turret resetting first before turn to a new attack target, now it should turn to new target immediately
   - Fix a bug that computer player record cannot be log normally in non English mode
   - Cloak Enhancement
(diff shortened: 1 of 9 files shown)


Coronia (approved):

DeathFishAtEase:
Tested by: https://www.phoboscn.top/t/topic/763/3?u=noble_fish

image

Starkku:
Is this actually a bug though

NetsuNegi:
> Is this actually a bug though

alright, I added a global switch for that

Purplexile:
> Is this actually a bug though

Its performance in the original is shown in the figure.
https://www.phoboscn.top/uploads/default/original/2X/6/6377d4ca05fc62cee64ea892ab4aa1884319e406.gif
Temporal PowerAnim bug
At least its performance on the PrismTower seems to be a bug.

mevitar:
Genetic Mutator power lines disappearing isn't a bug. The game simply switches to first frame of "SW fired" animation sequence when the building is powered down. Although making it customizable is a welcome addition.

However, the Prism Tower thing looks like an issue with its charge-up logic, as this doesn't occur when the tower isn't firing, only when frozen during charge-up. Looks like it switches to ActiveAnim= but the game doesn't make it transparent like the rest of the building.

Purplexile:
> Genetic Mutator power lines disappearing isn't a bug. The game simply switches to first frame of "SW fired" animation sequence when the building is powered down. Although making it customizable is a welcome addition.

The description here is not entirely accurate. In fact, when the Genetic Mutator in the "SuperAnimThree" stage is affected by being powered down, due to the existence of "SuperAnimThreePoweredEffect=true", it will switch to a dedicated powered-down animation (SuperLowPower=YAGNTC_P), rather than "simply switches to the first frame of the 'SW fired' animation sequence' as mentioned here".
Moreover, when it is attacked by the Temporal=yes warhead, this temporarily switched-out "SuperLowPower animation" will not correctly receive the transparent effect, which is contrary to the design of the Temporal warhead. Therefore, it should also be a defect. According to the design of the "freezing" of the target by the Temporal=yes warhead, it should cause the suspension of all 'building animations' of the target as the correct manifestation.

PR #2393, opened 2026-09-07
Veins palette fix

FS-21:
Fixed the palette that should use the veins. If I'm not wrong now works like in TS.

image


Coronia (approved):
since vein is not being actively used in RA2 mods, you can mark it as tested if you've done so by yourself

FS-21:
Yes I tested it in my mod because I use it in TS stuff but additonal testers are welcome.

Coronia:
@FS-21 this broke vanilla Tiberium's palette in a way that adding house color on top of it
image

FS-21:
fixed, should I create another pr?

PR #2396, opened 2026-09-08
Recursive conversion for spawners

FS-21:
Currently, when a spawner is converted into a type with a different Spawns, all spawned aircraft are destroyed and the manager is recreated from scratch.

Now spawned aircrafts are preserved across conversions and recursively converts existing spawned units (both in-flight and docked) to the new Spawns type instead of killing them.
In addition, if the spawner is actively attacking a target, newly ready or regenerated spawns will now automatically launch towards it instead of staying docked in cooldown.

spawners_improvement_01


Coronia:
if a missile spawn is converting to a jet during taking off it'll behave weirdly. But if it's already cruising on the sky then it'll behave normally
8108e17828bacafaf4372ba4566171b6

on the other hand, if a spawn unit is converting into a missile spawn one during taking off, the missile will stop on the sky and don't move, and it cannot regen missile either. They'll move again when converting back
18e2c73f8b6cd66e57d3d4322c84dc3e

either we should do more compatibility here or just kill all spawns like we did before

Coronia:
funnier thing, if a spawn unit is converting into a missile spawn one during the jet is returning, it'll change into missile and hit itself
613e4be7c9b6e04ee36e29d0fafc018a

and idk how to explain this. It's missile spawn to missile spawn this time
ce2f7ea87e08f24e4b14e5fcf3a7dd77

Coronia:
one more thing, if a spawn jet with Strafing weapon converts into another spawn jet without it, it might still perform a Strafing attack with the new spawn jet's weapon. This is more like a problem of conversion itself tho, but it would need to be more compatible to cooperate with this feature

TaranDahl:
Suggestion: Ignore all MissileSpawn. Do nothing when the from or to is MissileSpawn.

Fly-Star-him:
Perhaps you could add a `Convert.SpawnsConversion` tag to allow users to customize what can be converted.

FS-21:
I checked everyones feedback. Then both suggestions should be checked here right?  I mean:


- Ignore all MissileSpawn. Do nothing when the from or to is MissileSpawn.
- Convert.SpawnsConversion tag to allow users to customize what can be converted.


### Recursive conversion of spawned units

- When a spawner unit converts to another type with different `Spawns`, existing active and docked spawned units can now be recursively converted to the new `Spawns` type instead of being destroyed.
  - `Convert.SpawnsConversion` on `[TechnoType]` controls whether existing spawned units are recursively converted when the spawner undergoes type conversion. Defaults to `false`.
  - When either the old or the new spawner uses `MissileSpawn=yes`, spawn conversion is automatically bypassed to prevent suicidal pathing or frozen missile states.

In `rulesmd.ini`:
```ini
Convert.SpawnsConversion=false  ; boolean
```

fagonghaiwo:
5ec44b83e5af8d1c7828dbaa9ec73080

[CARRIER]
IsSimpleDeployer=yes
Convert.Deploy=LEVI
Convert.SpawnsConversion=1

not Convert but fall

PR #2398, opened 2026-09-09
Set the initial elite-level units that a country can produce

dh381-1:

## Description

- Now you can add `TechnoTypes` in the five lists. They will appear at an elite initial level when produced, and cloned infantry will also be affected by this effect.
- Use `Elite.AffectsAll` to configure this. If it is `true`, any unit in the list that appears by any means will Initial elite; if it is `false`, it will only affect produced units, cloned units, initial garrisons of buildings, and initial passengers of vehicles.

In `rulesmd.ini`:
```ini
[SOMECOUNTRYIES]        ; Country
EliteInfantry=          ; List of InfantryTypes
EliteUnits=             ; List of UnitTypes
EliteAircraft=          ; List of AircraftTypes
EliteBuildings=         ; List of Buildings
EliteDefenses=          ; List of Defenses
Elite.AffectsAll=true   ; boolean
```

- For each `TechnoType`, you can define `SpareCameo` and `SpareCameoPCX` to specify the elite‑level icon. When a `TechnoType` is added to the initial elite list, it will preferentially use the SpareCameo as its icon. Supports `SHP`/`PCX` formats.

In `artmd.ini`:
```ini
[SOMETECHNO]        ; TechnoType
SpareCameo=         ; filename
SpareCameoPCX=      ; filename - include ".pcx" extension
```


DeathFishAtEase:
The flag name design is somewhat unclear:

- If not deliberately aligning with vanilla flags, there is no need to set 3 new flags.
- If designed to align with vanilla flags, it should align with the dual-flag design of `Cameo` and `CameoPCX`, and only the PCX version needs to have the file extension filled in.

DeathFishAtEase (changes requested):
The current implementation has a problem that causes debug.log to output parsing errors.

In `rulesmd.ini`:
```ini
[French]
EliteInfantry=GGI
EliteUnits=MTNK
EliteAircraft=ORCA
```

In `debug.log`:
````
[Phobos] Failed to parse INI file content: [French]EliteInfantry=GGI
[Phobos] Failed to parse INI file content: [French]EliteUnits=MTNK
[Phobos] Failed to parse INI file content: [French]EliteAircraft=ORCA
```

DeathFishAtEase:
[Not a change request] I am wondering which would be better: adding 3 country-level lists in the manner aligned with vanilla, or extending Ares' existing Academy.Types feature set - such as adding country filtering, allowing icon changes, updating experience once when the unit's owner changes, etc.
If you have the energy to complete it before the first Pre-Release of v0.6, then it could replace both the current PR content and the existing 020d3425819bcc89075e844584e3873f444abe84.

dh381-1:
> [Not a change request] I am wondering which would be better: adding 3 country-level lists in the manner aligned with vanilla, or extending Ares' existing Academy.Types feature set - such as adding country filtering, allowing icon changes, updating experience once when the unit's owner changes, etc. If you have the energy to complete it before the first Pre-Release of v0.6, then it could replace both the current PR content and the existing [020d342](https://github.com/Phobos-developers/Phobos/commit/020d3425819bcc89075e844584e3873f444abe84).

I think this PR only needs to implement a list like the vanilla one; extending Ares' functionality can be done in a separate PR. Also, for `020d342`, do I need to open a new PR to add icon replacement functionality?

PR #2403, opened 2026-09-12
Fixed the issue#2361 that Phobos crashed when Antares drew Ebolt colors.

dh381-1:

## Description

When Antares and Phobos are both present at the same time, Antares causes Phobos to dereference a null pointer, resulting in a crash. It should now run normally.


DeathFishAtEase:
This PR is premature. Phobos has not yet been made compatible with Antares, and testing after Antares completes its refactoring of 3.0p1 has also not been completed (currently there is only some discussion on Discord about its performance being inferior to Ares 3.0p1).
In addition, Phobos has a large amount of content that interacts with Ares implemented through patch ares; they require the exact addresses of Ares 3.0(p1), while Antares obviously is not completely identical to Ares at the binary level (otherwise there would be no need to test the refactoring results). Therefore, using Phobos together with Antares will inevitably have a large number of incompatibility issues, not just EBolt.

Coronia:
put this PR in On Hold stat before we make the decision that Phobos should support Antares

